PR #196, opened 2013-03-30
[goolf/1.4.10, ictce/5.3.0] octave* (DROP)

fgeorgatos:
Only tested for some of the builds, but should not be far from final.

N.B. This version does *NOT* include MPITB, it is the plain versions.


Bart-CER:
Looks like octave will not make it into Easybuild?

fgeorgatos:
It (DROP) just means that this PR is practically orphan (from my side) and somebody else is welcome to take over it. (hm, maybe `(ORPHAN)` is a better tag... be my guest)

fgeorgatos:
@Bart-CER, it is quite certain that sooner or later an HPC site will have the *itch* and interest to provide this, hopefully along with the parallel components (`mpitb` with some older version, perhaps). I just don't have a thorn on my side for a while ;-)

fgeorgatos:
Hi, I'm closing this since it is an orphan PR and may better be followed up in #1366 and onwards ;
 
I had liked @boegel 's idea of using PETSc for exploiting common deps; hairy ball ensued; 
it's nice if dependencies versions will be aligned but feel free to provide whatever. OOMH.

PR #241, opened 2013-04-26
[ictce/5.3.0] ROOT, NCL, PyTables/python-meep, NWChem, DOLFIN, GATE, WIEN2k (DROP)

fgeorgatos:
as requested by @boegel

There are issues for these, see #229


fgeorgatos:
ROOT, and possibly GATE, have been fixed for later ictce toolchains of the same major (ie. 5.5.0), 
so I am inclined to close this issue, since it is by now lost in irrelevance.

PR #418, opened 2013-08-29
[goolf/1.4.10] octave/3.6.4 over qhull/2003.1 (DROP)

fgeorgatos:
these are meant as minimal builds, that don't depend on the multitude of deps that can be added on (it is the working subset of the already existing #196)


boegel:
@fgeorgatos: care to clarify the `DROP`?

fgeorgatos:
Hi, I'm closing this since it is an orphan PR and may better be followed up in #1366 and onwards ;
 
This PR included @rtmclay's spec from TACC which I certainly endorse for a full-fat build of Octave;
OOMH. Hoping somebody has the itch for this. Anno Domini 2015...

PR #494, opened 2013-11-02
add support for graphviz, HarfBuzz and Pango

boegel:
this builds upon the works done in #330 and #453


boegel:
I'm going to close this as since, since:

* it's using a very old toolchain
* easyconfigs for newer versions of both `cairo` and `Pango` have been included for a while

PR #539, opened 2013-11-15
Add new easyconfig for CP2K compiled with ictce-4.1.13. Modify patch.

apanteli:
I had to modify fix_compile_date_lastcvs.patch. Apparently the Makefile we have, is different from the one that was assumed in the existing patch.
Additionally I had to disable the testing function in the easyblock because it failed there, even though there is the `ignore_regtest_fails = True` in the config file.


A1ve5:
Sorry for reopening this...
@boegel Could you please make this file CP2K-20111205.tar.gz available again? Or how to produce it?
trying to compile CP2K-20111205-goolf-1.4.10-psmp.eb
TIA

boegel:
@A1ve5 Sorry, your question drowned in my overfull inbox...

Do you stil need that ancient version of CP2K?

I highly recommend using the latest versions, which have a proper versioning scheme (e..g `5.1`).

A1ve5:
Hi @boegel, we don't need it anymore. Thanks!

PR #699, opened 2014-02-10
contrib blasr configs.

pforai:
This is a config for BLASR from the PacBio SMRT 2.1 analysis toolkit plus patch.


boegel:
merged via #1662, thanks @pforai!

PR #701, opened 2014-02-13
Add support for SPM - needs work

georg-rath:
These configs need the spm easyblock in this PR: https://github.com/hpcugent/easybuild-easyblocks/pull/358 .

SPM is a MATLAB script, where the "performance-critical" parts have been compiled using the MATLAB-Compiler mcc. mcc needs a loaded GCC (4.4.x is supported, but it seems that 4.8.2. works fine).
This config/block uses the system GCC, which kinda sucks. What I would do is add MATLAB to the GCC toolchain (dummy right now) and solve the problem that way, but I am not sure if this is the right approach. Would be glad if someone who actually knows what they are doing could comment on this.

Stuff to do:
- Fix this dependency problem
- Do some testing (right now this builds and loads fine in MATLAB, but these are the only things I checked.)
- Fill in the right description

### DIFF
--- a/easybuild/easyconfigs/s/SPM/SPM-12b.5852.eb
+++ b/easybuild/easyconfigs/s/SPM/SPM-12b.5852.eb
@@ -27,12 +27,21 @@ source_urls = [
 
 sources = [ 'spm%(version_major)s.zip' ]
 
+mexfiles = [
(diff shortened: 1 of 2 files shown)


boegel:
@georg-rath: no, that doesn't work yet, but you can open a PR for it to https://github.com/hpcugent/easybuild-framework ;-)

migueldiascosta:
@georg-rath, this PR is being closed for the following reason(s): inactive.

PR #743, opened 2014-03-11
Add pbs_python to the growing suite of software supported by easybuild.

ehiggs:
This was used at the UGent site for installing pbs_python. It's a dep for hanythingondemand.


boegel:
easyconfig for pbs_python is included in #1520

PR #757, opened 2014-03-19
Add OpenCV without CUDA.

ehiggs:
This is OpenCV for the ictce 5.5.0 toolchain without CUDA. One of the challenges we've had is to make sure we have all the deps taken care of. Running `ldd` on the binaries in the installed directories and filtering out EasyBuild dependencies results in the following:
```
-bash-4.1$ ldd ~/.local/easybuild/software/OpenCV/2.4.8-ictce-5.5.0/bin/* | grep -v `whoami`| grep -v sandy | sort  | awk '{print $1}' | uniq
/lib64/ld-linux-x86-64.so.2
libc.so.6
libdl.so.2
libgcc_s.so.1
libglib-2.0.so.0
libgthread-2.0.so.0
libm.so.6
libpthread.so.0
librt.so.1
/$LIB/snoopy.so
libstdc++.so.6
linux-vdso.so.1
```

This required libjpeg (for which I've used libjpeg-turbo) and thus a toolchain version of NASM.

### DIFF
--- a/easybuild/easyconfigs/l/libjpeg-turbo/libjpeg-turbo-1.3.0-ictce-5.5.0.eb
+++ b/easybuild/easyconfigs/l/libjpeg-turbo/libjpeg-turbo-1.3.0-ictce-5.5.0.eb
@@ -1,0 +1,22 @@
+name = 'libjpeg-turbo'
+version = '1.3.0'
+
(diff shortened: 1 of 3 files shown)


ehiggs:
OpenCV was merged w/ #1088

PR #768, opened 2014-03-24
OpenBUGS-3.2.3.eb

pescobar:
As this app requires 32bit support, this easyconfig is forced to use the system compiler instead of an EB toolchain and verifies that glibc-devel.i686 is installed


fgeorgatos:
Test report by @fgeorgatos
**SUCCESS**
Build succeeded for 1 out of 1
Linux debian 6.0.10, Intel(R) Xeon(R) CPU           L5640  @ 2.27GHz, Python 2.6.6
See https://gist.github.com/4bbae4e3988f036479aa for a full test report.

PR #829, opened 2014-04-17
SNPhylo easybuild script

ehiggs:
This is the related scripts for hpcugent/easybuild-easyblocks#394
It includes all the dependent projects. I also have a script for MUSCLE which builds from source but I didn't include it since there is one which just takes the linux64 elf.


ehiggs:
Shall I close this?

boegel:
@ehiggs yes, I think this is of little use since we already have SNPhylo-20140701-foss-2016a.eb  & SNPhylo-20140701-intel-2016a.eb merged in

ehiggs:
Ok. Closing. :bomb:

PR #839, opened 2014-04-24
add easyconfigs for PerfExpert, ROSE, HPCToolkit + deps

boegel:
I've deliberately used a `dummy` toolchain here, to obtain a PerfExpert build that works with other toolchains too...

This is not airtight though, mostly because of the huge list of dependencies. If PerfExpert is to be used with a particular application that has a dependency in common with PerfExpert, PerfExpert will need to be built with the same compiler toolchain to make them compatible.

This might be troublesome, since e.g. ROSE requires a GCC 4.4 or older 4.x (so I should probably use a GCC-4.4 module as build dependency for ROSE to make sure that build works).

### DIFF
--- a/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69.eb
+++ b/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69.eb
@@ -1,0 +1,20 @@
+name = 'Autoconf'
+version = '2.69'
+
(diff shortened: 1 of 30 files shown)


migueldiascosta:
@boegel, this PR is being closed for the following reason(s): no activity for > 6 months.

PR #890, opened 2014-05-27
Orca 3.0.2

scintimandrion:
This is a new config file for ORCA 3.0.2. Bart Verleye suggested that I could request its addition to the repo.


easybuild/easyconfigs/o/ORCA/ORCA-3_0_2-linux_x86-64_openmpi_165.eb
```diff
@@ -0,0 +1,32 @@
+easyblock = "PackedBinary"
+
+name = "ORCA"
+version = '3_0_2-linux_x86-64'
```
boegel: you're missing a versionsuffix here, the easyconfig filename now doesn't match the contents, please fix (and try not to hardcode the OpenMPI version in the versionsuffix, only specify it once, see easyconfigs that have `-Python-2.7.8` as versionsuffix, for example)

scintimandrion:
Version suffixes updated.

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 3 (3 easyconfigs in this PR)
See https://gist.github.com/3eac8dba21a5c40c7c8a for a full test report.

easybuild/easyconfigs/o/ORCA/ORCA-3_0_0-linux_x86-64-OpenMPI-1.6.5.eb
```diff
@@ -1,20 +1,23 @@
 easyblock = "PackedBinary"
 
 name = "ORCA"
-version = '3_0_0-linux_x86-64_openmpi_165'
+version = '3_0_0-linux_x86-64'
```
boegel: We only have sources for ORCA v3.0.0 named `orca_3_0_0_linux_x86-64_openmpi_165.tbz `, so it seems like in *this* case the `_openmpi_165` should be part of the version after all...

boegel: fixed via #1449, this PR will be merged automatically when #1449 is merged in

boegel:
merged via #1449, thanks @scintimandrion!

Issue #605, opened 2013-12-10
GOOLF Toolchain OpenBLAS and core affinity is wrong with R and potentially others.


pforai:
There is an issue at least with R when it calls into OpenBLAS, OpenBLAS sets the thread's affinity to the first cpu core (0x1) and each subsequently call into OpenBLAS does the same, so effectively builds of R that run BLAS routines are pinned to the same core. 

Not sure if this problem persists with other software that uses OpenBLAS in the same manner as R. 

The problem is fixed if OpenBLAS is recompiled with NO_AFFINITY=1, which is not set by default. This might surprise some users, especially the ones that use the multicore package in R to do data parallel computing and R is compiled with OpenBLAS.

The default ICTCE builds don't have that issue, but for them the sequential MKL is linked in in the default block.

boegel:
Fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/916 by setting `NO_AFFINITY=1` when building OpenBLAS, so closing this.

PR #916, opened 2014-06-09
use NO_AFFINITY=1 to disable affinity being set by OpenBLAS itself by default

boegel:
This should fix (part of) the problem reported in #605.

### DIFF
--- a/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.2.6-cgmpich-1.1.6-LAPACK-3.4.2.eb
+++ b/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.2.6-cgmpich-1.1.6-LAPACK-3.4.2.eb
@@ -34,9 +34,8 @@ patches = [
 
 skipsteps = ['configure']
 
-threading = 'USE_THREAD=1'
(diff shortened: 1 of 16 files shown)


boegel:
Hmm, we never followed up on this...

@fgeorgatos, @pforai: should we still consider using `NO_AFFINITY=1` for all OpenBLAS versions?

@xianyi were any bugs/problems related to the OpenBLAS affinity solved recently, that may explain what we were seeing?

xianyi:
@boegel , we already disable affinity in OpenBLAS by default.

boegel:
@xianyi wait, so passing `NO_AFFINITY=1` doesn't make any difference? regardless of the OpenBLAS version?

xianyi:
For some old version (2 years old), you need pass `NO_AFFINITY=1`

ocaisa:
We ran into a similar issue with `cp2k.popt` in `CP2K`, with the current (default) settings of `OpenBLAS` in EB, OpenBLAS forks NCPU threads per MPI process that flood the node and kill performance.

@damianam configured OpenBLAS with `USE_OPENMP=1` and that seemed to fix the issue. That hint came from https://github.com/xianyi/OpenBLAS/issues/905

boegel:
follow up in #4524, I'm closing this PR since it's not worth fixing these ancient OpenBLAS easyconfigs imho

PR #929, opened 2014-06-14
add additional easyconfigs for FSL/fmri + deps

boegel:
depends on #908


boegel:
superseded by #1550 and #2775, etc., so closing as is

PR #932, opened 2014-06-18
Contrib goolf 1.4.20

fgeorgatos:
this goolf uses the same GCC compiler like #928 (GCC/4.7.4) and provides incremental upgrades for all the underlying components; this toolchain is expected to have maximal compatibility with debuggers.

The intention is to use it with recursive --try, to provide debugg-able builds.


fgeorgatos:
#1294 is far more modern than this PR, so closing...

PR #877, opened 2014-05-19
Cuda 6.0.37

pescobar:
added %(version_major)s_%(version_minor)s to source_urls so it works with future major releases

reviewing the install_dir I see these folders which contains a few libraries and binaries:

```
libnvvp  nvvm  open64
```

but those are not included in LD_LIBRARY_PATH, LIBRARY_PATH and PATH in the generated modulefile. There is also a new folder "pkgconfig" which was not present in cuda-5.x releases.  

Maybe we should improve the cuda easyblock to add this in the modulefile? 

maybe @ajdecon can give some advice about this?

PR #994, opened 2014-08-06
Adding CUDA v.6.0.37

tloizou:


fgeorgatos:
@boegel, @thekla: is this now a duplicate of #877? 

please check and act accordingly!

boegel:
Nice catch, yes, I think this one has to be closed, #877 covers it.

PR #1014, opened 2014-08-07
add easyconfig for intel 2014.08 toolchain, which is compatible with a hierarchical module naming scheme

boegel:
This update of the `intel` compiler toolchain not only updates the components to the latest available versions (except for `impi`, not sure if going for v5.0 is a good idea there), it also resolves a couple of problems w.r.t. compatibility with using a hierarchical module naming scheme.

In particular:

* the `GCC` easyconfig which serves as a dependency for `icc`, `ifort` is composed such that no `$MODULEPATH` extensions are included in the generated module file
* the `impi` easyconfig uses a non-dummy toolchain, i.e. the `iccifort` toolchain
* the `imkl` easyconfig uses a non-dummy toolchain, i.e. the (new, cfr.) `iimpi` toolchain

CC: @wpoely86, @kcgthb, @geimer, @ikirker


boegel:
This is no longer relevant; the issue with the GCC on which icc/ifort rely extending `$MODULEPATH` is fixed by @ocaisa introducing the concept of `GCCcore`

PR #1145, opened 2014-10-16
libint2 renamed to Libint

hajgato:
This renaming should be discussed. There is a corresponding easyblock PR. https://github.com/hpcugent/easybuild-easyblocks/pull/489


boegel:
merged via #1287, thanks @hajgato!

PR #1169, opened 2014-10-29
Easyconfigs for NCO and its dependency ANTLR

pm67nz:
Two simple new easyconfigs and a supporting patch file.


boegel:
merged via #1980, thanks @pm67nz!

PR #1175, opened 2014-10-30
Julia easyconfig

pm67nz:
This easyconfig requires the Julia easyblock which puts the configopts string into a file named Make.user.

edit: depends on https://github.com/hpcugent/easybuild-easyblocks/pull/499


boegel:
Jenkins: test this please

PR #1178, opened 2014-10-31
distutils_multiple-lib-dirs patch file for numpy 1.8.1

pm67nz:
Also known to work with numpy 1.8.2, this patch is equivalent to the similarly named one for numpy 1.7.1, but some lines have moved around in the target files.


boegel:
we already have this patch included now, via #1517, so ok to close

Sorry for not picking up your contribution @pm67nz, we missed it for some reason... :(

PR #1204, opened 2014-11-21
DEXSeq R packahe and its dependecies with intel-2014b

hajgato:


boegel:
Jenkins: test this please

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/76b2b7a9af610b8bd5e1 for a full test report.

boegel:
@hajgato: maybe it's better to use a `Bundle` for this, like we recently did in #1387, especially since you're also adding an extension to the (old) R easyconfig.

Let me know what you think.

hajgato:
@boegel the extra R-packages I am installing here have their external deps, except one. One also have its own EasyBlock. So your suggestion is to make a bundle with a bunch of external deps? Additionally I have to solve that XML uses its own EasyBlock within the bundle, and the others the R-package EasyBlock. Or you want me to bundle only those that depend only on each other, and not external libs? Anyway, what should be the name whatever case I will we choose?

boegel:
@hajgato: don't worry about R packages requiring a custom easyblock, `eb` takes care of that itself (it tries to find an easyblock matching the pkg name first, and then falls back to `RPackage`)

A bundle with the external deps listed makes sense, yes.

XML is already included in the `extra` R-bundle.

Having `DEXSeq` in the name would make sense?

@ehiggs did mention a problem with the bundles: you can't have multiple ones loaded at the same time, since they'll conflict on version.

We will need a solution for that (before EB v2.2).

hajgato:
@boegel But this means that I am not able to do a bundle, because I need the `extra` R-bundle, and the new `DEXSeq` R-bundle as well, right?

hajgato:
@boegel: OK, got it. I will do it.

boegel:
@hajgato: actually, depending on the `-extra` bundle will work, but only accidentally (you can't run into a conflict while still loading the module...)

Just take that road, we'll fix the issue with the R-bundle's all in one go.

You should probably go for a module like `R-bundle-DEXSeq/20150603-R-3.*`, which seems to be the solution we will be taking.

hajgato:
@boegel I thoght that I will do a Bundle with the name DEXSeq-1.10.8. Whats the use to make a R-bundle-DEXSeq/R-3* and a separare DEXSeq-1.10.8 Easyconfig if it is only loads R-bundle-DEXSeq/R-3*, if I can make thw whole ting in one easyconfig?

boegel:
@hajgato: that's true, but you'll need separate easyconfigs for the dependencies too (unless you make that a bundle *for* DEXSeq, that works too I guess)

migueldiascosta:
@hajgato, this PR is being closed for the following reason(s): no activity for > 6 months.

PR #1218, opened 2014-12-02
GROMACS 4.6.7 +  Plumed

wpoely86:
WARNING: do NOT merge. This needs work (a lot of it).

Plumed needs to be build alongside GROMACS because plumed has the needed patch on board using a dirty patch system. GROMACS also needs the object files from the plumed build to build itself.


pforai:
@wpoely86 any idea when you could continue to work on this? Would be interested in PLUMED. Could you maybe do a separate PR for PLUMED solo?

wpoely86:
@pforai I don't have time in the coming months. Basically the problem is that the `GROMACS-4.6.7-plumed.patch` is automatically generated by plumed using the gromacs build dir. This needs to be implementend in the easyblock or the patch needs to be dynamically adjusted by EB.

wpoely86:
Plumed split off to #1596

boegel:
@wpoely86 Is this still relevant? Cfr. the work done in https://github.com/hpcugent/easybuild-easyblocks/pull/1121

wpoely86:
Maybe, not sure. Probably not. Anyway, I don't have time for it. So closing.

PR #1226, opened 2014-12-05
add easyconfigs for recent Mesa with intel/2014b

boegel:

PR #1227, opened 2014-12-05
add easyconfig(s) for ROOT v5.34.23 with intel/2014b

boegel:
depends on #1226


boegel:
obsolete, no point in following up on this

PR #1233, opened 2014-12-09
Ruby 2.1.5. with intel2014b

hajgato:

PR #1258, opened 2014-12-16
Cuby 4

hajgato:
depends on #1233

### DIFF
--- a/easybuild/easyconfigs/c/Cuby/Cuby-4.eb
+++ b/easybuild/easyconfigs/c/Cuby/Cuby-4.eb
@@ -1,0 +1,27 @@
+easyblock = "PackedBinary"
+
+name = "Cuby"
+version = '4'
+
+homepage = 'http://cuby.molecular.cz/cuby4/'
+description = """Cuby is a computational chemistry framework written in ruby.
+ For users, it provides an unified access to various computational
+ methods available in difefrent software packages. For developers,
+ Cuby is much more - it is a complex framework that provides
+ object-oriented access to the data enetering the calculations and
+ to their results, making it easy to create new computational
+ protocols by combining existing blocks of the framework."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+#Source can be obtained via http://cuby.molecular.cz/?page=Downloads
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/c/Cuby/Cuby-4.eb
```diff
@@ -0,0 +1,27 @@
...
+homepage = 'http://cuby.molecular.cz/cuby4/'
+description = """Cuby is a computational chemistry framework written in ruby.
+ For users, it provides an unified access to various computational
+ methods available in difefrent software packages. For developers,
+ Cuby is much more - it is a complex framework that provides
+ object-oriented access to the data enetering the calculations and
+ to their results, making it easy to create new computational
+ protocols by combining existing blocks of the framework."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
```
wpoely86: why the `dummy` toolchain? Remember, if the version is also `dummy`, dependencies are not loaded

wpoely86:
ok to test

easybuild/easyconfigs/c/Cuby/Cuby-4.eb
boegel: +1

I would even go as far as using a non-dummy toolchain here, since you're picking a toolchain anyway when specifying the dependency

that'll make it clear that this module will also result in loaded `intel/2014b`

boegel:
Jenkins: test this please

easybuild/easyconfigs/c/Cuby/Cuby-4.eb
hajgato: I do not rally mind to add the toolchain, but anyway, it is a packed binary. So in theory it does not need a toochain. This is why I have chosen to add a dummy toolchain. Anyway, I will put intel/2014b.

boegel:
@hajgato: since there's no 'proper' version, should we add a checksum for the source tarball?

boegel:
@hagato informed me that licensing info is included in the tarball, so checksum is not an option

Fine, merging this in then, looks fine.

PR #1275, opened 2014-12-18
add easyconfig file for EasyBuild v1.16.0

boegel:


boegel:
working great now, so going in, thanks for the review @wpoely86, @fgeorgatos!

PR #1336, opened 2015-01-26
{cae}[dummy] ABAQUS v6.14.1 (REVIEW)

boegel:


easybuild/easyconfigs/a/ABAQUS/ABAQUS-6.14.1-linux-x86_64.eb
```diff
@@ -0,0 +1,15 @@
...
+homepage = 'http://www.simulia.com/products/abaqus_fea.html'
+description = "Finite Element Analysis software for modeling, visualization and best-in-class implicit and explicit dynamics FEA."
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = ['%s-%s.tgz' % (name, version.split('-')[0])]
+
+builddependencies = [('Java', '1.7.0_60')]
+
+preinstallopts = "export CHECK_DISK_SPACE=OFF && "
```
wpoely86: isn't `CHECK_DISK_SPACE=OFF ` enough?

boegel:
Test report by @boegel
**SUCCESS**
Linux SL 6.5, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/4c2afdd4842fc99ec3c9 for a full test report.

easybuild/easyconfigs/a/ABAQUS/ABAQUS-6.14.1-linux-x86_64.eb
boegel: it may not be, only exported variables are passed down into child shells:

```
[kehoste:~] $ FOO=bar /bin/bash -c "echo $FOO"

[kehoste:~] $ export FOO=bar && /bin/bash -c "echo $FOO"
bar
```

wpoely86: I know, but this seems to be a binary install? If not, better keep it.

boegel: the installation is done using a script, which may fork shells, so it should most probably stay (I seem to remember doing it without the export first)

boegel:
Going in, thanks for the review @wpoely86!

PR #1339, opened 2015-01-27
{devel}[intel/2015a] libX11 + deps (REVIEW)

gjbex:
Latest versions of
* libX11
* inputproto
* kbproto
* libXau
* libpthread-stubs
* libxcb
* xcb-proto
* xextproto
* xproto
* xtrans

PR #1340, opened 2015-01-27
{devel}[intel/2015a] Tk + fix for Tcl (depedency on PR 1339) (REVIEW)

gjbex:
Added latest Tk, and added required dependency on zlib to Tcl easyconfig (needed since Tcl 8.6.x).  This PR depends on PR 1339.

### DIFF
--- a/easybuild/easyconfigs/t/Tcl/Tcl-8.6.3-intel-2015a.eb
+++ b/easybuild/easyconfigs/t/Tcl/Tcl-8.6.3-intel-2015a.eb
@@ -12,6 +12,10 @@ toolchain = {'name': 'intel', 'version': '2015a'}
 source_urls = ["http://prdownloads.sourceforge.net/tcl"]
 sources = ['%(namelower)s%(version)s-src.tar.gz']
 
+dependencies = [
+    ('zlib', '1.2.8'),
+]
+
 configopts = '--enable-threads EXTRA_INSTALL="install-private-headers"'
 
 runtest = 'test'
--- a/easybuild/easyconfigs/t/Tk/Tk-8.6.3-intel-2015a.eb
+++ b/easybuild/easyconfigs/t/Tk/Tk-8.6.3-intel-2015a.eb
@@ -1,0 +1,26 @@
+easyblock = 'ConfigureMake'
+
+name = 'Tk'
+version = '8.6.3'
+
+homepage = 'http://www.tcl.tk/'
+description = """Tk is an open source, cross-platform widget toolchain that provides a library of basic elements for building
+ a graphical user interface (GUI) in many different programming languages."""
+
+toolchain = {'name': 'intel', 'version': '2015a'}
+
+source_urls = ["http://prdownloads.sourceforge.net/tcl"]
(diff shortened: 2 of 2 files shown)


easybuild/easyconfigs/t/Tcl/Tcl-8.6.3-intel-2015a.eb
```diff
@@ -12,6 +12,10 @@ toolchain = {'name': 'intel', 'version': '2015a'}
 source_urls = ["http://prdownloads.sourceforge.net/tcl"]
 sources = ['%(namelower)s%(version)s-src.tar.gz']
 
+dependencies = [
+    ('zlib', '1.2.8'),
```
boegel: Hmm, we would need to add this in the other Tcl easyconfigs too, but this can be a can of worms, since the zlib version used here has to match the zlib used by other entries in the dependency tree where `Tcl` is a part of...

@gjbex: Thoughts? Should we (try to) add this to the other Tcl easyconfigs too (separate PR, for sure)?

boegel:
Jenkins: ok to test

gjbex:
Test report by @gjbex
**SUCCESS**
Linux red hat enterprise linux computenode 6.4, Intel(R) Xeon(R) CPU E5-2680 v2 @ 2.80GHz, Python 2.7.6
See https://gist.github.com/f77d057ce636d73dce17 for a full test report.

gjbex:
Not fixing it implies that Tk/8.6.x won't build, at least not for intel/2015a and x == 3.

boegel:
@gjbex: I'm not opposing including the zlib dep in this Tcl easyconfig, I'm wondering whether zlib should also be added as a dep in the existing Tcl easyconfigs...

boegel:
@gjbex: missing easyconfig for `libX11/1.6.2-intel-2015a-Python-2.7.9`, according to Jenkins

boegel:
so, this depends on #1339? please indicate this in the pull request description

gjbex:
True, this won't build without PR 1339, apologies, forgot to mention.  Modified PR title accordingly.

boegel:
@gjbex: just mentioning it in the description is fine, no need to blow up the PR title imho

gjbex:
I understand what you mean by the remark on zlib.  The situation is probably similar as with Python/sqlite3.  It will build without problems, but fail to include a package that depends on the missing build dependency.  zlib in this case, sqlite3 for Python.
Everything works, as long as you don't need the zlib package, and apparently Tk does.

So indeed, to have any version of Tcl 8.6.x working properly, the easyconfigs should be modified, but as you indicate, it may be tricky.  My impression is that quite a number of easyconfigs for 2014b toolchains depend on zlib 1.2.7, rather than 1.2.8.

boegel:
@gjbex: the good news is that the unit tests will tell us if there's a conflict w.r.t. dependency versions somewhere, so it's pretty trivial to give it a shot; I'll look into that, in a separate PR

migueldiascosta:
@gjbex, this PR is being closed for the following reason(s): no activity for > 6 months.

PR #1344, opened 2015-01-27
{lang}[*] include zlib as a dependency in all Tcl easyconfigs (WIP)

boegel:
this syncs up existing easyconfigs like was done by @gjbex in #1340

the unit tests will check whether any conflicts were introduced at all (but I don't think there are)

@gjbex, @wpoely86: please review?


wpoely86:
looks fine

but maybe a more permanent solution is to add it to the toolchain?

boegel:
@wpoely86: that's a can of worms that we have to tackle carefully... but it's worth looking into, especailly for the common toolchains

wpoely86:
No bigger can of worms then adding as a dep on all kinds of easyconfigs?

boegel:
@wpoely86: I'm not sure... pushing zlib under GCC, or even binutils, is a lot lower in the stack, and affects *everything* being built with it

wpoely86:
well, they now silently fall back to system zlib. It's gonna give a lot more of `no version information`. That can potentially break some build systems but we already have patched several pieces of software for that.

boegel:
We can/should certainly look into this, but pulling zlib into the toolchain should also involve binutils, maybe even glibc (and more?)... Other common candidates are ncurses, libreadline, ...

wpoely86:
Definitely not glibc. That's a real can of worms. And the same reasoning behind openssl also holds here (see GHOST).

Anyway, we should probably open a dedicated bug report for that?

boegel:
@wpoely86: this one? https://github.com/hpcugent/easybuild-framework/issues/644

wpoely86:
Yes, and now actually do it :smiley:

boegel:
@wpoely86: Jenkins should be giving this a green light soon, last check? Good to go in?

easybuild/easyconfigs/z/zlib/zlib-1.2.5-ictce-5.5.0.eb
```diff
@@ -0,0 +1,24 @@
...
+ not covered by any patents -- lossless data-compression library for use on virtually any 
+ computer hardware and operating system."""
+
+toolchain = {'name': 'ictce', 'version': '5.5.0'}
+toolchainopts = {'optarch': True, 'pic': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = [('http://sourceforge.net/projects/libpng/files/zlib/%(version)s', 'download')]
+
+preconfigopts = 'LDSHARED="$CC -shared"'
```
wpoely86: drop this

wpoely86:
looks fine

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 27 out of 27 (27 easyconfigs in this PR)
See https://gist.github.com/54331d33d4920b70e9a8 for a full test report.

boegel:
Updated to include recent Tcl easyconfigs, and to fix remarks.

PR #1360, opened 2015-02-02
{toolchain} [iomkl/2015.01] (REVIEW)

pescobar:
with the previous openMPI easyconfig gfortran was being used instead of ifort when running "mpifort"

```
[escobar@maia ~]$ mpifort --version
GNU Fortran (GCC) 4.9.2
```

I have updated the configopts for OpenMPI and now ifort is being used

```
[pub@maia easybuild-easyconfigs]$ mpifort --version
ifort (IFORT) 15.0.1 20141023
```

@boegel can you review?

### DIFF
--- a/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.8.4-iccifort-2015.1.133-GCC-4.9.2.eb
+++ b/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.8.4-iccifort-2015.1.133-GCC-4.9.2.eb
@@ -16,6 +16,7 @@ configopts = '--with-threads=posix --enable-shared --enable-mpi-thread-multiple 
 configopts += '--enable-mpirun-prefix-by-default '  # suppress failure modes in relation to mpirun path
 configopts += '--with-hwloc=$EBROOTHWLOC '  # hwloc support
 configopts += '--disable-dlopen '  # statically link component, don't do dynamic loading
+configopts += ' CC=icc F77=ifort FC=ifort CXX=icpc '
 
 dependencies = [('hwloc', '1.10.0')]
 


easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.8.4-iccifort-2015.1.133-GCC-4.9.2.eb
```diff
@@ -16,6 +16,7 @@ configopts = '--with-threads=posix --enable-shared --enable-mpi-thread-multiple
 configopts += '--enable-mpirun-prefix-by-default '  # suppress failure modes in relation to mpirun path
 configopts += '--with-hwloc=$EBROOTHWLOC '  # hwloc support
 configopts += '--disable-dlopen '  # statically link component, don't do dynamic loading
+configopts += ' CC=icc F77=ifort FC=ifort CXX=icpc '
```
wpoely86: looks fine but doesn't easybuild already set those?

boegel: EasyBuild sets these in the environment, when building/installing something using the `iccifort` toolchain.

I'm not sure how this can fix anything...

@pescobar: are you saying that `mpifort` picks up `ifort` if `FC=ifort` was passed on the configure command line (which is different from just having `$FC` set to `ifort`)?

Are you sure you didn't have `$FC` defined when obtaining `ifort` via `mpifort --version`?

boegel:
This should also be applied to the other OpenMPI easyconfigs on top of `iccifort`...

fgeorgatos:
@pescobar, @boegel:
This rings a bell (or I think so); Is this in any way related to
https://github.com/hpcugent/easybuild-easyblocks/issues/100 ?
and esp. https://github.com/hpcugent/easybuild-easyblocks/issues/100#issuecomment-46543772

I'm still on stand-by, up to today, to hear a solid argument about which is the right way to do that and have the impression that people make choices on assumptions, which may not be universal.

fgeorgatos:
btw. 
I've found that a great litmus test is to consider such decisions on the assumption that you pass these components via a Continuous Integration/Delivery (CI/CD) mechanism. If doing so does not require a hack to do the intended thing (ie. you test/deliver what really should be tested/delivered) then the practice is very likely the correct one.

boegel:
@pescobar ping?

boegel:
@pescobar: ping again, is this still relevant?

pescobar:
@boegel I think this fix was needed in older easybuild versions. In the latest versions the required env vars are defined so this configopt is no longer needed

PR #1362, opened 2015-02-03
{devel}(goolf-1.5.14-Python-2.7.8) Instant 1.5.0 (REVIEW)

Bart-CER:
Only one 'small' issue, the link to download the src produces a very long and random file name when using wget. So I did the download outside the config/eb process.


Bart-CER:
Sorry, there is another issue, according to the doc, it also depends on NumPy, which is installed as part of our Python-2.7.8.
This should probably be added somehow too? However, how can it then be used when NumPy is part of the Python, without installing NumPy again separately?

easybuild/easyconfigs/i/Instant/Instant-1.5.0-goolf-1.5.14-Python-2.7.8.eb
```diff
@@ -0,0 +1,33 @@
+easyblock = "PythonPackage"
+
+name = 'Instant'
+version = '1.5.0'
+
+homepage = 'https://launchpad.net/instant'
+description = """Instant is a Python module that allows for instant inlining of 
+ C and C++ code in Python. It is a small Python module built on top of SWIG and Distutils.
+ It is part of the FEniCS Project (http://fenicsproject.org)."""
+toolchain = {'name': 'goolf', 'version': '1.5.14'}
```
wpoely86: add newline between description and toolchain

easybuild/easyconfigs/i/Instant/Instant-1.5.0-goolf-1.5.14-Python-2.7.8.eb
```diff
@@ -0,0 +1,33 @@
...
+
+#unfort. this doesn't work, as the file has a very long and strange file name when downloaded...
```
wpoely86: OK, we have the same problem with downloads from google drive. You add a comment where you have to download it

wpoely86:
ok to test

wpoely86:
If you use `python` as a dep, `numpy` is already there. There is no real need to specify it here that `numpy` is needed too.

boegel:
@Bart-CER: just maybe include a comment next to/above the `Python` dependency line mentioning that numpy is also required, and is expected to be delivered via this Python dep

boegel:
@Bart-CER: missing easyconfigs, see https://jenkins1.ugent.be/job/easybuild-easyconfigs-pr-builder/2234/console (follow the `Details` link in this PR page to obtain that link, then click the `Console output` link)

If these are provided via another PR, please mention that in the description

easybuild/easyconfigs/i/Instant/Instant-1.5.0-goolf-1.5.14-Python-2.7.8.eb
```diff
@@ -0,0 +1,35 @@
...
+
+toolchain = {'name': 'goolf', 'version': '1.5.14'}
+
+#download from https://bitbucket.org/fenics-project/instant/downloads/
+
+source_urls = ['']
+sources = ['']
+
+python = "Python"
+pythonversion = '2.7.8'
```
boegel: bump to 2.7.9?

easybuild/easyconfigs/i/Instant/Instant-1.5.0-goolf-1.5.14-Python-2.7.8.eb
```diff
@@ -0,0 +1,35 @@
...
+pythonshortversion = ".".join(pythonversion.split(".")[:-1])
+
+versionsuffix = '-%s-%s' % (python, pythonversion)
+
+#Depends on NumPy, which is included in this Python install
+dependencies = [
+    (python, pythonversion),
+    ('SWIG', '2.0.4', versionsuffix),
```
boegel: easyconfig for this is not available yet, is it in another PR somewhere?

easybuild/easyconfigs/i/Instant/Instant-1.5.0-goolf-1.5.14-Python-2.7.8.eb
Bart-CER: We'll have this problem on many eb's ... 
We use python 2.7.8, and at the moment we won't make the update yet. Not sure if you want me to have different sets of eb's, the ones that we use, and eb's for PRs? This would involve double testing as well. 
What is the EB idea in general?

easybuild/easyconfigs/i/Instant/Instant-1.5.0-goolf-1.5.14-Python-2.7.8.eb
Bart-CER: Added the eb file.

Bart-CER:
Added needed config files. 
Python 2.7.8 is part of PR #1365

PR #1364, opened 2015-02-03
{cae} (goolf-1.5.14-Python-2.7.8.eb) UFL 1.5.0 (REVIEW)

Bart-CER:
- part of my work towards upgrading dolfin
- same problem as PR 1362 when it comes to the download link.


Bart-CER:
should not be here, but develop.

PR #1349, opened 2015-01-29
{devel} [goolf-1.5.14-Python-2.7.6] PyQt4 SIP (WIP)

Bart-Ver:
This PR needs easyblock PR ~~https://github.com/hpcugent/easybuild-easyblocks/pull/540~~

SIP works and installs well.

PyQt doesn't make it through the sanity check, as it wants to import the module pyqt, but the module name is PyQt4. Using the 'options' doesn't seem to work.
Debug output: https://gist.github.com/Bart-CER/4b59e43db11f33e46b8d.

PR #1365, opened 2015-02-03
{cae} (goolf-1.5.14-Python-2.7.8.eb) UFL 1.5.0 (REVIEW)

Bart-CER:
- part of my work towards upgrading dolfin
- same problem as PR 1362 when it comes to the download link.


wpoely86:
ok to test

boegel:
@Bart-CER: looks good, but easyconfig for `Python/2.7.8-goolf-1.5.14` is missing, is that available via some other PR? If so, please let us know, and mention it in the description too.

Bart-CER:
I added the python eb file, however, with a lot of dependencies in the file. Not sure what to do with those, cf. PR #1349.

easybuild/easyconfigs/p/Python/Python-2.7.8-goolf-1.5.14.eb
```diff
@@ -0,0 +1,160 @@
...
+    ('ncurses', '5.9'),
+    ('LLVM', '3.2'),             # for Numba
+    ('SQLite', '3.8.4.3'),       # for IPython and general use
+    ('libxslt', '1.1.28'),       # for lxml.  brings in libxml2 and zlib dependencies.
+    ('CUDA', '6.0.37','', True), # for pycuda, duh.
+#   ('OpenSSL', '1.0.1f'),       # OS dependency should be preferred for security reasons
+
+    ('libjpeg-turbo','1.3.0'),   # For matplotlib and Pillow
+    ('libpng','1.6.10'),         # For matplotlib and Pillow
+    ('freetype', '2.5.3'),       # For matplotlib and Pillow
```
boegel: we try to stay away from excluding extra dependencies that are only required for Python packages...

see also the effort in https://github.com/hpcugent/easybuild-easyconfigs/pull/1387 to bundle multiple extensions that required extra dependencies together in a single `-extra` module

easybuild/easyconfigs/p/Python/Python-2.7.8-goolf-1.5.14.eb
```diff
@@ -0,0 +1,160 @@
...
+    }),
+    ('matplotlib', '1.3.1', {
+        'source_urls': ['https://pypi.python.org/packages/source/m/matplotlib/'],
+    }),
+    ('skimage', '0.10.1', {
+        'source_tmpl': 'scikit-image-%(version)s.tar.gz',
+        'source_urls': ['http://pypi.python.org/packages/source/s/scikit-image/'],
+    }),
+
+    ('lxml', '3.3.5', {
```
boegel: adding extra extensions is fine, as long as they don't introduce extra dependencies, but we should try and keep the different Python easyconfigs (for the same version of Python) in sync...

easybuild/easyconfigs/p/Python/Python-2.7.8-goolf-1.5.14.eb
```diff
@@ -0,0 +1,160 @@
...
+        'source_urls': ['https://pypi.python.org/packages/source/l/lxml/'],
+    }),
+    ('pycuda', '2014.1', {
```
boegel: this requires CUDA as a dep?

easybuild/easyconfigs/p/Python/Python-2.7.8-goolf-1.5.14.eb
```diff
@@ -0,0 +1,160 @@
...
+        'source_urls': ['https://pypi.python.org/packages/source/p/pycuda/'],
+    }),
+    #('pyopencl', '2014.1', {
+    #    'source_urls': ['https://pypi.python.org/packages/source/p/puopencl/'],
+    #}),
+    ('pysam', '0.8.0', {
```
boegel: requires SAMtools?

easybuild/easyconfigs/u/UFL/UFL-1.5.0-goolf-1.5.14-Python-2.7.8.eb
```diff
@@ -0,0 +1,32 @@
...
+ mathematical notation."""
+
+toolchain = {'name': 'goolf', 'version': '1.5.14'}
+
+majorversion = "%s.x" % ".".join(version.split('.')[:-1])
+source_urls = ['https://bitbucket.org/fenics-project/ufl/downloads/']
+sources = [SOURCELOWER_TAR_GZ]
+
+python = "Python"
+pythonversion = '2.7.8'
```
boegel: why not use the latest and greatest Python 2.7.9?

Bart-CER:
I made the python 2.7.8 config file 'equal' to the Python-2.7.8-intel-2014b.eb version.

Bart-CER:
Any more comments on this one?

boegel:
merged via #1467, thanks @Bart-CER!

PR #1367, opened 2015-02-03
{chem} [intel/2015a] Added git version of PSI4 (WIP)

wpoely86:
Depends on hpcugent/easybuild-easyblocks#545

Plugin stuff doens't work because they only halve-assed switched to CMake.


wpoely86:
Superseeded by #1431

PR #1376, opened 2015-02-04
R 3.0.2 modules sources update

sylmarien:
Some of the modules are not available anymore in their current version of the easyconfig file and/or not at the same URL. I just changed some minor version versions to fit with the ones currently available and/or the source URL from which the module is available.

These changes may be applicable to the R-3.0.2-ictce-5.5.0.eb easyconfig file too, but since I hadn't the ictce-5.5.0 toolchain to test it, I prefered to not take a wild guess that it would also work.

### DIFF
--- a/easybuild/easyconfigs/r/R/R-2.15.2-goalf-1.1.0-no-OFED.eb
+++ b/easybuild/easyconfigs/r/R/R-2.15.2-goalf-1.1.0-no-OFED.eb
@@ -134,7 +134,7 @@ exts_list = [
     ('R.methodsS3', '1.4.2', ext_options),
     ('R.oo', '1.13.0', ext_options),
     ('R.matlab', '1.7.0', ext_options),
-    ('Rniftilib', '0.0-32', ext_options),
+    ('Rniftilib', '0.0-32', rforge_options),
     ('gbm', '2.1', ext_options),
 
 ]
--- a/easybuild/easyconfigs/r/R/R-2.15.2-goolf-1.4.10.eb
+++ b/easybuild/easyconfigs/r/R/R-2.15.2-goolf-1.4.10.eb
@@ -134,7 +134,7 @@ exts_list = [
     ('R.methodsS3', '1.4.2', ext_options),
     ('R.oo', '1.13.0', ext_options),
     ('R.matlab', '1.7.0', ext_options),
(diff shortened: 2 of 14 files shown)


boegel:
@sylmarien: we generally stay away from bumping versions in existing easyconfigs, since that causes it to no longer match what people already have installed at their site...

The issue of disappearing sources is annoying, but needs to be dealt with differently, since this is not a permanent solution: the versions will likely be updated again in the future.

I have two options for you: either look into a recently merged easyconfig for R v3.1.2 (with no guarantees that you won't run into similar issues), and/or trying to obtain the source tarballs for the package versions as included in the original easyconfig via some other way.

In the long term, we want to look into setting up our own archive of sources that EasyBuild can consider as a backup, but that effort is still ongoing.

boegel:
Jenkins: ok to test

sylmarien:
@boegel Yes, at first I did get the tarball for the package versions in our old sources repository, but since it may not be able in the future to get them anymore, and since it was only hotfix changes I thought it was better to make it go upstream.

But I do get your point, so this pull request can be closed and not merged, but notice that there is also Rniftilib for which the source URL changed also.

boegel:
@sylmarien: can you please modify this PR to only change the Rniftilib source URL? Please do so in all affected R easyconfigs, regardless of whether you can test them or not (we can take care of that).

sylmarien:
@boegel Here are the changes, but it still won't download Rniftilib since the version changed too.
I have also reverted the version changes.

boegel:
@sylmarien `rforge_options` is not defined in all R easyconfigs yet, so Jenkins won't be happy with this because of the undefined variable

I'm also not sure whether this is worth the trouble... Fixing the source URL doesn't exactly help when the source tarball isn't available there anyway?

sylmarien:
@boegel Yes that's exactly what I meant. Sorry for not having checked that the variable actually existed in all of them though.

boegel:
@sylmarien: OK, then this PR can be closed? I prefer not to close any PRs issued by others, so please close it if you agree that it should be closed.

Hopefully we can look into a solution for this disappearing source tarballs issue soon... @fgeorgatos: I'm looking at you ;)

sylmarien:
It's ok, I close this PR.

Issue #1374, opened 2015-02-04
Distribution-related dependencies in some easyconfigs


sylmarien:
Hello,

I just wanted to point out that some easyconfig files contain system dependencies which names differ depending on the distribution used.

As an example in XCrySDen-1.5.53-goolf-1.4.10.eb, there is a dependency on the libXmu-devel package which is the name of the package on the RHEL-like distributions. Its Debian equivalent for example is libXmu-dev.

I also encountered this problem with Mesa (which is a dependency for XCrySDen) for a bit more packages. In order to help any other people that may encounter the same problem, here is a quick list of the Debian packages equivalents for this specific case:  
RHEL-like      ->    Debian equivalent  
libX11-devel  ->    libX11-dev  
xorg-x11-proto-devel -> x11proto-core-dev  
libXdamage-devel -> libXdamage-dev  
libXext-devel -> libXext-dev  
libXfixes-devel -> libXfixes-dev

Now, I don't know what could be done about that, I'm not sure that having "distribution-specific" easyconfig files would do any good, but it currently is "distribution-specific" already. A solution could be to have different options depending on the distribution (and then the checking at the beginning would use the correct one), but it could make the easyconfig files to just grow bigger and more complicated.

boegel:
As mentioned in https://github.com/hpcugent/easybuild-easyconfigs/pull/1377#issuecomment-72909848, we have a way of dealing with this, i.e. using a tuple of options in `osdependencies`, e.g.:

```python
osdependencies = [('libX11-devel', 'libX11-dev')]
```

If one of both packages can be found, EasyBuild will be happy. If need be, you can skip this check using `--ignore-osdeps` if you know the package is actually available, but under a different name.

Even better: if you can, replace these kind of OS deps with proper dependencies installed via EasyBuild.

If this is an adequate answer to your questions, please close this issue.

sylmarien:
@boegel Thank you for the answer, it indeed answers my question.

PR #1377, opened 2015-02-04
Resolving some distribution-related dependency problems

sylmarien:
This is the pull request for the issue #1374 about some distribution-related dependencies.

### DIFF
--- a/easybuild/easyconfigs/m/Mesa/Mesa-7.11.2-goalf-1.1.0-no-OFED-Python-2.7.3.eb
+++ b/easybuild/easyconfigs/m/Mesa/Mesa-7.11.2-goalf-1.1.0-no-OFED-Python-2.7.3.eb
@@ -30,22 +30,13 @@ dependencies = [
         ('libdrm', '2.4.27')
         ]
 
-if OS_NAME in ['debian', 'ubuntu']:
-  osdependencies = [
-              'libX11-dev', # Xlibs.h
-              'x11proto-core-dev', # X.h, glproto, xproto
-              'libXdamage-dev',
-              'libXext-dev',
-              'libXfixes-dev',
-           ]
-else:
-  osdependencies = [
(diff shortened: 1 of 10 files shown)


pescobar:
@sylmarien I think the os dependencies no longer require to specify the distribution (the IF sentences you include) since this PR was merged https://github.com/hpcugent/easybuild-framework/pull/1111

Now you just need to specify a list of tuples with the depencies and the framework will take care of it depending on the detected package management system. You can check an example here 
https://github.com/pescobar/easybuild-easyconfigs/blob/iomkl-2015.01/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.8.4-iccifort-2015.1.133-GCC-4.9.2.eb#L24

@boegel am I missing something which should also be remarked about this topic?

boegel:
@pescobar is right, there's no reason to resort to using `if`/`else` in easyconfig files

although it does work, anything else than variable assignments should be avoided in easyconfig files

so, replace:

```python
if OS_NAME in ['debian', 'ubuntu']:
  osdependencies = ['libXmu-dev']
else:
  # OK for OS_NAME == redhat, fedora, RHEL, SL, centos
  osdependencies = ['libXmu-devel']
```

with

```python
osdependencies = [('libXmu-dev', 'libXmu-devel')]
```

In the latter version, EasyBuild will check both package names, and if either of both is available, it will be happy.

Note that this OS deps requirement can also be skipped using the `--ignore-osdeps` command line option (at your own risk of running into a failed build).

boegel:
Jenkins: ok to test

sylmarien:
Thanks for the feedbacks, I didn't know about this syntax.

I modified my changes to what I understand is what you want. Let me know if I missed something.

migueldiascosta:
@sylmarien, this PR is being closed for the following reason(s): no activity for > 6 months, obsoleted by more recent PRs, uses an archived toolchain.

PR #1381, opened 2015-02-05
added Qt 5 for intel

JensTimmerman:


boegel:
i) target `deveop`
ii) please check https://github.com/hpcugent/easybuild/wiki/Policy-for-easyconfig-pull-requests

easybuild/easyconfigs/q/Qt/Qt-5.2.1-intel-2015a.eb
```diff
@@ -0,0 +1,21 @@
+easyblock = 'ConfigureMake'
```
boegel: why? we have an easyblock for `Qt` in place...

PR #1375, opened 2015-02-04
{perf}[foss/2015a] Scalasca + deps (REVIEW)

gjbex:
* Scalasca
* Cube
* Qt
* GLib
* gettext
* libffi
* OTF2
* OPARI2
* PDT
* Score-P
* PAPI

PR #1383, opened 2015-02-05
added Qt 5 for intel

JensTimmerman:
everything changed in qt 5, this now just works as configure make install.

### DIFF
--- a/easybuild/easyconfigs/q/Qt/Qt-5.2.1-intel-2015a.eb
+++ b/easybuild/easyconfigs/q/Qt/Qt-5.2.1-intel-2015a.eb
@@ -1,0 +1,21 @@
+easyblock = 'ConfigureMake'
+
+name = 'Qt'
+version = '5.2.1'
+
+homepage = 'http://qt-project.org/'
+description = "Qt is a comprehensive cross-platform C++ application framework."
+
+toolchain = {'name': 'intel', 'version': '2015a'}
+
+source_urls = [
+    'http://download.qt.io/archive/qt/%(version_major_minor)s/%(version)s/single/'
+]
+sources = ['%(namelower)s-everywhere-opensource-src-%(version)s.tar.gz']
+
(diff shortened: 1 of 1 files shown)


boegel:
again, please see https://github.com/hpcugent/easybuild/wiki/Policy-for-easyconfig-pull-requests

easybuild/easyconfigs/q/Qt/Qt-5.2.1-intel-2015a.eb
```diff
@@ -0,0 +1,21 @@
...
+
+source_urls = [
+    'http://download.qt.io/archive/qt/%(version_major_minor)s/%(version)s/single/'
+]
+sources = ['%(namelower)s-everywhere-opensource-src-%(version)s.tar.gz']
+
+configopts = "-static -qt-xcb -release -opensource -confirm-license -c++11 -no-qml-debug -no-gif -no-libpng -no-libjpeg"
+configopts += " -no-freetype -no-harfbuzz -no-openssl -no-xinput2 -no-xcb-xlib -no-widgets -no-nis -no-cups -no-dbus"
+configopts += " -no-eglfs -no-directfb -no-linuxfb -no-kms -no-opengl -nomake tools -nomake examples"
+configopts += " -platform linux-icc-64"
```
boegel: we have a Qt easyblock already, so you should update it to use these configure options in a block specific to Qt v5.x

for several of them, you probably need alternatives in case the required dependencies *are* provided (cfr. the Qt easyconfig in #1375)

JensTimmerman:
Could you add this policy to the contributing.md file? This would make it a lot easier to find when creating a new pull request.

boegel:
The policy is not official yet, but since people seem to like it, it will become official at some point.

The 'regulars' should already try to stick to it, though...

JensTimmerman:
@boegel any further remarks?

boegel:
Jenkins: test this please

boegel:
@JensTimmerman: remark w.r.t. Qt easyblock still stands?

PR #1408, opened 2015-02-16
{bio} [goolf-1.4.10] ms-20071014 (REVIEW)  

pescobar:


easybuild/easyconfigs/m/ms/ms-20071014-goolf-1.4.10.eb
```diff
@@ -0,0 +1,46 @@
...
+
+# as no version is provided in the tarball I verify checksum is fine
+checksums = [('md5', 'ef693dcc424fd5293b0bea4c68953700')]
+
+# delete provided binary
+postinstallcmds = ["rm -fr %(installdir)s/msdir/ms"]
+
+# here is where we do the compilation 
+# I got this line from the provided build script in %(installdir)s/msdir/clms
+postinstallcmds += ["cd %(installdir)s/msdir && $CC -O2 -o ms  ms.c  rand1.c streec.c -lm"]
```
boegel: why not make this `CmdCp` instead?

boegel:
@pescobar: please submit a test report too (but take a look at the remark first)

easybuild/easyconfigs/m/ms/ms-20071014-goolf-1.4.10.eb
pescobar: IMHO, this one is easier to read/understand than the CmdCp and anyway I was going to use files_to_copy = [ " * " ] so why not using TarBall easyblock? In any case, I think it's just a matter of taste, isn't it?

pescobar:
Test report by @pescobar
**SUCCESS**
Linux centos 6.5, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/f05fa9ce8e8d1fcd7c37 for a full test report.

easybuild/easyconfigs/m/ms/ms-20071014-goolf-1.4.10.eb
boegel: you're duplicating lots of stuff here that's already done by `CmdCp`: cleanup, you can copy stuff to `bin`

you're right that `CmpCp` should be enhanced though, it's usage is now rather strange

fgeorgatos: IMHO, `$CC` in `postinstallcmds` is no good should be avoided at all costs:

in the future, we might wish to transfer ownership of files to a separate user, after install step, as means of isolation.
I am already doing something of the like, in a manual step, in order to stabilise the builds and make the immutable.
Juelich is also doing something of the like, whereby one team passes buildset ownership to another.

In short, please do not change things after the install step, if you have a choice ;-)

boegel: Good argumentation to go with `CmdCp`?

PR #1422, opened 2015-02-24
add SLES11 package to openssl devel os dependency

dominikborkowski:
A lot of existing easyconfigs used 'openssl-devel' and 'libssl-dev' under osdependencies to indicate the need to have development files for openssl. SLES 11 uses 'libopenssl-devel', so we add that as a third option.


boegel:
Jenkins: ok to test

boegel:
going in, thanks @dominikborkowski, and congrats with your first EasyBuild contribution!

PR #1423, opened 2015-02-24
{bio}[dummy] CCP4 (REVIEW)

a-mazur:
That is a big X-ray software bundle. I have also integrated template module file they provide (see comments). Sources download needs registration.


boegel:
replaced by #1424

PR #1425, opened 2015-02-24
added perl as a build dependency for parallel

JensTimmerman:


JensTimmerman:
ok, so the tests fail because there are no perl's for the toolchains I added, however, I tested this, but this does actually breaks the build, I'm running into this issue: 
`https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=724123`
for some reason this does not occur with RHEL's perl, but it does on CENTOS and debian, so I actually consider the old parallel .eb's broken...

JensTimmerman:
Just tested this out, parallel build for version `20130122`  works with Perl-5.16.3, but fails with  Perl-5.20.1, so changing that.

JensTimmerman:
ok, I changed to an older perl where we had one, not sure what to do for old toolchains where we have no perl yet, do you want me to put time in adding perl to an old toolchain we're never going to use?

Also, what to do for the GCC one? Add a perl for it aswell? or remove it?

boegel:
add the missing easyconfigs for Perl, it's `-bare` anyway?

easybuild/easyconfigs/p/parallel/parallel-20141122-GCC-4.9.2.eb
```diff
@@ -11,6 +11,11 @@ source_urls = ['http://ftp.gnu.org/gnu/parallel']
 
 toolchain = {'name': 'GCC', 'version': '4.9.2'}
 
+# warning, parallel is a perl script and should depend on perl
```
boegel: drop the rather silly warning, add a Perl easyconfig for GCC toolchain

JensTimmerman:
I can look into adding the perls if I find some time later this year

boegel:
@JensTimmerman: adding those easyconfig is pretty trivial, no?

just add them, and we can fire a test report to see if things work out

PR #1427, opened 2015-02-24
{bio} [goolf-1.4.10] Trinity-2.0.4 (REVIEW)  

pescobar:
I have seen there is a custom easyblock for trinity but after checking the build procedure the "MakeCp" easyblock seems an easier approach IMHO  . maybe I am missing something?


Bart-CER:
I have installed 2.0.6 based on this config file, and the installation seemed to have run smoothly. I will pr the 2.0.6 config after the user tested the installation.

boegel:
Jenkins: test this please

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/485710d541cb93eaae29 for a full test report.

boegel:
I'd love to hear @JensTimmerman's thoughts on this, I know he spent some time with the Trinity easyblock...

JensTimmerman:
Does this correctly build jellyfish? This should end up in the sanity check, preferably also for the current easyconfigs.

pescobar:
@JensTimmerman is jellyfish included with trinity? in their docs they point to a different website to download jellyfish.....

https://trinityrnaseq.github.io/advanced_trinity_guide.html

boegel:
@pescobar: please enhance the sanity check to include jellyfish path?

JensTimmerman:
yes, they ship a prebuilt jellyfish, but because we know how to build it for our architecture, the easybclock was there to select the optimized one if it was a dependency instead of the prebuilt one.

boegel:
@JensTimmerman: according to @pescobar, `jellyfish` is being properly (re)built with this easyconfig, despite not using the easyblock we have for Trinity... 

Maybe `ConfigureMake` suffices for recent versions of Trinity? Should we enhance the easyblock to make it aware of that?

I'd still prefer dropping the `easyblock = 'ConfigureMake'` here, since we have a Trinity easyblock in place (which does an extensive sanity check, etc.).

JensTimmerman:
Can we confirm this for the intel build aswell?
But if everything works with a plain ConfigureMake now, kudo's tot he Trinity team for cleaning up their procedure. 
Not sure how to proceed with the ec vs block here. Might be worth it to update the block yes.

ehiggs:
>Maybe ConfigureMake suffices for recent versions of Trinity? Should we enhance the easyblock to make it aware of that?

@boegel: how would you want to do this? In the `__init__` function detect the version and if it's new style trinity set `self` to `ConfigureMake(*args, **kwargs)`. It seems awfully hacky. 

Another approach is to move the existing Trinity easyblock to become TrinityOld and make a TrinityNew that inherits ConfigureMake. Then the original Trinity easyblock can make the appropriate one at `__init__` and delegate all the functions to the impl.

My personal feeling on this is that the Trinity folks have taken their hacky build system and made it a more commodity package - so we can take advantage of this and just use `ConfigureMake`  -as @pescobar has done.

boegel:
There's no need to hack around at all, you can implement this cleanly in the Trinity easyblock by falling back to `ConfigureMake`.


```python
from distutils.version import LooseVersion
...
class EB_Trinity(EasyBlock):

    def configure_step(self):
         if LooseVersion(self.version) >= LooseVersion('2.0'):
             # new versions: use standard configure
             ConfigureMake.configure_step(self)
         else:
             # old versions
             ...
```

ehiggs:
As `EB_Trinity` doesn't inherit from `ConfigureMake` this risks serious issues if any method of `ConfigureMake` decides to refer to anything that is in `self.*` but not in `EasyBlock`. Also, I think this means that the version check will need to be done on each function call, forcing updates to rely on [shotgun surgery](https://en.wikipedia.org/wiki/Shotgun_surgery).

boegel:
@ehiggs: it's ok to also (or only) derive from `ConfigureMake`. And you'd only have the version check on configure/build/install, which is not too bad?

I can look into it if you're not up for it.

JensTimmerman:
Python has multiple inheratance, so yes, change
`class EB_Trinity(EasyBlock):` to `class EB_Trinity(EasyBlock, ConfigureMake):`

boegel:
@JensTimmerman: `ConfigureMake` is a subclass of `EasyBlock`, so there's no need to use multiple inheritance here, you just need to call out to `EasyBlock.foo(self, ...)` rather than using `super(EB_Trinity, self).foo()` in the right places

JensTimmerman:
indeed, it is a subclass at this point, but you never know what's going to happen in the future, so putting it there doesn't hurt.

boegel:
easyblock update tackled by @hajgato in https://github.com/hpcugent/easybuild-easyblocks/pull/802

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/37e25c091b21d20ef285 for a full test report.

boegel:
Going in, thanks @pescobar!

PR #1432, opened 2015-02-27
speed up unit tests by switching to fetch_parameters_from_easyconfig

boegel:


boegel:
Jenkins: test this please

test/easyconfigs/suite.py
```diff
@@ -45,6 +45,9 @@
 fancylogger.disableDefaultHandlers()
 fancylogger.setLogLevelError()
 
+# make sure no deprecated behaviour is triggered
```
wpoely86: wait what?

boegel: yes, the tests should never trigger deprecated (or no longer supported) behaviour...

wpoely86: a yeah, I was confused. This means sure it will always fail.

boegel: well, only if deprecated behaviour is triggered :)

boegel:
going in, thanks for the review @wpoely86!

PR #1436, opened 2015-03-03
Orca 3.0.3

scintimandrion:
New easyconfig for ORCA 3.0.3.


scintimandrion:
Yes.

boegel:
@scintimandrion: you're renaming/adjusting an existing easyconfig file, please just add a *new* easyconfig file instead

boegel:
Jenkins: ok to test

scintimandrion:
Ah, I had split ORCA-3.0.3 off from ORCA-3.0.2 rather than starting it from develop. I'll see if I can rectify that.

scintimandrion:
In fact, I'll just close this PR and start over.

PR #1451, opened 2015-03-06
use new 'Bundle' easyblock for HPCBIOS bundles, fix moduleclass

boegel:
depends on https://github.com/hpcugent/easybuild-framework/pull/1209


easybuild/easyconfigs/a/Autotools/Autotools-20150119-GCC-4.9.2.eb
```diff
@@ -14,4 +14,4 @@ dependencies = [
     ('libtool', '2.4.5'), # 20150119
 ]
 
-moduleclass = 'devel'
+moduleclass = 'bundle'
```
fgeorgatos: hm, is this really a bundle? just asking; because in that case also bodeps in possibly in the same group!

fgeorgatos:
+1 , overall good

wpoely86:
looks fine

boegel:
Thanks for the review @wpoely86, @fgeorgatos!

PR #1487, opened 2015-03-17
clean up OpenFOAM patch, don't patch symlink lnInclude/ptscotchDecomp.C

boegel:


boegel:
nothing much to review, patch was (sort of) broken

PR #1503, opened 2015-03-24
Create BuildEnv-1.0-intel-2015a-default.eb

ocaisa:
Add an easyconfig for the BuildEnv easyblock hpcugent/easybuild-easyblocks#583


easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
```diff
@@ -0,0 +1,13 @@
+easyblock = "BuildEnv"
+
+name = 'BuildEnv'
+version = '1.0'
+versionsuffix = '-default'
+
+
```
boegel: remove double newline

easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
```diff
@@ -0,0 +1,13 @@
+easyblock = "BuildEnv"
+
+name = 'BuildEnv'
+version = '1.0'
+versionsuffix = '-default'
```
boegel: what's the purpose of this versionsuffix? doesn't really give more information, I think?

easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
```diff
@@ -0,0 +1,13 @@
...
+
+
+homepage = 'None'
+description = """This module sets a set of environment variables for compilers, linkers, maths libraries, etc., that you can use to easily transition between toolchains when building your software."""
```
boegel: break up long description, 120 chars per line max, start continuation line with a single space

boegel:
Jenkins: ok to test

easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
ocaisa: The idea was that this file is for the default toolchain options, but you could set different toolchain options which would lead to different suffixes.

boegel: sure, but wouldn't just using an empty suffix here be sufficient?

easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
```diff
@@ -0,0 +1,14 @@
+easyblock = "BuildEnv"
+
+name = 'BuildEnv'
+version = '1.0'
```
boegel: will there ever be another version? (I'm not saying I have a better option for this, right now)

ocaisa: Originally I wanted to use ```default``` as the version but lmod doesn't like that, it only wants to see things that begin with numbers so I was forced to add a version number

easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
```diff
@@ -0,0 +1,13 @@
...
+
```
ocaisa: Ok, agreed.

easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
ocaisa: Ok agreed

boegel: What kind of problems did you run into exactly?

Lmod does support 'meta' modules (which are versionless)...

easybuild/easyconfigs/b/BuildEnv/BuildEnv-1.0-intel-2015a-default.eb
```diff
@@ -0,0 +1,13 @@
+easyblock = "BuildEnv"
```
boegel: same as 'software' name, so no need to include this explicitly

but maybe `name` should be something like `'eb-build-env'`?

ocaisa: I'd prefer to go the other way and rename the easyblock then, including eb in the name introduces a layer of obscurity for those who don't know what eb stands for (and they're the ones this is intended for).

ocaisa:
The error returned when ```default``` is used as the version number is
```
== 2015-03-25 12:28:50,730 main.Lmod INFO $MODULEPATH set based on list of module paths (via 'module use'): /tmp/eb-2O6SFE/tmp3GDhge/all:/usr/local/software/juropatest/Stage4/modules/all/Core:/usr/local/software/juropatest/Stage4/modules/all:/usr/local/software/juropatest/Stage4/modules/all/Toolchain/GCC/4.9.2:/usr/local/software/juropatest/Devel:/usr/local/software/juropatest/Stage4/modules/all/Toolchain/gpsolf/2015.01:/usr/local/software/juropatest/Stage4/modules/all/Toolchain/gpsmpi/2015.01:/usr/local/software/juropatest/modules/Linux:/usr/local/software/juropatest/modules/Core
== 2015-03-25 12:28:52,988 main.easyblock WARNING build failed (first 300 chars): Changing environment as dictated by module failed: name 'false' is not defined (stdout: os.environ['LMOD_DEFAULT_MODULEPATH'] = '/usr/local/software/juropatest/Stage4/modules/all/Core:/tmp/eb-2O6SFE/tmp3GDhge/all:/usr/local/software/juropatest/Stage4/modules/all:/usr/local/software/juropatest/Stage4
```
Turns out that the word ```default``` is the problem, simply using ```defaults``` allows it to build

boegel:
@rtmclay: it seems like Lmod has problems with a (meta)module being named something like `BuildEnv/default-intel-2015a` (am I getting that right @ocaisa?), while something like `BuildEnv/defaults-intel-2015a` is fine, is that expected?

boegel:
Jenkins: test this please

ocaisa:
I was thinking about this in the context of the subtoolchain searching and the versioning becomes a problem there because you will (potentially) want a BuildEnv in each and every toolchain. A more logical approach to versioning then would be to the use the toolchain name as a version, that will give you the uniqueness (at least in a hierarchical scheme)

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/fa7f934030b761a248be for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/6cd6f6be969d7766945c for a full test report.

boegel:
Jenkins: test this please

boegel:
good to now now is merged in https://github.com/hpcugent/easybuild-easyblocks/pull/583

boegel:
Thanks @ocaisa!

PR #1031, opened 2014-08-19
supermagic mpi sanity code

jordiblasco:

PR #1519, opened 2015-03-30
{tools}[GCC/4.9.2] numactl 2.0.10 (FIX)

jordiblasco:
The current version fails. The autogen.sh needs to be run before the configure.


fgeorgatos:
Hi, copying verbatim from #1031 (same story):
Load the autotools bundle as a builddep: `builddependencies = [('Autotools', 'XXX')]`

boegel:
This is a PR to `master`, this problem is already fixed in `develop` (see https://github.com/hpcugent/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/n/numactl/numactl-2.0.10-GCC-4.9.2.eb)

PR #1532, opened 2015-04-05
replace use of $root with %(installdir)s

boegel:
cfr. https://github.com/hpcugent/easybuild-framework/pull/1060 and https://github.com/hpcugent/easybuild-easyblocks/pull/590


pforai:
+1 good idea!

boegel:
going in, thanks @pforai for the review

PR #1535, opened 2015-04-07
[WIP] IPython 3.0 and deps

ehiggs:
Some of our users would like to use IPython notebook so here is a WIP of the easybuild files.


ehiggs:
#*@!% defaults submit PRs to master.

PR #1543, opened 2015-04-08
Toolchain goolf 1.5.16

jas02:


boegel:
Jenkins: ok to test

easybuild/easyconfigs/g/goolf/goolf-1.5.16.eb
```diff
@@ -0,0 +1,37 @@
+easyblock = "Toolchain"
+
+name = 'goolf'
+version = '1.5.16'
```
boegel: compared to the existing `goolf/1.5.14`, GCC is bumped from 4.8.2 to 4.8.3 and OpenBLAS is bumped from 0.2.8 to 0.2.13

Hence, I would bump the toolchain version here a bit more to better reflect the different in versions, e.g. to `1.5.20`?

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 6 out of 6 (6 easyconfigs in this PR)
See https://gist.github.com/5527b9db1fd357f9a986 for a full test report.

easybuild/easyconfigs/g/goolf/goolf-1.5.16.eb
jas02: Is that necessary? I would have to recompile all components and we already have GDAL software depend on this toolchain (in production), so recompile GDAL as well and ask customer to use different version. 

Let me know, I have prepared GDAL-1.9.2-goolf-1.5.16.eb to push.

boegel: Strictly speaking, no, it's not necessary.

And since changing the version now would cause problems on your end, let's not go there.

boegel:
Test report works out, Jenkins is happy, so merging this. Thanks @jas02!

easybuild/easyconfigs/g/goolf/goolf-1.5.16.eb
jas02: @boegel Thanks, it save me some time. Next time I will ask for best naming scheme in advance.

fgeorgatos: plus, leaving it as-is keeps it aligned with `gompi/1.5.16` which is a good thing, since they are tested together :+1:

PR #1568, opened 2015-04-26
updated bioconductor download urls

pescobar:
added http://www.bioconductor.org/packages/release/bioc/src/contrib/  as download url for bioconductor packages


fgeorgatos:
Please, consider this high priority PR, if you see it fixes broken stuff.
Isn't it worthy to delay the release by 1 day if it increases the quality for many/all?

fgeorgatos:
btw. before the debate ensues :), the original issue here is that bioconductor* should be weeded out of `R`! Keeping it separate invites no debates of if you should be deprived of R because of bio* malpractice.

boegel:
Thanks @pesp

PR #1419, opened 2015-02-23
{chem} [intel/2016b] Libint 2.1.0 (REVIEW)

wpoely86:
edit: requires ~~https://github.com/hpcugent/easybuild-easyblocks/pull/985~~

PR #1584, opened 2015-05-05
{math} [intel-2015a] graph-tools (WIP) 

hajgato:


hajgato:
Would need PR #1419 (GMP-6.0.0a) and PR #1585 (CGAL)
It seems that intel compiler have some C++11 issues (or the g++ have extra extension), but the Boost part cannot be compiled. 
The error message is something:
```C++
error: more than one instance of overloaded function "boost::end" matches the argument list:
            function template "boost::range_iterator<const T, void>::type boost::range_adl_barrier::end(const T &)"
            function template "auto std::end(const _Container &)->decltype((<expression>))"
            argument types are: (const std::vector<size_t, std::allocator<size_t>>)
              const_iterator ci2 = boost::end(const_range);
```
and similars, always with `::boost::begin` and `::boost::end`. I do not know whether is a boost issue or graph-tool issue, but graph tools does not support intel compiler. Anybody suggestions?

PR #1604, opened 2015-05-14
Easyconfig for CNVnator

pm67nz:


easybuild/easyconfigs/c/CNVnator/CNVnator-0.3-intel-2015a.eb
```diff
@@ -0,0 +1,32 @@
...
+sources = ['%(name)s_v%(version)s.zip']
+source_urls = ['http://sv.gersteinlab.org/cnvnator/']
+
+dependencies = [
+    ('ROOT', '5.34.30'),
+]
+
+start_dir = 'src'
+
+prebuildopts = "sed -i -e 's|(ROOTSYS)/[a-z]*|&/root|g' Makefile && (cd samtools && make lib) &&"
```
boegel: what is the sed magic doing here exactly?

is using SAMtools as a proper dependency an option, as opposed to building the SAMtools that is shipped with this?

boegel:
Jenkinks: ok to test

easybuild/easyconfigs/c/CNVnator/CNVnator-0.3-intel-2015a.eb
pm67nz: The easybuilt ROOT sets $ROOTSYS to point to itself (so it's the same as $EBROOTROOT), and CNVnator uses $ROOTSYS to find it.  The problem is that the ROOT easyblock installs its files into $ROOTSYS/lib/root, $ROOTSYS/include/root etc. while the CNVnator makefile looks in $ROOTSYS/lib, $ROOTSYS/include etc.  I don't know which of those is the correct convention for ROOT.  The sed command just adds /root on the end of those paths and could be done as a patch instead.

Yes using SAMtools as a dependency is an option, but it is only a static library that is being used (so I guess it would be a build dependency), and this way I didn't have to worry about matching the version of SAMtools or making another change to the CNVnator makefile.

migueldiascosta:
@pm67nz, this PR is being closed for the following reason(s): no activity for > 6 months, obsoleted by more recent PRs.

PR #1654, opened 2015-06-02
include Tk as dependency in Python 2.7.9 easyconfigs

boegel:
@JensTimmerman: please review


easybuild/easyconfigs/t/Tk/Tk-8.6.4-GCC-4.9.2-no-X11.eb
```diff
@@ -0,0 +1,24 @@
...
+version = '8.6.4'
+versionsuffix = '-no-X11'
+
+homepage = 'http://www.tcl.tk/'
+description = """Tk is an open source, cross-platform widget toolchain that provides a library of basic elements for building
+ a graphical user interface (GUI) in many different programming languages."""
+
+toolchain = {'name': 'GCC', 'version': '4.9.2'}
+
+source_urls = ["http://prdownloads.sourceforge.net/tcl"]
```
JensTimmerman: Why not use the `SOURCEFORGE` constant here?

boegel: URL doesn't match the software name (Tk vs `tcl`)

JensTimmerman:
I'm not sure you can just change these existing easyconfigs to add a new tk as a dependency, this breaks easyconfigs who now depend on python and a different version of tk..

boegel:
Hmm, good point...

boegel:
@JensTimmerman: including Tk 8.6.3 as a dep works, i.e. no conflicts are introduced in existing easyconfigs

I consider this a bug fix, so feel it's OK if this has effect on easyconfigs that haven't been merged centrally yet.

Thoughts?

JensTimmerman:
Sure, it works for our repository, but what about people's private EC's?

JensTimmerman:
Well, other people might not agree, it could break the build of private easyconfigs, I'm not sure what our policy there was, weren't we going to try not to break things?

boegel:
@JensTimmerman: we maintain backward compatibility, except when bug fixes are required, then (minor) breaking changes can be considered

I consider the missing Tk dep a bug, since things don't work because of it

People should move on to Python 2.7.10 anyway ;)

boegel:
@JensTimmerman: this will be clearly mentioned in the release notes, and people can still either use an older easyconfig that does not include the Tk dep, or use `--filter-deps=Tk`, if they feel they need to

boegel:
Also, I just realised there's no need for this whole discussion anyway: `Tcl` v8.6.3 is already an indirect dependency to Python, via SQLite, and since the `Tcl` version also dictates the `Tk` version, there's no room in practice for any other `Tk` versions.

JensTimmerman:
OK, good to go

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/42b2371430a154b35c3e for a full test report.

boegel:
Going in, thanks for the critical review @JensTimmerman!

PR #1753, opened 2015-06-26
Add support for https

ocaisa:
Add support for https


boegel:
Jenkins: ok to test

boegel:
@ocaisa: this introduces a dependency on OpenSSL, right?

In that case, an `osdependencies` line should be included like in https://github.com/hpcugent/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/p/Python/Python-2.7.9-intel-2015a.eb#L28 (maybe combined with a commented out OpenSSL dependency)

boegel:
merged via #1784, thanks @ocaisa!

Issue #638, opened 2013-12-23
some builds work on SL5, not on SL6 (EB v1.10)


boegel:
Various builds are known to work on SL5, but not on SL6. This issue is intended to group those together, to keep track of what needs to be fixed.
- [ ] cairo: #639 (+ NCL/6.0.0 because of dep)
- [x] FDTD Solutions: #640
- [ ] HH-suite: #414 (+ HPCBIOS_BioInfo because of dep)
- [ ] CLHEP/VTK: #641 (+ DOLFIN, GATE because of dep)
- [ ] XML-LibXML: #643 (+ XML-Simple because of dep)
- [ ] Sphinx: #644
- [x] ~~PLINK #645 (fixed in #594)~~

Micket:
Long since outdated/resolved

PR #1773, opened 2015-07-07
Glob the lib so.

ehiggs:
[Requires #638 from easyblocks](https://github.com/hpcugent/easybuild-easyblocks/pull/638)

### DIFF
--- a/easybuild/easyconfigs/h/Hadoop/Hadoop-2.5.0-cdh5.3.1-native.eb
+++ b/easybuild/easyconfigs/h/Hadoop/Hadoop-2.5.0-cdh5.3.1-native.eb
@@ -26,7 +26,7 @@ dependencies = [(java, javaver)]
 
 build_native_libs = True
 
-extra_native_libs = [('snappy', 'lib/libsnappy.so')]
+extra_native_libs = [('snappy', 'lib/libsnappy.so*')]
 
 parallel = 1
 


easybuild/easyconfigs/h/Hadoop/Hadoop-2.5.0-cdh5.3.1-native.eb
```diff
@@ -26,7 +26,7 @@ dependencies = [(java, javaver)]
 
 build_native_libs = True
 
-extra_native_libs = [('snappy', 'lib/libsnappy.so')]
+extra_native_libs = [('snappy', 'lib/libsnappy.so*')]
```
boegel: wouldn't it be better to let the easyblock handle this?

ehiggs: I think it's best here in case any of the supported native libs require particular patterns. It's nicer to not need to update the block.

fgeorgatos: yeah, basically the easyblocks become a distraction in relation to easyconfigs; 
it would be ideal if there was a way to have both of them next to each other
(I even consider the crazy idea of passing easyblocks in docstrings inside easyconfigs... :-P )

ehiggs: I actually think the easyblocks mostly work along the wrong axis. There should be blocks for fetching, unpacking, building, installing, testing, etc; and the easyconfigs specify which blocks to assemble and use.

But that's not really the discussion I want to have in a PR where I'm just adding a single character to an existing eb file.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/d6e88ff0eeea48b5c3cc for a full test report.

boegel:
Tested on top of https://github.com/hpcugent/easybuild-easyblocks/pull/638, looks good, thanks @ehiggs!

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/b24a3c3155faafaac35c for a full test report.

Issue #1797, opened 2015-07-14
Rmpi 0.6-5 segfaults when built with goolf/1.7.20


boegel:
Rmpi 0.6-5 segfaults (both with and without applying the patch available for it), when built with `goolf/1.7.20` on SL6/sandybridge:

```
installing to /user/scratchdelcatty/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/R/3.2.0-goolf-1.7.20/lib64/R/library/Rmpi/libs
** R
** demo
** inst
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** testing if installed package can be loaded
sh: line 1: 31332 Segmentation fault      '/user/scratch/gent/vsc400/vsc40023/easybuild_REGTEST/SL6/sandybridge/software/R/3.2.0-goolf-1.7.20/lib64/R/bin/R' --no-save --slave 2>&1 < '/tmp/eb-obfAi2/RtmpHWrJMr/file79321af8f070'

*** caught segfault ***
address (nil), cause 'memory not mapped'
ERROR: loading failed
(at easybuild/tools/run.py:360 in parse_cmd_output)
```

@pescobar is not seeing this problem though...

Rmpi is excluded from the R-3.2.0-goolf-1.7.20.eb easyconfig (see #1573) because of this

boegel:
I'm now seeing Rmpi segfault with all `goolf` R easyconfigs...

This must be related to the patch for OpenMPI that was included in #1785.

@besserox: Is it possible that an error like this now suddenly pops up now that your OpenMPI patch is included, while it didn't before?

boegel:
@pescobar: can you check if you can reproduce this issue after rebuilding OpenMPI with the patch from #1785 included (on a test system, of course).

besserox:
One easy way to test if the OpenMPI patch is causing the bug is to set the environment variable OMPI_MCA_btl="^openib". This disables the Infiniband, and then the codepath related to this patch.

boegel:
From the discussion at https://github.com/open-mpi/ompi/issues/638, it seems like these changes are required too (especially the latter): https://github.com/open-mpi/ompi/commit/2f137ff151e68834fd18cc6342d7c081965240f5 and https://github.com/jsquyres/ompi-release/commit/8be602996671dc34337aa6dcd3754c739583a093

boegel:
[...]

PR #1785, opened 2015-07-11
Infiniband patch for OpenMPI < 1.8.6

boegel:
wrapper PR for #1678 by @ULHPC / @sylmarien 

this basically only fixing conflicts with current `develop` so the PR can be merged

PR #1798, opened 2015-07-15
revert #1785, patch for OpenMPI is incomplete and causes segfaults

boegel:
The OpenMPI patch included in #1785 is incomplete, and causes segfaults.

See discussion in #1797 and https://github.com/open-mpi/ompi/issues/638.

It's possible that the patch can be enhanced to make it work, but that will be for an EasyBuild version later than v2.2.0

cc @ULHPC, @sylmarien, @besserox


boegel:
alternative patch for this: #1799, #1800

PR #1799, opened 2015-07-15
add patch for OpenMPI 1.7.x and 1.8.x to fix memalign issue (WIP)

boegel:
fixed version of #1785 for OpenMPI 1.7.x and 1.8.x, cfr. #1797 

@besserox: please review?

### DIFF
--- a/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.7.3-GCC-4.8.2.eb
+++ b/easybuild/easyconfigs/o/OpenMPI/OpenMPI-1.7.3-GCC-4.8.2.eb
@@ -11,7 +11,10 @@ toolchain = {'name': 'GCC', 'version': '4.8.2'}
 sources = [SOURCELOWER_TAR_GZ]
 source_urls = ['http://www.open-mpi.org/software/ompi/v%(version_major_minor)s/downloads']
 
-patches = ['OpenMPI-%(version)s-vt_cupti_events.patch']
+patches = [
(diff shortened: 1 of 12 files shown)


boegel:
Hmm, hold on, the Rmpi segfault is still occurring on top of OpenMPI built with this patch... :(

fgeorgatos:
visual review OK; what a spectacular bug was that... it flied in the face of all rpath'ers :)

besserox:
Sorry, I was busy in the last 2 days.

I'm a bit lost now. What is the status of this? 
Is there still a segfault of Rmpi with this patch?

boegel:
@besserox sorry for not getting back... yes, Rmpi still segfaults with this (last time I checked)

besserox:
@boegel If Rmpi crashes with this, that kind of a problem :-( 
Unfortunately I don't have time to look into this.

Regarding this patch itself, it improves the performance code with OpenMPI, but it's still bad compared to other MPIs. So, this patch is no so urgent for me.

That's why I also prefer to focus on other toolchains like ictce/intel or gmvoolf for now.

boegel:
In that case, I suggest we just close this (and any related issues/PRs)...

There's little point in spending time on this if it only slightly helps with the performance of *old* OpenMPI versions.

besserox:
@boegel Yes, this can be closed for now. 

I will check (someday) what is the status with the recent version of OpenMPI.

PR #1826, opened 2015-08-05
{bio}[intel/2015a] Bowtie 1.1.2 (REVIEW)

ehiggs:
This also includes a patch to the Makefile so it doesn't force GCC when we're building with ICC.


easybuild/easyconfigs/b/Bowtie/Bowtie-1.1.2-intel-2015a.eb
```diff
@@ -0,0 +1,30 @@
...
+It aligns short DNA sequences (reads) to the human genome.
+"""
+
+toolchain = {'name': 'intel', 'version': '2015a'}
+toolchainopts = {'optarch': True, 'pic': True}
+
+sources = ['%(namelower)s-%(version)s-src.zip']
+source_urls = ['http://download.sourceforge.net/bowtie-bio/']
+
+patches = ['compiler-agnostic.patch', 'int64typedef.patch']
```
boegel: I would prefer using `buildopts` rather than including a patch to remove the compiler variables from the Makefile (a patch will break sooner or later):

```python
buildopts = 'CC="$CC" CXX="$CXX" CPP="$CXX"'
```

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/037c893b7ed8b6eaa679 for a full test report.

boegel:
Thanks @ehiggs!

PR #1833, opened 2015-08-07
Add easyconfig for PGI 15.7 using GCC 4.9.2 libraries.

bartoldeman:
See also hpcugent/easybuild-easyblocks#658 and hpcugent/easybuild-framework#1342

### DIFF
--- a/easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
+++ b/easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
@@ -1,13 +1,12 @@
 name = 'pgi'
 version = '15.7'
-year = '2015'
 
 homepage = 'http://www.pgroup.com/'
 description = "C, C++ and Fortran compilers from The Portland Group - PGI"
 
-toolchain = {'name': 'dummy', 'version': 'dummy'}
+toolchain = {'name': 'dummy', 'version': ''}
 
-sources = ['pgilinux-'+year+'-%(version_major)s%(version_minor)s-x86_64.tar.gz']
+sources = ['pgilinux-20%(version_major)s-%(version_major)s%(version_minor)s-x86_64.tar.gz']
 
 gnu = 'GNU'
 gnuver = '4.9.2-2.25'
--- a/easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
+++ b/easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
@@ -1,0 +1,22 @@
+name = 'pgi'
+version = '15.7'
+year = '2015'
+
+homepage = 'http://www.pgroup.com/'
+description = "C, C++ and Fortran compilers from The Portland Group - PGI"
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = ['pgilinux-'+year+'-%(version_major)s%(version_minor)s-x86_64.tar.gz']
+
+gnu = 'GNU'
+gnuver = '4.9.2-2.25'
+versionsuffix = '-%s-%s' % (gnu, gnuver)
+
+dependencies = [(gnu, gnuver)]
+
+# license file
+import os
+license_file = os.path.join(os.getenv('HOME'), "licenses", "pgi", "license.dat")
+
+moduleclass = 'compiler'
--- a/easybuild/easyconfigs/p/PGI/PGI-15.7-GNU-4.9.2-2.25.eb
+++ b/easybuild/easyconfigs/p/PGI/PGI-15.7-GNU-4.9.2-2.25.eb
@@ -1,4 +1,4 @@
-name = 'pgi'
+name = 'PGI'
 version = '15.7'
 
 homepage = 'http://www.pgroup.com/'
--- a/easybuild/easyconfigs/p/PGI/PGI-15.7-GNU-4.9.3-2.25.eb
+++ b/easybuild/easyconfigs/p/PGI/PGI-15.7-GNU-4.9.3-2.25.eb
@@ -1,4 +1,4 @@
-name = 'pgi'
+name = 'PGI'
 version = '15.7'
 
 homepage = 'http://www.pgroup.com/'
--- a/easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.3-2.25.eb
+++ b/easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.3-2.25.eb
@@ -1,0 +1,21 @@
+name = 'pgi'
+version = '15.7'
+
+homepage = 'http://www.pgroup.com/'
+description = "C, C++ and Fortran compilers from The Portland Group - PGI"
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['pgilinux-20%(version_major)s-%(version_major)s%(version_minor)s-x86_64.tar.gz']
+
+gnu = 'GNU'
+gnuver = '4.9.3-2.25'
+versionsuffix = '-%s-%s' % (gnu, gnuver)
+
+dependencies = [(gnu, gnuver)]
+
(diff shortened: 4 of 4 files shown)


easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
```diff
@@ -0,0 +1,22 @@
+name = 'pgi'
```
boegel: PGI?

easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
```diff
@@ -0,0 +1,22 @@
...
+version = '15.7'
+year = '2015'
```
boegel: only used in one place, so no reason to use a separate variable?

does it make sense to make the version `2015-15.7`?

easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
```diff
@@ -0,0 +1,22 @@
...
+
+homepage = 'http://www.pgroup.com/'
+description = "C, C++ and Fortran compilers from The Portland Group - PGI"
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = ['pgilinux-'+year+'-%(version_major)s%(version_minor)s-x86_64.tar.gz']
```
boegel: just use `%(version)s`?

easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
```diff
@@ -0,0 +1,22 @@
...
+
+gnu = 'GNU'
+gnuver = '4.9.2-2.25'
+versionsuffix = '-%s-%s' % (gnu, gnuver)
+
+dependencies = [(gnu, gnuver)]
+
+# license file
+import os
```
boegel: it's time we introduce an additional constant `HOME`, so we can get rid of these `import` statements...

boegel:
Jenkins: ok to test

bartoldeman:
Actually the year is related to the major version number, see:
http://www.pgroup.com/support/download_releases.php
so one can just use 20%(version_major)s-%(version_major)s%(version_minor)s
The tarball does not have a . at 15.7, so I had to use %(version_major)s%(version_minor)s

bartoldeman:
I also changed the toolchain version from 'dummy' to '' so that the GCC module is loaded and its libraries are then picked up by the makelocalrc script.

easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
boegel: @bartoldeman: we try to stick to the 'official' software name; I think `PGI` is a lot more common than `pgi`, so can we change this (along with the easyconfig filename)?

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 0 out of 0 (2 easyconfigs in this PR)
See https://gist.github.com/ba5792fd6d04266b0799 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/b9917b69ec14d3d91565 for a full test report.

easybuild/easyconfigs/p/pgi/pgi-15.7-GNU-4.9.2-2.25.eb
bartoldeman: My reasoning was as follows (see the comment thread for the easyblock)

The compiler itself ships with a sample modulefile named "pgi/15.7", so people are used to loading modulefiles for this compiler with the non-capitalized pgi (you can do a google search for "pgi module" and see what other sites use). 


On the other hand this also seems to go with an all-lowercase convention that many HPC sites but evidently not easybuild use.

boegel: You have full control over module names in EasyBuild, so using the official name `PGI` here doesn't mean you can't obtain a module named `pgi/...`. It does mean you'll get `PGI/...` by default though, that's true.

A quick look at http://www.pgroup.com/ makes it clear to me that `PGI` is the official name; `pgi` is not mentioned at all on the main page (I didn't look beyond that).

Personally, I'm not convinced that going with `pgi` here because most sites seem to use `pgi` as a module name is a good reason to diverge from our policy of using 'official' software names in easyconfig files, especially since `eb` supports using a custom module naming scheme to consistently make all of your modules lowercase if you want to.
But, I don't want to turn this into a bikeshedding discussion either...

@rjeschmi: thoughts?

rjeschmi: I think pgi themselves distribute modulefiles and I think they have the
lowercase name. (Away from my computer right now)

So in that sense the official name is what they use there.

boegel: The only reason people pick lowercase module names is to dance around the problem that the most common modules tool has a case-sensitive `module avail`, because it's broken. ;-)

And they're even shooting themselves in the foot there, since I would use `module avail PGI` as a first attempt to see if PGI compilers are available (since `PGI` is how it's spelled all over their website)...

I'd like to hear back from @bartoldeman: do you feel strongly about sticking to `pgi`? If not, I would prefer going with `PGI` instead.

rjeschmi: Is there a way to override the module name in the default naming scheme?

boegel: @rjeschmi: not 100% sure what you mean, but I think the answer is no (currently). I do see a use for it though, e.g.:

```
name = 'hanythingondemand'
...
module_name = 'hod'
```

boegel: @rjeschmi: support for overriding the `name` bit of the module name is implemented in https://github.com/hpcugent/easybuild-framework/pull/1389 (still WIP)

bartoldeman: I am happy with using PGI internally everywhere, but leaving the option open to use a "pgi" lowercase modulename, as the PGI compiler supplies itself.

I am in a bit of a crunch now trying to get our easybuild structure usable by users for testing (see my list email to come about lmod lua files) -- I will have more time next week if that is ok.

boegel: The option is always there to make EB name the module any way you want, the question is just whether having it produce a module named `PGI/...` by default is OK...

In my view, lowercase module names (for specific modules, or all modules) is a site-specific policy.

The default policy we maintain in EB is to use the 'official' software name (i.e., the one most prevalent on the homepage), but it's not limiting in any way since we support custom module naming schemes.

I'm not sure which Lmod/Lua issue you're referring to. EasyBuild or Lmod mailing list?

boegel: W.r.t. timing: no pressure there, we still have loads of time to get PGI support included for the next EB release (planned for end of Oct'15).

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/47c405c2bde5d19c5e82 for a full test report.

boegel:
Jenkins: test this please

boegel:
Retested on top of https://github.com/hpcugent/easybuild-easyblocks/issues/658, good to go.

Thanks for looking into this @bartoldeman!

PR #1835, opened 2015-08-07
put dummy template values in place to make comparison with dumped easyconfig more robust

boegel:
This is required once https://github.com/hpcugent/easybuild-framework/pull/1345 is merged to make the dump aspect of the easyconfig unit tests pass, unless more easyconfigs are tweaked to use templates like `%(name)s` and/or `%(namelower)s`.

For easyconfigs using the `CmdCp` easyblock, it's a bit more tricky since there additional templates (`%(source)s` and `%(target)s`) are supported by the `CmdCp` easyblock itself, which get in the way when trying to resolve templates (since they're not defined (yet) after just parsing the easyconfig).

That's not too big of an issue, it just means that the unit tests are a bit strict w.r.t. template values for these easyconfigs, and may require that templates like `%(name)s` is used (where applicable) in strings where `%(source)s` and/or `%(target)s` are used as well.

A couple of easyconfigs using `CmdCp` are tweaked to deal with this.

### DIFF
--- a/easybuild/easyconfigs/f/FastTree/FastTree-2.1.7-goolf-1.4.10.eb
+++ b/easybuild/easyconfigs/f/FastTree/FastTree-2.1.7-goolf-1.4.10.eb
@@ -16,7 +16,7 @@ sources = ['%(name)s-%(version)s.c']
 
 skipsteps = ['source']
 
-cmds_map = [('FastTree.*.c', '$CC -DOPENMP $CFLAGS $LIBS %(source)s -o FastTree')]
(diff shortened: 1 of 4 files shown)


boegel:
I'm merging this since it's pretty trivial, and currently blocking me from making progress.

@rjeschmi: let me know if you have any remarks on this

rjeschmi:
Looks fine

It does seem arbitrary to replace the last instance with (name)s and not the source file instance, but it is arbitrary I guess

PR #1761, opened 2015-07-02
Dump tests

Caylo:
requires https://github.com/hpcugent/easybuild-framework/pull/1311

PR #1840, opened 2015-08-11
Update texinfo-5.2-GCC-4.8.2.eb

ocaisa:
Added dependency on texlive. Completed the build process to make necessary installations in custom ```TEXMFHOME``` and have module set that variable. This should only really be used as a build dependency since it will overwrite any default/user value for ```TEXMFHOME```

### DIFF
--- a/easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
+++ b/easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
@@ -17,7 +17,7 @@ preinstallopts = "make TEXMF=%(installdir)s/texmf install-tex && "
 
 # This will overwrite a users TEXMFHOME so this module is best used as a build dependency
 modextravars = {'TEXMFHOME': '%(installdir)s/texmf'}
-modloadmsg    = """\\n\\nThis module will set/overwrite any existing value for the environment variable TEXMFHOME."""
+modloadmsg    = """\\n\\nThis module will set/overwrite the value for the environment variable TEXMFHOME."""
 modloadmsg    += """\\nIf you use a custom texmf directory (such as \~/texmf) then you should copy files found in the"""
 modloadmsg    += """\\nnew TEXMFHOME to your custom directory and reset the value of TEXMFHOME to point to that space:"""
 modloadmsg    += """\\n\\tcp -r \$TEXMFHOME /path/to/your/texmf/.."""
--- a/easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
+++ b/easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
@@ -20,7 +20,7 @@ modextravars = {'TEXMFHOME': '%(installdir)s/texmf'}
 modloadmsg    = """\n\nThis module will set/overwrite the value for the environment variable TEXMFHOME."""
 modloadmsg    += """\nIf you use a custom texmf directory (such as ~/texmf) then you should copy files found in the"""
 modloadmsg    += """\nnew TEXMFHOME to your custom directory and reset the value of TEXMFHOME to point to that space:"""
-modloadmsg    += """\n\tcp -r $TEXMFHOME /path/to/your/texmf/.."""
+modloadmsg    += """\n\tcp -r $TEXMFHOME/* /path/to/your/texmf"""
 modloadmsg    += """\n\texport TEXMFHOME=/path/to/your/texmf\n\n"""
 
 sanity_check_paths = {
--- a/easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
+++ b/easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
@@ -17,11 +17,11 @@ preinstallopts = "make TEXMF=%(installdir)s/texmf install-tex && "
 
 # This will overwrite a users TEXMFHOME so this module is best used as a build dependency
 modextravars = {'TEXMFHOME': '%(installdir)s/texmf'}
-modloadmsg    = """\\n\\nThis module will set/overwrite the value for the environment variable TEXMFHOME."""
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: ok to test

easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
```diff
@@ -16,6 +11,14 @@ sources = [SOURCELOWER_TAR_XZ]
 
 toolchain = {'name': 'GCC', 'version': '4.8.2'}
 
+osdependencies = ['texlive']
+
+preinstallopts = "mkdir -p %(installdir)s/texmf/{doc,tex}/{context,eplain,generic,latex,lualatex,plain,xelatex,xetex} && "
```
boegel: the `install-tex` make target doesn't take care of this? :(

easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
```diff
@@ -16,6 +11,14 @@ sources = [SOURCELOWER_TAR_XZ]
...
+preinstallopts += "make TEXMF=%(installdir)s/texmf install-tex && "
+
+# This will overwrite a users TEXMFHOME so this module should only really be used as a build dependency
+modextravars = {'TEXMFHOME': '%(installdir)s/texmf'}
```
boegel: so, maybe it shouldn't be set by the module then?

where texinfo is used as a build dep, you can redefine `$TEXMFHOME` during the build (via `(pre)buildopts`, for example)

easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
ocaisa: I don't know, I didn't try it. I just assumed it needed the proper structure.

easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
ocaisa: Hmm, maybe but that's a pretty big gotcha for someone who tries to use it, without setting that variable it simply doesn't work properly because the necessary texinfo.tex file won't be found.

ocaisa: I can add a modloadmsg as a warning for people who use it directly.

boegel: What kind of issues might people be running into if this module is loaded with setting `$TEXMFHOME`? Any idea?

easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
boegel: Let's only include it if it's needed.

easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
ocaisa: If they use custom style sheets in their own ```~/texmf``` then they won't be available to ```tex``` any more. The new warning will help them address it.

ocaisa:
Hmm, I tested the warning and it installs fine for me but causes problems in the testsuite...am I missing something?

boegel:
@ocaisa: the easyconfig unit tests were made stricter in #1761 to verify that the `EasyConfig.dump()` method isn't changing the values in the dumped easyconfig file

It seems like the `\\n` values in your `modloadmsg` are being replaced with just `\n`.

Why are you escaping '\n`, shouldn't be needed?

And if it is needed to obtain the result you want, we'll need to fix the issue in `dump` in framework.

ocaisa:
Don't really know why, that's what I saw somewhere else and the output looked good so I figured it was ok. I'll retest without any escaping.

ocaisa:
Removed all escaping

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/c2c36ce9c578aca4d97f for a full test report.

boegel:
@ocaisa: the existing easyconfig for `bc` includes `texinfo` as a (non-build) dependency, should that be updated to import it as a build dep instead?

boegel:
https://github.com/ocaisa/easybuild-easyconfigs/pull/3 fixes some (very minor) style issues

easybuild/easyconfigs/t/texinfo/texinfo-5.2-GCC-4.8.2.eb
```diff
@@ -16,6 +11,18 @@ sources = [SOURCELOWER_TAR_XZ]
...
+modloadmsg    = """\n\nThis module will set/overwrite the value for the environment variable TEXMFHOME."""
+modloadmsg    += """\nIf you use a custom texmf directory (such as ~/texmf) then you should copy files found in the"""
+modloadmsg    += """\nnew TEXMFHOME to your custom directory and reset the value of TEXMFHOME to point to that space:"""
+modloadmsg    += """\n\tcp -r $TEXMFHOME/* /path/to/your/texmf"""
+modloadmsg    += """\n\texport TEXMFHOME=/path/to/your/texmf\n\n"""
+
 sanity_check_paths = {
     'files': ['bin/info', 'bin/makeinfo', 'bin/pod2texi', 'bin/texi2pdf'],
     'dirs': [],
```
boegel: enhance the sanity check too, to check for (some of) the stuff that was missing but should now be there?

ocaisa:
Withing looking at the config output of ```bc``` I would say yes, it should only be a build dep. It is only required when building gnu documentation, you won't need it at runtime.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/2840e0921d57a04886cb for a full test report.

boegel:
Thanks @ocaisa! Going in...

PR #1853, opened 2015-08-14
stop using 'unresolved_deps' in easyconfigs unit tests

boegel:
required because of https://github.com/hpcugent/easybuild-framework/pull/1356


boegel:
This introduced a bug: conflict detection was broken because of this, seehttps://github.com/hpcugent/easybuild-easyconfigs/pull/1862#issuecomment-131315860.

Fixed via #1870

PR #1862, opened 2015-08-14
{vis}[intel/2015a] Graphviz 2.38.0 + pygraphviz 1.3rc2 w/ Python 2.7.10 (REVIEW)

boegel:
nothing too special other that version updates

requires ~~#1861~~

PR #1780, opened 2015-07-09
{vis} [intel/2015a] Gtkmm and deps (pango, GLib, cairo, Atkmm) (REVIEW)

JensTimmerman:

PR #1870, opened 2015-08-15
fix conflict detection in unit tests

boegel:
Conflict detection is currently broken, see for example #1862 and #1780.

This is caused by the change made in #1853, apparently the value for `spec['dependencies']` doesn't include everything after all...

Related: https://github.com/hpcugent/easybuild-framework/pull/1359

note: this requires https://github.com/hpcugent/easybuild-framework/pull/1360


rjeschmi:
Looks reasonable. I guess this was a bug eh?

boegel:
@rjeschmi: it worked fine when `unresolved_deps` was being used (cfr. #1853), I think (not 100% sure though, it may have missed stuff if `hiddendependencies` were involved)

Thanks for the review!

PR #1920, opened 2015-08-27
fix OpenFOAM patch for v2.3.1 (WIP)

boegel:
fix for https://github.com/hpcugent/easybuild-easyconfigs/issues/1842

cc @FokkeDijkstra

### DIFF
--- a/easybuild/easyconfigs/o/OpenFOAM/cleanup-OpenFOAM-2.3.1.patch
+++ b/easybuild/easyconfigs/o/OpenFOAM/cleanup-OpenFOAM-2.3.1.patch
@@ -138,11 +138,12 @@ diff -ru OpenFOAM-2.3.1.orig/src/parallel/decompose/scotchDecomp/Make/options Op
 diff -ru OpenFOAM-2.3.1.orig/wmake/rules/linux64Gcc/c OpenFOAM-2.3.1/wmake/rules/linux64Gcc/c
 --- OpenFOAM-2.3.1.orig/wmake/rules/linux64Gcc/c	2014-12-04 18:52:43.000000000 +0100
 +++ OpenFOAM-2.3.1/wmake/rules/linux64Gcc/c	2015-03-17 11:10:09.815948216 +0100
-@@ -2,7 +2,7 @@
+@@ -2,7 +2,8 @@
  
  cWARN        = -Wall
  
 -cc          = gcc -m64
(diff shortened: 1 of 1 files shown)


boegel:
Still testing this, the patches for the other OpenFOAM versions need to be changed too.

@wpoely86: please review?

wpoely86:
I don't like this. The values used by EB should be substituted in the Makefile, so the same compiler flags (and versions) are used for all extensions/plugins/...

I've patched the build system of PSI to do this.

boegel:
@wpoely86: you mean, apply runtime patching in the OpenFOAM easyblock to inject the actual value of `$MPICC`, `$CFLAGS` in these files?

wpoely86:
yes, so the same values are used for the core and the plugins.

boegel:
Makes sense, and shouldn't be too hard to implement.

Then we can basically kick out part of the patches we're using now, which I'm definitely for...

Thoughts @FokkeDijkstra for tackling this via the OpenFOAM easyblock itself, and injecting the actual values used by EB rather than using `$(MPICC)` in these files?

FokkeDijkstra:
The advantage of using the OpenFOAM easyblock is that you have better control over the compiler and flags used by OpenFOAM for building the user plugins.
The main disadvantage is that if this easyblock breaks for future OpenFOAM versions, it may be harder to debug (for an easybuild user) what is causing the easybuild failure. This assuming that the original patch would have failed in the patch phase.

I would say implement it in the easyblock, but make it clear in the easybuild output what the block is trying to achieve, and give a clear message if that process fails.

wpoely86:
You can go for a dual approach if you like: add a patch in the easyconfig that replaces the values in the makefile with python strings and let the easyblock fill in the actual values. If something changes in the makefile, only the patch needs updating.

boegel:
I think it's a lot easier to patch in the OpenFOAM easyblock than trying to keep a patch file working, since with runtime patching we have a lot more flexibility...

Making sure clear error messages are present is not a problem, and we can also make the patching optional (but enabled by default), so people can easily disable it if it would fail for future OpenFOAM versions...

boegel:
changes to OpenFOAM easyblock to use runtime patching implemented in https://github.com/hpcugent/easybuild-easyblocks/pull/688, closing this, let's follow up in the easyblock PR

PR #1945, opened 2015-09-03
Molpro

valtandor:
New easyconfigs for a precompiled binary version of Molpro, including a patch file.


valtandor:
Yes.

boegel:
@valtandor: please target the `develop` branch when sending PRs

also, GitHub is telling me that the patch file you've included is a binary file?

Specifying `EB_Molpro` as an easyblock doesn't make sense if this is a binary install either (also, why do a binary install if EB now has support for building Molpro from source?).

valtandor:
@boegel: Regarding the develop branch, I apologise - I must have selected master as the target by mistake.

The last line of the patch is a binary string, it's true. Most of it is plain text, however. The patch affects very close to the last line of the text component before it goes into a bundled tarball.

As for why do a binary install: Some users only receive the binary build of Molpro, as their licence doesn't cover source code.

valtandor:
This pull request has been closed and a new one opened into the develop branch.

boegel:
see #1952

PR #1950, opened 2015-09-07
Added new Go versions 1.4.2 and 1.5

cfenoy:
…ons can only be built with Go compilers. 1.4.2 is the latest version that can be compiled with gcc


cfenoy:
The 1.5 needs hpcugent/easybuild-easyblocks#682

boegel:
@cfenoy: please target `develop` when issuing PRs, see also https://github.com/hpcugent/easybuild-easyblocks/pull/682#issuecomment-138330332

PR #1952, opened 2015-09-08
Molpro

valtandor:
edit: requires https://github.com/hpcugent/easybuild-easyblocks/pull/684


valtandor:
Yes

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
```diff
@@ -0,0 +1,23 @@
...
+homepage = 'https://molpro.net'
+description = """Molpro is a complete system of ab initio programs for molecular electronic structure calculations."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+#toolchain = {'name': 'intel', 'version': '2015a'}
+#toolchainopts = {'usempi': True}
+
+# no source URL available, requires registration to download
+sources = ['%(namelower)s-%(versionprefix)s%(version)s%(versionsuffix)s.sh']
+patches = ['%(namelower)s-%(versionprefix)s%(version)s%(versionsuffix)s.patch']
```
boegel: @valtandor: you explained in #1945 what's going on with the binary patch, but I'm not very comfortable with it

the patch looks like this (GitHub doesn't even support showing it partially), without the binary part:

```diff
--- Molpro/molpro-mpp-2010.1.23-Linux_x86_64.sh	2015-09-02 12:57:08.239724674 +1200
+++ Molpro/molpro-mpp-2010.1.23-Linux_x86_64.sh	2015-09-02 13:01:18.998587159 +1200
@@ -162,7 +162,8 @@
  echo Token sucessfully downloaded and installed
  exit 0
 fi
-molpro_error "Token downloaded unsucessful"
+echo "Token downloaded unsucessful"
+exit 0
 _EOF_
```

Can the patch be done a little bit differently to avoid hitting the binary bit, e.g. by fixing/changing the condition in the `if` above?

Why is this needed b.t.w., what's going on here exactly?

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
```diff
@@ -0,0 +1,23 @@
...
+precompiled_binaries = True
+
+# license file
+#import os
+#license_file = os.path.join(os.getenv('HOME'), 'licenses', name, 'license.lic')
```
boegel: this should stay?

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
```diff
@@ -0,0 +1,23 @@
...
+versionprefix = 'mpp-'
+version = '2010.1.23'
+versionsuffix = '-Linux_x86_64'
+
+homepage = 'https://molpro.net'
+description = """Molpro is a complete system of ab initio programs for molecular electronic structure calculations."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+#toolchain = {'name': 'intel', 'version': '2015a'}
+#toolchainopts = {'usempi': True}
```
boegel: please remove commented out lines

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
```diff
@@ -0,0 +1,23 @@
+name = 'Molpro'
+easyblock = 'EB_Molpro'
```
boegel: no need for this, `eb` will pick this automagically based on the `name`

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
```diff
@@ -0,0 +1,23 @@
...
+versionprefix = 'mpp-'
```
boegel: remove, since it's overwritten below?

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
valtandor: If it's to be installed at a site that has a licence token, it might be helpful for them to see how the licence file can be supplied.

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
valtandor: I fail to see how this is overwritten?

valtandor:
Hi @boegel,

The purpose of the patch was to ensure that the installation succeeds even if no licence token has been provided at install time. Please note that in such cases, either an admin must later install a token (for all users) or each user must supply an appropriate token. Accordingly, as far as I can tell, no licences were violated in the creation of this patch.

I have prepared a new version of the patch that entirely omits any binary content, and hope you find it satisfactory. It has been tested and works on an installation.

boegel:
@valtandor: my concern had nothing to do with license issues of the patch, but making sure it's clear what is going on, i.e. that the patch shows up in the diff here rather than just as 'binary content'; if git recognises it as being binary, any potential future changes to the patch are also troublesome

Patching out the requirement for a license token at installation time makes sense to me (as long as the installation itself doesn't require it to be there, e.g. when checking whether a valid license is available).

Thanks a lot for the changes, I'll try and take a look ASAP.

boegel:
Jenkins: ok to test

boegel:
Jenkins: test this please

easybuild/easyconfigs/m/Molpro/Molpro-mpp-2010.1.23-Linux_x86_64.eb
```diff
@@ -0,0 +1,21 @@
+name = 'Molpro'
+versionprefix = 'mpp-'
+version = '2010.1.23'
+versionsuffix = '-Linux_x86_64'
+
+homepage = 'https://molpro.net'
```
boegel: this needs to be `https://www.molpro.net/`

boegel: cfr. #2297

boegel:
continued in #2329

@valtandor: I had to change the source filename (`.Linux` instead of `-Linux`)

PR #1990, opened 2015-09-24
add HBase/1.0.2

molden:
necessary for new HBase support in hanythingondemand

### DIFF
--- a/easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
+++ b/easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
@@ -9,7 +9,7 @@ description = """Apache HBase. is the Hadoop database, a distributed, scalable, 
 toolchain = {'name': 'dummy', 'version': ''}
 
 sources = ['%(namelower)s-%(version)s-bin.tar.gz']
-source_urls = ['http://www.us.apache.org/dist/hbase/hbase-1.0.2/']
+source_urls = ['http://www.us.apache.org/dist/hbase/hbase-%(version)s/']
 
 dependencies = [('Java', '1.7.0_80')]
 
@@ -17,7 +17,7 @@ parallel = 1
 
 sanity_check_paths = {  
     'files': ["bin/hbase"],  
-    'dirs': ["bin","conf","docs","hbase-webapps","lib"]  
+    'dirs': ["conf", "docs", "hbase-webapps", "lib"]  
 }  
 
 modextravars = {'HBASE_HOME': '%(installdir)s'}
--- a/easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
+++ b/easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
@@ -1,0 +1,25 @@
+easyblock = 'PackedBinary'
+
+name = 'HBase'
+version = '1.0.2'
+
+homepage = 'http://hbase.apache.org/'
+description = """Apache HBase. is the Hadoop database, a distributed, scalable, big data store. """
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['%(namelower)s-%(version)s-bin.tar.gz']
+source_urls = ['http://www.us.apache.org/dist/hbase/hbase-1.0.2/']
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
```diff
@@ -0,0 +1,25 @@
...
+name = 'HBase'
+version = '1.0.2'
+
+homepage = 'http://hbase.apache.org/'
+description = """Apache HBase. is the Hadoop database, a distributed, scalable, big data store. """
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['%(namelower)s-%(version)s-bin.tar.gz']
+source_urls = ['http://www.us.apache.org/dist/hbase/hbase-1.0.2/']
```
boegel: please use `%(version)s` here too, to avoid having the version hard-coded in multiple places:

```
source_urls = ['http://www.us.apache.org/dist/hbase/hbase-1.0.2/']
```

boegel: sorry, this:

```
source_urls = ['http://www.us.apache.org/dist/hbase/hbase-%(version)s/']
```

easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
```diff
@@ -0,0 +1,25 @@
...
+
+name = 'HBase'
+version = '1.0.2'
+
+homepage = 'http://hbase.apache.org/'
+description = """Apache HBase. is the Hadoop database, a distributed, scalable, big data store. """
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['%(namelower)s-%(version)s-bin.tar.gz']
```
boegel: would building this ourselves from source be an option too? just wondering...

easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
```diff
@@ -0,0 +1,25 @@
...
+source_urls = ['http://www.us.apache.org/dist/hbase/hbase-1.0.2/']
+
+dependencies = [('Java', '1.7.0_80')]
+
+parallel = 1
+
+sanity_check_paths = {  
+    'files': ["bin/hbase"],  
+    'dirs': ["bin","conf","docs","hbase-webapps","lib"]
```
boegel: spaces after `,` would be nice (and including `bin` here isn't needed, since we already check that `bin/hbase` is there):

```python
sanity_check_paths = {  
    'files': ['bin/hbase'],
    'dirs': ['conf', 'docs', 'hbase-webapps', 'lib'],
}
```

boegel:
Jenkins: ok to test

easybuild/easyconfigs/h/HBase/HBase-1.0.2.eb
molden: would be an option but not always necessary as it is basically Java code. Maybe we could make a distinction in the module name with a suffix? In case we want to add a built version from source?

boegel: well, it would be sort of clear from the toolchain being used

if it's basically Java code, then it's good as is

I assume the `bin/hbase` is then a bash script or somesuch?

molden: Yes that is a bash script. However I we would like to build it from source we would need to add an easyblock (like the hadoop easyblock) I assume and not use a toolchain, as it will use Maven tool to build the sources.

boegel: That sounds like... fun. ;-)

boegel:
@molden: if you can fix the (other) minor remarks, this should be good to go

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/c042cb4037e1ebf39d1d for a full test report.

boegel:
Going in, thanks @molden!

PR #1996, opened 2015-09-25
{devel}[dummy,GCC/5.2.0] Boost 1.59.0 with full icu support

torbjoernk:
The 'parallel' option is matching an addition in the Boost easyblock (see hpcugent/easybuild-easyblocks#701).

Also had to add an update for the bzip2 config.

[Review on Reviewable](https://reviewable.io/reviews/hpcugent/easybuild-easyconfigs/1996)

### DIFF
--- a/easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
+++ b/easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
@@ -1,0 +1,25 @@
+name = 'Boost'
+version = '1.59.0'
+
+homepage = 'http://www.boost.org/'
+description = """Boost provides free peer-reviewed portable C++ source libraries."""
+
+toolchain = {'name': 'GCC', 'version': '5.2.0'}
+toolchainopts = {'pic': True}
+
+source_urls = [SOURCEFORGE_SOURCE]
+sources = ['%%(namelower)s_%s.tar.gz' % '_'.join(version.split('.'))]
+
+dependencies = [('bzip2', '1.0.6')]
+
+boost_mpi = False
+
+configopts = '--without-libraries=python --with-icu'
+
+toolset = 'gcc'
+
+osdependencies = [('zlib-devel', 'zlib1g-dev'), ('libicu-devel', 'icu-devel', 'libicu-dev')]
+
+parallel = 8
+
+moduleclass = 'devel'
--- a/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-GCC-5.2.0.eb
+++ b/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-GCC-5.2.0.eb
@@ -1,0 +1,19 @@
+name = 'bzip2'
+version = '1.0.6'
+
+homepage = 'http://www.bzip.org/'
+description = """bzip2 is a freely available, patent free, high-quality data compressor. It typically 
+compresses files to within 10% to 15% of the best available techniques (the PPM family of statistical 
+compressors), whilst being around twice as fast at compression and six times faster at decompression."""
+
+toolchain = {'name': 'GCC', 'version': '5.2.0'}
+toolchainopts = {'pic': True}
+
+sources = [SOURCE_TAR_GZ]
+source_urls = ['http://www.bzip.org/%(version)s/']
+
+import os
+os.environ['CC'] = 'gcc'
+os.environ['CFLAGS'] = '-O3 -fPIC'
+
+moduleclass = 'tools'


easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-GCC-5.2.0.eb
```diff
@@ -0,0 +1,19 @@
...
+
+toolchain = {'name': 'GCC', 'version': '5.2.0'}
+toolchainopts = {'pic': True}
+
+sources = [SOURCE_TAR_GZ]
+source_urls = ['http://www.bzip.org/%(version)s/']
+
+import os
+os.environ['CC'] = 'gcc'
+os.environ['CFLAGS'] = '-O3 -fPIC'
```
wpoely86: this is done by EB itself. It should not be done in the easyconfig (run with the `-l` flag to see the variables being set).

easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
```diff
@@ -0,0 +1,25 @@
...
+
+boost_mpi = False
+
+configopts = '--without-libraries=python --with-icu'
+
+toolset = 'gcc'
+
+osdependencies = [('zlib-devel', 'zlib1g-dev'), ('libicu-devel', 'icu-devel', 'libicu-dev')]
+
+parallel = 8
```
wpoely86: EB will by default utilize the maximal number of cores in the current cpuset/cgroup. It you want to limit this to maximally 8, use `maxparallel`. Else this line can go.

easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
```diff
@@ -0,0 +1,25 @@
...
+
+dependencies = [('bzip2', '1.0.6')]
+
+boost_mpi = False
+
+configopts = '--without-libraries=python --with-icu'
+
+toolset = 'gcc'
+
+osdependencies = [('zlib-devel', 'zlib1g-dev'), ('libicu-devel', 'icu-devel', 'libicu-dev')]
```
wpoely86: split this line, 1 dep per line (for example, look in the other easyconfigs for Boost).

easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
```diff
@@ -0,0 +1,25 @@
...
+source_urls = [SOURCEFORGE_SOURCE]
+sources = ['%%(namelower)s_%s.tar.gz' % '_'.join(version.split('.'))]
+
+dependencies = [('bzip2', '1.0.6')]
+
+boost_mpi = False
+
+configopts = '--without-libraries=python --with-icu'
+
+toolset = 'gcc'
```
wpoely86: why should this do?

easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-GCC-5.2.0.eb
torbjoernk: I've added that bzip2 config as EB was complaining that it couldn't resolve the bzip2 dependency when I ran `eb Boost-1.59.0-GCC-5.2.0.eb -r`. Thus I've just copied `bzip2-1.0.6-GCC-4.9.2.eb` and adjusted GCC to 5.2.0.

easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
```diff
@@ -0,0 +1,25 @@
+name = 'Boost'
+version = '1.59.0'
+
+homepage = 'http://www.boost.org/'
+description = """Boost provides free peer-reviewed portable C++ source libraries."""
+
+toolchain = {'name': 'GCC', 'version': '5.2.0'}
```
wpoely86: We don't use the GCC toolchain anymore directly. Use `GNU` which is a combination of GCC with binutils. There reason for this is that RHEL 6 has a binutils that is too old to support Haswell hardware.

wpoely86:
ok to test

easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
torbjoernk: `toolset` is an option from the Boost easyblock for specifying the so called _toolset_ for b2.

easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
torbjoernk: Didn't knew that. Going to change that.

easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-GCC-5.2.0.eb
wpoely86: This should only be present in the `dummy` bzip2 easyconfig. 

But this easyconfig can go if you switch to the `GNU` toolchain. It already has a bzip2 easyconfig.

easybuild/easyconfigs/b/Boost/Boost-1.59.0-GCC-5.2.0.eb
boegel: why not list `zlib` and `libicu` as EasyBuild dependencies?

boegel:
@torbjoernk: first things first: welcome to EasyBuild! Let us know if something isn't clear to you. Or maybe @ocaisa or @geimer can help or answer your questions. ;-)

torbjoernk:
Thanks guys. I'll rework this PR a little more ontop of #2001 and try to get libicu into EB.

PR #2008, opened 2015-09-28
New easyconfig for Google Test (gtest) 1.7.0

tovrstra:
From the commit message:

New easyconfig for Google Test 1.7.0

Some differences from the existing easyconfig for Google Test 1.6.0:

- New version

- GTEST_ROOT environment variable is defined for the Module file such that CMake build scripts automatically pick up the right location of Google Test.

- Shared libraries are installed, in addition to the static link libraries.

### DIFF
--- a/easybuild/easyconfigs/g/gtest/gtest-1.7.0-intel-2015a.eb
+++ b/easybuild/easyconfigs/g/gtest/gtest-1.7.0-intel-2015a.eb
@@ -1,0 +1,36 @@
+easyblock = "MakeCp"
+
+name = 'gtest'
+version = '1.7.0'
+
+homepage = 'https://code.google.com/p/googletest/'
+description = "Google's framework for writing C++ tests on a variety of platforms"
+
+toolchain = {'name': 'intel', 'version': '2015a'}
+
+sources = [SOURCE_ZIP]
+source_urls = ['https://googletest.googlecode.com/files']
+
+with_configure = True
+
+so_files = [
+    'lib/.libs/libgtest.so',
+    'lib/.libs/libgtest.so.0',
+    'lib/.libs/libgtest.so.0.0.0',
+    'lib/.libs/libgtest_main.so',
+    'lib/.libs/libgtest_main.so.0',
+    'lib/.libs/libgtest_main.so.0.0.0',
+]
+files_to_copy = ['lib', 'include', (so_files, 'lib')]
+
+sanity_check_paths={
+    'files': ['lib/libgtest.la', 'lib/libgtest_main.la'],
+    'dirs': ['include'],
+}
+
+# Environment variable to help CMake's FindGTest module:
+modextravars = {
+    'GTEST_ROOT': '%(installdir)s'
+}
+
+moduleclass = 'devel'


boegel:
Jenkins: ok to test

easybuild/easyconfigs/g/gtest/gtest-1.7.0-intel-2015a.eb
```diff
@@ -0,0 +1,36 @@
+easyblock = "MakeCp"
+
+name = 'gtest'
+version = '1.7.0'
+
+homepage = 'https://code.google.com/p/googletest/'
```
boegel: use https://github.com/google/googletest ?

easybuild/easyconfigs/g/gtest/gtest-1.7.0-intel-2015a.eb
```diff
@@ -0,0 +1,36 @@
+easyblock = "MakeCp"
+
+name = 'gtest'
```
boegel: hmm, should we rename to googletest or GoogleTest, which seems to be the official name of the project?

easybuild/easyconfigs/g/gtest/gtest-1.7.0-intel-2015a.eb
```diff
@@ -0,0 +1,36 @@
...
+version = '1.7.0'
+
+homepage = 'https://code.google.com/p/googletest/'
+description = "Google's framework for writing C++ tests on a variety of platforms"
+
+toolchain = {'name': 'intel', 'version': '2015a'}
```
boegel: (also) for `intel/2015b` would be nice

tovrstra:
Thanks for all the feedback. Most is easily fixed and I'll take care of it.

I would go for googletest as new name, because that is the name that is also used in all their filenames. The other name used in the project is "Google Test". I would think that two capital letters and one space are more than enough reasons to just stick to "googletest".

tovrstra:
After reading some more about Google Test, I'd recommend to remove easyconfigs for this library, unless there are historic reasons to keep them. This FAQ explains why:

https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-is-it-not-recommended-to-install-a-pre-compiled-copy-of-google-test-for-example-into-usrlocal

boegel:
@tovrstra: it sounds like they should fix there code, rather than recommending people to not install it just once. That's pretty nuts if you ask me. :-)

We're not going to remove existing easyconfigs, since that would break backward compatibility.

So, do you still want this PR merged, or not?

torbjoernk:
What about pushing that update and add a little user warning/info when installing/loading the googletest module about that issue @tovrstra linked?

boegel:
We can use `modloadmsg` for that, good idea @torbjoernk.

tovrstra:
Thanks for all the feedback. I also find this limitation a little disturbing. I'm not sure if it is worth considering another c++ unit testing framework for my own codes that doesn't have this limitation. Any suggestions?

We probably can't fix the cause for the weird recommendation in the Google Test FAQ. This is likely due to some poor design choices that can't be fixed easily. (Otherwise, they would have fixed it already.) I'm not sure if you even want to have a new easyconfig for Google Test, given this situation. I am not going to use it and I'll adapt my CMake files instead as to download and build Google Test locally. (CMake supports such things.) I just don't want to run into the potential weird problems mentioned in the FAQ.

If someone still wants an easybuild config, the easiest way is to write a patch for the CMakeLists.txt files to include an install target. (Support for building Google Test with autotools is not being maintained.)

torbjoernk:
[off-topic]
@tovrstra Have a look at https://github.com/philsquared/Catch. It's header-only and more "C++ agnostic". The only thing which keeps me from using it is it's (current) lack of generators and mocking (though GoogleMock could be used there).
[/off-topic]

tovrstra:
[off-topic]
@torbjoernk Thanks for the suggestion. I extensively rely on generators. They seem to be supportedin Catch but it just isn't documented yet: https://github.com/philsquared/Catch/blob/c51e86819dc993d590e5d0adaf1952f4b53e5355/projects/SelfTest/GeneratorTests.cpp It may not be fully functional yet.
[/off-topic]

boegel:
@tovrstra ping on remarks?

tovrstra:
I'm no longer using this test framework through easybuild. Instead, my cmake files download and build google test as part of the tests, because this is the method recommended by the googletest developers. (Some previous comments.)

If it works, it may still be helpful for others though.

wpoely86:
I suggest to drop this. Googletest should be part of the project itself and not be external. I don't see any advantage and only lots of potential trouble for this.

boegel:
OK, let's close.

@tovrstra please close (we try to stay away from closing issues/PRs by others)

tovrstra:
Agreed. Closing.

PR #2040, opened 2015-10-08
{math}[dummy] Eigen3: 3.2.5 & 3.2.6

torbjoernk:
recent Eigen3 releases 3.2.5 and 3.2.6

Side question: why are there any compiler/toolchain specific versions of Eigen3? Eigen3 is header-only and only the test suite requires a compiler (which isn't used in the Eigen easyblock).

### DIFF
--- a/easybuild/easyconfigs/e/Eigen/Eigen-3.2.5.eb
+++ b/easybuild/easyconfigs/e/Eigen/Eigen-3.2.5.eb
@@ -1,0 +1,18 @@
+name = 'Eigen'
+version = '3.2.5'
+
+homepage = 'http://eigen.tuxfamily.org/index.php?title=Main_Page'
+description = """Eigen is a C++ template library for linear algebra:
+ matrices, vectors, numerical solvers, and related algorithms."""
+
+toolchain = {'name': 'dummy', 'version': ''}
+toolchainopts = {'optarch': True, 'pic': True}
+
+source_urls = [BITBUCKET_SOURCE]
+sources = ['%(version)s.tar.bz2']
+
+checksums = [
+    '21a928f6e0f1c7f24b6f63ff823593f5'
+]
+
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: ok to test

boegel:
@torbjoernk: I guess it's mostly to keep things 'clean' when including Eigen as a dependency somewhere else, i.e. to avoid

```python
dependencies = [
    ('Eigen', '3.2.5', '', True),
]
```

and use the more common

```python
dependencies = [
    ('Eigen', '3.2.5'),
]
```

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/d17e9aa74f9f18b90b28 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/e63f0be7352562cacc94 for a full test report.

torbjoernk:
I see the convenience for the easyconfigs. However, from a user's perspective, I find it rather disturbing having various occurrences of the same version of a compiler- and toolchain-independent library under different toolchains (e.g. with `module avail`).

boegel:
@torbjoernk: I see your point.

We (well, mostly @ocaisa for now) have been working on making the dependency resolution mechanism (a.k.a. robot) subtoolchains aware (see https://github.com/hpcugent/easybuild-framework/pull/1306).

Once that is in place, `eb` will also consider `Eigen-3.2.5.eb` when trying to resolve a dependency spec like `('Eigen', '3.2.5')`, after considering all the other subtoolchains first.

In fact, this use case is a good reason to also (always/by default) consider easyconfigs using the `dummy` toolchain; we were considering not doing this by default at least, but this case it clear we should (cc @ocaisa).

boegel:
Going in, thanks @torbjoernk!

PR #2083, opened 2015-10-26
Drop GNU

ocaisa:
The argument goes (from @geimer originally) that since `binutils` uses machine code, it doesn't matter how you build binutils. In that case, there's no need to introduce the obfuscation layer `GNU`, simply explicitly list `binutils` as a dependency of `GCC`. 

This allows the `GCC` module to work out-of-the-box when building arch-optimised code.

### DIFF
--- a/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-GCC-4.9.3-binutils-2.25.eb
+++ b/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-GCC-4.9.3-binutils-2.25.eb
@@ -1,0 +1,26 @@
+easyblock = 'ConfigureMake'
+
+name = 'Autoconf'
+version = '2.69'
+
+homepage = 'http://www.gnu.org/software/autoconf/'
+description = """Autoconf is an extensible package of M4 macros that produce shell scripts
+ to automatically configure software source code packages. These scripts can adapt the
+ packages to many kinds of UNIX-like systems without manual user intervention. Autoconf
+ creates a configuration script for a package from a template file that lists the
+ operating system features that the package can use, in the form of M4 macro calls."""
+
+toolchain = {'name': 'GCC', 'version': '4.9.3-binutils-2.25'}
+
+source_urls = [GNU_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [('M4', '1.4.17')]
+
+sanity_check_paths = {
+    'files': ["bin/%s" % x for x in ["autoconf", "autoheader", "autom4te", "autoreconf", "autoscan",
+                                     "autoupdate", "ifnames"]],
+    'dirs': [],
+}
+
+moduleclass = 'devel'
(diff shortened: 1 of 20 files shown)


boegel:
Jenkins: ok to test

boegel:
This yields conflicts for all easyconfigs that use the `GNU` toolchain, so the easyconfigs for the GNU toolchain itself should be modified too?

ocaisa:
The conflict is from a fundamental change to the GCC easyconfig `builddependencies -> dependencies`. The only try way around it is to have a new GCC easyconfig.

boegel:
@ocaisa: Well, the issue is that GNU includes a binutils next to GCC, while GCC now also loads binutils as a dependency itself.

So, GNU should be redefined to only load the GCC, or to include the same binutils as GCC does?

ocaisa:
Yeah, that's the debate. To keep GNU working as it did, I'd have to create a new GCC easyconfig that still uses the builddep...not very hard to do. Changing GNUs binutils dep changes GNU (however superficially) so I'm probably more inclined to just create the new GCC and update the GCC dep in GNU.

ocaisa:
Agree?

boegel:
I don't see the problem with changing the dependencies in `GNU`, as long as it still delivers GCC + binutils (the latter directly or indirectly).

I wouldn't add more GCC easyconfigs, it would only add to the confusion?

ocaisa:
Ok, then the answer is trivial, I just delete the binutils dependency since it comes directly from the GCC easyconfig anyway. This makes GNU just a pseudo-layer over GCC. It will probably trigger a few other dependency errors, I'll fix them when they pop up.

boegel:
@ocaisa: maybe make it explicit and change the version of the binutils dep in GNU instead? (along with adding a comment that that binutils version is also loaded as a dep via GCC)?

wpoely86:
I'm not entirely convinced that this is a good idea (not rebuilding binutils). Replacing GNU by GCC is perfectly fine by me. But I would keep the rebuild.

ocaisa:
But that's not possible (right?), if GCC has a bintuils dependency (not build dependency), you can't use the GCC easyconfig to build binutils

ocaisa:
You'd need to make the binutils build part of GCC itself (like someone suggested in the mailing list) and revert back to a binutils builddep

wpoely86:
@ocaisa you would need an intermediary GCC (but not named so) easyconfig. Not pretty.

ocaisa:
@wpoely86 Yeah, we had discussed internally about whether to have a `GCC-base`. This is not pretty but has a nice side effect for us: `GCC-base` can sit underneath all compiler 'families' (GCC, iccifort, PGI,...) and gives all of them an equivalence. This means I can install common tools like cmake at the GCC-base level rather than dummy. If others thought this was a good idea I'd be happy to implement, with our toolchain-based naming scheme it makes my life even easier.

ocaisa:
So yes, definitely I would be +1 for an intermediate GCC

boegel:
so, https://github.com/hpcugent/easybuild-easyconfigs/pull/2108 is preferred over this, right? so let's close this one? @ocaisa?

ocaisa:
Yes, #2108 is preferred...closing

PR #2085, opened 2015-10-27
{vis}[intel/2015b] pycairo 1.10.0 for cairo 1.14.2 w/ Python 2.7.10 (REVIEW)

boegel:
requires https://github.com/hpcugent/easybuild-easyblocks/pull/722

PR #2086, opened 2015-10-27
{vis}[intel/2015b] p4vasp 0.3.29 + PyGTK 2.24.0 (WIP)

boegel:
requires ~~#2085~~


easybuild/easyconfigs/g/GObject-Introspection/GObject-Introspection-1.46.0-intel-2015b.eb
```diff
@@ -0,0 +1,37 @@
...
+
+toolchain = {'name': 'intel', 'version': '2015b'}
+
+source_urls = [FTPGNOME_SOURCE]
+sources = [SOURCELOWER_TAR_XZ]
+
+patches = ['GObject-Introspection-%(version)s_env-python.patch']
+
+dependencies = [
+    ('Python', '2.7.10'),
```
wpoely86: if you want to stick to your system, this means a version suffix

boegel: OK, will change.

easybuild/easyconfigs/p/p4vasp/p4vasp-0.3.29-intel-2015b-Python-2.7.10.eb
```diff
@@ -0,0 +1,36 @@
...
+sources = [SOURCE_TGZ]
+
+dependencies = [
+    ('PyGTK', '2.24.0', versionsuffix),
+]
+
+skipsteps = ['configure', 'build']
+
+# redefine $HOME to specify installation prefix O_o
+preinstallopts = "HOME=%(installdir)s make local &&"
```
wpoely86: if that `$HOME` is only need for the make local, maybe put it in a subshell to avoid trouble?

wpoely86:
lgtm

easybuild/easyconfigs/p/p4vasp/p4vasp-0.3.29-intel-2015b-Python-2.7.10.eb
boegel: well, it is local, since the install command is run in a subshell anyway, and the change to `$HOME` is really only there for the `make local` command (no `export`, not a separate command to redefine `$HOME` but part of the `make local` command being run)

wpoely86: Yes it will leak to whatever happens after the `&&`

boegel: I understand your concern, but it won't leak, the `HOME=...` part is only used in the `make local` command, not beyond.
See:

```
 cat Makefile 
all:
	echo ${HOME}
$ HOME=foo make && make
echo foo
foo
echo /Users/kehoste
/Users/kehoste
```

boegel:
Still WIP: installation works, but trying to use `p4v` yields `"ImportError: No module named gtk"`...

boegel:
Jenkins: test this please

boegel:
Hmm, Jenkins seems to refuse retesting this... Will open a new PR.

boegel:
see #2328, didn't help with the Jenkins issue though... >_<

PR #2090, opened 2015-10-28
MPICH

valtandor:
New Easyconfigs for MPICH (version 3.0.4, principally) with various toolchains and with support for the Slurm scheduler.


easybuild/easyconfigs/m/MPICH/MPICH-3.0.4-GCC-4.8.1-slurm.eb
```diff
@@ -16,7 +16,7 @@ source_urls = ['http://www.mpich.org/static/tarballs/%(version)s']
 preconfigopts = 'export FC="$F90"; export FCFLAGS="$F90FLAGS"; unset F90; unset F90FLAGS; '
 
 # Build shared libraries
-configopts = '--enable-shared'
+configopts = '--enable-shared --with-slurm=/usr/lib64/slurm --with-pmi=pmi2 '
```
boegel: you're changing an existing easyconfig here, please don't, i.e. make sure MPICH-3.0.4-GCC-4.8.1.eb is not being removed

boegel:
Jenkins: ok to test

easybuild/easyconfigs/m/MPICH/MPICH-3.0.4-GCC-4.8.1-slurm.eb
```diff
@@ -0,0 +1,30 @@
...
+toolchain = {'name': 'GCC', 'version': '4.8.1'}
+
+sources = ['mpich-%(version)s.tar.gz']
+source_urls = ['http://www.mpich.org/static/tarballs/%(version)s']
+
+# MPICH configure wants F90/F90FLAGS to be renamed to FC/FCFLAGS.
+preconfigopts = 'export FC="$F90"; export FCFLAGS="$F90FLAGS"; unset F90; unset F90FLAGS; '
+
+# Build shared libraries
+configopts = '--enable-shared --with-slurm=/usr/lib64/slurm --with-pmi=pmi2 '
```
boegel: hmm, I'm not sure if we want site-specific (specific to sites that are running SLURM) easyconfigs in the central easyconfigs repo like this...

valtandor: Either way, it's understandable. I'll leave it to you to make the policy decision.

valtandor:
@boegel: Did you end up deciding whether you want these in the central repo? If you don't, let me know and I can close this PR.

boegel:
@valtandor: since you're using a proper versionsuffix, I'm keen on merging them in, but they're going to be a pain when testing, since we're not a SLURM site (and I expect the configuration step will fail if I'm not on a SLURM system)

I can just skip these when regression testing of course...

valtandor:
Makes sense.

easybuild/easyconfigs/m/MPICH/MPICH-3.0.4-GCC-4.8.1-slurm.eb
kcgthb: Especially since users may have installed Slurm in a different directory.

valtandor:
@boegel @kcgthb I can always cancel the pull request if that would be preferred. I had uploaded it in the hope that it would be useful.

boegel:
@valtandor: Don't get me wrong, I'm not saying this is utterly useless, but we have tried to refrain from included easyconfigs that have site-specific things in them.

Like @kcgthb pointed out, the location to the SLURM installation may be (very) different for other sites.

I feel that this should be documented instead, somewhere on http://easybuild.readthedocs.org.

There are a couple of site-specific changes that can be made to OpenMPI as well, which could be documented on the same page.

Thoughts?

valtandor:
@boegel: I have no objection. Would you like me to close this PR, therefore, or will you do it once you've finished documenting things?

valtandor:
OK, I'll close this PR. Would it be helpful for me to cast my eyes over related documentation?

PR #2092, opened 2015-10-28
don't use 'beta' in commented out line in hanythingondemand easyconfigs

boegel:
matters little since it's commented out, but we were bitten by this in our local install after just commenting out these lines...


ehiggs:
:+1:

boegel:
Thanks for the review @ehiggs!

PR #2108, opened 2015-10-30
GCCcore

ocaisa:
GCCcore is a reworking of the GCC/GNU concept. GCCcore is an intermediary GCC build that allows us to build binutils with a non-system compiler. 
The proposed implementation is that GCCcore will sit underneath all compilers including GCC. This will allow us to construct GCC in the same way we do intel, PGI (on top of an underlying gcc compiler). We can then safely place GCC/intel/PGI in the same lmod family.
The added advantage is that one can build common applications like CMake in GCCcore and have the same installation available in all naming schemes to all members of the compiler family.
It also allows to have different versions of gcc for GCCcore and GCC, meaning we can keep a universally compatible gcc version underneath all toolchains (which would currently be 4.9.3).
This PR is paired with hpcugent/easybuild-framework#1451


ocaisa:
This is an alternative approach to #2083 and one that I would much prefer (personally)

boegel:
Jenkins: ok to test

geimer:
The naming, i.e., `GCC` providing a "full" compiler including the required `binutils` is in my opinion much less confusing than the `GNU` thing.  But the approach is essentially the same.  (And I'm still not convinced that rebuilding `binutils` gives you any benefit.)

One thing to note is that if you intend to build common stuff using `GCCbase`, you have to remember adding `binutils` as a build dependency to *every* such package, as `GCCbase` itself cannot simply depend on the `binutils` module.  But here is an idea that just crossed my mind which could potentially solve this problem: What if `GCCbase` would depend on `binutils` and `GCC` would `module swap` it with the bootstrapped version?  This is what Lmod does behind the scenes and I verified that it also works with Tcl/C modules.  The question is "just" how to tell this to EasyBuild...

But it seems that in your approach `GNU` is intended for building the common stuff?  If so, this would not be backwards-compatible (in the strict sense), as it currently combines `GCC` with a different `binutils` build.

BTW: Would this bootstrapping also work with the subtoolchain searching enabled?  I could easily imagine that the bootstrapped `binutils` build would be skipped since a version built with a subtoolchain is already available.

One final comment (which is more general): It is rediculous to rebuild Autoconf and Automake for every toolchain.  Their sources don't include a single line of C/C++/Fortran code that get's compiled during installation.  It's all M4, shell, and perl scripts...

ocaisa:
> One thing to note is that if you intend to build common stuff using GCCbase, you have to remember adding binutils as a build dependency to every such package, as GCCbase itself cannot simply depend on the binutils module. 

This is true, you can see I've done that for the `icc` and `ifort` easyconfs in the PR. Without some logic like you described to do some magic like a `module swap` there's no effective way around this other than creating a `bundle` easyconfig that packages the two.

> But it seems that in your approach GNU is intended for building the common stuff? If so, this would not be backwards-compatible (in the strict sense), as it currently combines GCC with a different binutils build.

No. In the mapping I've created 
```
GCC -> GCCcore
GNU -> GCC
```
`GCCcore` is the underlying GCC compiler instance across all compilers (`GCC`, `iccifort`, `PGI`,...) and common tools across toolchains are therefore built with `GCCcore`. It is true that this is not the same as the previous incarnation where `GNU` had this role. To me the old model (where `iccifort` was built on top of `GNU`) is a little broken because it doesn't allow me to place `GNU` and `iccifort` modules in the same `family`. With this more flexible model (where `GCC` is actually a complete new build of the GCC compiler on top of `GCCcore`) there is no (direct) dependence between compilers and I can even have different GCC versions for `GCCcore` and `GCC`

> BTW: Would this bootstrapping also work with the subtoolchain searching enabled? I could easily imagine that the bootstrapped binutils build would be skipped since a version built with a subtoolchain is already available.

This is actually nicely solved in this setup. Subtoolchain searching has a specific option to enable/disable including dummy in your toolchain hierarchy (though still allowing you to explicitly include anything in an easyconfig, which you would need to build `GCCcore`). The proposed default is to exclude dummy, meaning the binutils in GCCcore would be the one found by default.

> One final comment (which is more general): It is rediculous to rebuild Autoconf and Automake for every toolchain. Their sources don't include a single line of C/C++/Fortran code that get's compiled during installation. It's all M4, shell, and perl scripts...

Agreed. When subtoolchain searching is included these can take a default placing in GCCcore.

geimer:
> No. In the mapping I've created
> ```
GCC -> GCCcore
GNU -> GCC
```
>GCCcore is the underlying GCC compiler instance across all compilers (GCC, iccifort, PGI,...) and common tools across toolchains are therefore built with GCCcore. 

Hmm... In the `GNU-4.9.3-2.25.eb` file you replaced the dependency on `GCC` by `GCCcore`, no?  Am I missing something here?

**Comment by ocaisa** [comment]
Ah ok, yeah that was just to keep GNU working in a backwards compatible way. GNU is not part of the setup at all, I just wanted it to keep working as it did before.

File: easybuild/easyconfigs/g/GCC/GCC-4.9.3-binutils-2.25.eb [hunk 1]
```diff
@@ -25,7 +26,12 @@ sources = [
     'mpc-1.0.2.tar.gz',
 ]
 
-builddependencies = [('binutils', binutilsver)]
+dependencies = [
+    # GCC built on top of (dummy-built) binutils
+    ('GCCcore', gccver, '-binutils-%s' % binutilsver),
```
JensTimmerman: can't this be a build dependency? (but this probably doesn't make any difference since it will be a dependency trough binutils anyway?)

JensTimmerman:
+1 on this

easybuild/easyconfigs/g/GCC/GCC-4.9.3-binutils-2.25.eb
ocaisa: Yeah, I can clean this out. This is a hangover from the toolchain-based naming scheme that I use. With the latest release I can improve the intelligence of the naming scheme so this isn't necessary.

easybuild/easyconfigs/g/GNU/GNU-4.9.3-2.25.eb
```diff
@@ -12,9 +12,9 @@ toolchain = {'name': 'dummy', 'version': 'dummy'}
 
 dependencies = [
     # GCC built on top of (dummy-built) binutils
-    ('GCC', gccver, '-binutils-%s' % binutilsver),
```
wpoely86: shouldn't we just load `GCC` here?

wpoely86:
This looks good. I would suggest that `GCCcore` should be a hidden module. The end-user should never load it directly. For building build tools with `GCCcore` we should indeed add a bundle combining `GCCcore` and `binutils`.

JensTimmerman:
I created https://github.com/hpcugent/easybuild-easyconfigs/pull/2135 that basically does the same as this (If I understand everyting correctly) , but here the naming is better, less suffixes...

easybuild/easyconfigs/g/GNU/GNU-4.9.3-2.25.eb
ocaisa: I'm trying to make this work exactly like it did before, with the renaming that this PR does this means GCC -> GCCcore

boegel:
Jenkins: test this please

JackPerdue:
What's the status of this at this point?  I'm rewriting my IBM XL compiler stuff, which relies upon a multilib GCC, and I'm wondering if I should (or can) rebuild my GCCs using base/binutils/etc.  If so, what easyconfigs or PRs should I use as a model for what's to come (or is there such a model yet)?

boegel:
@JackPerdue: I need to take a good look at this (and https://github.com/hpcugent/easybuild-framework/pull/1451), but this is planned to go in very soon (cfr. https://github.com/hpcugent/easybuild/issues/170).

For now, I'd say hold on until you see an update from me here, I haven't had the change yet to take a detailed look at this and evaluate the possible repercussions (there should be none, ideally, for people who already have modules installed).

JackPerdue:
OK...  on a semi-related note (feature request to gum up the works)... the configure scripts for the XL compilers are hard-coded so that it expect to find ld and as in the same directory as gcc.  Obviously that's not the case with EB at present.  As such, I'm using this hack for now:

<pre>
# toolchain = gcc, dependencies = binutils and xlcbase
preconfigopts = """
    gdir=$(dirname $(which gcc))
    if [ -L $gdir/as ] ; then rm $gdir/as ; fi
    ln -vs `which as` $gdir
    if [ -L $gdir/ld ] ; then rm $gdir/ld ; fi
    ln -vs `which ld` $gdir
    ...
</pre>

with corresponding rm's in the postinstallcmds.  Obviously this won't work if the user doesn't have write perms to GCC's bin dir (it so happens I do for now).  If there was a GCC-binutils that combined the bin from GCC and binutils, that would be fabulous (and eliminate the cheese above).

boegel:
@JackPerdue: maybe we should simply add an option in the GCC easyblock to make it install the symlinks to `$(which as)` and `$(which ld)` properly, when installing GCC? Doesn't seem to crazy of a requirement.

Doing this from another (non-GCC) easyconfig is very dirty imho, since you're changing things in another installation prefix...

JackPerdue:
That would be great... I agree its dirty and was quite sure it wouldn't be acceptable in the long run but am working around things for now.  If I could set 'xllinks= True' or 'ibmxllinks=True'  (default False) in the GCC-multilib easyconfig , that would be ideal.

boegel:
@JackPerdue: can you issue a feature request for that in the easyblocks repo? shouldn't be too hard to implement

ocaisa:
Hi all,

I think this will work well in GCCcore and shouldn't have any ill effects. Once GCCcore is installed you build binutils with it and therefore if you use that as a dependency (which you should) of the target compiler, it will take precedence anyway over what is in GCCcore.

Alan

On 2 December 2015 at 22:29, Kenneth Hoste <notifications@github.com<mailto:notifications@github.com>> wrote:

@JackPerdue<https://github.com/JackPerdue>: maybe we should simply add an option in the GCC easyblock to make it install the symlinks to $(which as) and $(which ld) properly, when installing GCC? Doesn't seem to crazy of a requirement.

Doing this from another (non-GCC) easyconfig is very dirty imho, since you're changing things in another installation prefix...

—
Reply to this email directly or view it on GitHub<https://github.com/hpcugent/easybuild-easyconfigs/pull/2108#issuecomment-161439305>.



--
Dr. Alan O'Cais
Application Support
Juelich Supercomputing Centre
Forschungszentrum Juelich GmbH
52425 Juelich, Germany

Phone: +49 2461 61 5213
Fax: +49 2461 61 6656
E-mail: a.ocais@fz-juelich.de<mailto:a.ocais@fz-juelich.de>
WWW:    http://www.fz-juelich.de/ias/jsc/EN


------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------
Forschungszentrum Juelich GmbH
52425 Juelich
Sitz der Gesellschaft: Juelich
Eingetragen im Handelsregister des Amtsgerichts Dueren Nr. HR B 3498
Vorsitzender des Aufsichtsrats: MinDir Dr. Karl Eugen Huthmacher
Geschaeftsfuehrung: Prof. Dr.-Ing. Wolfgang Marquardt (Vorsitzender),
Karsten Beneke (stellv. Vorsitzender), Prof. Dr.-Ing. Harald Bolt,
Prof. Dr. Sebastian M. Schmidt
------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------

easybuild/easyconfigs/i/icc/icc-2015.3.187-GCCcore-4.9.3-binutils-2.25.eb
```diff
@@ -0,0 +1,27 @@
...
+homepage = 'http://software.intel.com/en-us/intel-compilers/'
+description = "C and C++ compiler from Intel"
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = ['l_ccompxe_%(version)s.tgz']
+
+gcc = 'GCCcore'
+binutilsver = '2.25'
+gccver = '4.9.3-binutils-%s' % binutilsver
```
boegel: I would keep using `GCC` in the versionsuffix rather than `GCCcore`?

`GCCcore` should be an internal thing as much as possible (maybe even a hidden module by default?)

easybuild/easyconfigs/g/GCCcore/GCCcore-4.9.3-binutils-2.25.eb
```diff
@@ -0,0 +1,47 @@
+easyblock = "EB_GCC"
+name = "GCCcore"
+version = '4.9.3'
+
+binutilsver = '2.25'
+versionsuffix = '-binutils-%s' % binutilsver
```
boegel: do we need `binutils` here in the versionsuffix?

boegel: or even the version? since it's only a build dep, it doesn't matter?

easybuild/easyconfigs/i/intel/intel-2015b.eb
```diff
@@ -9,7 +9,7 @@ description = """Intel Cluster Toolkit Compiler Edition provides Intel C/C++ and
 toolchain = {'name': 'dummy', 'version': 'dummy'}
 
 compver = '2015.3.187'
-gccsuff = '-GNU-4.9.3-2.25'
+gccsuff = '-GCCcore-4.9.3-binutils-2.25'
```
boegel: maybe this should just be `-GCC-4.9.3-2.25`?

ocaisa: You mean a dependency directly on GCC? Or just a different name? Definitely don't agree on depending in GCC directly since that wrecks the whole concept of having GCC/iccifort/PGI in the same `family`

boegel: Just the name, I see no use in 'exposing' the 'GCCcore' label.

easybuild/easyconfigs/i/icc/icc-2015.3.187-GCCcore-4.9.3-binutils-2.25.eb
ocaisa: Happy to hide it but I have a sneaking feeling that you can't hide toolchains (if you can, how do you indicate that the toolchain module is hidden in the easyconfig?)

ocaisa:
Some notes for future self...it may be the case that you want to build GCC 5 on top of a GCCcore 4.X that is common to all your compiler family (GCC/PGI/iccifort/...). In that case you can use GCCcore as a home for utilities like git, CMake, M4, Autotools,...and in particular vis libraries like libX11 (especially wxwidgets which will only build with GCC anyway). This will ensure a single installation that will also automatically appear for all toolchains when you use a module hierarchy.

There are two potential gotchas, one is that your GCC 5.x build should be configured to use the old C++ ABI by default to ensure compatibility. You can do this via the environment variable [`_GLIBCXX_USE_CXX11_ABI`](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html) or by (re-)configuring the default behaviour for this variable with [`--with-default-libstdcxx-abi=gcc4-compatible`](https://gcc.gnu.org/onlinedocs/libstdc++/manual/configure.html)

The other gotcha is how to deal with the a Python requirement in GCCcore.  Probably you want your runtime python instances to come from the compiler level but some vis tools  (like libX11) require python for whatever reason. Putting the same python version available at GCCcore and compiler level will cause problems with `--minimal-toolchains` (since it will always default to GCCcore) so you will need to ensure a versionsuffix is always there for a python at GCCcore.  The same could be true in other cases too but this is the only one where I see an immediate issue.

easybuild/easyconfigs/i/icc/icc-2015.3.187-GCCcore-4.9.3-binutils-2.25.eb
boegel: good point

there's a problem with this easyconfig though, it doesn't provide what it claims to (based on the filename):

```
ERROR: easyconfig file icc-2015.3.187-GCCcore-4.9.3-binutils-2.25.eb does not contain module icc/2015.3.187-GCCcore-4.9.3-binutils-2.25 (mods: ['icc/2015.3.187'])
```

missing `versionsuffix`?

@ocaisa: I'll play around with this a bit, and send you a PR soonish.

boegel:
@ocaisa: this is the situation as is in this PR, after I fixed two versionsuffix issues: http://users.ugent.be/~kehoste/intel-foss-2015b-GCCcore.pdf

easybuild/easyconfigs/i/icc/icc-2015.3.187-GCCcore-4.9.3-binutils-2.25.eb
boegel: fix in https://github.com/ocaisa/easybuild-easyconfigs/pull/14

easybuild/easyconfigs/i/intel/intel-2015b.eb
boegel: I'm reconsidering this, since it would cause too much confusing between `GCC-4.9.3-binutils-2.25` and `GCCcore-4.9.3-binutils-2.25`

boegel:
this is the current situation (as in EasyBuild v2.4.0): http://users.ugent.be/~kehoste/intel-foss-2015b-eb240.pdf

easybuild/easyconfigs/b/binutils/binutils-2.25-GCCcore-4.9.3-binutils-2.25.eb
```diff
@@ -6,7 +6,7 @@ version = '2.25'
 homepage = 'http://directory.fsf.org/project/binutils/'
 description = "binutils: GNU binary utilities"
 
-toolchain = {'name': 'GCC', 'version': '4.9.3-binutils-%s' % version}
+toolchain = {'name': 'GCCcore', 'version': '4.9.3-binutils-%s' % version}
```
boegel: existing easyconfigs should not be removed/renamed, instead a new one should be added instead

JackPerdue:
I like the dependency graph... I don't feel so much like I'm going in circles with this issue... :)

FWIW, here is my current take:

<pre>
# NOTE: 'base' builds depend upon https://github.com/hpcugent/easybuild-easyblocks/pull/750
# NOTE2: 'Version's below include suffix

1) document system compiler:
    Name: GCC
    Version: system
    Alternative Version(w/suff): .4.4.7-11.el6  (-2.20.51.0.2-5.42 [for recording binutils version])
    [NOTE: system will be more flexible... the actual version is recorded in the repo for reproducibility later]
    easyblock = SystemCompiler 
    toolchain = dummy/dummy
    builddeps = none
    hiddendeps = none
    deps = none
    hidden = True
1a) flex/2.5.39-GCC-system
1b) Bison/3.0.4-GCC-system w/ M4/1.4.17-GCC-system
1c) zlib/1.2.8-GCC-system
2) build newer binutils with system compiler
    Name: binutilsbase0
    Version: 2.25-GCC-system
    easyblock = ConfigureMake
    toolchain = dummy/NULL
    hiddendeps = GCC/flex/Bison/zlib (above)
    builddeps = hiddendeps
    deps = none
    hidden = True
3) build new gccbase using binutils above
    Name: GCC
    Version: 4.8.4-binutilsbase0-2.25-GCC-system.eb
    easyblock = EB_GCC
    toolchain = dummy/NULL
    builddeps = none
    hiddendeps = binutilsbase0/2.25-GCC-system
    deps = hiddendeps
    hidden = True
3a) flex
    hidden = True
3b) Bison w/M4
    hidden = True
3c) zlib
    hidden = True
4) build new binutils using newer compiler
    Name: binutilsbase
    Version: 2.25-gccbase-4.8.4-binutilsbase-2.25-GCC-system
    easyblock = ConfigureMake
    toolchain = dummy/NULL
    hiddendeps = gccbase/4.8.4-binutilsbase0-2.25-GCC-system,flex/Bison/M4
    builddeps = hiddendeps
    deps = none
    hidden = True
5) (Finally) build GCC (w/binutils support)
    Name: GCC
    Version: 4.8.4 (or 4.8.4-binutilsbase-225???)
    easyblock = unspecified (will use EB_GCC by default)
    toolchain = dummy/dummy
    deps = binutilsbase-2.25-gccbase-4.8.4-binutilsbase-2.25-GCC-system
    hidden = False
6) (additionally) build unhidden binutils with this compiler
    Name: binutils
    Version: 2.25-4.8.4-binutilsbase-225
    easyblock = ConfigureMake
    toolchain = GCC/4.8.4 (or GCC/4.8.4-binutilsbase-225)
    deps = False
</pre>

But that changes from day to day.  Still torn between GCC vs gccbase (and using hax from     https://github.com/hpcugent/easybuild-easyblocks/pull/750) and how much of the versions I need to preserver.  I think hidden build deps helps a lot but haven't had much luck with my patchy testing of it so far.

But I _have_ been pondering it all (there isn't a trivial one-true-answer to this quandry).

boegel:
There's a couple of things I don't like in this PR, as it is currently.

The major issue is that it's fiddling with the `2015b` toolchains, which are close to the end of their 'lifetime'; preferably before 2016 we should define the `2016a` update, so that's a way better time to make changes like this (#2194 is a likely candidate to become `intel/2016a`).

My other main issue is that *two* GCC versions get built for `foss/2015b`, and they're both loaded at runtime, which is nuts imho, and there's no need for it (if anything, it's a recipe for disaster with ABI compatibility, cfr. @ocaisa's comment above).
This can be avoided though, by making `GCC` a toolchain (or bundle) with just `GCCcore` and `binutils` as deps, rather than another GCC *build*.

Also, the `binutils` in the `GCCcore` versionsuffix bothers me, because `binutils` is not a runtime dependency of `GCCcore`. As soon as both `GCCcore` and `binutils` are runtime deps (in `GCC`, `icc` or `ifort`, then it makes sense to me to include `binutils` in the versionsuffix.

In fact, to shorten things, I'm move in favour of actually dropping the `binutils` tag, and just to include the binutils version, i.e. use `GCC-4.9.3-2.25` rather than `GCC-4.9.3-binutils-2.25`.

http://users.ugent.be/~kehoste/intel-foss-2015b-GCCcore-bis.pdf shows an tweaked version of the `foss/2015b` and `intel/2015b` DAGs in this PR.
I'm still using `GCC-4.9.3-binutils-2.25` for the `foss/2015b` part of this, because I haven't fully made up my mind yet, but it makes sense to me to also go with `GCC-4.9.3-2.25` there.

Extra bonus: going with `GCC_4.9.3-2.25` avoids fiddling with the existing `GCC-4.9.3-binutils-2.25.eb`, and to add `GCC-4.9.3-2.25.eb` instead (which would be a *bundle* with `GCC/4.9.3` and `binutils/2.25-GCCcore-4.9.3`, as opposed to an actual GCC build).

boegel:
I propose to go with #2214 instead, as a way forward.

boegel:
@ocaisa: please close, since #2214 is favoured over this?

ocaisa:
Agreed, closing

PR #2109, opened 2015-11-01
fix libunistring sanity check: include/unistring is a directory

boegel:
went undetected because of bug in sanity check implementation, fixed by https://github.com/hpcugent/easybuild-framework/pull/1436


boegel:
Merging as trivial.

PR #2122, opened 2015-11-09
updated homepage url for chip-seq easyconfigs

pescobar:
@boegel I just noticed I missed the url :P


boegel:
@pescobar: I should have noticed during the review... Thanks for fixing this!

PR #2132, opened 2015-11-11
Paraview -> ParaView

ocaisa:
Make the name change to match the releases


boegel:
Jenkins: ok to test

boegel:
@ocaisa: you should consider changes the name of `Paraview` to `ParaView` wherever it's used as a dep?

boegel:
and then we can probably get rid of the easyconfigs using the faulty name `Paraview`?

ocaisa:
I can't find this as a dependency anywhere...it's mentioned in an OpenFOAM patch that's all

boegel:
Some of the new PRs have it:

* https://github.com/hpcugent/easybuild-easyconfigs/pull/2142/files#diff-1ac4947119851357e1e6597e321211e9R37
* https://github.com/hpcugent/easybuild-easyconfigs/pull/2048/files#diff-905cfb22d1be95155079af3382ea5199R42
* https://github.com/hpcugent/easybuild-easyconfigs/pull/2023/files#diff-0f8fe653b4f48acd9d745f3bf8db2df3R42

So, now's a good time to rename them...

@hajgato: thoughts?

boegel:
Going in, thanks @ocaisa!

PR #2135, opened 2015-11-12
{devel} [GCC] GCC made binutils a dependency for GCC instead of a buildependency

JensTimmerman:
g++ is broken if this binutils isn't loaded
```
-bash-4.1$ cat test.cpp
#include <cstdlib>

int main()
{
return 0;
}
-bash-4.1$ module load GCC
-bash-4.1$ module list
Currently Loaded Modulefiles:
  1) cluster/delcatty(default)   2) GCC/4.9.3-binutils-2.25
-bash-4.1$ g++ test.cpp 
collect2: fatal error: cannot find 'ld'
compilation terminated.
-bash-4.1$ module load binutils
-bash-4.1$ module list
Currently Loaded Modulefiles:
  1) cluster/delcatty(default)               2) GCC/4.9.3-binutils-2.25                 3) binutils/2.25-GCC-4.9.3-binutils-2.25
-bash-4.1$ g++ test.cpp 
```

### DIFF
--- a/easybuild/easyconfigs/g/GCC/GCC-4.9.3-binutils-2.25-GCC-4.9.3.eb
+++ b/easybuild/easyconfigs/g/GCC/GCC-4.9.3-binutils-2.25-GCC-4.9.3.eb
@@ -2,7 +2,8 @@ name = "GCC"
 version = '4.9.3'
 
 binutilsver = '2.25'
-versionsuffix = '-binutils-%s' % binutilsver
+binutilsversuffix = '-%s-%s' % (name, version)
+versionsuffix = '-binutils-%s%s' % (binutilsver, binutilsversuffix)
 
 homepage = 'http://gcc.gnu.org/'
 description = """The GNU Compiler Collection includes front ends for C, C++, Objective-C, Fortran, Java, and Ada,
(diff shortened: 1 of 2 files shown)


JensTimmerman:
jenkins: test this please

JensTimmerman:
https://github.com/hpcugent/easybuild-easyconfigs/pull/2108 is probably the way to go here.

boegel:
The intention is that people use the `GNU` toolchain, but that's indeed not clear. #2108 was created to resolve exactly this problem, and can be applied to existing installations with minimal impact.

So, I agree that #2108 is probably a better solution.

Thoughts, @ocaisa?

ocaisa:
Yes, this is solved in #2108 but I'm not sure how easily it can be applied to existing installations. For that to work in all cases you'd have to edit all existing `GCC` easyconfigs that have a binutils build dependency to use the new method and force a rebuild. I haven't done that in my easyconf PR, I only did it for the stack of most recent `foss` and `intel` toolchains (so the PR is more forward looking).

boegel:
since there's usually a binutils on the OS that avoids this problem, and that we're fixing this for the `2015b` toolchains (and future ones) in #2108, I'm closing this

PR #1973, opened 2015-09-18
Added icc, ifort, iccifort 2016.0.109 files

DirkdeDraak:
Added support for installing icc, ifort and toolchain iccifort version 2016.0.109.

edit: requires https://github.com/hpcugent/easybuild-easyblocks/pull/691

PR #2148, opened 2015-11-20
Add Intel Parallel Studio 2016, update 1

jas02:
Updated version of Intel Parallel Studio 2016 packages.

### DIFF
--- a/easybuild/easyconfigs/i/icc/icc-2016.1.150-GCC-4.9.3.eb
+++ b/easybuild/easyconfigs/i/icc/icc-2016.1.150-GCC-4.9.3.eb
@@ -8,7 +8,7 @@ description = "C and C++ compiler from Intel"
 
 toolchain = {'name': 'dummy', 'version': 'dummy'}
 
-sources = ['parallel_studio_xe_%(version_major)s_composer_edition_for_cpp_%(version)s.tgz']
+sources = ['parallel_studio_xe_%(version_major)s_composer_edition_for_cpp_update%(version_minor)s.tgz']
 
 checksums = ['4b93b0ff549e6bd8d1a8b9a441b235a8']
 
--- a/easybuild/easyconfigs/h/HPL/HPL-2.1-intel-2016.01.eb
+++ b/easybuild/easyconfigs/h/HPL/HPL-2.1-intel-2016.01.eb
@@ -1,0 +1,18 @@
+name = 'HPL'
+version = '2.1'
+
+homepage = 'http://www.netlib.org/benchmark/hpl/'
+description = """HPL is a software package that solves a (random) dense linear system in double precision (64 bits) arithmetic 
+ on distributed-memory computers. It can thus be regarded as a portable as well as freely available implementation of the 
+ High Performance Computing Linpack Benchmark."""
+
+toolchain = {'name': 'intel', 'version': '2016.01'}
+toolchainopts = {'optarch': True, 'usempi': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = ['http://www.netlib.org/benchmark/%(namelower)s']
+
+# fix Make dependencies, so parallel build also works
+patches = ['HPL_parallel-make.patch']
+
+moduleclass = 'tools'
--- a/easybuild/easyconfigs/i/ifort/ifort-2016.1.150-GCC-4.9.3.eb
+++ b/easybuild/easyconfigs/i/ifort/ifort-2016.1.150-GCC-4.9.3.eb
@@ -8,7 +8,7 @@ description = "C and C++ compiler from Intel"
 
 toolchain = {'name': 'dummy', 'version': 'dummy'}
 
-sources = ['parallel_studio_xe_%(version_major)s_composer_edition_for_fortran_%(version)s.tgz']
+sources = ['parallel_studio_xe_%(version_major)s_composer_edition_for_fortran_update%(version_minor)s.tgz']
 
 checksums = ['1e848c8283cf6a0210bce1d35ecd748b']
 
--- a/easybuild/easyconfigs/i/iimpi/iimpi-2016.01-GCC-4.9.3.eb
+++ b/easybuild/easyconfigs/i/iimpi/iimpi-2016.01-GCC-4.9.3.eb
@@ -2,7 +2,7 @@
 easyblock = "Toolchain"
 
 name = 'iimpi'
-version = '2016.00'
+version = '2016.01'
 versionsuffix = '-GCC-4.9.3'
 
 homepage = 'http://software.intel.com/en-us/intel-cluster-toolkit-compiler/'
--- a/easybuild/easyconfigs/i/imkl/imkl-11.3.1.150-iimpi-2016.01-GCC-4.9.3.eb
+++ b/easybuild/easyconfigs/i/imkl/imkl-11.3.1.150-iimpi-2016.01-GCC-4.9.3.eb
@@ -9,7 +9,7 @@ description = """Intel Math Kernel Library is a library of highly optimized,
  applications that require maximum performance. Core math functions include
  BLAS, LAPACK, ScaLAPACK, Sparse Solvers, Fast Fourier Transforms, Vector Math, and more."""
 
-toolchain = {'name': 'iimpi', 'version': '2016.00-GCC-4.9.3'}
+toolchain = {'name': 'iimpi', 'version': '2016.01-GCC-4.9.3'}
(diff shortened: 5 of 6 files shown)


boegel:
Jenkins: ok to test

easybuild/easyconfigs/i/icc/icc-2016.1.150-GCC-4.9.3.eb
```diff
@@ -0,0 +1,27 @@
...
+
+name = 'icc'
+version = '2016.1.150'
+
+homepage = 'http://software.intel.com/en-us/intel-compilers/'
+description = "C and C++ compiler from Intel"
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = ['parallel_studio_xe_%(version_major)s_composer_edition_for_cpp_%(version)s.tgz']
```
boegel: wait, the tarball still only has `2016`, the same as the previous release?

please tell me that's not the case...

boegel: ah, no, you have both `%(version_major)s` and `%(version)s` in there, nevermind :)

easybuild/easyconfigs/i/intel/intel-2016.00.eb
```diff
@@ -0,0 +1,22 @@
...
+name = 'intel'
+version = '2016.00'
+
+homepage = 'http://software.intel.com/en-us/intel-cluster-toolkit-compiler/'
+description = """Intel Cluster Toolkit Compiler Edition provides Intel C/C++ and Fortran compilers, Intel MPI & Intel MKL."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+compver = '2016.1.150'
+gccsuff = '-GCC-4.9.3'
```
boegel: usually, the `intel` toolchains are `GNU` based, but we're redefining `GNU` and `GCC` in other PRs so that binutils is included in both cases, so I need to take another look at that to have a full pictures here

@ocaisa: thoughts?

boegel:
@jas02: please add an HPL easyconfig that uses this new toolchain?

easybuild/easyconfigs/i/intel/intel-2016.00.eb
```diff
@@ -0,0 +1,22 @@
+# This is an easyconfig file for EasyBuild, see http://hpcugent.github.io/easybuild
+easyblock = "Toolchain"
+
+name = 'intel'
+version = '2016.00'
```
boegel: hmm, not sure if this is the best version, maybe we should stick to `2016.1.150`?

that is: the first update of the 2016 version of these tools; the `150` part comes back in the `impi` and `imkl` versions to indicate that all of these 'match'

wpoely86:
Test report by @wpoely86
**FAILED**
Build succeeded for 0 out of 7 (7 easyconfigs in this PR)
See https://gist.github.com/01393f7b95e1fc075220 for a full test report.

easybuild/easyconfigs/i/intel/intel-2016.00.eb
jas02: `impi` is in version 5.1.2.150 (update 2). Other components of intel toolchain are update1.  So, there is not consistency from Intel side.

This is just maintenance release, changing every time release of toolchain is not what our users want, I think. They want to use latest updated versions of all Intel tools, but in familiar toolchain name (intel-2016.00). What do you think?

wpoely86:
Test report by @wpoely86
**FAILED**
Build succeeded for 0 out of 7 (7 easyconfigs in this PR)
See https://gist.github.com/c54a0911f990136510bf for a full test report.

easybuild/easyconfigs/i/intel/intel-2016.00.eb
boegel: Hmm, I see your point. The `.00` just seems weird to me, what does that mean? initial release?

I don't feel really strongly about it, just as long as we can motivate whatever we come up with

boegel: Ah, but this isn't the initial release, this is update 1, so it should be `2016.01`?

jas02: I used convention from #1973. You had no objections against naming convention and I thought that it will be new toolchain name.

boegel: well, except that #1973 deals with icc/ifort 2016.0.109, while here you have 2016.1.150... The bump in the compiler version should be reflected in the toolchain version, no?

Now both this PR and #1973 propose to use `2016.00`... We want to support installing both the 2016.0.109 and 2016.1.150 versions (for benchmarking/testing, for example).

wpoely86:
@jas02 do these work? You need changes in the easyblock?

I'm gonna try to fixes the easyblock using hpcugent/easybuild-easyblocks#691 and your current branch.

easybuild/easyconfigs/i/intel/intel-2016.00.eb
jas02: OK, how to solve it? Should I rename version to `2016.01`? And every other "major" update will be increment of this number? We have to be consistent in the future.

boegel: Using `2016.01` makes sense to me, but we can discuss this during the conf call this week (I'll fix the date/time for that ASAP).

easybuild/easyconfigs/i/icc/icc-2016.1.150-GCC-4.9.3.eb
boegel:
```python
sources = ['parallel_studio_xe_%(version_major)s_composer_edition_for_cpp_update%(version_minor)s.tgz']
```

easybuild/easyconfigs/i/imkl/imkl-11.3.1.150-iimpi-2016.00-GCC-4.9.3.eb
```diff
@@ -0,0 +1,38 @@
...
+
+moduleclass = 'numlib'
+
+postinstallcmds = [
+    # extract the examples
+    'tar xvzf %(installdir)s/mkl/examples/examples_cluster.tgz -C %(installdir)s/mkl/examples/',
+    'tar xvzf %(installdir)s/mkl/examples/examples_core_c.tgz -C %(installdir)s/mkl/examples/',
+    'tar xvzf %(installdir)s/mkl/examples/examples_core_f.tgz -C %(installdir)s/mkl/examples/',
+    'tar xvzf %(installdir)s/mkl/examples/examples_f95.tgz -C %(installdir)s/mkl/examples/',
+    'tar xvzf %(installdir)s/mkl/examples/examples_mic.tgz -C %(installdir)s/mkl/examples/'
```
boegel: hmm, it would be better to handle this via the imkl easyblock (do-once-and-forget)? thoughts @jas02?

easybuild/easyconfigs/i/impi/impi-5.1.2.150-iccifort-2016.1.150-GCC-4.9.3.eb
```diff
@@ -0,0 +1,31 @@
...
+
+moduleclass = 'mpi'
+
+postinstallcmds = [
+    'ln -s %(installdir)s/lib64/libmpi.so  %(installdir)s/lib64/libmpich.so',
+    'ln -s %(installdir)s/lib64/libmpigc4.so  %(installdir)s/lib64/libmpichcxx.so',
+    'ln -s %(installdir)s/lib64/libmpigf.so %(installdir)s/lib64/libfmpich.so',
+    'ln -s %(installdir)s/lib64/libmpigf.so %(installdir)s/lib64/libmpichf90.so',
+    'ln -s %(installdir)s/lib64/libmpi.so %(installdir)s/lib64/libmpl.so',
+    'ln -s %(installdir)s/lib64/libmpi.so %(installdir)s/lib64/libopa.so'
```
boegel: @jas02: why is this needed? this should also be handled via the easyblock (where we can check on things, for example make sure that the target doesn't exist yet, etc.)?

boegel:
@jas02: please merge https://github.com/jas02/easybuild-easyconfigs/pull/2 to fix (part of) the remarks

boegel:
Jenkins: test this please

boegel:
will be merged via #2219

PR #2157, opened 2015-11-26
{lang}[GCC/4.9.2] Julia 0.3.7 (REVIEW)

tvandera:
Julia language easyconfig


boegel:
@tvandera: you'll get the hang of it after a while ;-)

pro tip: always keep an eye on the diff that GitHub shows you when opening the PR, you'll notice that you have the wrong target branch straight away then

PR #2162, opened 2015-11-26
{lang}[GCC/4.9.2] Julia 0.3.7 (WIP)

tvandera:
Julia language easyconfig

### DIFF
--- a/easybuild/easyconfigs/j/Julia/Julia-0.3.7-GCC-4.9.2-bare.eb
+++ b/easybuild/easyconfigs/j/Julia/Julia-0.3.7-GCC-4.9.2-bare.eb
@@ -1,0 +1,25 @@
+name = 'Julia'
+version = '0.3.7'
+versionsuffix = '-bare'
+
+easyblock = 'ConfigureMake'
+
+homepage = 'http://julialang.org/'
+description = """Julia is a high-level, high-performance dynamic programming
+language for technical computing, with syntax that is familiar to users of
+other technical computing environments."""
+
+toolchain = {'name': 'GCC', 'version': '4.9.2'}
+
+source_urls = ['https://github.com/JuliaLang/julia/archive/']
+sources = ['v%(version)s.zip']
+
+skipsteps = ['configure']
+
+installopts = 'prefix=%(installdir)s'
+
+sanity_check_paths = {
+    'files': ["bin/julia"],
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/j/Julia/Julia-0.3.7-GCC-4.9.2-bare.eb
```diff
@@ -0,0 +1,25 @@
+name = 'Julia'
+version = '0.3.7'
+versionsuffix = '-bare'
+
+easyblock = 'ConfigureMake'
```
boegel: this should go first, separated from `name =` with a blank line

boegel:
@tvandera: there's a WIP easyblock for Julia, see https://github.com/hpcugent/easybuild-easyblocks/pull/499

Does that still make sense, you think? Would your easyconfig benefit from that too?

boegel:
Jenkins: ok to test

tvandera:
There are two options for Julia: go with the system's libraries for BLAS, LAPACK, FFWT or let the julia build process build its own version. This PR goes with the former, PR #499 goes with the latter. In the end PR #499 seems like the better solution.

boegel:
@tvandera: #499 is on PyQuante?

Also, why would letting Julia build its own BLAS/LAPACK/FFTW be better? It's not easy to do that right, in particular w.r.t. MPI etc. And EasyBuild already knows about how to link to the BLAS/LAPACK/FFTW included in the compiler toolchain being used.

If Julia is highly dependent on BLAS/LAPACK/FFTW, then you should use a full toolchain (e.g. `foss` or `intel`) here rather than `GCC`, possible with extra configure options to make sure Julia picks up the BLAS/LAPACK/FFTW libraries provided by the toolchain.

tvandera:
I agree with previous poster: I'll make this a config linking with existing BLAS/LAPACK/... instead of building its own.

boegel:
@tvandera ping?

tvandera:
Yes. It's still on my stack. Unfortunately, the current EasyConfig we are using works fine for us, but is not suited for general inclusion since we compile BLAS, etc ourselves instead of using the versions from the toolchain...

migueldiascosta:
@tvandera, this PR is being closed for the following reason(s): no activity for > 6 months.

PR #2194, opened 2015-12-08
{*}[intel/2015.08] HPL 2.1 (REVIEW)

boegel:
latest version of the 2015 Intel compilers + latest Intel MPI and MKL

requires ~~#2108~~ #2214

PR #2209, opened 2015-12-10
{*}[intel/2016.00] HPL 2.1 (REVIEW)

boegel:
requires https://github.com/hpcugent/easybuild-easyblocks/issues/751 and https://github.com/hpcugent/easybuild-easyblocks/pull/773

wrapper PR for https://github.com/hpcugent/easybuild-easyconfigs/pull/1973 which includes https://github.com/DirkdeDraak/easybuild-easyconfigs/pull/1, and list the components for icc/ifort to avoid a big fat installation that also includes mkl/tbb/ipp/etc

PR #2214, opened 2015-12-12
{bundle}[dummy] GCC-4.9.3-2.25: bundle of GCCcore 4.9.3 and binutils 2.25

boegel:
requires ~~https://github.com/hpcugent/easybuild-framework/pull/1451~~ and https://github.com/hpcugent/easybuild-easyblocks/pull/773

This sort of replaces #2108, and deals with a coupe of problems there, in particular:

* avoids building GCC twice to construct the `foss` toolchain; the `GCC` is just a bundle of `GCCcore` and `binutils`
* drop `binutils` from versionsuffix of `GCCcore` (since it's only a build dep there)
* drop `-binutils-2.25` as versionsuffix for `binutils-2.25`; it looks stupid, and which binutils was used to build binutils with is most likely irrelevant

As opposed to #2108, these additional easyconfigs do not affect existing toolchain, in particular the `2015b` ones, which are close to 'end of life', and thus should not be fiddled with.

The intention is to use the `GCC/4.9.3-2.25` bundle as a base for the `2016a` toolchains (#2194 will be adjusted to use it, since `intel/2015.08` a likely candidate for `intel/2016a`).

There's a `GCC/4.9.3-binutils-2.25` too, which is an actual GCC build rather than a bundle. This may be confusing right now, but since there's no GCC 4.9 or binutils update, that's the way it is.

@ocaisa, @geimer, @wpoely86: thoughts?

### DIFF
--- a/easybuild/easyconfigs/b/Bison/Bison-3.0.4-GCCcore-4.9.3.eb
+++ b/easybuild/easyconfigs/b/Bison/Bison-3.0.4-GCCcore-4.9.3.eb
@@ -1,0 +1,27 @@
+easyblock = 'ConfigureMake'
+
+name = 'Bison'
+version = '3.0.4'
+
+homepage = 'http://www.gnu.org/software/bison'
+description = """Bison is a general-purpose parser generator that converts an annotated context-free grammar
+ into a deterministic LR or generalized LR (GLR) parser employing LALR(1) parser tables."""
+
+toolchain = {'name': 'GCCcore', 'version': '4.9.3'}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = [GNU_SOURCE]
+
+builddependencies = [
+    ('M4', '1.4.17'),
+    # use same binutils version that was used when building GCCcore toolchain
+    ('binutils', '2.25', '', True),
+]
+
+
+sanity_check_paths = {
+    'files': ["bin/%s" % x for x in ["bison", "yacc"]] + ["lib/liby.a"],
+    'dirs': [],
+}
+
+moduleclass = 'lang'
--- a/easybuild/easyconfigs/b/binutils/binutils-2.25-GCCcore-4.9.3.eb
+++ b/easybuild/easyconfigs/b/binutils/binutils-2.25-GCCcore-4.9.3.eb
@@ -1,0 +1,40 @@
+easyblock = 'ConfigureMake'
+
+name = 'binutils'
+version = '2.25'
+
+homepage = 'http://directory.fsf.org/project/binutils/'
+description = "binutils: GNU binary utilities"
+
+toolchain = {'name': 'GCCcore', 'version': '4.9.3'}
+
+sources = [SOURCE_TAR_GZ]
+source_urls = [GNU_SOURCE]
+
+builddependencies = [
+    ('flex', '2.5.39'),
+    ('Bison', '3.0.4'),
+    # zlib required, but being linked instatically, so not a runtime dep
+    ('zlib', '1.2.8'),
+    # use same binutils version that was used when building GCC toolchain, to 'bootstrap' this binutils
+    ('binutils', version, '', True)
+]
+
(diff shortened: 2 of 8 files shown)


boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/111589ed1a1216c2c8c5 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/54dbc3c7542ea28efb28 for a full test report.

wpoely86:
Looks fine, but preferably (but not possible currently), the `GCCcore` module is hidden (or spits out a warning that it shouldn't be loaded directly

boegel:
Hiding a toolchain isn't possible yet, but we can add a `modloadmsg`?

The problem is, it will always be printed, I'm not sure we can actually detect a 'manual' load...

boegel:
@wpoely86: also, hiding modules is considered a site-specific choice, it shouldn't be hardcoded somewhere...

Lmod will hopefully soon allow to make modules invisible (simply not show them in `module avail` without having to rename them), that's the way to go imho.

wpoely86:
Well, we create a module that never should be loaded directly by the user. That makes hiding it no longer a site specific choice.

boegel:
@wpoely86: well, shouldn't, but they will get a GCC that works, they're just going to rely on the OS-provided binutils in that case.

The current situation is worse, where people should load GNU, rather than GCC.

Anyway, support for hiding toolchains will be added at some point, and I'm willing to have the discussion of hiding `GCCcore` by default then.

ocaisa:
I don't like that you intend to use GCC as a base for iccifort, this means I can't put iccifort and GCC in the 'compiler' family since I need them both loaded at once. Is it not ok to explicitly list GCCcore and bintuils for those cases?
I guess that's a site specific choice but anyone with our case is probably not uncommon (for some software packages at least), we have the same MPI implementation for GCC and iccifort, and in that case we have a reverse race condition...whoever gets loaded last dictates which MPI gets picked up.

boegel:
@ocaisa: you're right, I shouldn't be using `GCC` as a base for icc/ifort, I should be using `GCCcore`, since the whole point of using `GCCcore` is to be able to use it as a base for the `intel` toolchain without having `GCC` around as well... Thanks a lot for bringing that up.
I'll change that in #2194, #2209 and #2219 for the new `intel` toolchains.

Besides that, do you agree that this `GCC` bundle is a better approach than having two full `GCC` builds around for `foss` (which is what you're doing in #2108).

ocaisa:
I'm happy with this. The setup is completely flexible so works in all scenarios. A bundle is fine for the use case here and the setup in the framework still gives me the freedom to build a different GCC version and use that instead...that's my target use case, GCCcore 4.9.3 for all my compilers (Intel, PGI, GCC) with the GCC version being 5.X (configured to use the old C++ ABI be default for compatbility with  software installed using GCCcore).

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/9a597457bd4e4738637a for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/9690b61507d70dd1f752 for a full test report.

boegel:
As agreed with @ocaisa (who created #2108) and @geimer, I'm merging this one rather than #2108, for a number of reasons:

* doesn't affect existing `2015b` toolchains (which are sort of end-of-life)
* avoids building/installing GCC *twice* to install the `foss` toolchain
  * this is overkill since both the `GCC` and `GCCcore` are using the same GCC version
  * it's still possible to build GCC version B on top of GCC version A (typically with B > A), taking into account ABI compatibility issues as discussed in https://github.com/hpcugent/easybuild-easyconfigs/pull/2108#issuecomment-163284578, which is a valid use case
* shorter versionsuffixes

Thanks for the review @wpoely86, and @ocaisa and @geimer for the feedback!

PR #2048, opened 2015-10-12
{cae} [intel-2015b] OpenFOAM 2.4.0 with external dependenices (REVIEW)

hajgato:

PR #2023, opened 2015-10-05
{cae} [intel-2015a] OpenFOAM 2.3.1 with external dependencies (REVIEW) 

hajgato:
Needs PR ~~#1627~~

PR #2226, opened 2015-12-15
rename patch files for OpenFOAM according to (our) 'standards'

boegel:
workaround for https://github.com/hpcugent/easybuild-framework/issues/1516, so #2023 and #2048 can be properly tested via `--from-pr`


wpoely86:
lgtm

boegel:

ocaisa:
Why were the patches not copied rather than renamed? This breaks working eb files for us...and took a while to solve

boegel:
@ocaisa: at some point, we need to clean up mistakes we made in the past and work towards something 'uniform'...

Thanks for bringing this up though. How did the problem manifest itself, and why did it take you a while to solve?

This also goes back to your feature request to make sure everything that is needed to reproduce a build is archived somewhere (potentially including the used easyblock(s)...).

ocaisa:
It wasn't a guru like me who was solving it which was maybe why it took a while :)

They tried to reinstall an existing eb file with EB 2.5 but the patch file could no longer be found.

boegel:
Where did you get the easyconfig file from? The easyconfig archive maintained by EasyBuild itself? Or an existing install directory?

ocaisa:
Custom easyconfig with custom toolchain and configuration (used for a previous installation)

boegel:
In that case, either never removing the patch files or copying them next to the sources could have saved you...

ocaisa:
Yup, exactly

Issue #2188, opened 2015-12-05
Problem building GCC/4.9.3-binutils-2.25


verdurin:
On a F23 system with EB 2.4.0:

```
== building and installing GCC/4.9.3-binutils-2.25...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
 == FAILED: Installation ended unsuccessfully (build directory: /home/adam/.local/easybuild/build/GCC/4.9.3/dummy--binutils-2.25): build failed (first 300 chars): cmd " make -j 4  bootstrap " exited with
exitcode 2 and output:
echo stage3 > stage_final
make[1]: Entering directory '/home/adam/.local/easybuild/build/GCC/4.9.3/dummy--binutils-2.25/gcc-4.9.3/obj'
make[2]: Entering directory '/home/adam/.local/easybuild/build/GCC/4.9.3/dummy--binutils-2.25/gcc-4.9
== Results of the build can be found in the log file /tmp/eb-flSkAK/easybuild-GCC-4.9.3-20151204.233536.pbHgs.log
ERROR: Build of /home/adam/.local/easybuild/software/EasyBuild/2.4.0/lib/python2.7/site-packages/easybuild_easyconfigs-2.4.0-py2.7.egg/easybuild/easyconfigs/g/GCC/GCC-4.9.3-binutils-2.25.eb failed (err: 'build failed (first 300 chars): cmd " make -j 4  bootstrap " exited with exitcode 2 and output:\necho stage3 > stage_final\nmake[1]: Entering directory \'/home/adam/.local/easybuild/build/GCC/4.9.3/dummy--binutils-2.25/gcc-4.9.3/obj\'\nmake[2]: Entering directory \'/home/adam/.local/easybuild/build/GCC/4.9.3/dummy--binutils-2.25/gcc-4.9')
```

There are errors like this in the log:

```
make[1]: Entering directory '/home/adam/.local/easybuild/build/GCC/4.9.3/dummy--binutils-2.25/gcc-4.9.3/obj'
Checking multilib configuration for libgomp...
Checking multilib configuration for libssp...
Checking multilib configuration for libquadmath...
mkdir -p -- x86_64-unknown-linux-gnu/libssp
Configuring in ./fixincludes
mkdir -p -- x86_64-unknown-linux-gnu/libquadmath
Checking multilib configuration for libatomic...
[...]

PR #2228, opened 2015-12-15
Harden dummy:dummy binutils

wpoely86:
By adding the system library path using the `rpath` option, the
resulting binaries should be 'harden' to work in an environment with a
other libstdc++ etc. The rpath will make sure it always finds the
correct library.

Fix for #2188

### DIFF
--- a/easybuild/easyconfigs/b/binutils/binutils-2.25.eb
+++ b/easybuild/easyconfigs/b/binutils/binutils-2.25.eb
@@ -19,8 +19,10 @@ builddependencies = [
 ]
 
 # statically link with zlib, to avoid runtime dependency on zlib
-preconfigopts = 'LIBS="$EBROOTZLIB/lib/libz.a"'
-prebuildopts = 'LIBS="$EBROOTZLIB/lib/libz.a"'
+# further, add the system library path in the rpath: this should 'harden' the
+# resulting binutils to bootstrap GCC (no trouble when other libstdc++ is build etc)
+preconfigopts = 'LIBS="$EBROOTZLIB/lib/libz.a -Wl,-rpath=/lib64 -Wl,-rpath=/usr/lib64 -Wl,-rpath=/lib -Wl,-rpath=/usr/lib"'
+prebuildopts = preconfigopts
 
 # make sure that system libraries are also considered by ld and ld.gold is also built
 # --enable-plugins should be used whenever --enable-gold is used, see http://llvm.org/docs/GoldPlugin.html


wpoely86:
Test report by @wpoely86
**SUCCESS**
See https://gist.github.com/b854fb1a1211da7fcce9 for a full test report.

wpoely86:
Test report by @wpoely86
**SUCCESS**
See https://gist.github.com/a2d490feff7db105af82 for a full test report.

wpoely86:
Verfication by `objdump -x <binary>`:
```
Dynamic Section:
  NEEDED               libdl.so.2
  NEEDED               libc.so.6
  RPATH                /lib64:/usr/lib64:/lib:/usr/lib
```

wpoely86:
Test report by @wpoely86
**SUCCESS**
Linux debian 8.1, Intel(R) Xeon(R) CPU            5130  @ 2.00GHz, Python 2.7.9
See https://gist.github.com/536c3bc22998030be340 for a full test report.

easybuild/easyconfigs/b/binutils/binutils-2.25.eb
```diff
@@ -19,8 +19,10 @@ builddependencies = [
 ]
 
 # statically link with zlib, to avoid runtime dependency on zlib
-preconfigopts = 'LIBS="$EBROOTZLIB/lib/libz.a"'
-prebuildopts = 'LIBS="$EBROOTZLIB/lib/libz.a"'
+# further, add the system library path in the rpath: this should 'harden' the
+# resulting binutils to bootstrap GCC (no trouble when other libstdc++ is build etc)
+preconfigopts = 'LIBS="$EBROOTZLIB/lib/libz.a -Wl,-rpath=/lib64 -Wl,-rpath=/usr/lib64 -Wl,-rpath=/lib -Wl,-rpath=/usr/lib"'
```
boegel: you can't combine all the `rpath` chunks in a single one?

```
-Wl,-rpath=/lib64:/usr/lib64:/lib:/usr/lib"
```

Haven't checked, but I would fine it weird if this is not supported.

This is screaming for support in framework btw. ;-)

boegel:
I'm not sure whether it's a good idea to include something like this so close to the release...
Especially since this is waaaaay down in the stack.

On the other hand, I guess just building this + GCC on top of its is going to tell us whether it causes any problems?

@geimer, @ocaisa: do you think something like this may have fallout in one way or another?

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/e65fef3cbe0db35e570c for a full test report.

boegel:
built on top of this binutils (all easyconfigs that use a dummy-built binutils): https://gist.github.com/ea03dd15f457abefdc3f

easybuild/easyconfigs/b/binutils/binutils-2.25.eb
wpoely86: The shortest form I can find is:
```
-Wl,-rpath=/lib64,-rpath=/usr/lib64,-rpath=/lib,-rpath=/usr/lib
```

wpoely86: For the framework support, follow up in hpcugent/easybuild-framework#51

boegel: @wpoely86: let's leave it as is for now, since this is tested thoroughly, and should be good to merge

We can clean it up later.

I'd love to get feedback from @geimer on this though, before merging it...

boegel:
Just made very sure that this doesn't "trickle up" to stuff build on top of this binutils, it doesn't (seem to):

```
-bash-4.2$ objdump -x $EASYBUILD_PREFIX/software/binutils/2.25/bin/ld| egrep 'Dynamic section|NEEDED|RPATH'
  NEEDED               libdl.so.2
  NEEDED               libc.so.6
  RPATH                /lib64:/usr/lib64:/lib:/usr/lib
-bash-4.2$ objdump -x $EASYBUILD_PREFIX/software/GCC/4.9.2-binutils-2.25/bin/gcc  | egrep 'Dynamic section|NEEDED|RPATH'
  NEEDED               libm.so.6
  NEEDED               libc.so.6
  NEEDED               ld-linux-x86-64.so.2
```

cc @ocaisa

ocaisa:
Given the way the mainline toolchains (foss, intel) are constructed in the next release (with a new binutils built on top of the compiled GCC) I can't see any problem for them. Also since the dummy binutils libs are  not dynamic there's no issue there. I guess the only impact would be if this somehow affected the behaviour of the linker...but from GCC at least that doesn't seem to be the case.

boegel:
I rebuilt more than just GCC on top of this, see https://gist.github.com/ea03dd15f457abefdc3f .

So, should be good to go, I'm merging this in, so @geimer can come back to us later to tell us this is the worst idea ever. ;-)

boegel:
Thanks @wpoely86!

geimer:
This relies on `libstdc++.so` residing in `(/usr)?/lib(64)?`, right?  If so, it may work on Fedora, but may break on Debian and derivatives where `libstdc++.so` lives in `/usr/lib/gcc/x86_64-linux-gnu/<system_gcc_version>`. In any case, rpath'ing system library directories somehow feels wrong to me...

But to be honest, I also can't think of a better alternative at the moment.

wpoely86:
@geimer yes, I'm aware. I've been playing with `-static-libgcc -static-libstdc++` but that relies on the fact that the static libraries are installed. It would be cool if we could add the `rpath` after compiling and just for the libraries that are used.

And, it's indeed quite dirty. But this binutils should only be used once for building GCC and then thrown away.

PR #2222, opened 2015-12-14
added new Cray* toolchain versions with dependency pinning

gppezzi:
This PR adds eb files for the new CrayGNU, CrayIntel and CrayCCE toolchains (with dep. pinning). For validation, it adds also easyconfig files for HPL.

Requires: EB 2.5.0 and [hpcugent/easybuild-easyblocks/pull/766](https://github.com/hpcugent/easybuild-easyblocks/pull/766)

PR #2239, opened 2015-12-17
Added sample HPL easyconfig files for new CrayGNU-2015 TC

gppezzi:


boegel:
@gppezzi just adding these to #2222 would be better imho, since it allows to submit a test report for both the toolchain and HPL on top of it in a single place

gppezzi:
ok, I'll move it there.

PR #2240, opened 2015-12-18
autodetects SSE2 vs Altivec (IBM Power) vs none

JackPerdue:
We need to do what we can to remove SSE stuff where we can. (/me speaking plurally ;)

I was told to put this in an easyblock at some point but, really, my python is lame and my bash is at least readable (and easily updated).

Personally... I think the ConfigureMake easyblock rocks and we should convert all the easyblock.py stuff to .eb files.   It worked really well with Trinity lately: http://www.siliconslick.com/easybuild/ebfiles_repo_cleaned/ada/Trinity/Trinity-2.1.1-intel-2015B.eb

Anyway... with this change:

1) it still configures/compiles the same on our x86_64 systems
2) it attempts to use the altivec feature of Power processors (which is new)

Tear it apart (with kindness please... tis the season to be kind... hope yours is kind to you)

### DIFF
--- a/easybuild/easyconfigs/f/FFTW/FFTW-3.3.4-gompi-2015b.eb
+++ b/easybuild/easyconfigs/f/FFTW/FFTW-3.3.4-gompi-2015b.eb
@@ -17,10 +17,10 @@ preconfigopts = """
     grep sse2 /proc/cpuinfo >& /dev/null
     if [ $? -eq 0 ] ; then 
       echo "Found SSE2"
-      export CLINE1="--enable-single --enable-sse2 --enable-mpi"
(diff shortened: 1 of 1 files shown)


boegel:
With this kind of complexity, yes, we should look into an easyblock instead.
Thanks for sharing and letting us know your Python-fu is too low to tackle this yourself.

It should probably be fairly easy to handle this in an easyblock, but someone will have to tackle it.

Please (re)ping me on this if you don't hear back.

migueldiascosta:
@JackPerdue @boegel is this still relevant?

boegel:
I don't think so, and if it is, is should be handled via the `FFTW` easyblock, so closing

PR #2244, opened 2016-01-03
160101 dylib

citibob:
Mass changes of '.so' to '.%s' % SHLIB_EXT in easyconfigs.  They have been spot checked, and the script to do them (included) was somewhat conservative.


citibob:
yes

fgeorgatos:
sry: I really meant "scripts" subdirectory...

boegel:
Jenkins: ok to test

boegel:
@citibob: as @fgeorgatos already pointed out, the easyconfigs repository is not the right location for scripts like this.
Also, even though I appreciate the effort, it probably reasonably trivial to achieve the same with a single `sed` command?

In addition, I notice you (accidentally?) removed an easyconfig file, i.e. `a2ps-4.14-goalf-1.1.0-no-OFED.eb`.

citibob:
Kenneth,



On Sun, Jan 3, 2016 at 10:29 AM, Kenneth Hoste <notifications@github.com>
wrote:

> @citibob <https://github.com/citibob>: as @fgeorgatos
> <https://github.com/fgeorgatos> already pointed out, the easyconfigs
> repository is not the right location for scripts like this.
>

I'll move it to where they suggested.


> Also, even though I appreciate the effort, it probably reasonably trivial
> to achieve the same with a single sed command?
>

My first draft was a single sed command, but the script is a little more
sophisticated.  The problem is, what happens when you find something like :

     "stuff/directory-%s/lib-%s.so" % (mydir, libname)

If you just use sed naively, you'll end up replacing it garbage like:
     "stuff/directory-%/lib-%s.%s" % SHLIB_EXT % (mydir, libname)

The alternative would be to use '+' instead of %.  But as you pointed out,
that's not idiomatic.  So my python script just avoids messing with the
string if there's already a '%s' in it.  I didn't know how to do that with
straight sed.

> In addition, I notice you (accidentally?) removed an easyconfig file, i.e.
> a2ps-4.14-goalf-1.1.0-no-OFED.eb.
>
Thanks, I'll put it back.

-- Elizabeth

citibob:
Looks like my script has a couple of bugs still.  Working on it...

citibob:
The Jenkins test results have been incorporated into a new iteration, this PR is being closed.

boegel:
replaced with #2245

PR #2245, opened 2016-01-03
dylib fix applied by script.5

citibeth:
I'd like to see how Jenkins feels about this commit.

### DIFF
--- a/easybuild/easyconfigs/a/APR-util/APR-util-1.5.4-foss-2015a.eb
+++ b/easybuild/easyconfigs/a/APR-util/APR-util-1.5.4-foss-2015a.eb
@@ -20,7 +20,7 @@ dependencies = [
 configopts = "--with-apr=$EBROOTAPR/bin/apr-1-config --with-sqlite3=$EBROOTSQLITE --with-expat=$EBROOTEXPAT "
 
 sanity_check_paths = {
-    'files': ["bin/apu-1-config", "lib/libaprutil-1.so", "lib/libaprutil-1.a"],
+    'files': ["bin/apu-1-config", "lib/libaprutil-1.%s" % SHLIB_EXT, "lib/libaprutil-1.a"],
     'dirs': ["include/apr-1"],
 }
 
--- a/easybuild/easyconfigs/a/APR/APR-1.5.2-foss-2015a.eb
+++ b/easybuild/easyconfigs/a/APR/APR-1.5.2-foss-2015a.eb
@@ -12,7 +12,7 @@ source_urls = ['http://archive.apache.org/dist/apr/']
 sources = [SOURCELOWER_TAR_GZ]
 
 sanity_check_paths = {
-    'files': ["bin/apr-1-config", "lib/libapr-1.so", "lib/libapr-1.a"],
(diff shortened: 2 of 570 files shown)


citibeth:
yes

boegel:
Jenkins: ok to test

boegel:
@citibeth: sorry for taking so long to follow up on this

boegel:
I'll give this a detailed look, but it looks fine at first sight.

I opened PR #2298 to enhance the unit tests to check for the use `.so` in easyconfigs.

boegel:
Reviewing a PR this big is an issue (especially since GitHub feels it's too big), but I managed to wrestle through it.
I'll follow-up with a PR to fix some style issues I ran into, but after merging this one.

Thanks for your efforts on this @citibeth!

lgtm

boegel:
Going in, thanks @citibeth!

citibeth:
Kenneth,

Thank you, I really appreciate your efforts on this.  I noticed there are a
lot of PR's pending, and you need to prioritize your time.

The new unit tests for '.so' seem like a good idea.  However, there are a
few corner cases where .so IS appropriate to be seen in an .eb file ---
usually when you're making Python modules.  Without looking at the new unit
test code, I'm wondering if this is taken into account.

Take care,
-- Elizabeth

boegel:
@citibeth: no, it's not taken into account right now... Can you give an example, or provide more details?

Jenkins is now reporting 146 easyconfigs which are still using `.so` in one way or another, see https://jenkins1.ugent.be/job/easybuild-easyconfigs-pr-builder/5566/console.

I'm wondering if you did the changes on top of `develop` or `master`?

Also, how much of these easyconfigs using `.so` can/should be ignored?

boegel:
@citibeth: OK, I ran into an example while handling the remaining usages of `.so` in #2298, see 90d4a80 .
Note that it should be OK to simplify the check a bit, to check for a non-empty `lib/python2.7/site-packages` directory; the `import` check done for Python packages generally already makes sure it's installed as it should, so checking for the `.so` explicitly is overkill; besides, it may also be located in a `.egg` directory or even a zipped `.egg` *file*, so just checking for a non-empty parent directory is easier anyway...

citibeth:
Kenneth,

Regarding remaining .so stuff.... this merge request is only for
automatically applied fixes.  Unlike with the easyblocks, I have not
manually gone through and fixed the remaining stuff, although that would
not be too hard.  Here's what I use to find what remains, I'm getting about
190 instances right now:
   grep '\.so' `find . -name '*.eb'` | grep -v sourceforge | less

Known reasons for these include:
  1. If there was already a '%' in the string, I leave it alone.  This is
to avoid the non-idomatic change of 'lib%s.so' % lib_name --> ('lib%s' +
SHLIB_EXT) % lib_name, preferring a manual solution instead.

  2. Python modules are named .so on Macintosh

  3. The grep statement picks up some false positives.

I'm unlikely to use any of the remaining unfixed .eb files, and am unable
to do justify the work required to fix them at this time.

-- Elizabeth

boegel:
@citibeth: ok, thanks for the info, I'll handle the remaining ones in #2298, where we have an actual check for it in place too

PR #2249, opened 2016-01-04
{tools}[dummy/dummy] hub 2.2.2 (REVIEW)

nathanhaigh:
Easyconfig for installing hub 2.2.2.

### DIFF
--- a/easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
+++ b/easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
@@ -1,0 +1,22 @@
+easyblock = 'PackedBinary'
+
+name = 'hub'
+version = '2.2.2'
+versionsuffix = '-linux-amd64'
+
+homepage = 'https://hub.github.com/'
+description = """hub is a command-line wrapper for git that makes you better at GitHub."""
+
+sources = ['%%(namelower)s-%s-%%(version)s.tgz' % versionsuffix[1:]]
+
+source_urls = ['https://github.com/github/hub/releases/download/v%s/' % version]
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sanity_check_paths = {
+    'files': ['bin/hub'],
+    'dirs': []
+}
+
+
+moduleclass = 'tools'
--- a/easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
+++ b/easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
@@ -7,9 +7,8 @@ versionsuffix = '-linux-amd64'
 homepage = 'https://hub.github.com/'
 description = """hub is a command-line wrapper for git that makes you better at GitHub."""
 
-sources = ['%%(namelower)s-%s-%%(version)s.tgz' % versionsuffix[1:]]
-
-source_urls = ['https://github.com/github/hub/releases/download/v%s/' % version]
+source_urls = ['https://github.com/github/hub/releases/download/v%(version)s/']
+sources = ['%(namelower)s%(versionsuffix)s-%(version)s.tgz']
 
 toolchain = {'name': 'dummy', 'version': 'dummy'}
 
@@ -17,6 +16,5 @@ sanity_check_paths = {
     'files': ['bin/hub'],
     'dirs': []
 }
-
 
 moduleclass = 'tools'
--- a/easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
+++ b/easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
@@ -12,6 +12,8 @@ sources = ['%(namelower)s%(versionsuffix)s-%(version)s.tgz']
 
 toolchain = {'name': 'dummy', 'version': 'dummy'}
 
+postinstallcmds = ['chmod -x %(installdir)s/install']
+
 sanity_check_paths = {
     'files': ['bin/hub'],
     'dirs': []


easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
```diff
@@ -0,0 +1,22 @@
+easyblock = 'PackedBinary'
+
+name = 'hub'
+version = '2.2.2'
+versionsuffix = '-linux-amd64'
+
+homepage = 'https://hub.github.com/'
+description = """hub is a command-line wrapper for git that makes you better at GitHub."""
+
+sources = ['%%(namelower)s-%s-%%(version)s.tgz' % versionsuffix[1:]]
```
boegel: maybe change this to:

```python
source_urls = ['https://github.com/github/hub/releases/download/v%(version)s/']
sources = ['%(namelower)s%(versionsuffix)s-%(version)s.tgz']
```

easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
```diff
@@ -0,0 +1,22 @@
...
+source_urls = ['https://github.com/github/hub/releases/download/v%s/' % version]
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sanity_check_paths = {
+    'files': ['bin/hub'],
+    'dirs': []
+}
+
+
```
boegel: drop the duplicate empty line please

boegel:
@nathanhaigh: have you tried building it from source? We already have easyconfigs for Go, so may be easy?

boegel:
Jenkins: ok to test

nathanhaigh:
I did try building from source, but hub recommends using the binaries. I haven't any experience with Go and only marginally more with EasyBuild! I wasn't sure what EasyBlock to use and if there are existing tools using it which I could learn from.

Happy to submit a PR for an easyconfig building from source if I can get a bit of guidance.

boegel:
@nathanhaigh: ok, no problem, it's just that we try to stay away from binary packages when we can (not sure if it matters a lot for tools written in Go though); let's follow this up in the issue you opened (#2251)

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/15f2a858b0445acb9e55 for a full test report.

boegel:

nathanhaigh:
Hold off merging this for the moment. I'm experiencing some odd behaviour once I've loaded the module and then run `eb` to install a tool.

What I've noticed is that during the `configuring` step of `eb blah.eb`  I get 10's of thousands of hub processes being spawned.

nathanhaigh:
This is what `module show` ouputs:

```
$ module show hub/2.2.2-linux-amd64
-------------------------------------------------------------------------------------------------------------------------
   /mnt/bioinf-7/easybuild/modules/tools/hub/2.2.2-linux-amd64:
-------------------------------------------------------------------------------------------------------------------------
whatis("Description: hub is a command-line wrapper for git that makes you better at GitHub. - Homepage: https://hub.github.co
m/ ")
conflict("hub")
prepend_path("MANPATH","/opt/shared_apps/easybuild/software/hub/2.2.2-linux-amd64/share/man")
prepend_path("PATH","/opt/shared_apps/easybuild/software/hub/2.2.2-linux-amd64/bin")
setenv("EBROOTHUB","/opt/shared_apps/easybuild/software/hub/2.2.2-linux-amd64")
setenv("EBVERSIONHUB","2.2.2")
setenv("EBDEVELHUB","/opt/shared_apps/easybuild/software/hub/2.2.2-linux-amd64/easybuild/hub-2.2.2-linux-amd64-easybuild-deve
l")
prepend_path("PATH","/opt/shared_apps/easybuild/software/hub/2.2.2-linux-amd64")
help([[ hub is a command-line wrapper for git that makes you better at GitHub. - Homepage: https://hub.github.com/
    
]])
```

Notice that last `prepend_path` adds `/opt/shared_apps/easybuild/software/hub/2.2.2-linux-amd64` to the `PATH`. Well, in that directory is an executable called `install` which is confusing `eb` during the `configuring` step. I don't think this should be added to the path.

Other than the `hub` binary under the `bin` directory, there is also `man` pages under `share` and some auto-completion config files under `etc`:

```
$ ls -1 /opt/shared_apps/easybuild/software/hub/2.2.2-linux-amd64
bin
easybuild
etc
install
LICENSE
README.md
share
```

So I'm wondering how best to deal with all of these things.

boegel:
Hmm, for some reason I ended up not hitting the merge button on this on yet, even though I intended to.

The problem you ran into is indeed worrying.

It's caused by the `Binary` easyblock (from which the `PackedBinary` used here) also considering `''` (a path relative to the install directory) for `$PATH`. It probably shouldn't be doing that blindly, but conditionally, so we can make sure the install dir is not prepended to `$PATH` in this case?

nathanhaigh:
The reason this is so bad is that the install script provided by `hub` actually calls `install` which it expects on the PATH. Since the `install` provided by `hub` is on the PATH, we get recursive calls, ugh!

Not adding the install dir to PATH would work - can you make those changes @boegel?

nathanhaigh:
Damn, deleted my branch by mistake - restored it now :)

nathanhaigh:
@boegel the latest commit addresses these issues without the need to modify what gets prepended to `PATH`. Although the latter would be a better solution, at least this easyconfig won't cause these issues now.

boegel:
Jenkins: ok to test

easybuild/easyconfigs/h/hub/hub-2.2.2-linux-amd64.eb
```diff
@@ -0,0 +1,22 @@
...
+homepage = 'https://hub.github.com/'
+description = """hub is a command-line wrapper for git that makes you better at GitHub."""
+
+source_urls = ['https://github.com/github/hub/releases/download/v%(version)s/']
+sources = ['%(namelower)s%(versionsuffix)s-%(version)s.tgz']
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+postinstallcmds = ['chmod -x %(installdir)s/install']
```
boegel: hmm, are you sure this is OK? Why is this script there at all?

nathanhaigh: Yep, here's more info https://github.com/hpcugent/easybuild-easyblocks/issues/801

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/f35de541b5fe44046230 for a full test report.

boegel:
lgtm

boegel:

PR #2253, opened 2016-01-05
{bio}[foss/2015b] Bowtie 1.1.2 (REVIEW)

nathanhaigh:
Easyconfig for installing Bowtie-1.1.2-foss-2015b

### DIFF
--- a/easybuild/easyconfigs/b/Bowtie/Bowtie-1.1.2-foss-2015b.eb
+++ b/easybuild/easyconfigs/b/Bowtie/Bowtie-1.1.2-foss-2015b.eb
@@ -1,0 +1,19 @@
+#Modified from existing version by:
+#Robert Schmidt <rjeschmi@gmail.com>
+#Ottawa Hospital Research Institute - Bioinformatics Team
+name = 'Bowtie'
+version = '1.1.2'
+
+homepage = 'http://bowtie-bio.sourceforge.net/index.shtml'
+description = """Bowtie is an ultrafast, memory-efficient short read aligner.
+It aligns short DNA sequences (reads) to the human genome.
+"""
+
+toolchain = {'name': 'foss', 'version': '2015b'}
+#toolchainopts = {'optarch': True, 'pic': True}
+
+sources = ['%(namelower)s-%(version)s-src.zip']
+source_urls = ['http://download.sourceforge.net/bowtie-bio/']
+
+moduleclass = 'bio'
+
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/b/Bowtie/Bowtie-1.1.2-foss-2015b.eb
```diff
@@ -0,0 +1,19 @@
...
+name = 'Bowtie'
+version = '1.1.2'
+
+homepage = 'http://bowtie-bio.sourceforge.net/index.shtml'
+description = """Bowtie is an ultrafast, memory-efficient short read aligner.
+It aligns short DNA sequences (reads) to the human genome.
+"""
+
+toolchain = {'name': 'foss', 'version': '2015b'}
+#toolchainopts = {'optarch': True, 'pic': True}
```
boegel: any particular reason you commented this out?

enabling `pic` may be important in places where Bowtie is used as a dependency (`optarch` is enabled by default, so that's kind of useless enabling it explicitly)

easybuild/easyconfigs/b/Bowtie/Bowtie-1.1.2-foss-2015b.eb
```diff
@@ -0,0 +1,19 @@
...
+
+sources = ['%(namelower)s-%(version)s-src.zip']
+source_urls = ['http://download.sourceforge.net/bowtie-bio/']
+
+moduleclass = 'bio'
+
```
boegel: drop trailing empty line

boegel:
Jenkins: ok to test

easybuild/easyconfigs/b/Bowtie/Bowtie-1.1.2-foss-2015b.eb
nathanhaigh: Only commented them out due to my ignorance about what they are/do and the fact I don't think I've done this when previously building bowtie from source.

In fact the only thing I usually do is to run `make WITH_TBB=1` to include support for multithreading as per http://bowtie-bio.sourceforge.net/manual.shtml#building-from-source

boegel: Using `WITH_TBB` implies including `tbb` as a dependency, so it's a little bit more complicated...

There is a free version of TBB available though, so it's worth considering.

@nathanhaigh: can you clean up this line so we can get this merged?

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/07f897e35f3a1aeb9241 for a full test report.

boegel:

PR #2282, opened 2016-01-08
{bio}(dummy/dummy) GATK 3.3-0, GATK 3.5 (REVIEW)

nathanhaigh:


easybuild/easyconfigs/g/GATK/GATK-3.3-0-Java-1.8.0_66.eb
```diff
@@ -0,0 +1,42 @@
...
+description = """The Genome Analysis Toolkit or GATK is a software package developed at the Broad Institute 
+ to analyse next-generation resequencing data. The toolkit offers a wide variety of tools,
+ with a primary focus on variant discovery and genotyping as well as strong emphasis on 
+ data quality assurance. Its robust architecture, powerful processing engine and 
+ high-performance computing features make it capable of taking on projects of any size."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+# download manually from http://www.broadinstitute.org/gatk/download
+sources = ['GenomeAnalysisTK-%(version)s-g37228af.tar.bz2']
```
boegel: what's this about? other GATK 3.3.0 easyconfigs don't have this

boegel: I mean, the `g37228af` label in the source filename

boegel:
Jenkins: ok to test

easybuild/easyconfigs/g/GATK/GATK-3.3-0-Java-1.8.0_66.eb
nathanhaigh: It seems that older versions are downloaded with a filename suffix like this one. Only the latest (v3.5) is doenloaded without this suffix. Not sure what happens when the next version is released - v3.5 may also get a similar suffix appended to the download filename.

Although all versions of GATK seem to retern a similar suffix when you request the version from the commandline.

boegel: Ugh, that sucks...

Since the download must be done manually anyway, I would stick to the name without the commit ID in it, and thus require people to rename after downloading in case they're downloading old versions?

boegel:
lgtm

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/188b4a894caf92d06512 for a full test report.

boegel:

PR #2298, opened 2016-01-10
check for use of .so in easyconfigs unit tests

boegel:
Enhancement for unit tests so the effort done in ~~#2245~~
 is not in vain.

This will require that ~~#2245~~ is merged in order to make the unit tests pass.


boegel:
This is clearly old & stale, I'll look into a better way of doing this later...

Issue #2338, opened 2016-01-19
CrayPGI support


jgphpc:
Trying new easyconfig for PGI support => easyconfigs/c/CrayPGI/CrayPGI-5.2.82.eb

This is related to:
* https://github.com/hpcugent/easybuild-framework/issues/1536
* ~~https://github.com/hpcugent/easybuild-framework/pull/1342~~

This is my very first time doing this, apologies in advance if i am mistaking.

boegel:
hpcugent/easybuild-framework#

boegel:
let's follow up in #2927

PR #2340, opened 2016-01-19
pgi support

jgphpc:
Issue #2338


jgphpc:
Will continue in https://github.com/hpcugent/easybuild-easyconfigs/pull/2927

PR #2377, opened 2016-01-27
Grass

valtandor:
A new EasyConfig for GRASS GIS. Requires #3976 in its current form.


valtandor:
Please note the change to the libreadline easyconfig, in which I updated the version of the ncurses dependency from 5.9 to 6.0. I'm not sure if this breaks any rules.

valtandor:
Note that I have updated the pull request by replacing references to ncurses 6.0 with ncurses 5.9, however these changes haven't shown up in the pull request, possibly as a result of GitHub's outage earlier today.

boegel:
Large PRs are a PITA, experience has learned that it takes a lot of effort/time to get them ready to merge... Can you flesh out some of this into one or more additional PRs? The Python easyconfigs (+ deps) is a good candidate at first glance.

boegel:
Jenkins: ok to test

valtandor:
@boegel: I'll endeavour to do that.

boegel:
@valtandor: thanks for looking into splitting off various PRs to make this one more manageable!

For future endeavours like this: don't include the dependencies like CMake, Python, etc. in multiple PRs.
That makes it more labor-intensive to figure out which PRs should be merged first.

If the dependencies are missing but provided through other PRs, Jenkins will tell us what's ready to merge and what not, since it'll merge your branch with current `develop` when (re)testing the PR.

I've been working on getting the PRs merged, but I hit a couple of roadblocks w.r.t. keeping the easyconfigs in sync with existing ones, cfr. #2411 and #2411. I submitted PRs your way to sync them up, all you need to do is review/merge them. :-)

A couple of others need a sync with current `develop` to get a "clear view", i.e. #2418, #2419, #2420, #2421, #2422.
(this sort of 'bookkeeping' would not be needed if you didn't include deps in those PRs, but we'll go with what we have now and get them merged ASAP)

valtandor:
OK. I wasn't aware of what Jenkins would or wouldn't do to help out. Based on this, I'll endeavour to stick with the "one package per PR" rule from now on.

boegel:
Jenkins: test this please

boegel:
GitHub claims merge with `develop` is broken, but it seems to work just fine... @valtandor can you merge https://github.com/valtandor/easybuild-easyconfigs/pull/11 ?

valtandor:
@boegel Done. Thanks for your patience.

boegel:
@valtandor missing dependencies for `GRASS-6.4.4-goolf-1.5.14.eb` (is that easyconfig still relevant?)

```
libpng/1.6.10-goolf-1.5.14, GDAL/1.11.1-goolf-1.5.14, libspatialite/4.1.1-goolf-1.5.14, freetype/2.5.3-goolf-1.5.14, ffmpeg/2.3.3-goolf-1.5.14
```

valtandor:
@boegel Probably not. I've removed it for the time being. I can always put it back in if it seems necessary.

easybuild/easyconfigs/g/GRASS/GRASS-7.0.2-foss-2015a.eb
```diff
@@ -0,0 +1,64 @@
...
+    --without-glw \
+    --without-tcltk \
+    --with-python \
+    --without-wxwidgets \
+    --without-x \
+    --with-png \
+    --with-png-libs=$EBROOTLIBPNG/lib
+"""
+
+prebuildopts = "sed -i -e '/^\s*gui\s*\\\s*$/d' Makefile && rm -rf ./gui && "
```
valtandor: @boegel: It seems Travis has an issue with the `prebuildopts` line:

```
"sed -i -e '/^\\s*gui\\s*\\\\s*$/d' Makefile && rm -rf ./gui && " != "sed -i -e '/^\\s*gui\\s*\\s*$/d' Makefile && rm -rf ./gui && "
```

What's the proper way to handle backslashes and metacharacters?

boegel: @valtandor the tests are failing because the `prebuildopts` value isn't the same after the parsed easyconfig file has been dumped again... this may be a bug in the dumping, but it does indicate that nasty things are going on

What are you trying to do here exactly?

easybuild/easyconfigs/g/GRASS/GRASS-7.0.2-foss-2015a.eb
```diff
@@ -0,0 +1,64 @@
...
+
+sanity_check_paths = {
+    'files': [],
+    'dirs': ["."]
```
boegel: @valtandor this needs work...

easybuild/easyconfigs/g/GRASS/GRASS-7.0.2-foss-2015a.eb
valtandor: @boegel: I got the original easyconfig here from a colleague, and the aim is to produce a CLI-only version of GRASS, i.e., without the "gui" directory or any building therein. So the prebuildopts aims to remove the gui directory from the build and the reference to it in the Makefile.

Is there a more effective, or at least less suspicious, way of doing this?

boegel: I took a look at the Makefile, and the `sed` command aims to remove this from the `DIRS` variable:

```
    gui \ 
```

Since that by itself should be enough, I would drop the `rm -rf gui`?

Also, I would simplify the `sed` a little bit, to:

```python
# remove gui from list of directories
prebuildopts = "sed -i '' '/gui /d' Makefile"
```

This works since `gui<space>` only appears in one place in the `Makefile`, so it should be fine...

boegel:
@valtandor I started working on this (cfr. https://github.com/valtandor/easybuild-easyconfigs/pull/12), but I hit a blocker of which I'm not sur how you got around it:

```
checking whether to use BLAS... yes                                                                                                                                            checking for location of BLAS includes...                                                                                                                                      checking for location of BLAS library...                                                                                                                                       checking for cblas.h... yes                                                                                                                                                    checking for dnrm2_ in -lblas... no                                                                                                                                            checking for dnrm2_ in -lblas... no                                                                                                                                            configure: error: *** Unable to locate BLAS library.   
```

The `configure` script is hard checking with `-lblas`, but this can never work with OpenBLAS (even with also specifying `--with-blas-libs`) since it doesn't provide `libblas.*`.

Did this actually work for you? And if so, do you have `libblas.*` somewhere on the system?

valtandor:
@boegel: My apologies for the slow response. Apparently, we have `/usr/lib64/libblas.*` on our machines. This seems to be installed by means of an RPM and presumably comes with the OS in our case.

I suppose this means that the ideal EB config will not have --with-blas.

valtandor:
@boegel OK, I've uploaded a new version (for GRASS 7.0.2 and GRASS 7.0.4). Also some consequential changes to NetCDF - their nc-config script seemed a little broken and had to be patched to incorporate the actual NetCDF install paths. (This is only needed for GRASS though?) `--with-blas` has also been taken out in the hope that the new EB files will work on your test machines. I don't warrant that they're going to work everywhere, nor that they have the "correct" set of features for the set of use cases needed by EB users, but I can work with you to tweak them further if needed.

boegel:
@valtandor still missing deps for this one, unless I'm missing a PR where this is available already?

```
Irresolvable dependencies encountered: GDAL/2.1.0-foss-2015a
```

boegel:
@valtandor ping?

valtandor:
@boegel Thanks for your patience. An updated GDAL easyconfig has been submitted as #3976.

migueldiascosta:
@valtandor, this PR is being closed for the following reason(s): no activity for > 6 months, obsoleted by more recent PRs.

PR #2385, opened 2016-01-28
Style fixes with autopep8: f

wpoely86:
Generated by:
find -name '*.eb' -print0 | xargs -0 autopep8 --max-line-length=120 -i -v

### DIFF
--- a/easybuild/easyconfigs/f/FIAT/FIAT-1.1-ictce-5.2.0-Python-2.7.3.eb
+++ b/easybuild/easyconfigs/f/FIAT/FIAT-1.1-ictce-5.2.0-Python-2.7.3.eb
@@ -10,7 +10,8 @@ arbitrary order instances of Jacobi-type quadrature rules on the same element sh
 
 toolchain = {'name': 'ictce', 'version': '5.2.0'}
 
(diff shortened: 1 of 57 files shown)


easybuild/easyconfigs/f/FIAT/FIAT-1.1-ictce-5.2.0-Python-2.7.3.eb
```diff
@@ -10,7 +10,8 @@ arbitrary order instances of Jacobi-type quadrature rules on the same element sh
 
 toolchain = {'name': 'ictce', 'version': '5.2.0'}
 
-source_urls = ['https://launchpad.net/%(namelower)s/%(version_major_minor)s.x/release-%(version_major_minor)s/+download/']
+source_urls = [
```
wpoely86: @boegel what do you think about this? We can make an exception for this by appending `#noqa`?

boegel: no, we format it like

```python
source_urls = [
    'https://launchpad.net/%(namelower)s/%(version_major_minor)s.x/release-%(version_major_minor)s/+download/',
]
```

boegel:
merged via #2492, thanks @wpoely86!

PR #2442, opened 2016-02-04
removed fftw from the new Cray* XC toolchains

gppezzi:
To be in sync with https://github.com/hpcugent/easybuild-framework/pull/1585

(now we need to rebuild all apps using XC  to test where we need fftw...)


gppezzi:
I've created a new PR to replace this, since removing fftw doesn't seem to be enough.

PR #2452, opened 2016-02-05
Adding intel compiler 15.6 and create intel-2016c

hapoo:
This commit adds following easyconfigs for
- icc-2015.6
- ifort-2015.6
- imkl-11.2.4
- impi-5.0.3
  and bundles them in intel-2015c

### DIFF
--- a/easybuild/easyconfigs/i/imkl/imkl-11.2.4.223-iimpi-7.6.5-GCC-4.9.3-2.25.eb
+++ b/easybuild/easyconfigs/i/imkl/imkl-11.2.4.223-iimpi-7.6.5-GCC-4.9.3-2.25.eb
@@ -1,0 +1,24 @@
+# This is an easyconfig file for EasyBuild, see http://hpcugent.github.io/easybuild
+
+name = 'imkl'
+version = '11.2.4.223'
+
+homepage = 'http://software.intel.com/en-us/intel-mkl/'
+description = """Intel Math Kernel Library is a library of highly optimized,
+ extensively threaded math routines for science, engineering, and financial
+ applications that require maximum performance. Core math functions include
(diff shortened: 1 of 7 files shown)


easybuild/easyconfigs/i/impi/impi-5.0.3.049-iccifort-2015.6.233-GCC-4.9.3-2.25.eb
```diff
@@ -0,0 +1,24 @@
+# This is an easyconfig file for EasyBuild, see http://hpcugent.github.io/easybuild
+
+name = 'impi'
+version = '5.0.3.049'
```
boegel: why not bump this to 5.1.2.150?

easybuild/easyconfigs/i/imkl/imkl-11.2.4.223-iimpi-7.6.5-GCC-4.9.3-2.25.eb
```diff
@@ -0,0 +1,24 @@
+# This is an easyconfig file for EasyBuild, see http://hpcugent.github.io/easybuild
+
+name = 'imkl'
+version = '11.2.4.223'
```
boegel: why not bump this to 11.3.1.150?

boegel:
Jenkins: ok to test

boegel:
@hapoo Looks good, just a couple of small remarks/questions w.r.t. versioning.

We typically also include an easyconfig file for HPL that uses this toolchain when we define a new toolchain, by means of test case. Would you mind adding that too?

boegel:
@hapoo: ping?

PR #2443, opened 2016-02-05
Add support for PyGTK with intel/2016a toolchain

ocaisa:
I solved this by following the recipes available for [homebrew](https://github.com/Homebrew/homebrew)

PR #2468, opened 2016-02-10
Update naming of GTK+, PyCairo and Gdk-Pixbuf

boegel:
fleshed out from #2443

@ocaisa: I prefer making changes like this in a separate PR (it makes #2443 needlessly complicated, and deserves it's own entry in the release notes)


boegel:
Thanks @ocaisa!

PR #2478, opened 2016-02-12
{tools} [dummy] git-lfs-1.1.1 (REVIEW)

pescobar:


easybuild/easyconfigs/g/git-lfs/git-lfs-1.1.1.eb
```diff
@@ -0,0 +1,34 @@
...
+
+homepage = 'https://git-lfs.github.com/'
+description = """Git Large File Storage (LFS) replaces large files such as audio samples, videos, 
+ datasets, and graphics with text pointers inside Git, while storing the file contents on a remote server 
+ like GitHub.com or GitHub Enterprise."""
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+source_urls = ['https://github.com/github/git-lfs/releases/download/v%(version)s/']
+sources = ['%(name)s-linux-amd64-%(version)s.tar.gz']
```
boegel: why binary?

boegel: oh, it's Go...

easybuild/easyconfigs/g/git-lfs/git-lfs-1.1.1.eb
```diff
@@ -0,0 +1,34 @@
...
+    'dirs': [""]
+}
+
+# add the installation dir to PATH
+modextrapaths = {
+      'PATH': '',
+      }
+
+# delete to provide installation script
+postinstallcmds = ["rm -fr %(installdir)s/install.sh"]
```
boegel: why is this being removed?

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/845372829c2363eddd3e for a full test report.

easybuild/easyconfigs/g/git-lfs/git-lfs-1.1.1.eb
pescobar: the install.sh script has execution permission so when adding the install_dir to PATH it would be accessible for users and I prefer to avoid it

Also, this is the content of this script which tries to install to /usr/local 

```
#!/usr/bin/env bash
set -eu

prefix="/usr/local"

if [ "${PREFIX:-}" != "" ] ; then
  prefix=${PREFIX:-}
elif [ "${BOXEN_HOME:-}" != "" ] ; then
  prefix=${BOXEN_HOME:-}
fi

mkdir -p $prefix/bin

rm -rf $prefix/bin/git-lfs*
for g in git*; do
  install $g "$prefix/bin/$g"
done

PATH+=:$prefix/bin
git lfs install
```

I was thinking about adding a ```modloadmsg``` line like 'to initialize git-lfs execute git lfs install. This is only needed once' but I decided not to add it. Do you think it's worth?

boegel: How about just removing execution permissions on the script instead?

```python
# users should not run this script
postinstallcmds = ["chmod og-x %(installdir)s/install.sh"]
```

boegel: @pescobar ping on `chmod`?

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/ba0e2238270727b534339855ea66123f for a full test report.

boegel:
lgtm

boegel:
Going in, thanks @pescobar!

PR #2399, opened 2016-01-29
Easyconfig files adapted to use the new MPICH easyblocks

damianam:
See https://github.com/hpcugent/easybuild-easyconfigs/issues/2334 and https://github.com/hpcugent/easybuild-easyblocks/pull/818

PR #2505, opened 2016-02-17
Updated mpich derivatives easyconfigs

damianam:
As suggested in https://github.com/hpcugent/easybuild-easyconfigs/pull/2399#issuecomment-184796911

MVAPICH2 easyconfigs don't need to be updated, as all the changes happen transparently in the easyblock


boegel:
Jenkins: ok to test

easybuild/easyconfigs/m/MPICH/MPICH-3.2-GCC-4.9.3.eb
```diff
@@ -0,0 +1,19 @@
...
+toolchain = {'name': 'GCC', 'version': '4.9.3'}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = ['http://www.mpich.org/static/tarballs/%(version)s']
+
+sanity_check_paths = {
+    'files': ['bin/mpiexec', 'bin/mpirun',
+              'include/mpi.h', 'include/mpif.h'],
+    'dirs': [],
+}
```
boegel: this can be removed too? is done by easyblock?

easybuild/easyconfigs/m/MPICH2/MPICH2-1.1-GCC-4.8.1.eb
```diff
@@ -12,9 +12,6 @@ toolchain = {'name': 'GCC', 'version': '4.8.1'}
 sources = ['mpich2-%(version)s.tar.gz']
 source_urls = ['http://www.mpich.org/static/tarballs/%(version)s']
 
-# MPICH configure wants F90/F90FLAGS to be renamed to FC/FCFLAGS.
-preconfigopts = 'export FC="$F90"; export FCFLAGS="$F90FLAGS"; unset F90; unset F90FLAGS; '
-
 sanity_check_paths = {
```
boegel: also remove?

easybuild/easyconfigs/m/MPICH/MPICH-3.2-GCC-4.9.3.eb
damianam: It is partly done by the easyblock. Not all the files are checked there, since not all the derivatives generate all the files. Example: PSMPI doesn't generate mpiexec and mpirun. The headers are currently not checked in the easyblock, but I guess it makes sense to add them.

boegel: ok, but this is not a derivative?

And for the derivates, should we support `custom_sanity_check_paths`?

damianam: Right, this is not a derivative, but the MPICH easyblock, that is the base class for `EB_MVAPICH2`, has by default the common ones. Whatever is not common (like `mpiexec`), goes in the `sanity_check_paths` of the easyconfig.

I don't think we need `custom_sanity_check_paths`, whatever is in `sanity_check_paths` is transparently added to the common checks.

damianam:
@boegel I am closing this PR also (why there were 2?) and I'll make a new one with the changes needed for https://github.com/hpcugent/easybuild-easyblocks/pull/844

boegel:
@damianam the MPICH easyconfigs are updated in #2589, but not the `psmpi` ones (for which you're working on an easyblock, I think?)

boegel:
@damianam as to why there are two PRs: you probably merged #2399 in here accidentally at some point?

PR #2526, opened 2016-02-18
{bio} [foss, 2015b] VCFtools-0.1.14

verdurin:
Update VCFtools to 0.1.14, for foss-2015b

Upstream has moved to GitHub and is now using autotools.

depends on #2531


boegel:
Jenkins: ok to test

boegel:
@verdurin: missing dep?

```
Irresolvable dependencies encountered: Perl/5.16.3-foss-2015b
```

easybuild/easyconfigs/v/VCFtools/VCFtools-0.1.14-foss-2015b.eb
```diff
@@ -0,0 +1,40 @@
...
+ genetic variation data in the form of VCF files."""
+
+toolchain = {'name': 'foss', 'version': '2015b'}
+toolchainopts = {'optarch': True}
+
+source_urls = ['https://github.com/%(namelower)s/%(namelower)s/releases/download/v%(version)s/']
+sources = ['%(namelower)s-%(version)s.tar.gz']
+
+perl = 'Perl'
+perlver = '5.16.3'
```
boegel: bump to 5.20.2, easyconfig is already available?

verdurin: Yes, looking at it now. 

I do think we need better documentation of these things. It's hard to tell when you didn't make the original decisions why particular EBs have particular deps like this.

boegel: Well, since we're also updating `$PERL5LIB`, it's because Perl scripts are included in this pkg?

verdurin: Yes, they are, but I meant the comment more generally.

There's no record in the EB itself why e.g. this or that version of Perl or Python has been chosen.

boegel:
@verdurin: did you intend to include `R-bundle-Bioconductor` in here?

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/0aff5e26e247878bdd75 for a full test report.

easybuild/easyconfigs/v/VCFtools/VCFtools-0.1.14-foss-2015b.eb
```diff
@@ -0,0 +1,41 @@
...
+toolchainopts = {'optarch': True}
+
+source_urls = ['https://github.com/%(namelower)s/%(namelower)s/releases/download/v%(version)s/']
+sources = ['%(namelower)s-%(version)s.tar.gz']
+
+perl = 'Perl'
+perlver = '5.20.2'
+perlsuffix = '-bare'
+dependencies = [
+    (perl, perlver, perlsuffix),
```
boegel: @verdurin: are you sure a bare Perl install is sufficient?

since it's a dependency, why not include the full Perl install (so this can be combined with other Perl tools that require non-standard Perl modules to be available)?

Also, why no versionsuffix anymore?

verdurin: @boegel - this is what I was getting at in the earlier comment.

It's not clear to me what the policy is, if any, of when we need to be explicit about these suffixes and what the basis is for choosing them in the first place.

Maybe it's just because I'm used to RPM spec files.

In this case, I didn't want to add yet another large .eb for my small one (there's no non-bare recent Perl for this toolchain), but I suppose that is the cleanest approach.

boegel: We typically try to use the latest version for dependencies too.

The other thing to keep in mind is that the modules that get installed may need to be used together with other modules. So, some effort to try and avoid possible conflicts is good.

It's impossible to avoid all conflicts though, but not using a `-bare` version gives more options.
We really only use `-bare` versions as a build dependency (since a full-blown install with extensions is typically not needed then).

boegel:
Jenkins: test this please

boegel:
@verdurin: easyconfig should be renamed to `VCFtools-0.1.14-foss-2015b-Perl-5.20.3.eb` (cfr. versionsuffix)

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/c91919736d973d3178c5 for a full test report.

boegel:
Going in, thanks @verdurin!

PR #2530, opened 2016-02-19
fix determining list of easyconfigs, don't assume locations are correct

boegel:


test/easyconfigs/easyconfigs.py
```diff
@@ -338,19 +338,21 @@ def template_easyconfig_test(self, spec):
 
 def suite():
     """Return all easyblock initialisation tests."""
```
fgeorgatos: easyblock really?

PR #2544, opened 2016-02-22
{bio}[foss/2015b] NextClip (REVIEW)

nathanhaigh:


nathanhaigh:
The script directory contains Perl scripts - should Perl then be added as a dependency?

boegel:
Jenkins: ok to test

boegel:
@nathanhaigh: I'd say that depends on how important those scripts are... Are they considered to be a main part of this software, or just helper scripts that happen to be included?

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/e18fdb5dfaf3cb6b70ca for a full test report.

boegel:
@nathanhaigh: let me know what you'll do w.r.t. the Perl dep, otherwise good to go

nathanhaigh:
I'd say they were more like helper scripts. On a slightly closer loo, there are also R scripts there as well as Perl scripts.

boegel:
Changing the script to use `/usr/bin/env` would be good indeed...

easybuild/easyconfigs/n/NextClip/NextClip_scripts_shebang.patch
```diff
@@ -0,0 +1,60 @@
...
+ 
+ # Script:  nextclip_make_report.pl
+ # Purpose: Produce PDF report of NextClip analysis
+diff --git a/scripts/nextclip_plot_graphs.sh b/scripts/nextclip_plot_graphs.sh
+index f6a3a2b..d4b809e 100755
+--- a/scripts/nextclip_plot_graphs.sh
++++ b/scripts/nextclip_plot_graphs.sh
+@@ -1,4 +1,4 @@
+-#!/bin/bash
++#!/usr/bin/env bash
```
boegel: hmm, I think this is very uncommon?

easybuild/easyconfigs/n/NextClip/NextClip_scripts_shebang.patch
```diff
@@ -0,0 +1,60 @@
+diff --git a/scripts/index_bwa.sh b/scripts/index_bwa.sh
+index 2862e42..472d7f1 100755
+--- a/scripts/index_bwa.sh
++++ b/scripts/index_bwa.sh
+@@ -1,4 +1,4 @@
+-#!/bin/bash
++#!/usr/bin/env bash
```
boegel: very uncommon, just keep `/bin/bash`?

easybuild/easyconfigs/n/NextClip/NextClip_scripts_shebang.patch
```diff
@@ -0,0 +1,60 @@
+diff --git a/scripts/index_bwa.sh b/scripts/index_bwa.sh
```
boegel: @nathanhaigh: please add a line of comment above this line describing what this patch does, and add yourself as patch author


```
Fix hardcoding of `/usr/bin/perl`, use `perl` available through `$PATH`
author: Nathan Haigh (<affiliation>)
```

easybuild/easyconfigs/n/NextClip/NextClip_scripts_shebang.patch
```diff
@@ -0,0 +1,60 @@
...
+ 
+ source bwa-0.6.1
+ 
+diff --git a/scripts/nextclip_index_reference.pl b/scripts/nextclip_index_reference.pl
+index 03810b3..b7f7649 100755
+--- a/scripts/nextclip_index_reference.pl
++++ b/scripts/nextclip_index_reference.pl
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl -w
++#!/usr/bin/env perl -w
```
boegel: Does this still work? I'm not sure if passing options to `perl` in a shebang line is possible...

fgeorgatos: yes, i've seen this working well - it needs more confirmation about it working *everywhere* though

nathanhaigh: Actually, this doesn't work unless `perl -w` is quoted. Otherwise you get:

    /usr/bin/env: perl -w: No such file or directory

A closer look at those scripts shows they all us the `use warnings;` pragma. So, I'll strip this `-w` flag using the patch.

easybuild/easyconfigs/n/NextClip/NextClip_scripts_shebang.patch
nathanhaigh: It may be uncommon but I was aiming for consistent shebang line usage of `env`.

boegel: well, ok, but we don't build our own bash?

I would really prefer keeping `/bin/bash` here (and below), I don't think consistency should overrule common practice (with the exception of `/usr/bin/perl` being wrong for technical reasons)

easybuild/easyconfigs/n/NextClip/NextClip_scripts_shebang.patch
```diff
@@ -0,0 +1,51 @@
+Fix hardcoding of /usr/bin/perl, use perl available through $PATH
+Fix hardcoding of /bin/bash, use bash available through $PATH
```
boegel: drop this comment if you revert those parts of the patch?

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/71e965091f9fa6bdecc9 for a full test report.

boegel:
Thanks @nathanhaigh!

PR #2551, opened 2016-02-23
{lib} [goolf-1.7.20]  libxml2-2.9.3 (REVIEW)

pescobar:
I had to modify the default libxml2 easyconfig to include this in the configopts

```
--with-zlib=$EBROOTZLIB
```

before adding ``` --with-zlib=$EBROOTZLIB``` the libxml2 build procedure was using the system zlib (I have zlib-devel rpm installed) and then I was getting weird errors like this one compiling another application which uses libxml2 module. I loaded libxml2 module which loads zlib-1.2.8 as dependency but libxml2.so looks for system zlib

```
/scicore/soft/apps/libxml2/2.9.3-goolf-1.7.20/lib/libxml2.so: undefined reference to `gzopen64@ZLIB_1.2.3.3'
```

I think this fix should be added in every libxml2 easyconfig to avoid issues in systems where zlib-devel is installed

### DIFF
--- a/easybuild/easyconfigs/l/libxml2/libxml2-2.9.3-goolf-1.7.20.eb
+++ b/easybuild/easyconfigs/l/libxml2/libxml2-2.9.3-goolf-1.7.20.eb
@@ -1,0 +1,29 @@
+easyblock = 'ConfigureMake'
+
+name = 'libxml2'
+version = '2.9.3'
+
+homepage = 'http://xmlsoft.org/'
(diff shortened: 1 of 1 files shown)


boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/3447b17e82cec05ea0f0 for a full test report.

boegel:
lgtm, thanks @pescobar 

syncing up existing libxml2 easyconfigs is done in #2555

PR #2561, opened 2016-02-24
fix pcre config options. Required for future R lang releases

pescobar:


easybuild/easyconfigs/p/PCRE/PCRE-8.38-intel-2016a.eb
```diff
@@ -15,6 +15,6 @@ toolchainopts = {'optarch': True, 'pic': True}
 source_urls = [SOURCEFORGE_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
 
-configopts = "--with-pic --disable-cpp --enable-utf8 --enable-unicode-properties"
+configopts = "--with-pic --disable-cpp --enable-utf --enable-unicode-properties"
```
boegel: should enable both utf and utf8?

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 12 out of 12 (12 easyconfigs in this PR)
See https://gist.github.com/f9c9af75e01e9b0bf113 for a full test report.

easybuild/easyconfigs/p/PCRE/PCRE-8.38-intel-2016a.eb
pescobar: ```--enable-utf8``` is not needed. this is the output of ```./configure --help```

```
  --enable-utf8           another name for --enable-utf. Kept only for
                          compatibility reasons
```

pescobar:
for the record...this fix is needed because when compiling R-devel version and loading the PCRE module in easybuild compiled without ```--enable-utf``` then you get this error:

```
checking if PCRE version >= 8.10, < 10.0 and has UTF-8 support... no
checking whether PCRE support suffices... configure: error: pcre >= 8.10 library and headers are required
```

when recompiling PCRE with the ```--enable-utf``` option this issue is fixed. 

In previous R release PCRE was included in the R tarball so PCRE was not a required dependency in the R easyconfig but in coming R version PCRE has to be included as a dependency because R tarball no longer includes it

I did the tests with this R tarball: https://stat.ethz.ch/R/daily/R-devel_2016-02-23.tar.gz

pescobar:
after recompiling PCRE with ``` --enable-utf --enable-unicode-properties ``` it works.

```
checking if PCRE version >= 8.10, < 10.0 and has UTF-8 support... yes
checking if PCRE version >= 8.32... yes
checking whether PCRE support suffices... yes
```

easybuild/easyconfigs/p/PCRE/PCRE-8.38-intel-2016a.eb
boegel: ah, ok, great

boegel:
This is good to go imho, thanks @pescobar!

PR #2549, opened 2016-02-23
{vis}[intel-2016b] libX11 1.6.3 (REVIEW)

hajgato:
The following packages were moved to builddeps:
`xcb-proto`, `xextproto`, `inputproto`, `kbproto`, and `ibpthread-stubs`

- `xcb-proto` is one pkconfig file, a bunch of Python files and a couple of xml files.
- `xextproto` is one pkconfig file, a bunch of header files, and a couple of xml files.
- `inputproto` is one pkconfig file, and a couple of header files. 
- `kbproto` is  one pkconfig file, a couple of header files, and a bunch of xml and svg files.
- `libpthread-stubs` is one pkconfig file.

We have to check this new configuration on a completely X-less machine, and we have to check softwares depend on libX11, they might need (some of) the above modules as builddependencies.

Only `xcp-proto` depends on Python, and as it is now only a builddependency, `libX11` is finally Python free!

PR #2562, opened 2016-02-24
{vis}[intel-2016b] libXdmcp 1.1.2 (REVIEW)

hajgato:
Depends on #2549 (xproto)


boegel:
@hajgato: why close?

hajgato:
@boegel: moved to #2549 because libxcb can use libXdmcp

PR #2589, opened 2016-02-29
Update easyconfig files for MPICH and MVAPICH2 and their toolchains

besserox:
I would like to have an equivalent of the foss toolchain but with MVAPICH2 instead of OpenMPI.

- I have updated the MVAPICH2 easyblock to be able to compile the latest versions of MPVACH2 (cf https://github.com/hpcugent/easybuild-easyblocks/pull/844)
- I'm proposing a new toolchain named **foss_mv2**, same as **foss** except that OpenMPI is replaced by MVAPICH2. Feel free to propose another name that better suits you.
- I have started discussing with MVAPICH2 team so they will keep their source tarballs available online in the future. I'd like to get this pull request merged first to demonstrate them how all this works.


boegel:
Jenkins: ok to test

easybuild/easyconfigs/f/foss_mv2/foss_mv2-2016a.eb
```diff
@@ -0,0 +1,35 @@
+easyblock = 'Toolchain'
+
+name = 'foss_mv2'
```
boegel: you're reinventing the existing `gmvolf` toolchain :)

the `foss` toolchain deliberately is not composed of the first letter of every component (like `goolf`, `gmvolf`, ...), to retain the freedom of changing some components at some point in time (e.g. change `OpenMPI` to `MVAPICH2` in future release).

So, maybe stick to `gmvolf` as naming?

easybuild/easyconfigs/f/foss_mv2/foss_mv2-2016a.eb
```diff
@@ -0,0 +1,35 @@
...
+version = '2016a'
```
boegel: the `<year>[ab]` naming is kind of reserved for the "common toolchains" (`foss` and `intel`) which we redefine twice a year (Jan/Jul)

since we now have a pretty meaningless versioning of `gmvolf`, maybe we can switch to `<year>.<month>` there too, like we do for additional definitions of `intel` and `foss` that are not considered 'common'

This is not perfect either though, since you could easily define two variants of `gmvolf` in the same month (one using GCC v4, the other using GCC v5 for example)...

I'm open to suggestions here...

boegel:
@besserox: thanks for reaching out to the MVAPICH2 maintainers to keep the source tarballs for older releases available, this has been pretty annoying...

easybuild/easyconfigs/f/foss_mv2/foss_mv2-2016a.eb
besserox: Sure, it can stay `gmvolf`.

easybuild/easyconfigs/f/foss_mv2/foss_mv2-2016a.eb
besserox: My idea here is to (try to) keep the same software/version of all the components of `foss` (except OpenMPI being replaced by MVAPICH2 of course). That why I believe it would make sense to use the same `<year>[ab]` version numbering.

boegel:
I tested the MVAPICH2 easyconfigs included in here on top of https://github.com/hpcugent/easybuild-easyblocks/pull/844 (with https://github.com/besserox/easybuild-easyblocks/pull/2 included), works like a charm.

boegel:
@besserox: yes, the easyblock not being merged is the problem with the failing tests (since there's no MPICH easyblock yet)

besserox:
Should we also update the easyconfig file `m/MPICH2/MPICH2-1.1-GCC-4.8.1.eb`?

boegel:
@besserox: let's update MPICH2 in a separate PR?

boegel:
Test report by @boegel
**SUCCESS**
Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/fd1312097e0dc93e1499 for a full test report.

boegel:
Jenkins: test this please

easybuild/easyconfigs/g/gmpolf/gmpolf-2016a.eb
```diff
@@ -0,0 +1,35 @@
...
+comp_mpi_tc_name = 'gmpich'
+comp_mpi_tc = (comp_mpi_tc_name, version)
+
+# compiler toolchain depencies
+# we need GCC and OpenMPI as explicit dependencies instead of gompi toolchain
+# because of toolchain preperation functions
+# For binutils, stick to http://wiki.osdev.org/Cross-Compiler_Successful_Builds
+dependencies = [
+    ('GCC', gccver),
+    ('MVAPICH2', '2.2b', '', ('GCC', gccver)),
```
boegel: @besserox this should be `MPICH`?

easybuild/easyconfigs/g/gmvapich2/gmvapich2-2016a.eb
```diff
@@ -0,0 +1,25 @@
...
+ including MVAPICH2 for MPI support."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+compname = 'GCC'
+compver = '4.9.3-2.25'
+comp = (compname, compver)
+
+mpilib = 'MVAPICH2'
+mpiver = '2.2b'
```
boegel: these are used in only one place, so just drop `mpilib` and `mpiver`, and include directly in `dependencies`?

boegel:
@besserox we usually include `HPL` easyconfigs for new toolchains, by means of testcase... Can you add an `HPL` easyconfig for both the new `gmpolf` and `gmvolf` toolchain definitions?

besserox:
@boegel Fixed

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 16 out of 16 (16 easyconfigs in this PR)
See https://gist.github.com/3679705cf3afbc4145ec for a full test report.

boegel:
Good to go, thanks @besserox!

PR #2590, opened 2016-02-29
add easyconfig for Moses (WIP)

boegel:
requires https://github.com/hpcugent/easybuild-easyblocks/pull/845


boegel:
Old & stale, using deprecated toolchain, so closing...

PR #2593, opened 2016-02-29
add easyconfigs for Xmipp 20150615 (WIP)

boegel:


boegel:
Closing this since it's using an ancient toolchain version, will pick this up again with a newer toolchain if the need arises.

PR #2594, opened 2016-02-29
add easyconfigs for Scipion (WIP)

boegel:
depends on #2593 
requires https://github.com/hpcugent/easybuild-easyblocks/pull/846


boegel:
Closing this since it's using an ancient toolchain version, will pick this up again with a newer toolchain if the need arises.

PR #2600, opened 2016-03-01
Ignore log files created by easybuild via gc3pie

nathanhaigh:
Installing software using gc3pie generated `eb-*.log` files; let git ignore these files.


boegel:
Jenkins: ok to test

boegel:
OK to include since we shouldn't ever commit `*.log` files in this repo.

PR #2612, opened 2016-03-02
Update files that have dummy as toolchain, are not a toolchain themse…

ocaisa:
…lves and have dependencies. Without an empty version the dependencies do not get loaded at build time.

To get the list of files:
```
for f in `grep -r "^toolchain.*version.*dummy" easybuild | cut -f1 -d:`; do grep "dependencies[ ]*=" $f > /dev/null; if [ $? -eq 0 ]; then echo $f; fi; done
```
The result can be cleaned of toolchains. To make the changes:
```
cat cleaned_list |xargs -i sed -i 's/'\''dummy'\''}/'\'''\''}/' {}
```

### DIFF
--- a/easybuild/easyconfigs/a/ABAQUS/ABAQUS-6.12.1-linux-x86_64.eb
+++ b/easybuild/easyconfigs/a/ABAQUS/ABAQUS-6.12.1-linux-x86_64.eb
@@ -5,7 +5,7 @@ versionsuffix = '-linux-x86_64'
 homepage = 'http://www.simulia.com/products/abaqus_fea.html'
 description = "Finite Element Analysis software for modeling, visualization and best-in-class implicit and explicit dynamics FEA."
 
-toolchain = {'name': 'dummy', 'version': 'dummy'}
+toolchain = {'name': 'dummy', 'version': ''}
 
 sources = [SOURCE_TGZ]
 
--- a/easybuild/easyconfigs/a/ABAQUS/ABAQUS-6.13.5-linux-x86_64.eb
+++ b/easybuild/easyconfigs/a/ABAQUS/ABAQUS-6.13.5-linux-x86_64.eb
@@ -5,7 +5,7 @@ versionsuffix = '-linux-x86_64'
 homepage = 'http://www.simulia.com/products/abaqus_fea.html'
 description = "Finite Element Analysis software for modeling, visualization and best-in-class implicit and explicit dynamics FEA."
 
-toolchain = {'name': 'dummy', 'version': 'dummy'}
+toolchain = {'name': 'dummy', 'version': ''}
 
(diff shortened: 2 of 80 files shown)


boegel:
Test report by @boegel
**FAILED**
Build succeeded for 76 out of 80 (80 easyconfigs in this PR)
node2059.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/b16e7d9d2255d7192303 for a full test report.

boegel:
@ocaisa:

* the MATLAB fails are not caused by the change made here, but because the required license key is missing (I didn't provide it when running the test to avoid that it leaks into the test report)
* the fail for `picard-2.1.0.eb` also occurs with `develop`, and is thus not caused by the change being made here
* the fail for `imkl-10.2.6.038-32bit.eb` is likely related to the change, but I don't consider this a blocker to merge this in since it's a very old version of `imkl` that is not used anywhere in a toolchain; I'll deal with that one in a separate PR

So, going in, thanks for looking into this!

boegel:
confirmed that the MATLAB easyconfigs were *not* broken

follow-up on picard issue in https://github.com/hpcugent/easybuild-easyconfigs/pull/2572#issuecomment-191398549, installation fails because https://github.com/hpcugent/easybuild-easyblocks/pull/796 is not merged yet

boegel:
I looked into the problem with imkl, it's a nice one...

The problem is that the `install.sh` script doesn't deal well when the build directory contains a directory with `--` in it. And that's exactly what we get when installing `imkl` with `dummy` toolchain, using `''` as version *and* using `-32bit` as a `versionsuffix`; these things combined leads to `dummy--32bit` as one of the directories part of the build dir.

The `install.sh` script splits on `--` at some point, with the intention to split on a long option name like `--nonrpm`; but this breaks because `--` is also somewhere else.

Oh boy... :)

boegel:
This patch dances around the issue:

```diff
--- pset/install.sh.orig	2016-03-02 22:29:20.049306187 +0100
+++ pset/install.sh	2016-03-02 22:28:15.257923543 +0100
@@ -1281,7 +1281,7 @@
 	
 	[ $err -eq ${ERR_OK} ] || return 1
 		
-        RS=$(echo $CMD_STR | sed s/.*--$cmd[[:blank:]]*//g | sed 's/[[:blank:]]*--.*$//g')
+        RS=$(echo $CMD_STR | sed s/.*--$cmd[[:blank:]]*//g | sed 's/[[:blank:]]*--[^3].*$//g')
         [[ -z "$RS" ]] && return 1
         echo $RS
 	
```

boegel:
imkl issue fixed in #2634

PR #2621, opened 2016-03-03
{lang}[intel-2016a] Python 2.7.11 with libX11 1.6.3 (REVIEW)

hajgato:
Needs #2614

PR #2624, opened 2016-03-03
{lang}[intel-2016a] R 3.2.3 with libX11 1.6.3 (REVIEW)

hajgato:
Needs #2621


easybuild/easyconfigs/r/R/R-3.2.3-intel-2016a-libX11-1.6.3.eb
```diff
@@ -0,0 +1,453 @@
...
+    ('ncurses', '6.0'),
+    ('libpng', '1.6.21'),  # for plotting in R
+    ('libjpeg-turbo', '1.4.2'),  # for plottting in R
+    ('LibTIFF', '4.0.6'), # for plotting in R
+    ('cairo', '1.14.6', versionsuffix), # for plotting in R
+    ('Java', '1.8.0_72', '', True),  # Java bindings are built if Java is found, might as well provide it
+    ('Tcl', '8.6.4'),  # for tcltk
+    ('Tk', '8.6.4', versionsuffix),  # for tcltk
+    ('cURL', '7.47.0'),  # for RCurl
+    ('libxml2', '2.9.3', '-Python-2.7.11%s' % versionsuffix),  # for XML
```
boegel: does this need libxml *with* Python bindings enabled?

if we can get around including Python as an (indirect) dep for R, I think we should?

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 5 (5 easyconfigs in this PR)
node2634.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/f851233238f58296fae4 for a full test report.

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 5 (5 easyconfigs in this PR)
node2044.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/e98b3723685a483a474d for a full test report.

easybuild/easyconfigs/r/R/R-3.2.3-intel-2016a-libX11-1.6.3.eb
hajgato: @boegel Not, but `cairo` have a `GLib` dependency, and `GLib` have a `libxml2-Python` dependency. So all the chain should have `libxml2-Python` dependency to avoid conflicts (or we have to delete the `Python` dependency of the existing `GLib` easyconfig.

hajgato: We should have a "dontload@build" option in case of Dependencies, so that dep is not loaded during the build. I mean, in case of `GLlib` `Python` is only a builddep, but we have to use `libxml2` with `Python` capabilities.  `libxml2` with `Python` could be builddep, and normal `libxml2` could be the dep, with "dontload@build" 
This should help us to easier handle circular dependencies.

boegel: I don't think using two different libxml2 builds as build dep/runtime dep is a good idea...

Who knows what happens during the build, it may remember where the libxml2 it was built on top of is installed.

Why do we have to use libxml2 w/ Python in GLib? And how can Python only be a build dep then?

hajgato: @boegel: I am not able to answer to that question. Most of the `GLib` easyconfigs have `Python` as builddep, and normal `libxml2` as dep. The latest version (2.47.5) has a 'libxml2-python` dep.
@ocaisa put it there, do not know the reason. (maybe needed only for testing?)

hajgato: maybe `libxml2` should also be a builddep.

ocaisa: With the latest libxml easyblock there's no way to turn off the Python dependency, it's required to build the bindings. I don't know if the bindings are used, you'd have to check the configure output of glibc, if not then we'd need to build a switch into the libxml easyblock

hajgato: checking the `GLib` configure.log, only `xmlcatalog` needed. And Indeed, `Python` must be there to build `GLib` there is no other way. Anyway, I am not 100% sure that we need `libxml2` as dep for `GLib`, but I also feel that this issue should not block this PR. (maybe I am wrong here)

ocaisa: I checking everything in bin and lib of a build I have of `GLib/2.47.5` and there doesn't seem to be any dependency on libxml, I think it will work as a build dep.

hajgato: I have the feeling that `GLib` does not need `libxml2` at all. The only xml stuff I found there, is xmlcatalog. xml catalogs are pointing to xml documents, but anyway, we are not keen to install documentation. Moreover, not giving the `--with-xml-catalog=CATALOG` option to `configure` it defaults to `/etc/xml/catalog`
So probably libxml2 is not a real dep for `GLib` 

How shoud we proceed?

damianam:
Sorry for the interruption here, but I have a question:

Are we sure that the required dependency is libX11 and not libXt? I was building R these days also, and the easyconfig in the repo couldn't get built because of a missing libXt header.

hajgato:
@damianam if you want X11 support, you need `libX11`, `libXt`, and `libXmu`  
You probably took easyconfig from the WIP (Work In Progress) version, check the current one.

damianam:
@hajgato, no, I took it from master, but it didn't build. However, after adding `libXt` it did. `libXt` loaded `libX11` though. `libXmu` wasn't loaded. I didn't test the build, so maybe there is a hidden problem in my setup.

easybuild/easyconfigs/r/R/R-3.2.3-intel-2016a-libX11-1.6.3.eb
boegel: I guess we could try and see if GLib works without libxml2, on a setup where libxml2 is not provided by the OS?

easybuild/easyconfigs/c/cairo/cairo-1.14.6-intel-2016a-libX11-1.6.3.eb
```diff
@@ -0,0 +1,56 @@
+easyblock = 'ConfigureMake'
+
+name = 'cairo'
+version = '1.14.6'
+
+libx11 = 'libX11'
+libxver = '1.6.3' 
+versionsuffix = '-%s-%s' % (libx11, libxver)
```
boegel: @hajgato: you could use the enhanced `cairo` easyconfig from #2630, instead of adding one with a versionsuffix?

hajgato: @boegel: I am on it, but because the `GLib`-`libxm2` issue, I need a `GLib` with  `Pyhon-libX11`

easybuild/easyconfigs/r/R/R-3.2.3-intel-2016a-libX11-1.6.3.eb
boegel: GLib uses `xmllint` at build time, which is provided by `libxml2`, cfr. https://mail.gnome.org/archives/gtk-list/2013-April/msg00010.html

so, `libxml2` is definitely a build dep

hajgato:
@boegel finished, `Python` dependency removed

boegel:
@hajgato: can you also include the recently added `doSNOW` in the list of extensions, cfr. #2649

I'm retesting this, expect a test report soon.

boegel:
Test report by @boegel
**SUCCESS**
node2090.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/41a85bffd68236993ce9 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2668.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/85c159b23da33d7d7a37 for a full test report.

hajgato:
@boegel `doSNOW` added.
It would be nice, if we could maintain extension lists  for multiple easyconfigs, for example `R`and `R with X`, `Python` and `Python with X` etc.

boegel:
Test report by @boegel
**SUCCESS**
node2668.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/bdfc144875c4721dfef4 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2090.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/3185523be1924cefdb8e for a full test report.

boegel:
Going in, thanks @hajgato!

PR #2680, opened 2016-03-09
SuiteSparse use LD_LIBRARY_PATH pointing to shared libraries

hajgato:


easybuild/easyconfigs/s/SuiteSparse/SuiteSparse-4.5.1-intel-2016a-METIS-5.1.0.eb
```diff
@@ -21,4 +21,6 @@ preinstallopts = 'MY_METIS_LIB="$EBROOTMETIS/lib/libmetis.so" '
 
 maxparallel = 1
 
+modextrapaths = {'CPATH': 'include', 'LD_LIBRARY_PATH': 'lib'}
```
boegel: this smells like a fix for the easyblock?

hajgato:
made an EasyBlock instead: https://github.com/hpcugent/easybuild-easyblocks/pull/862

PR #2696, opened 2016-03-10
Add GitPython to dummy (with keyring) to allow use of the EB git feat…

ocaisa:
…ures.


ocaisa:
@boegel solves the requirements for GitHub integration and also handles keyring (and the backend requirement of keyring)

easybuild/easyconfigs/g/GitPython/GitPython-1.0.2.eb
```diff
@@ -0,0 +1,52 @@
...
+    }),
+    ('gitdb', '0.6.4', {
+        'source_urls': ['https://pypi.python.org/packages/source/g/gitdb/'],
+    }),
+    ('GitPython', '1.0.2', {
+        'source_urls': ['https://pypi.python.org/packages/source/G/GitPython'],
+        'modulename': 'git',
+    }),
+    # Add the requirements for GitHub integration for EasyBuild
+    ('keyrings.alt', '1.1.1', {
```
ocaisa: Not sure if there are security issues with this keyring backend

easybuild/easyconfigs/g/GitPython/GitPython-1.0.2.eb
```diff
@@ -0,0 +1,52 @@
...
+exts_filter = ('python -c"import %(ext_name)s"', '')
+
+# allow use of system Python
+allow_system_deps = [('Python', SYS_PYTHON_VERSION)]
+
+exts_list = [
+    ('setuptools', '18.1', {
+        'source_urls': ['https://pypi.python.org/packages/source/s/setuptools/'],
+    }),
+    ('pip', '7.1.0', {
```
boegel: are both `setuptools` and `pip` *runtime* dependencies of any of the ones included below?

since you're using the system Python here, you're also at the mercy of the system Python to provide what's required to install Python packages...

boegel:
Test report by @boegel
**FAILED**
node2653.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/9b4989178d74afc2a1be for a full test report.

boegel:
Test report by @boegel
**FAILED**
node2111.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/af13281f5eedaadc35da for a full test report.

boegel:
@ocaisa: the tests are failing becayse `keyrings.alt` is only compatible with Python 2.7.x, and the system Python I was using is 2.6.x...

So, I think we need a better way of dealing with this...

For the optional dependencies for EasyBuild specifically, there are a number of options we can look into, like trying to install the Python packages for the optional dependencies, but not failing the installation if something went wrong.

easybuild/easyconfigs/g/GitPython/GitPython-1.0.2.eb
```diff
@@ -0,0 +1,52 @@
...
+    }),
+    ('GitPython', '1.0.2', {
+        'source_urls': ['https://pypi.python.org/packages/source/G/GitPython'],
+        'modulename': 'git',
+    }),
+    # Add the requirements for GitHub integration for EasyBuild
+    ('keyrings.alt', '1.1.1', {
+        'source_urls': ['https://pypi.python.org/packages/source/k/keyrings.alt/'],
+    }),
+    ('keyring', '8.4.1', {
```
boegel: last version still compatible with Python 2.6 is 5.7.1, see https://github.com/jaraco/keyring/blob/master/CHANGES.rst, see stick to that?

easybuild/easyconfigs/g/GitPython/GitPython-1.0.2.eb
boegel: this package provides additional (potentially less secure) backends, but I don't think that should hold us back from including it

boegel:
@ocaisa since `keyring` is not related to `GitPython` at all, maybe we should give this bundle another name?

migueldiascosta:
@ocaisa @boegel is this still relevant?

boegel:
It sort of still is, but we haven't figured out a way to deal with the `keyrings.alt` issue...

ocaisa:
There are quite a few additional deps being pulled in with this that need to be included

ocaisa:
I think this is really useful but probably the name is wrong, it should maybe be something like `EB-git-integration` and maybe there should be an easyblock so that you can check the system python to see if it is `2.6`

boegel:
This never really went anywhere, and I don't think it will...

I've opened #11509 which adds an easyconfig for `GitPython` using the system toolchain, but `keyring` is another piece of work, because it requires a bunch of deps that are not trivial to install without having control of the Python version (or the compiler, which comes into play with `cryptography`).

There's no point in keeping this open, if we really need/want a `keyring` easyconfig we can open a new PR for it.

So, closing this...

PR #2705, opened 2016-03-11
{chem}[intel-2015b] QuantumESPRESSO 5.2.1 with netCDF-Fortran 4.4.2 (REVIEW)

hajgato:
Yambo can use netCDF, needed for some functionality (for example read the examples)

### DIFF
--- a/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-5.2.1-intel-2015b-netCDF-Fortran-4.4.2.eb
+++ b/easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-5.2.1-intel-2015b-netCDF-Fortran-4.4.2.eb
@@ -1,0 +1,72 @@
+name = 'QuantumESPRESSO'
+version = '5.2.1'
+netcdf = 'netCDF-Fortran'
+netcdfver = '4.4.2'
+versionsuffix = '-%s-%s' % (netcdf, netcdfver)
+
+homepage = 'http://www.pwscf.org/'
+description = """Quantum ESPRESSO  is an integrated suite of computer codes
+ for electronic-structure calculations and materials modeling at the nanoscale.
+ It is based on density-functional theory, plane waves, and pseudopotentials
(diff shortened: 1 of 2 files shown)


boegel:
@hajgato Is the Yambo built on top of netCDF crippled in any way? If not, why don't we just enhance the existing QE easyconfig?

hajgato:
@boegel This is what I do not know. In theory, if `netCDF` is there, then `Yambo` uses it, otherwise standard fortran IO. I am not 100% sure that it is still interfaced perfectly with `QuantumESPRESSO`. It has to be investigated.

boegel:
@hajgato ping?

hajgato:
@boegel pong. Was on holiday

boegel:
@hajgato have you figured out whether including netCDF by default makes sense?

hajgato:
From the `Yambo` homepage:
```
One of the most useful features of yambo is its ability to carry out all (or most of!) the database I/O using the NetCDF libraries. If NetCDF support is not possible, yambo will handle the database I/O using the native binary Fortran format.
```
According to my understanding, `Yambo` handles files either via `netCDF` or native-Fortran. I have no idea whether it can do both (maybe yes). Our user were not complaining, so I guess the `QuantumESPRESSO`-`Yambo` interface is fine.
from `QuantumESPRESSO` website:
```
There is then a shell of external packages, which typically read data produced by Quantum ESPRESSO but do not need it to work. Some of them (notably Yambo, Wannier90, WanT) can be automatically downloaded and installed from the core distribution using make.
```
 There might be problems if you have native-Fortran style `Yambo` files, and you would like to use with the `netCDF` enabled `Yambo`, but I am not sure that this is a possibility in combination with `QuantumESPRESSO`. In the case of standalone `Yambo` (which we do not have), it would be more realistic.

hajgato:
@boegel ^

boegel:
@hajgato does it make sense to keep this open, since you mentioned #3357 replaces this?

hajgato:
For future versions, PR #3357 should be used.

PR #2714, opened 2016-03-12
Drop -XC suffix on Cray toolchain

gppezzi:
see https://github.com/hpcugent/easybuild-easyconfigs/issues/2711

Thanks @brettbode for the test report on XE/XK!

Thanks @pforai for the TC conversion py script!
 (I fixed a small bug so I'm posting it here https://gist.github.com/gppezzi/e039598e8a6bcb68f49e)


pforai:
lgtm

pforai:
And thx for the fix

boegel:
Thanks @gppezzi!

PR #2653, opened 2016-03-07
{bio}[foss-2015b] Relion-1.4-WIP

verdurin:
I'm afraid this brings with it essentially the whole of X11...

WIP because it builds in local testing but testing of running it has been minimal so far, notably not the GUI yet.

Happy to split this up into multiple PRs if that's easier to review, though a single one appears to make testing easier.

PR #2720, opened 2016-03-15
{vis}[foss-2015b] X11 WIP

verdurin:
First run at providing X11 for foss-2015b, as required by #2653


boegel:
Jenkins: ok to test

verdurin:
Okay, what's the best way to proceed here, @boegel? I fixed an inconsistency with `ncurses` 5.9 versus 6.0, but now that problem is occurring throughout `foss-2015b` when testing this PR.

Ought I to re-work them all to use 5.9?

boegel:
@verdurin we switched to ncurses 6.0 in the `2016a` toolchains; for `foss/2015b` you should stick to 5.9...

boegel:
@verdurin Do you think this is still relevant/useful?

verdurin:
@boegel - no, this has been superseded now. Closing.

PR #2738, opened 2016-03-19
fix perl extensions download urls

pescobar:


easybuild/easyconfigs/p/Perl/Perl-5.22.1-foss-2016a.eb
```diff
@@ -153,7 +153,7 @@ exts_list = [
     }),
     ('Archive::Extract', '0.76', {
         'source_tmpl': 'Archive-Extract-0.76.tar.gz',
-        'source_urls': ['https://cpan.metacpan.org/authors/id/ / "/ "BINGOS"'],
+        'source_urls': ['http://search.cpan.org/CPAN/authors/id/B/BI/BINGOS'],
```
boegel: @pescobar Why not keep the metacpan URL?

boegel:
Thanks @pescobar!

PR #2748, opened 2016-03-21
{bio} [goolf-1.4.10] BayeScan-2.1 (REVIEW)

pescobar:


easybuild/easyconfigs/b/BayeScan/BayeScan-2.1-goolf-1.4.10.eb
```diff
@@ -0,0 +1,33 @@
...
+easyblock = 'MakeCp'
+
+name = 'BayeScan'
+version = '2.1'
+
+homepage = 'http://cmpg.unibe.ch/software/BayeScan/'
+description = """BayeScan aims at identifying candidate loci under natural selection from genetic data, 
+ using differences in allele frequencies between populations."""
+
+# BayeScan makefile is hardcoded to use g++. In case you want to use a non-gcc toolchain you will need to patch the makefile
```
boegel: I would rather see this problem being dealt with, otherwise it'll be missed easily when someone uses `--try-toolchain` on this...

I think this would work:

```python
prebuildopts = "sed -i .bk 's/g\+\+/${CC} ${CFLAGS}/g' Makefile && "
```

Note that if you're not using `$CFLAGS`, you'll end up using the default for GCC, which is `-O0` (i.e. *NO* optimisations). You really don't want that...

boegel: @pescobar cfr. #3356

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/ed9e0aef105fef22c1c281b3bef0d0e4 for a full test report.

boegel:
lgtm

boegel:
Going in, thanks @pescobar!

PR #2750, opened 2016-03-21
Caffe

FokkeDijkstra:
I've installed caffe using the foss-2016a toolchain. As caffe has several dependencies I needed to include these as well. These dependencies (mainly OpenCV) introduce several extra dependencies.

Since all of these are needed to install Caffe, I've included all dependencies in this PR. If another way of getting these changes in is preferred, please let me know.


FokkeDijkstra:
Yes.

wpoely86:
ok to test

easybuild/easyconfigs/a/ant/ant-1.9.6-Java-1.8.0_74.eb
```diff
@@ -0,0 +1,26 @@
...
+ as targets and extension points dependent upon each other. The main known usage of Ant is the build of Java applications."""
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['apache-%(name)s-%(version)s-src.tar.gz']
+source_urls = ['http://archive.apache.org/dist/%(name)s/source/']
+
+java = 'Java'
+javaver = '1.8.0_74'
+versionsuffix = '-%s-%s' % (java, javaver)
```
wpoely86: In 2.7.0 we introduced a couple of new string templates to make this easier:
```
    %(javashortver)s: short version for Java (<major>.<minor>)
    %(javaver)s: full version for Java
```
Just load `Java` under dependencies and then this templates will work.

easybuild/easyconfigs/f/FFmpeg/FFmpeg-2.8.6-foss-2016a.eb
```diff
@@ -0,0 +1,30 @@
...
+    ('x264', '20160114'),
+]
+
+configopts = '--enable-pic --enable-shared --enable-gpl --enable-version3 --enable-nonfree --cc="$CC" --cxx="$CXX" '
+configopts += '--enable-libx264'
+
+sanity_check_paths = {
+    'files': ['bin/ff%s' % x for x in ['mpeg', 'probe', 'server']] +
+             ['lib/lib%s.%s' % (x, y) for x in ['avdevice', 'avfilter', 'avformat', 'avcodec', 'postproc',
+                                                'swresample', 'swscale', 'avutil'] for y in ['so', 'a']],
```
wpoely86: use `SHLIB_EXT` constant instead of `so`.

easybuild/easyconfigs/g/GLib/GLib-2.47.5-foss-2016a.eb
```diff
@@ -0,0 +1,23 @@
...
+name = 'GLib'
+version = '2.47.5'
+
+homepage = 'http://www.gtk.org/'
+description = """GLib is one of the base libraries of the GTK+ project"""
+
+toolchain = {'name': 'foss', 'version': '2016a'}
+toolchainopts = {'optarch': True, 'pic': True}
+
+source_urls = ['http://ftp.gnome.org/pub/gnome/sources/glib/%(version_major_minor)s/']
```
wpoely86: use `FTPGNOME_SOURCE` ?

easybuild/easyconfigs/g/gflags/gflags-2.1.2-foss-2016a.eb
```diff
@@ -0,0 +1,29 @@
+easyblock = 'CMakeMake'
+
+name = 'gflags'
+version = '2.1.2'
+
+homepage = 'https://github.com/gflags/gflags'
+description = """
+The gflags package contains a C++ library that implements commandline flags processing. It includes built-in support for standard types such as string and the ability to define flags in the source file in which they are used.
```
wpoely86: put line breaks if longer then 120 chars.

easybuild/easyconfigs/g/gflags/gflags-2.1.2-foss-2016a.eb
```diff
@@ -0,0 +1,29 @@
...
+toolchainopts = {'optarch': True, 'pic': True}
+
+configopts = '-DBUILD_SHARED_LIBS=on -DBUILD_STATIC_LIBS=on'
+
+sources = ['v%(version)s.tar.gz']
+source_urls = ['https://github.com/gflags/gflags/archive/']
+
+sanity_check_paths = {
+    'files': ['bin/gflags_completions.sh'] + 
+             ['lib/%s' % x for x in ['libgflags.so', 'libgflags_nothreads.so', 'libgflags.a', 'libgflags_nothreads.a']] +
```
wpoely86: use `SHLIB_EXT`

easybuild/easyconfigs/g/gflags/gflags-2.1.2-foss-2016a.eb
```diff
@@ -0,0 +1,29 @@
...
+             ['include/gflags/gflags_completions.h'],
+    'dirs': [],
+}
+
+builddependencies = [
```
wpoely86: put this before `sanity_check_paths`

easybuild/easyconfigs/s/snappy/snappy-1.1.3-foss-2016a.eb
```diff
@@ -0,0 +1,28 @@
...
+description = """Snappy is a compression/decompression library. It does not aim
+for maximum compression, or compatibility with any other compression library;
+instead, it aims for very high speeds and reasonable compression."""
+
+toolchain = {'name': 'foss', 'version': '2016a'}
+
+source_urls = ['https://github.com/google/snappy/archive']
+sources = ['%s.tar.gz' % commit]
+
+preconfigopts = "./autogen.sh && "
```
wpoely86: build dependency on `Autotools`?

FokkeDijkstra:
wpoely86, thanks for the suggestions. I will implement and test these.

FokkeDijkstra:
When checking the actual dependencies of the caffe binary there is a whole bunch of libraries taken from the system. Is there a policy about what libraries are supposed to be provided by the OS and which should come from easybuild modules?

The list is:
libgtk-x11-2.0.so.0
libgdk-x11-2.0.so.0
libatk-1.0.so.0
libpangoft2-1.0.so.0
libpangocairo-1.0.so.0
libgdk_pixbuf-2.0.so.0
libcairo.so.2
libpango-1.0.so.0
libfreetype.so.6
libfontconfig.so.1
libgstbase-0.10.so.0
libgstreamer-0.10.so.0
libgstvideo-0.10.so.0
libgstapp-0.10.so.0
libgstriff-0.10.so.0
libgstpbutils-0.10.so.0
libX11.so.6
libXfixes.so.3
libXext.so.6
libXrender.so.1
libXinerama.so.1
libXi.so.6
libXrandr.so.2
libXcursor.so.1
libXcomposite.so.1
libXdamage.so.1
libpng12.so.0
libpixman-1.so.0
libgstaudio-0.10.so.0
libgsttag-0.10.so.0
libjpeg.so.62
libxcb.so.1
libgstinterfaces-0.10.so.0
libXau.so.6

boegel:
@FokkeDijkstra generally, we try and provide as many dependencies as possible through EasyBuild rather than relying on the OS. This makes it more likely that the installation will work out-of-the-box for others (which may not have the required dependencies available on the OS), and protects you from things disappearing or being updated in the OS.

That being said, this can require a lot of effort...
It highly depends on how much logic is supported in the build procedure of `Caffe`.

If you're not up to tackling all dependencies through EasyBuild, then it would be nice to specify which OS dependencies are required, by listing the corresponding package names in `osdependencies`.

boegel:
@FokkeDijkstra Are you planning to open a separate pull request for Caffe itself? This PR is already pretty big, so it's probably better to avoid making it even bigger and opening one or more additional PRs (and specifying that they depend on this, etc.).

FokkeDijkstra:
Caffe itself should have been included, but was missing. The whole reason for adding all these in one PR, was preventing the creation of interdependent pull requests.
Caffe is now included.

easybuild/easyconfigs/c/caffe/caffe-rc3-foss-2016a-CUDA-7.5.18.eb
```diff
@@ -0,0 +1,48 @@
...
+    ('gflags', '2.1.2'),
+    ('OpenCV', '3.1.0'),
+    ('CUDA', cudaversion, '', True),
+    ('cuDNN', '4.0', '', True),
+    ('LMDB', '0.9.18'),
+    ('leveldb', '1.18'),
+    ('snappy', '1.1.3'),
+    ('protobuf', '2.6.1'),
+    ('HDF5', '1.8.16','-serial'),
+    ('Boost', '1.59.0','-Python-2.7.11')
```
boegel: @FokkeDijkstra so, what will we do with the X11 deps?

FokkeDijkstra:
I hoped for some advice on which packages to take from the OS, and which not. I see several packages in the repository which could be included. I will start with adding these to the packages which depend on them. This will take some time though.
It will also add new easyconfigs to this PR. Do you have any idea how difficult it would be to split up this PR into smaller pieces, which have a dependency chain?

boegel:
@FokkeDijkstra I don't think making this PR bigger than it already is is a good idea; if anything, we should flesh out some things here already in separate, smaller PRs, and get them merged.

Good candidates are easyconfigs that are just a different toolchain/software version compared to existing easyconfigs (those can be packed in a single PR, since we have a good way of reviewing those).

The general rule is to keep PRs smaller than 10 easyconfigs, preferably smaller than 5, to make them easy to handle (review/test/merge).

FokkeDijkstra:
When I find some spare time, I will recreate this PR in smaller chunks. I guess this one can be rejected.

FokkeDijkstra:
I've submitted most of this PR in smaller chunks already. This one should be rejected.

PR #2755, opened 2016-03-21
{data}[intel/2016a] HDF5 v1.8.16

hajgato:
(created using `eb --new-pr`)


hajgato:
wrong PR, the right one is PR #2756

boegel:
@hajgato: it seems like we need protection in `--from-pr` to avoid opening empty PRs? Please open an issue on that in the framework repo.

hajgato:
@boegel you mean `--new-pr` right?

PR #2781, opened 2016-03-25
Gnuplot

jgphpc:
just a new gnuplot compiled with gcc.


boegel:
@jgphpc Please target `develop` rather than `master`, or use `eb --new-pr` (which will make sure `develop` is targeted for you)...

jgphpc:
will redo

PR #2800, opened 2016-03-30
Contribution: dca and magma 2.0.0 easyonfigs

gbalduzz:
Added DCA++ bundle file for building the Dynamical Cluster Approximation software.
 The necessary easyconfig files where added, notably magma-2.0.0 and its patch file for linking to libcusparse.


gbalduzz:
yes

boegel:
Jenkins: ok to test

@gppezzi Can you review/test?

easybuild/easyconfigs/d/DCA/DCA++-CMake-CrayGNU-2015.11.eb
```diff
@@ -0,0 +1,25 @@
+easyblock = 'Bundle'
+
+name = 'DCA++'
+version = '1.0'
```
gppezzi: I'd suggest to use the same version as the toolchain (2015.11)

gppezzi: You could also add 
```
versionsuffix = '-cuda' 
```
to make the dependency on gpu explicit (or `-gpu`). As opposed to the `-dora` that actually means cpu-only.

easybuild/easyconfigs/g/gtest/gtest-1.7.0-CrayGNU-2015.11.eb
```diff
@@ -0,0 +1,45 @@
+# Built with EasyBuild version 2.5.0 on 2016-01-22_09-08-42
```
gppezzi: Please remove this comment.

easybuild/easyconfigs/g/gtest/gtest-1.7.0-CrayGNU-2015.11.eb
```diff
@@ -0,0 +1,45 @@
...
+source_urls = ['https://googletest.googlecode.com/files']
+
+sanity_check_paths={
+    'files': ['CMakeLists.txt', ],
+    'dirs': ['include', 'src'],
+}
+
+moduleclass = 'devel'
+
+# Build statistics
```
gppezzi: Please also remove these build statistics (should be kept on local builds, it doesn't make sense to upload this to the repo).

easybuild/easyconfigs/c/cmake/CMake-3.3.1-CrayGNU-2015.11.eb
```diff
@@ -0,0 +1,24 @@
...
+
+name = 'CMake'
+version = "3.3.1"
+
+homepage = 'http://www.cmake.org'
+description = """CMake, the cross-platform, open-source build system.
+ CMake is a family of tools designed to build, test and package software."""
+
+toolchain = {'name': 'CrayGNU', 'version': '2015.11'}
+toolchainopts = {'dynamic': True}
```
gppezzi: Dynamic is now true by default, you can remove this line.

easybuild/easyconfigs/c/cmake/CMake-3.3.1-CrayGNU-2015.11.eb
```diff
@@ -0,0 +1,24 @@
+# Built with EasyBuild version 2.5.0 on 2016-01-22_09-06-09
```
gppezzi: Please remove this comment.

easybuild/easyconfigs/d/DCA/DCA++-CMake-CrayGNU-2015.11.eb
```diff
@@ -0,0 +1,25 @@
...
+#preinstallopts: change build_directory
+#disable clean build directory
+
+homepage = 'http://www.itp.phys.ethz.ch/people/schulthess'
+description = """This modules bundles all the dependencies for DCA++, the best DCA ever"""
+
+toolchain = {'name': 'CrayGNU', 'version': '2015.11'}
+
+dependencies = [
+    ('CMake', '3.3.1'),
```
gppezzi: Magma already uses CMake 3.5.0: you could use 3.5.0 also here (if DCA++ is OK with that) or downgrade CMake version on Magma to match 3.3.1 (in order to avoid compiling CMake twice).

easybuild/easyconfigs/m/magma/magma-2.0.0.patch
```diff
@@ -0,0 +1,106 @@
...
+ #
+ #GPU_TARGET ?= Fermi Kepler
++GPU_TARGET = Kepler
+ 
+-# --------------------
+-# programs
+-
+-CC        = gcc
+-CXX       = g++
++CXX       = CC
```
gppezzi: Can you please use the following variables for defining the compilers (instead of hardcoding them)?

```
CXX      := $(CXX)
CC       := $(CC)
NVCC     := nvcc
FORT     := $(F90)
```

gppezzi:
@gbalduzz can you fix those remarks? I can fix it for you if you prefer, just let me know.

gbalduzz:
I agree with all the fixes, if you can implement them I would be very
grateful, otherwise let me know and I will do it (probably tomorrow).
Cheers, Giovanni

easybuild/easyconfigs/d/DCA/DCA++-CMake-CrayGNU-2015.11.eb
pforai: Also CMake should be moved to `builddependencies=[]` as there is no need for the CMake module to be loaded if a user wants to use the software, this is only required during compilation.

boegel:
@gbalduzz did you mean to close this PR?

PR #2804, opened 2016-03-30
Gtool 0.7.5

rtripath89:
Created an Easyconfig file for gtool v0.7.5 .


boegel:
@rtripath89 since you're targeting `master` here, and you already have #2805 in place that targets `develop` as expected, please close this PR

PR #2819, opened 2016-04-01
Replace bitstring 3.1.3 with 3.1.4 in Python 2.7.11

stachon:
Version 3.1.3 can no longer be downloaded, see https://github.com/scott-griffiths/bitstring/issues/159


boegel:
@stachon I would strongly prefer not bumping the version, we keep the extension versions fixed once the easyconfig has been merged in.

Once option is to change the download URL though, since v3.1.3 is still available from https://github.com/scott-griffiths/bitstring/archive/bitstring-3.1.3.tar.gz...

However, it should be possible to get the tarball back up there too, I think...

boegel:
@stachon No luck so far with getting 3.1.3 back on PyPI... So, let's go with the alternate GitHub URL?

boegel:
@stachon ah, also, this PR was opened to `master`; only when a new EasyBuild release is done, we update `master`...

I'll look into another PR to `develop` for changing the bitstring URL myself, unless you beat me to it...

boegel:
cfr. #2880

PR #2820, opened 2016-04-02
Qt/486 with dummy toolchain

jgphpc:
https://github.com/hpcugent/easybuild-easyblocks/pull/881

Remark:
gcc/434 failed to build qt/4.8.7
gcc/434 too old for any qt/5.x


boegel:
replaced by #2821

PR #2840, opened 2016-04-08
Builddependencies fixed for GObject-introspection

hajgato:
Check:
https://packages.debian.org/source/sid/gobject-introspection
and
https://packages.debian.org/sid/gobject-introspection


boegel:
merged via #2852, thanks @hajgato

PR #2871, opened 2016-04-12
HTSlib for foss 2016a

RvDijk:

PR #2872, opened 2016-04-12
Bcftools foss2016a

RvDijk:
Depends on #2871


boegel:
@RvDijk maybe you should take a look at the (experimental) support for `eb --new-pr` that is included in EasyBuild v2.7.0...

There's no documentation yet though, I'm working on that (see boegel-eb.readthedocs.org/en/new_update_pr/Integration_with_GitHub.html)

RvDijk:
@boegel I don't understand how --new-pr is used, but I'm looking forward to the documentation.

PR #2876, opened 2016-04-13
GSL-2.1 for foss 2016a

RvDijk:

PR #2877, opened 2016-04-13
{numlib}[foss/2016a} GSL 2.1 (REVIEW)

hajgato:


hajgato:
oops, check #2876 (missed with 2 hrs)....

PR #2906, opened 2016-04-18
Wannier90-1.2 and 2.0.1 using toolchain intel-2016.02-GCC-4.9

migueldiascosta:
Using "MakeCp" easyblock since there is no configure step, still needs an ugly hack to skip the need for a make.sys file...

### DIFF
--- a/easybuild/easyconfigs/w/Wannier90/Wannier90-1.2-intel-2016.02-GCC-4.9.eb
+++ b/easybuild/easyconfigs/w/Wannier90/Wannier90-1.2-intel-2016.02-GCC-4.9.eb
@@ -13,7 +13,7 @@ source_urls = ['http://www.wannier.org/code']
 
 # ugly hack
 prebuildopts = 'rm make.sys; touch make.sys; '
-prebuildopts += 'F90=ifort FCOPTS="-O2" LDOPTS="-O2" LIBDIR=$LAPACK_LIB_DIR LIBS=$LIBLAPACK '
+prebuildopts += 'F90=ifort FCOPTS="$FFLAGS" LDOPTS="$FFLAGS" LIBDIR="$LAPACK_LIB_DIR" LIBS="$LIBLAPACK" '
 
 files_to_copy = [(['wannier90.x'], 'bin')]
 
--- a/easybuild/easyconfigs/w/Wannier90/Wannier90-1.2-intel-2016.02-GCC-4.9.eb
+++ b/easybuild/easyconfigs/w/Wannier90/Wannier90-1.2-intel-2016.02-GCC-4.9.eb
@@ -15,10 +15,13 @@ source_urls = ['http://www.wannier.org/code']
 prebuildopts = 'rm make.sys; touch make.sys; '
 prebuildopts += 'F90=$F90 FCOPTS="$FFLAGS" LDOPTS="$FFLAGS" LIBDIR="$LAPACK_LIB_DIR" LIBS="$LIBLAPACK" '
 
-files_to_copy = [(['wannier90.x'], 'bin')]
+# build program and library
+buildopts = 'all'
+
+files_to_copy = [(['wannier90.x'], 'bin'), (['libwannier.a'], 'lib')]
 
 sanity_check_paths = {
-    'files': ['bin/wannier90.x'],
+    'files': ['bin/wannier90.x', 'lib/libwannier.a'],
     'dirs': []
 }
 
--- a/easybuild/easyconfigs/w/Wannier90/Wannier90-1.2-intel-2016.02-GCC-4.9.eb
+++ b/easybuild/easyconfigs/w/Wannier90/Wannier90-1.2-intel-2016.02-GCC-4.9.eb
@@ -13,7 +13,7 @@ source_urls = ['http://www.wannier.org/code']
 
 # ugly hack
 prebuildopts = 'rm make.sys; touch make.sys; '
-prebuildopts += 'F90=ifort FCOPTS="$FFLAGS" LDOPTS="$FFLAGS" LIBDIR="$LAPACK_LIB_DIR" LIBS="$LIBLAPACK" '
+prebuildopts += 'F90=$F90 FCOPTS="$FFLAGS" LDOPTS="$FFLAGS" LIBDIR="$LAPACK_LIB_DIR" LIBS="$LIBLAPACK" '
 
 files_to_copy = [(['wannier90.x'], 'bin')]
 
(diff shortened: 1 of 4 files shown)


easybuild/easyconfigs/w/Wannier90/Wannier90-2.0.1-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,26 @@
...
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = ['http://www.wannier.org/code']
+
+# ugly hack
+prebuildopts = 'rm make.sys; touch make.sys; '
+prebuildopts += 'F90=ifort COMMS=mpi MPIF90=mpiifort FCOPTS="-O2" LDOPTS="-O2" LIBDIR=$LAPACK_LIB_DIR LIBS=$LIBLAPACK '
```
boegel: why is this more extensive than the one included in the other easyconfig?

also, take into account that `$LIBLAPACK` likely contains spaces, so you need `LIBS="$LIBLAPACK`

in addition, please consider not hardcoding `-O2`, nad using `$FFLAGS` instead with `FCOPTS="$FFLAGS"`; I guess you can also use `$FFLAGS` for `LDOPTS`

easybuild/easyconfigs/w/Wannier90/Wannier90-2.0.1-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,26 @@
...
+description = """A tool for obtaining maximally-localised Wannier functions"""
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = ['http://www.wannier.org/code']
+
+# ugly hack
+prebuildopts = 'rm make.sys; touch make.sys; '
```
boegel: can you explain why this is needed exactly? can we figure out a less ugly solution?

boegel:
Jenkins: ok to test

easybuild/easyconfigs/w/Wannier90/Wannier90-2.0.1-intel-2016.02-GCC-4.9.eb
migueldiascosta: regarding the difference between the easyconfigs, wannier 2.x supports MPI, 1.x doesn't (however, some applications still only support 1.x; 1.2 is the last release of the 1.x series, 2.0.1 is the latest releast of 2.x)

regarding LIBLAPACK quoting, will do, thanks!

regarding the optimizations flags, of course, sorry!

easybuild/easyconfigs/w/Wannier90/Wannier90-2.0.1-intel-2016.02-GCC-4.9.eb
migueldiascosta: well, this (and other) package don't have a configure script but instead expect you to alter a file (that is included in the Makefile) with the architecture dependent options. 

In this case I removed the default file and defined the options in the command line. I could have used something like sed to alter the file instead, but I don't know if that would be less ugly.

Maybe in these cases a different easyblock is needed to create/alter this file before building?

boegel: @migueldiascosta is `make.sys` a file that gets included in the `Makefile`? There may be no need to remove/empty it at all, since what you specify on the command line via `buildopts` should get preference over what's in the file...

easybuild/easyconfigs/w/Wannier90/Wannier90-1.2-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,25 @@
...
+description = """A tool for obtaining maximally-localised Wannier functions"""
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = ['http://www.wannier.org/code']
+
+# ugly hack
+prebuildopts = 'rm make.sys; touch make.sys; '
+prebuildopts += 'F90=ifort FCOPTS="$FFLAGS" LDOPTS="$FFLAGS" LIBDIR="$LAPACK_LIB_DIR" LIBS="$LIBLAPACK" '
```
boegel: @migueldiascosta please use `$F90` rather than hardcoding `ifort`, that will make the use of `--try-toolchain` works as expected: `F90="$F90"`

This may look a bit silly, but it makes sense if the `Makefile` (or `make.sys`) has a hardcoded value for `F90` in place; defining an environment variable `$F90` (which is done by the toolchain mechanism in the EasyBuild framework) and specifying `F90=` on the `make` command line are not equivalent...

easybuild/easyconfigs/w/Wannier90/Wannier90-2.0.1-intel-2016.02-GCC-4.9.eb
migueldiascosta: yes, it's included in the Makefile, and yes, the command line options should get preference, but they don't... :(

maybe because it's included in ./src/Makefile, which gets called from ./Makefile... (?)

boegel: OK, it's possible they're not taking care of making sure that the options specified on the command line trickle down properly... I'm not sure how to fix that easily.

How about patching the `include` line from the `src/Makefile` instead, either with an actual patch file (since it's likely the patch will keep working for future versions)?

easybuild/easyconfigs/w/Wannier90/Wannier90-2.0.1-intel-2016.02-GCC-4.9.eb
boegel: OK, thanks for clarifying the 1.x vs 2.x, I missed that.

easybuild/easyconfigs/w/Wannier90/Wannier90_1x_ignore_makesys.patch
```diff
@@ -0,0 +1,11 @@
+--- src/Makefile.orig	2016-04-19 15:26:27.373047000 +0800
```
boegel: @migueldiascosta please include a line of comment explaining what this patch does, why it is needed, and mention yourself as an author (any lines above the `---` line are comments); please do so in both patches

```
avoid including make.sys, which contains hardcoding settings we don't need/want
author: Miguel Dias Costa (National University of Singapore)
```

boegel:
Test report by @boegel
**SUCCESS**
node2614.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/c3f9c3b3fd5f901a84968584ce2823c3 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/acd673ba81308581f3d602f7b69cc33c for a full test report.

boegel:
lgtm

boegel:

PR #2811, opened 2016-03-31
Use `env` wherever `prebuildopts is used to set environmental variables.

riccardomurri:
As requested by @boegel in PR #2807

PR #2925, opened 2016-04-20
BerkeleyGW 1.0.6 and 1.1-beta2 using intel-2016.02-GCC-4.9 toolchain

migueldiascosta:
Used patch to disable including arch.mk and defined variables in prebuildopts.

The only difference between the two eb files, besides the version, is the checksum.

(Using the MakeCp easyblock because there is no configure, but BGW does support make install. Is it possible to skip the configure step in the ConfigureMake? Or is there a block with just build and install?)


easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,35 @@
...
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+source_urls = ['http://www.berkeleygw.org/releases/']
+sources = ['BGW-%(version)s.tar.gz']
+checksums = ['8740562da41e87a62eb8d457675e4bfd']
+
+patches = ['BerkeleyGW_ignore_archmk.patch']
+
+files_to_copy = ['*']
```
boegel: @migueldiascosta using `ConfigureMake` together with `skipsteps = ['configure']` makes more sense, assuming you can somehow specify what the installation prefix should be...

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,35 @@
...
+
+prebuildopts = 'COMPFLAG=-DINTEL PARAFLAG=-DMPI MATHFLAG=-DUSESCALAPACK FCPP="cpp -ansi" F90free="$MPIF90 -free" '
```
boegel: please use `env`, cfr. #2811

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,35 @@
...
+prebuildopts += 'LINK="$MPIF90" FOPTS="$FFLAGS" MOD_OPT="-module " INCFLAG="-I" C_PARAFLAG=-DPARA CC_COMP="$MPICXX" '
+prebuildopts += 'C_COMP="$MPICC" C_LINK="$MPICXX" C_OPTS="$CFLAGS" REMOVE="/bin/rm -f" MKLPATH="$MKLROOT" '
```
boegel: is the `REMOVE` necessary?

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,35 @@
...
+prebuildopts += 'FFTWLIB="$MKLROOT/lib/intel64/libfftw2xf_double_intel.a" FFTWINCLUDE="$FFTW_INC_DIR" '
```
boegel: Hmm, you need to pass the FFTW2 library here?

It's unfortunate you can't use `$LIBFFT`...

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,35 @@
...
+prebuildopts += 'LAPACKLIB="$LIBLAPACK" BLACSDIR="$BLACS_LIB_DIR" BLACS="$LIBBLACS" SCALAPACKLIB="$LIBSCALAPACK" '
+
+buildopts = 'all-flavors'
+
+sanity_check_paths = {
+    'files': [],
+    'dirs': ['bin']
```
boegel: how many binaries are included? maybe make this a bit more specific?

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,35 @@
+easyblock = 'MakeCp'
+
+name = 'BerkeleyGW'
+version = "1.0.6"
+
+homepage = 'http://www.berkeleygw.org'
+description = """The BerkeleyGW Package is a set of computer codes that calculates the quasiparticle
+                 properties and the optical responses of a large variety of materials from bulk
+                 periodic crystals to nanostructures such as slabs, wires and molecules."""
```
boegel: no need for the long indent, this will probably make the output of `module help BerkeleyGW` looks rather weird...

just use a single space to indent the continuation lines

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,35 @@
...
+source_urls = ['http://www.berkeleygw.org/releases/']
+sources = ['BGW-%(version)s.tar.gz']
+checksums = ['8740562da41e87a62eb8d457675e4bfd']
+
+patches = ['BerkeleyGW_ignore_archmk.patch']
+
+files_to_copy = ['*']
+
+prebuildopts = 'COMPFLAG=-DINTEL PARAFLAG=-DMPI MATHFLAG=-DUSESCALAPACK FCPP="cpp -ansi" F90free="$MPIF90 -free" '
+prebuildopts += 'LINK="$MPIF90" FOPTS="$FFLAGS" MOD_OPT="-module " INCFLAG="-I" C_PARAFLAG=-DPARA CC_COMP="$MPICXX" '
```
boegel: it looks like some of these are fairly standard, maybe avoiding the include of `arch.mk` is not a good option here, and we should only override what we need to?

boegel:
Jenkins: ok to test

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: ok, will try that

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: right, sorry

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: well, it's not defined, and I think it doesn't have a default, but will check

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: also, I think $LIBFFT doesn't have the fortran wrappers, does it?

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: quite a few :) but some are more important than others, I'll include those

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: an arch.mk will need to be copied from a few provided ones, and manipulated in order to at least remove the options that we want to define here, since the include seems to take preference

boegel: if you pass changes via `buildopts`, so they goign *after* `make` in the command line being used, overriding values should work?

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: is there a way to reproduce in an interactive shell the exact environment that is created when running eb with a specific easyblock and easyconfig?

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
boegel: OK, it's indeed better to include a couple of specific binaries rather than just specifying that a non-empty `bin` directory should be there

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
boegel: You can check with `--extended-dry-run` or `-x` to see how `$LIBFFT` will be defined, but it's good to do that on a system where the toolchain you're using is actually available (since the definition depends on which libraries are actually present on the filesystem):

```
$ eb --from-pr 2925 -x | grep LIBFFT
  export LIBFFT="-Wl,-Bstatic -Wl,--start-group -lfftw3xc_intel -lfftw3x_cdft_lp64 -lmkl_cdft_core -lmkl_blacs_intelmpi_lp64 -lmkl_intel_lp64 -lmkl_sequential -lmkl_core -Wl,--end-group -Wl,-Bdynamic"
```

Note that these are the FFTW 3.x wrappers...

If this is not sufficient/correct, then you can't use it obviously, it just came to mind.

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: you're right, it should, will try

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
boegel: Also, with `eb --dump-env-script` you can create a script that can be sourced to set up the environment in which EasyBuild operates, for a specific easyconfig file.

migueldiascosta: regarding the environment variables set by easybuild, yes, I've been doing that, or using the logs, just wondered if there was a better way to interactively test/debug

regarding fftw3, in version 1.1-beta2 of BGW, one could use FFTW 3 instead of 2 (by adding -DFFTW3), but one would still need the Fortran wrapper (libfftw3xf...), those are for C

migueldiascosta: regarding the environment, sorry, somehow your answer with dump-env-script option didn't appear when I replied, yes, that's exactly what I was looking for :)

boegel: Not including the Fortran wrapper is actually not intention, so probably a bug in the EasyBuild framework...

cfr. https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/toolchains/fft/intelfftw.py#L67

easybuild/easyconfigs/b/BerkeleyGW/BerkeleyGW-1.0.6-intel-2016.02-GCC-4.9.eb
migueldiascosta: ok, the patch is not needed if in `prebuildopts` we copy one of the example files to `arch.mk`  and then override with the variables in `buildopts`

choosing example file we need to change less, this part of the eb file could be simplified to

```
prebuildopts = 'cp config/stampede.tacc.utexas.edu.mk arch.mk; '

buildopts = 'DEBUGFLAG="" FFTWLIB="$MKLROOT/lib/intel64/libfftw2xf_double_intel.a" FFTWINCLUDE="$FFTW_INC_DIR" '
buildopts += 'LAPACKLIB="$LIBLAPACK" BLACSDIR="$BLACS_LIB_DIR" BLACS="$LIBBLACS" SCALAPACKLIB="$LIBSCALAPACK" '
```

boegel: that looks a lot better indeed

which choices are available? using a very system-specific one like the one for stampede seems weird... isn't there a more generic one available (that is more likely to not disappear)?

migueldiascosta: yes, it's weird, that's why I didn't commit :)

that is the one that we need to change less, since it's the only example that uses the intel compilers/mk/mpi stack. 

There is a `generic.mpi.linux.mk` file, but then we need to override almost all the other options again (well, at least we can skip `FCPP`, `REMOVE`, ...)

boegel: Well, I'd stick to using `generic.mpi.linux.mk` rather than relying on an example `.mk` file that happens to specify some things that match the toolchain being used (since that may change)

also, it will limit the changes required when using a totally different toolchain (for example via `--try-toolchain=foss,...`)

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/b38478316ff96493829da6bcd1aa7fd7 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2401.golett.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/9f6be251d7868721971c1451d7a2cd4c for a full test report.

boegel:
lgtm

boegel:

PR #2928, opened 2016-04-20
Score-p + CrayGNU

jgphpc:
@geimer i am opening this PR to start the discussion about Score-p on the Cray, in relation with PR
https://github.com/hpcugent/easybuild-easyconfigs/pull/2892

edit: depends on https://github.com/hpcugent/easybuild-easyblocks/pull/889 and ~~https://github.com/hpcugent/easybuild-easyblocks/pull/894~~

### DIFF
--- a/easybuild/easyconfigs/c/CrayGNU/CrayGNU-2016.03.eb
+++ b/easybuild/easyconfigs/c/CrayGNU/CrayGNU-2016.03.eb
@@ -1,0 +1,19 @@
+easyblock = 'CrayToolchain'
+
+name = 'CrayGNU'
+version = '2016.03'
+
+homepage = 'http://docs.cray.com/books/S-9408-1603/'
+description = """Toolchain using Cray compiler wrapper, using PrgEnv-gnu module (PE release: November 2015).\n"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+dependencies = [
+    # PrgEnv version is not pinned, as Cray recommends to use the latest (default) version
+    ('PrgEnv-gnu', EXTERNAL_MODULE),
+    ('gcc/4.9.3', EXTERNAL_MODULE),
+    ('cray-libsci/16.03.1', EXTERNAL_MODULE),
+    ('cray-mpich/7.3.2', EXTERNAL_MODULE),
+]
+
+moduleclass = 'toolchain'
--- a/easybuild/easyconfigs/l/libunwind/libunwind-1.1.eb
+++ b/easybuild/easyconfigs/l/libunwind/libunwind-1.1.eb
@@ -1,0 +1,24 @@
+easyblock = 'ConfigureMake'
+
+name = "libunwind"
+version = "1.1"
+
+homepage = 'http://www.nongnu.org/libunwind/'
+description = """The primary goal of libunwind is to define a portable and efficient C programming interface
+ (API) to determine the call-chain of a program. The API additionally provides the means to manipulate the
+ preserved (callee-saved) state of each call-frame and to resume execution at any point in the call-chain
+ (non-local goto). The API supports both local (same-process) and remote (across-process) operation.
+ As such, the API is useful in a number of applications"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+dependencies = [ ('binutils', '2.25.1'), ]
+
+sources = [SOURCE_TAR_GZ]
+source_urls = [GNU_SAVANNAH_SOURCE]
+
+sanity_check_paths = {
+    'files': ["include/libunwind.h", "lib/libunwind.%s" % SHLIB_EXT],
+    'dirs': []
+}
+
+moduleclass = 'lib'
--- a/easybuild/easyconfigs/b/binutils/binutils-2.25.1.eb
+++ b/easybuild/easyconfigs/b/binutils/binutils-2.25.1.eb
@@ -28,6 +28,17 @@ prebuildopts = preconfigopts
 # --enable-plugins should be used whenever --enable-gold is used, see http://llvm.org/docs/GoldPlugin.html
 configopts = '--with-sysroot=/ --enable-gold --enable-ld=default --enable-plugins'
 
+# libtool library
+# /home/users/p01991/EB/easybuild/software/binutils/2.25.1/bin/ld: 
+# /home/users/p01991/EB/easybuild/software/binutils/2.25.1/lib/libbfd.a(     format.o): relocation R_X86_64_32 against `.rodata.str1.1' 
+# can not be used when making a shared object; recompile with -fPIC
+# /home/users/p01991/EB/easybuild/software/binutils/2.25.1/lib/libbfd.a: error adding symbols: Bad value
+# collect2: error: ld returned 1 exit status
+# make[4]: *** [libscorep_measurement.la] Error 1
+configopts += ' --enable-shared --with-pic'
+configopts += ' LDFLAGS=-L../libiberty/pic'
+# TODO: check lib/libbfd.so
+
 binlist = ['addr2line', 'ar', 'as', 'c++filt', 'elfedit', 'gprof', 'ld', 'ld.bfd', 'ld.gold', 'nm',
            'objcopy', 'objdump', 'ranlib', 'readelf', 'size', 'strings', 'strip']
 sanity_check_paths = {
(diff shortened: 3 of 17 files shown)


jgphpc:
I am trying to build Cube/434 using gcc/434:

https://github.com/hpcugent/easybuild-easyconfigs/pull/2928/commits/9ed5c87258f3197e6b49db21f9764c707c8ec428 => easyconfigs/c/Cube/Cube-4.3.4.eb

but it fails with:
```
Compiler family DUMMY not supported yet (only: GCC, Intel, IBMXL)
```

@geimer could you help me modify the easyblock/scorep.py ?

```
module rm PrgEnv-cray/5.2.82 craype/2.5.4
eb binutils-2.25.1.eb -r        # dummy toolchain
eb libunwind-1.1.eb             # dummy
eb SIONlib-1.6.1-tools.eb       # dummy
eb OPARI2-2.0.eb                # dummy
eb Qt-4.8.6.eb                  # dummy / (gcc/434 does not support Qt/487 or Qt/5x)
eb Cube-4.3.4.eb                # dummy / qt/486
```

geimer:
@jgphpc: Please take a look at https://github.com/hpcugent/easybuild-easyblocks/pull/889/commits/860ad6ce9d7f5a217553e070cc7477effaa2c233 This should do the trick for using the `dummy` toolchain.

boegel:
Jenkins: ok to test

jgphpc:
@geimer Thanks. I would need the same patch for easyblocks/p/pdt.py

geimer:
@jgphpc: See https://github.com/hpcugent/easybuild-easyblocks/pull/894  There was actually another bug in there (the configure flag for the compiler was determined, but never passed on to configure).

jgphpc:
@geimer i have a few remaining issues, first one is about _libscorep_measurement.la_:

* eb Score-P-2.0.1-CrayGNU-2016.03.eb
```
== processing EasyBuild easyconfig 
~/p01991/pug.git/scorep/easybuild_easyconfigs/CRAYSWAN/Score-P-2.0.1-CrayGNU-2016.03.eb
== building and installing Score-P/2.0.1-CrayGNU-2016.03...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== testing...
== installing...
== FAILED: Installation ended unsuccessfully (build directory: 
/dev/shm/p01991/ScoreP/2.0.1/CrayGNU-2016.03): 
build failed (first 300 chars): cmd " make install " exited with exitcode 2
```

The log says:
```
build-backend:
libtool:   error: cannot install 'libscorep_measurement.la' to a directory not ending in 
/home/users/p01991/EB/easybuild/software/OTF2/2.0/lib/backend
```

My workaround is to manually edit build-backend/libscorep_measurement.la and replace the libdir= line with:
```
libdir='path_to_install_dir/lib/backend'
instead of:
libdir='/home/users/p01991/EB/easybuild/software/OTF2/2.0/lib/backend'
```

Do you know how to fix this from the easyconfig file ?

geimer:
@jgphpc: I currently have no clue how to fix this in the easyconfig.  However, the error suggests to me that something went wrong during `configure`, as Cray is supported by the Score-P build system and should just work out-of-the-box (if the Cray is detected correctly). Likewise, it should not be necessary to pass `CC` and `CXX` via `configopts`. Could you provide me access to the EasyBuild logs to take a closer look?

jgphpc:
Here is the log showing the issue with the .la file, i believe i asked
scorep support in the past but i am unable to find it in my emails.

I use configopts for a different issue with the configure:
        gcc: error: unrecognized command line option '-craype-verbose'

boegel:
@jprorama to deal with the latter problem, you can consider unsetting `$CFLAGS` and `$CXXFLAGS` through `preconfigopts`...

EasyBuild is defining those assuming that the Cray compiler wrappers will be used.

geimer:
@jgphpc: It's unclear why the Score-P build is trying to install something into the OTF2 installation directory.  I will try to reproduce manually this with a non-EB build on daint.

The `-craype-verbose` option is most likely induced by EasyBuild, see https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/toolchains/compiler/craype.py#L76, so I think this issue has to be addressed there...

boegel:
We probably need to support (back) for specifying that the Cray wrappers should be bypassed...

cc @pforai

jgphpc:
@boegel the dependency seems to be ignored (qt not loaded => qmake not found):
```
dependencies = [ ('Qt', '4.8.6'), ]
preconfigopts = 'export PATH=$EBROOTQT/bin:$PATH && '
```
Is this expected because of the dummy toolchain ? 
Is there a way to workaround it (apart from loading Qt from the command line) ?

jgphpc:
Hi @markus , i installed the dependencies for daint: 
```module use /apps/common/UES/SLES11/easybuild/modules/all```.
The log file for:
```eb /apps/common/UES/sandbox/jgp/easybuild.forkjg/easybuild-easyconfigs/easybuild/easyconfigs/easyconfigs/s/Score-P/Score-P-2.0.1-CrayGNU-2015.11.eb```
can be found in: 
```
== FAILED: Installation ended unsuccessfully 
(build directory: /dev/shm/piccinal/ScoreP/2.0.1/CrayGNU-2015.11): 
build failed (first 300 chars): Sanity check failed: 
no file of ('lib/libscorep_adapter_mpi_event.a', 'lib64/libscorep_adapter_mpi_event.a') in 
/apps/daint/UES/5.2.UP04/easybuild/software/Score-P/2.0.1-CrayGNU-2015.11, 

== Results of the build can be found in the log file 
/tmp/eb-ge9x3S/easybuild-Score-P-2.0.1-20160424.165817.jtLyw.log
```

**Comment by geimer** [comment]
@jgphpc: The issue with the `-craype-verbose` is indeed a bug in the Score-P build system.  Building the `scorep-score` tool should ignore the `CFLAGS` and `CXXFLAGS` variables, as it may be (and in this case actually is) built with a totally different compiler.  We ran into a similar issue on Jureca recently, where Cube was built with the `dummy` toolchain and Score-P should be built with Intel. We've already fixed this in Score-P trunk and will include this fix in the next bugfix release.  Until then, overriding `CFLAGS` and `CXXFLAGS` is the only viable choice.

The failing sanity check can be fixed by checking for `lib/backend/libscorep_...` rather than `lib/libscorep_...` (and potentially `lib64` as well).

BTW: I was not yet able to try the build myself, as some of the external prerequisites (papi/5.4.3.1, cray-libsci/16.03.1, cray-mpich/7.3.2) don't seem to be publicly available on daint.

**Comment by geimer** [comment]
@jgphpc: With the latest https://github.com/hpcugent/easybuild-easyblocks/pull/889 and https://github.com/hpcugent/easybuild-easyconfigs/pull/2892, the Cray build should now actually work even w/o messing around with `CC`, `CFLAGS` & Co. in the Score-P easyconfig file.  Please update the sanity check as outlined in my last comment and try again.  Thanks!

File: easybuild/easyconfigs/c/CrayGNU/CrayGNU-2016.03.eb [hunk 1]
```diff
@@ -0,0 +1,19 @@
+easyblock = 'CrayToolchain'
+
+name = 'CrayGNU'
+version = '2016.03'
```
boegel: @jgphpc I'd prefer is this is added via it's own PR, with an HPL easyconfig accompanying it

c @gppezzi

easybuild/easyconfigs/l/libunwind/libunwind-1.1.eb
```diff
@@ -0,0 +1,24 @@
...
+ preserved (callee-saved) state of each call-frame and to resume execution at any point in the call-chain
+ (non-local goto). The API supports both local (same-process) and remote (across-process) operation.
+ As such, the API is useful in a number of applications"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+dependencies = [ ('binutils', '2.25.1'), ]
+
+sources = [SOURCE_TAR_GZ]
+source_urls = [GNU_SAVANNAH_SOURCE]
+
```
boegel: @jgphpc libunwind depends on XZ, see also https://github.com/geimer/easybuild-easyconfigs/pull/4 and https://github.com/hpcugent/easybuild-easyconfigs/pull/3055

jgphpc:
moving to a new PR: https://github.com/hpcugent/easybuild-easyconfigs/pull/3501

PR #2944, opened 2016-04-23
also test with Lmod as modules tool on Travis

boegel:
just like https://github.com/hpcugent/easybuild-easyblocks/pull/897


boegel:
except that here, no bugs were uncovered.. :-)

PR #2950, opened 2016-04-26
[phys]{intel/2016.02-GCC-4.9} VASP 5.4.1

migueldiascosta:
Not sure if you're interested in this one (if not, just go ahead and reject the pull request).

On one hand, it's proprietary, so you can only test it if you have access to the package and put it the sources folder (but there are other cases like this in the repository).

On the other hand, again, I couldn't get completely away from some nasty overrides.

### DIFF
--- a/easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
+++ b/easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
@@ -1,0 +1,31 @@
+easyblock = 'MakeCp'
+
+name = 'VASP'
+version = '5.4.1'
+
+homepage = 'http://www.vasp.at'
+description = """The Vienna Ab initio Simulation Package (VASP) is a computer program for atomic scale
+materials modelling, e.g. electronic structure calculations and quantum-mechanical molecular dynamics,
+from first principles."""
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+
+prebuildopts = 'cp arch/makefile.include.linux_intel ./makefile.include; '
+prebuildopts += 'env LIBS="" '
+
+buildopts = 'all BLACS="$LIBBLACS" '
+buildopts += 'OBJECTS="fftmpiw.o fftmpi_map.o fftw3d.o fft3dlib.o $FFTW_LIB_DIR/libfftw3xf_intel.a"'
+
+parallel = 1
+
+files_to_copy = [(['bin/vasp_std', 'bin/vasp_gam', 'bin/vasp_ncl'], 'bin')]
+
+sanity_check_paths = {
+    'files': ['bin/vasp_std', 'bin/vasp_gam', 'bin/vasp_ncl'],
+    'dirs': []
+}
+
+moduleclass = 'phys'
--- a/easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
+++ b/easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
@@ -17,7 +17,7 @@ sources = [SOURCELOWER_TAR_GZ]
 prebuildopts = 'cp arch/makefile.include.linux_intel ./makefile.include && '
 
 # path to libfftw3xf_intel.a is hardcoded in makefile.include
-prebuildopts += 'sed -i "s|\$(MKLROOT)/interfaces/fftw3xf|\$FFTW_LIB_DIR|" makefile.include && '
+prebuildopts += 'sed -i "s|\$(MKLROOT)/interfaces/fftw3xf|\$(FFTW_LIB_DIR)|" makefile.include && '
 
 # VASP uses LIBS as a list of folders
 prebuildopts += 'unset LIBS && '
--- a/easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
+++ b/easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
@@ -16,10 +16,13 @@ sources = [SOURCELOWER_TAR_GZ]
 
 prebuildopts = 'cp arch/makefile.include.linux_intel ./makefile.include && '
 
-buildopts = 'all LIBS="lib" BLACS="$LIBBLACS" '
+# path to libfftw3xf_intel.a is hardcoded in makefile.include
+prebuildopts += 'sed -i "s|\$(MKLROOT)/interfaces/fftw3xf|\$FFTW_LIB_DIR|" makefile.include && '
 
-# override path to libfftw3xf_intel.a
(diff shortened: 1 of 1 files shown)


boegel:
Jenkins: ok to test

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,31 @@
...
+description = """The Vienna Ab initio Simulation Package (VASP) is a computer program for atomic scale
+materials modelling, e.g. electronic structure calculations and quantum-mechanical molecular dynamics,
+from first principles."""
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+
+prebuildopts = 'cp arch/makefile.include.linux_intel ./makefile.include; '
```
boegel: please use `&&` rather than `;` to glue things together

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,31 @@
...
+prebuildopts += 'env LIBS="" '
+
+buildopts = 'all BLACS="$LIBBLACS" '
+buildopts += 'OBJECTS="fftmpiw.o fftmpi_map.o fftw3d.o fft3dlib.o $FFTW_LIB_DIR/libfftw3xf_intel.a"'
```
boegel: @migueldiascosta I think this one needs a bit of explaining? can you include a comment on the line above that explains why this is needed?

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,31 @@
...
+materials modelling, e.g. electronic structure calculations and quantum-mechanical molecular dynamics,
+from first principles."""
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+
+prebuildopts = 'cp arch/makefile.include.linux_intel ./makefile.include; '
+prebuildopts += 'env LIBS="" '
```
boegel: same here, what's this about? please explain it

boegel:
@migueldiascosta although this concerns licensed software, it's still useful for people who have access to it

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,31 @@
...
+
+homepage = 'http://www.vasp.at'
+description = """The Vienna Ab initio Simulation Package (VASP) is a computer program for atomic scale
+materials modelling, e.g. electronic structure calculations and quantum-mechanical molecular dynamics,
+from first principles."""
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+sources = [SOURCELOWER_TAR_GZ]
```
boegel: maybe include a comment explaining why we don't auto-download this, including a URL with more information on getting access to the sources?

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
migueldiascosta: well, before including a comment, the issue is that the provided makefile.include has the path to libfftw3xf_intel.a hardcoded in the same variable as the other fft objects,

```
OBJECTS    = fftmpiw.o fftmpi_map.o fftw3d.o fft3dlib.o \
             $(MKLROOT)/interfaces/fftw3xf/libfftw3xf_intel.a
```

so I had to override the whole variable. Alternatively, I could have used sed on the makefile.include. Is there a better way?

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
migueldiascosta: easybuild sets `LIBS=-liomp5 -lpthread` but vasp's makefile uses LIBS as a list of folders

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
boegel: There would be a better way if we supported sed-like regex's in the `patches` parameter, but we don't yet.

Using `sed` on `makefile.include` would make the change more clear, and would protect against other changes to `OBJECTS`, so I think we should go with that?

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
boegel: ok, please add a comment to make that clear

also, do you want to use `unset LIBS &&` instead?

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,33 @@
...
+
+toolchain = {'name': 'intel', 'version': '2016.02-GCC-4.9'}
+toolchainopts = {'usempi': True}
+
+# Vasp is proprietary software, see http://www.vasp.at/index.php/faqs on how to get access to the code
+sources = [SOURCELOWER_TAR_GZ]
+
+prebuildopts = 'cp arch/makefile.include.linux_intel ./makefile.include && '
+
+buildopts = 'all LIBS="lib" BLACS="$LIBBLACS" '
```
boegel: is the `LIBS='lib'` correct? now I'm lost... :)

migueldiascosta: it is, but I suppose it's better to use `unset LIBS` in case what's in the makefile changes

boegel: yep, I think so

easybuild/easyconfigs/v/VASP/VASP-5.4.1-intel-2016.02-GCC-4.9.eb
fgeorgatos: hey, am I correct to point that, as regards makefiles, the canonical way would be to feed it as parameter (ie. use `buildopts`) to the make command? give it a shot...

boegel: @fgeorgatos unsetting `$LIBS` is definitely better here (since passing an empty value with `LIBS=''` would still result in overwriting...)

boegel:

PR #2966, opened 2016-04-28
{chem}[intel/2016.02-GCC-4.9] QuantumESPRESSO 5.3.0

migueldiascosta:


boegel:
@migueldiascosta this probably needs the updated easyblock from https://github.com/hpcugent/easybuild-easyblocks/pull/904, just so you know :)

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-5.3.0-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,42 @@
+easyblock = 'ConfigureMake'
```
boegel: @migueldiascosta we have an easyblock dedicated to QuantumESPRESSO, why not use that?

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-5.3.0-intel-2016.02-GCC-4.9.eb
```diff
@@ -0,0 +1,42 @@
...
+
+source_urls = ['http://qe-forge.org/gf/download/frsrelease/204/912/']
+sources = ['espresso-%(version)s.tar.gz']
+checksums = ['6848fcfaeb118587d6be36bd10b7f2c3']
+
+configopts = 'DFLAGS="-D__INTEL -D__MPI -D__PARA -D__DFTI -D__SCALAPACK" '
+configopts += 'FFT_LIBS="$LIBFFT" --with-scalapack=intel '
+configopts += '--prefix=%(installdir)s/bin'
+
+buildopts = 'all'
```
boegel: we usually build QE 'all in', i.e. with `buildopts = 'all w90 want yambo plumed xspectra' `

boegel:
Jenkins: ok to test

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-5.3.0-intel-2016.02-GCC-4.9.eb
migueldiascosta: @boegel ah, that explains a few things :) But in the other easyconfig files no mention is made to that easyblock, is it automatically chosen from the easyconfig name?

boegel:
@migueldiascosta yes, if not `easyblock` spec is present, EasyBuild goes hunting for an easyblock based on the software name (in this case, `quantumespresso.py` that provides the `EB_QuantumESPRESSO` class, i.e. https://github.com/hpcugent/easybuild-easyblocks/blob/master/easybuild/easyblocks/q/quantumespresso.py)

migueldiascosta:
using the updated easyblock, 5.3.0 fails because the package is missing upftools and 5.4.0 fails because p2y is not being built in yambo...

boegel:
@migueldiascosta so, does the easyblock need further updates?

migueldiascosta:
@boegel either create an exception for 5.3.0 when checking upftools in the easyblock, or in the 5.3.0 easyconfig replace `all` with `pw ph pp pwcond neb cp ld1 tddfpt` (without `upf`, 'all' or 'pwall'). In 5.4.0 'all' can be used gain.

Also in 5.3.0, when linking with FFTW3 (not a problem when using DFTI), a patch is needed to remove  fftw_version from FFTXlib (http://qe-forge.org/pipermail/q-e-developers/2016-January/001088.html). No longer needed in 5.4.0.

Additionaly, the configure options for yambo in install/plugins_makefile now (since 5.3.0) target the ones in yambo-3.4.2 (--with-X-libs instead of --with-X). However, the yambo compilation is failing on netcdf missing dependencies, although it built it's own internal netcdf...

migueldiascosta:
@boegel regarding the last point in the previous comment, the problem was with the name mangling in yambo's internal netcdf. It can be fixed by patching that netcdf's Makefile to use `CPPFLAGS=-DpgiFortran` (which also applies to ifort), but I don't know why ifort is not being correctly detected...

boegel:
Test report by @boegel
**SUCCESS**
node2012.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/48bbcce86017645228ba1e2d71c5f27a for a full test report.

boegel:
lgtm

boegel:
Test report by @boegel
**SUCCESS**
node2601.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/5bd35b4e7453a0ae38017898b4de3dda for a full test report.

boegel:

migueldiascosta:
@boegel fyi, we ended up downgrading everything to intel-2016.01 because there is a MKL bug in 2016.02 and 2016.03 [1] that eventually (for large systems) showed up in all the DFT and related apps we use (Espresso, Siesta, Vasp, BerkeleyGW, etc). Intel says it will be fixed in 2016.04.

[1] https://software.intel.com/en-us/forums/intel-clusters-and-hpc-technology/topic/634884

boegel:
@migueldiascosta maybe it's if we adjust this easyconfig to use the `intel/2016.01` easyconfig instead then?

boegel:
@migueldiascosta ping? should we change the toolchain to `intel/2016.01` for this easyconfig, since `2016.02` and `2016.03` are known not to work?

migueldiascosta:
@boegel sorry for the delay. well, I don't know - this bug probably affects any application that calls mkl's scalapack a sufficient number of times (and there is no problem if it doesn't, I suppose it's simply not freeing the communicators), if people are using `2016.02` and `2016.03` maybe the problem is not so frequent. we decided to downgrade, but there may be valid reasons to use the more recent versions, if one is not affected by this bug. my intention was merely to let people now that if the "too many communicators" error appears, the workaround for now is to use `2016.01`.

PR #3017, opened 2016-05-09
flash with foss2016a

RvDijk:
Flash with foss 2016a


boegel:
@RvDijk did you mean to add an easyconfig named `FLASH-1.2.11-foss-2016a.eb`, rather than changing the `toolchain` spec in an existing one?

boegel:
Jenkins: ok to test

RvDijk:
@boegel I'll make a new PR.

PR #3036, opened 2016-05-11
clean up dummy bzip2 easyconfig, define buildopts rather than defining $CC and $CFLAGS via os.environ

boegel:
mainly done to avoid the `import os`...

this requires ~~https://github.com/hpcugent/easybuild-easyblocks/pull/910~~ (with https://github.com/pescobar/easybuild-easyblocks/pull/5 included)

### DIFF
--- a/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb
+++ b/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb
@@ -12,8 +12,8 @@ toolchainopts = {'pic': True}
 sources = [SOURCE_TAR_GZ]
 source_urls = ['http://www.bzip.org/%(version)s/']
 
-import os
-os.environ['CC'] = 'gcc'
-os.environ['CFLAGS'] = '-O3 -fPIC'
+buildopts = "CC=gcc CFLAGS='-Wall -Winline -O3 -fPIC -g $(BIGFILES)'"
+
+with_shared_libs = False
 
 moduleclass = 'tools'
--- a/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb
+++ b/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb
@@ -14,6 +14,7 @@ source_urls = ['http://www.bzip.org/%(version)s/']
 
 buildopts = "CC=gcc CFLAGS='-Wall -Winline -O3 -fPIC -g $(BIGFILES)'"
 
+# disable building of shared libraries, doesn't work everywhere (e.g. on OS X, where 'gcc' is actually Clang...)
 with_shared_libs = False
 
 moduleclass = 'tools'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb
```diff
@@ -12,8 +12,9 @@ toolchainopts = {'pic': True}
 sources = [SOURCE_TAR_GZ]
 source_urls = ['http://www.bzip.org/%(version)s/']
 
-import os
-os.environ['CC'] = 'gcc'
-os.environ['CFLAGS'] = '-O3 -fPIC'
+buildopts = "CC=gcc CFLAGS='-Wall -Winline -O3 -fPIC -g $(BIGFILES)'"
+
+# disable building of shared libraries, doesn't work everywhere (e.g. on OS X, where 'gcc' is actually Clang...)
```
wpoely86: sorry, but not working on OS X is not a good enough reason to disable it by default.

boegel: well, it was disabled already (without the enhanced of the bzip2 easyblock)

I prefer being able to install bzip2 without shared libs over not installing it, especially since the shared libraries are rarely needed anyway (you can tell by i) the way they need to be installed, ii) a `libbz2.so` isn't even installed by default)

wpoely86: If it doesn't work on any Linux system fine, but only OS X is not good enough.

boegel: So, you want me to patch the Makefile for the .so instead so it works on OS X too?

I'm just avoiding a regression, and the `.so`'s are basically pointless for the dummy installation of bzip2

boegel: how about this gem then?

```python
with_shared_libs = OS_TYPE == 'Linux'
```

boegel:
Test report by @boegel
**SUCCESS**
boegel.ugent.be - Darwin UNKNOWN UNKNOWN, Intel(R) Core(TM) i5-4258U CPU @ 2.40GHz, Python 2.7.10
See https://gist.github.com/d0286e8fadc655857a61f576d38cdb7b for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2571.golett.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/bde29874f2dcc029eb085512007989f5 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2014.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/6ebcba6ee137927e04616b8017143c40 for a full test report.

boegel:
Going in, thanks for the review @wpoely86!

PR #3124, opened 2016-05-25
ADD HDF/4.2.11-foss-2015a

EvanFishbein:


boegel:
@EvanFishbein this is probably not the pull request you intended to make, see also https://github.com/hpcugent/easybuild-easyconfigs/pull/3120#issuecomment-221687614

PR #3168, opened 2016-06-02
Adding new easyconfigs + a few minor fixes

besserox:
Please let me know if you prefer these changes in different PRs

New software:
- Bullet
- Voro++

New software versions:
- Eigen 3.2.8
- SuiteSparse 4.5.2
- SuperLU 5.1.1 (based on SuperLU easyblock)
- SuperLU 4.3 (based on ConfigureMake easyblock)
- GDB 7.11
- Valgrind 3.11.0
- R 3.2.3 bare
- pkgconfig

Minor fixes:
- h5py (fix `pyver` variable issue, strangely before it crashed on some machines and not others)
- Paraview 4.4.0 with foss failed to compile with shipped HDF5 -> Properly use HDF5 dependencies instead and update Paraview 4.4.0 with intel in the same way

### DIFF
--- a/easybuild/easyconfigs/b/Bullet/Bullet-2.83.7-foss-2016a.eb
+++ b/easybuild/easyconfigs/b/Bullet/Bullet-2.83.7-foss-2016a.eb
@@ -1,0 +1,29 @@
+easyblock = 'CMakeMake'
+
+name = 'Bullet'
+version = '2.83.7'
+
+homepage = "http://bulletphysics.org/"
+description = "Bullet professional 3D Game Multiphysics Library provides state of the art collision detection, soft body and rigid body dynamics."
+
+toolchain = {'name': 'foss', 'version': '2016a'}
+toolchainopts = {'usempi': False, 'pic': True}
+
+source_urls = ['https://github.com/bulletphysics/bullet3/archive/']
+sources = ["%(version)s.tar.gz"]
+
+builddependencies = [('CMake', '3.5.2')]
+
+# build shared libraries
+configopts = "-DBUILD_SHARED_LIBS=ON"
+
+sanity_check_paths = {
+    'files': ['include/bullet/btBullet%sCommon.h' % x for x in ['Collision', 'Dynamics']]
+           + ['lib/libBullet%s.%s' % (x,SHLIB_EXT) for x in ['Collision', 'Dynamics']],
+    'dirs': [],
+}
+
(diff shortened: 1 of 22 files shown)


boegel:
@besserox That's quite a lot of stuff in a single PR, including things that are not related to each other...

The issue with `pyver` in the `h5py` easyconfig is probably because `%(pyver)s` requires a recent version of EasyBuild (v2.7.0 or newer).

Would you mind opening separate PRs for the new software, and also for the Paraview HDF5 fix?

If the others are only new software versions for existing toolchains, they can be packed together since that's easy to review via `--review-pr`. You will need to undo the `pyver` stuff though...

I know that having multiple PRs in flight can be a pain, but maybe you can use the experimental `--new-pr` and `--update-pr` support, see http://boegel-eb.readthedocs.io/en/new_update_pr/Integration_with_GitHub.html for up-to-date documentation (about to be merged in the stable docs).

besserox:
@boegel No problem, I will separate this in different PRs.

One question: Why do you want me to undo the `pyver` stuff. I believe that's better to have something that works everywhere, no? Also the cluster where it was failing uses Python 2.7.3 (and has also other Python version installed). Could it be that EasyBuild picks the wrong version of Python?

boegel:
@besserox Using new (or updated) easyconfigs with older versions of EasyBuild is not guaranteed to work, i.e. is not forward-compatible (we do try hard to be backward-compatible though).

The `%(pyver)s` template is resolved to the version of Python that is included as a dependency, so it should never pick up the wrong version; see https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/framework/easyconfig/templates.py#L196

besserox:
I was on recent version of master branch to test this, so I guess EasyBuild 2.8.0

Also, my Python path only contains paths to EB directories: `easybuild-easyconfigs:easybuild-easyblocks:easybuild-framework:vsc-install/lib:vsc-base/lib`. (normaly all updated to master branch)

And also:
```
 $ python --version
Python 2.7.3
```

OK, I will try again and open a dedicated issue I still have it.

boegel:
@besserox you shouldn't experience issues w.r.t. `%(pyver)s` with EasyBuild v2.7.0; if you do, you're hitting a bug (or something else is going on). Please do open an issue if you're still seeing it.

besserox:
PR split in PR #3174, #3175, #3177, #3178.

boegel:
Thanks @besserox!

PR #3173, opened 2016-06-04
EasyConfigs for Anaconda and Conda Env

jerowe:
This is a requirement of https://github.com/hpcugent/easybuild-easyblocks/pull/950.

Easyconfigs for Anaconda and CondaCreate.


cfenoy:
Due to the way EasyBuild guesses LDFLAGS directories the lib directories of Anaconda will be added to the LDFLAGS of any software depending on Anaconda. Have you found any issue when compiling software with Anaconda as a dependency?

The "problematic" code is in:
easybuild-framework-2.7.0/easybuild/tools/toolchain/toolchain.py:620

jerowe:
@cfenoy, I didn't realize that LDFLAGS was there, but I haven't had any problems. Anaconda only encounters problems at LD_LIBRARY_PATH, or at least as far as I know.

For instance, loading the anaconda module allows for conda install -c bioconda perl-threaded, which installs gcc and compiles perl with no difficulties.

easybuild/easyconfigs/c/CondaCreate/requirements.txt
```diff
@@ -0,0 +1 @@
+perl-app-cpanminus
```
fgeorgatos: this to be deleted?

easybuild/easyconfigs/c/CondaCreate/condacreate-1.0.eb
```diff
@@ -0,0 +1,37 @@
...
+
+builddependencies = [('anaconda', '2-4.0.0')]
+
+sanity_check_paths = {
+    'files': ["bin/cpanm"],
+    'dirs': []
+}
+
+moduleclass = 'tools'
+##Trailing comment
```
fgeorgatos: delete line

easybuild/easyconfigs/c/CondaCreate/condacreate-1.0.eb
```diff
@@ -0,0 +1,37 @@
...
+##
+
+easyblock = 'EB_CondaCreate'
+
+name = "condacreate-env"
+version = "1.0"
+variant = "Linux-x86_64"
+
+homepage = "https://github.com/bioconda/bioconda-recipes"
+description = """ Conda is a platform- and language-independent package manager that sports easy distribution, installation and version management of software. The bioconda channel is a Conda channel providing bioinformatics related packages. This repository hosts the corresponding recipes. """
```
fgeorgatos: break along several lines, for readability.

otherwise, lgtm!

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
```diff
@@ -0,0 +1,42 @@
...
+
+easyblock = 'EB_anaconda'
+
+name = "anaconda"
+version = "2-4.0.0"
+variant = "Linux-x86_64"
+
+homepage = "https://www.continuum.io/"
+description = """Built to complement the rich, open source Python community,
+the Anaconda platform provides an enterprise-ready data analytics platform
```
fgeorgatos: 1-space indent helps readability

cfenoy:
@jerowe I meant, If you've tried to compile an application that depends on python and other libraries with easybuild. We've found issues when compiling applications that depend on python and cairo. We've found other issues but I can not remember now, I'll try to find an example and give it to you so you can test if it is failing

easybuild/easyconfigs/c/CondaCreate/requirements.txt
jerowe: Its up to you. I wanted to include a sample of a requirements file, but all its just a line separated by line of conda packages to install.

If you prefer it could be an option supplied by EB_condaCreate

`conda_packages: ['perl-app-cpanminus', 'picard', 'samtools']` ?

I went with requirements.txt because it seems to be a convention of the python community, but I also like the idea of having the whole thing contained in the eb config.

jerowe:
@cfenoy, I haven't run into this issue, at least not with anaconda. I used to have issues using just python to compile things, but I haven't tried the easybuild python.

If you give me an example I'll test it out.

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
jerowe: Will do.

easybuild/easyconfigs/c/CondaCreate/requirements.txt
boegel: @jerowe I think this suggestion makes sense, keep it contained to the easyconfig file itself

jerowe: I'm trying toimplement that, but don't I have to have a source?

boegel: @jerowe not sure what you mean... how `conda_packages` is interpreted is up to the `EB_CondaCreate` easyblock?

jerowe: Any update? Is there a way to not require a source?

jerowe:
Any updates? Is there any chance this would be merged by mid August?

easybuild/easyconfigs/c/CondaCreate/requirements.txt
boegel: @jerowe you can update the easyblock to just skip the extract step, by adding:

```python
def extract_step(self):
    """No sources provided, so nothing to unpack."""
    pass
```

I'm a bit weary of this though, since that means this will require you to be online all the time.

Also, shouldn't we enforce people to specify the packages to install with a specific version?

boegel:
@jerowe sorry for not getting back to you on this...

I haven't had time to take a detailed look at this yet, let alone play around with it myself.

I've tagged this for EasyBuild v2.9.0, which is planned for Sept'16 (rough timeline).

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
```diff
@@ -0,0 +1,42 @@
+##
+# This is an easyconfig file for EasyBuild, see https://github.com/hpcugent/easybuild
+#
+# Authors::   Jillian Rowe <jillian.e.rowe@gmail.com>
+#
+# This work is based on experiences from the UNITE project
+# http://apps.fz-juelich.de/unite/
+##
+
+easyblock = 'EB_anaconda'
```
boegel: you can drop this, EB will derive this from the software `name`

easybuild/easyconfigs/c/CondaCreate/condacreate-1.0.eb
```diff
@@ -0,0 +1,37 @@
...
+#
+# This work is based on experiences from the UNITE project
+# http://apps.fz-juelich.de/unite/
+##
+
+easyblock = 'EB_CondaCreate'
+
+name = "condacreate-env"
+version = "1.0"
+variant = "Linux-x86_64"
```
boegel: use `versionsuffix`?

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
```diff
@@ -0,0 +1,42 @@
...
+
+name = "anaconda"
+version = "2-4.0.0"
+variant = "Linux-x86_64"
```
boegel: use `versionsuffix`?

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
```diff
@@ -0,0 +1,42 @@
...
+
+homepage = "https://www.continuum.io/"
+description = """Built to complement the rich, open source Python community,
+the Anaconda platform provides an enterprise-ready data analytics platform 
+that empowers companies to adopt a modern open data science analytics architecture.
+"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+anaconda_version = 2
```
boegel: not used, so please remove?

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
```diff
@@ -0,0 +1,42 @@
...
+
+source_urls = ["http://repo.continuum.io/archive/Anaconda%s-%s.sh" % (version, variant)]
+sources = ["Anaconda%s-%s.sh" % (version, variant)]
+checksums = [ '31ed3ef07435d7068e1e03be49381b13' ]
+
+license_file = 'LICENSE.txt'
+
+sanity_check_paths = {
+    'files': ["bin/pydoc"],
```
boegel: that's the only binary that is shipped?

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
```diff
@@ -0,0 +1,42 @@
...
+    'dirs': []
+}
+
+modextravars = {
+    'ANACONDA_LICENSE': license_file,
+}
+
+moduleclass = 'lang'
+##Trailing comment
```
boegel: please remove...

easybuild/easyconfigs/c/CondaCreate/requirements.txt
jerowe: Thanks for the point in the right direction. ;-)

I think if people are creating conda environments they know being on the internet is required.

I'm not sure what you mean by that.

```
conda install my-package #defaults to latest
conda install my-package=1.2 #install a specific version
```

Actually, let me make sure the config and easyblocks I have here are up to date. I did some work on them recently and I'm not sure I pushed that work.

easybuild/easyconfigs/c/CondaCreate/condacreate-1.0.eb
jerowe: Instead of variant?

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
jerowe: Actually, I wanted to ask about this.

I want my users to be able to specify a a version, and in my EB I would have a dict of appropriate URLs.

2 would default to the latest 2.0 version (2.4 as of now)
3 would default to the latest 3.0 version (3.4 as of now)

Otherwise they could specify the exact version.

What do you think?

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
jerowe: There are a few more that are shipped. I'll update this.

easybuild/easyconfigs/c/CondaCreate/requirements.txt
boegel: I think we should avoid or at the very least discourage people to let conda install the latest version, part of the idea behind EasyBuild is that you can easily reproduce previous installations...

easybuild/easyconfigs/c/CondaCreate/condacreate-1.0.eb
boegel: yes, because `versionsuffix` means something in EasyBuild, while `variant` is just a local variable in the easyconfig

You'll need to adjust the easyconfig filename accordingly to make the test suite happy.

easybuild/easyconfigs/a/Anaconda/anaconda-2.4.0.eb
boegel: This doesn't match the EB way of doing things... Today, if you want to install a different version of Anaconda, you need to compose a different easyconfig file and adjust the `version` spec.

jerowe: Alright then, that's fine. Lets stick to the EB way of doing things. ;-)

jerowe:
Sorry, I have a ton of case conflicts in here somehow. Let me work on those and I'll update the other things.

jerowe:
I think I killed this. Should I just open up a new branch?

jerowe:
I accidentally closed and reopened it here.

https://github.com/hpcugent/easybuild-easyconfigs/pull/3337

PR #3199, opened 2016-06-10
Easy config: NAMD-2.11-intel-2016a-mpi.eb

jhein32:
Proposal for building a NAMD 2.11 cpu version, parallelised using MPI. Using the Intel 2016a toolchain.
Using input from Alan O'Cais, Jülich Supercomputing centre, communicated via the EasyBuild mailing list.


jhein32:
@boegel Like this?

boegel:
@jhein32 :+1:

boegel:
Test report by @boegel
**FAILED**
node2497.golett.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/c13e3721754e74d864e5c966ccc0ceba for a full test report.

jhein32:
@boegel You manually need to put the tarball in place.  Access is password protected - you can register free of charge.  Similar like an intel compiler.

boegel:
@jhein32 it is, already fixing it, thanks

boegel:
Test report by @boegel
**SUCCESS**
node2062.delcatty.os - Linux SL 6.7, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/229c5313d2c1058fc107fdf42e1ce609 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2496.golett.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/d8873c7bd52792e681208bdd5ccf1337 for a full test report.

jhein32:
You need to do all these tests manually?

boegel:
@jhein32: the unit tests for the easyconfigs (which check for things like dependencies, version conflicts, existence of patch files) are run automatically by Travis

The test reports for the actual test builds are triggered manually, yes, but all we basically do is `eb --from-pr 3199 -dfr --upload-test-report`, and EasyBuild takes care of reporting back the test report into the PR.

boegel:
@jhein32 test builds work, that's 90% of the work

Please see https://github.com/jhein32/easybuild-easyconfigs/pull/1 for some trivial style fixes; I'm assuming the hardcoded `parallel = 12` is not required to make the build actually work?
EasyBuild will pick a proper setting for `parallel` by default based on the number of available cores and `ulimit` settings.

But, there are a couple of things I'm not entirely happy with here, maybe @ocaisa can shed some light on them:

* disabling the tests using `runtest = False`; is that because the tests are broken, or just to speed up the installation?

* the hardcoded `-ltcl8.6`, and actually just the fact that `Make.config` is 'fixed'; this should be tackled in the NAMD easyblock instead, once and for all? Why is this needed exactly anyway? It may be an indication of something else going wrong...

jhein32:
Hi Kenneth, @boegel 

As an input for your bullet point 1, I commented the runtest line and then it fails the test on our system.  I get and error 139, which seems to point to seg-fault - not so nice.  I assume some work is needed here.

For the test it seems to want to run an mpi job on our frontend, using a quite complex mpi joblaunch command.  Where/how in the EB set up do I control the MPI job launcher?  I am currently hoping it is something as trivial as that.

fgeorgatos:
@jhein32: please merge PR https://github.com/jhein32/easybuild-easyconfigs/pull/1

boegel:
@jhein32 controlling the MPI job launcher to use is likely very specific to NAMD (some configure option), there's no generic way in EasyBuild to do this...

jhein32:
@boegel Somehow the setup needs to start the namd executable to do the test.  In case of an MPI-code, such as NAMD, one needs a job launcher to start the executable.  I am aware that if none is given the "system" typically pulls one in for you.   How is EasyBuild starting the exec?  Just executable name?  I was wondering whether there is (or we can have) an environment variable or command line option to explicitly specify an mpi job launcher?

jhein32:
@boegel @fgeorgatos Regarding the merging, could you please be more specific here?  I am still on a learning curve regarding git (CVS user myself).  Alternatively could kenneth do this as the repository owner?

boegel:
@jhein32 if you visit https://github.com/jhein32/easybuild-easyconfigs/pull/1, you can merge my pull request to *your* branch, after reviewing the changes I made (see https://github.com/jhein32/easybuild-easyconfigs/pull/1/files)

I can not merge it for you, since I'm issuing a PR to *your* repo in this case.

W.r.t. the MPI launcher: apparently the NAMD easyblock is indeed trying to set up a NAMD test run itself, cfr. https://github.com/hpcugent/easybuild-easyblocks/blob/master/easybuild/easyblocks/n/namd.py#L122 .

It is using the `mpi_cmd_for` function provided by the EasyBuild framework for this (see https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/tools/toolchain/mpi.py#L169), which may need some love/work in this particular case, it's not terribly advanced...

jhein32:
@boegel @fgeorgatos Ok, I have merged this now.  It is currently testing (according to the webpage).  

Regarding MPI and testing, it is a bigger issue and easybuild needs to be flexible.  Looking at section of mpi.py, it seems to me, that depending on MPI library it has a default, which is fine.  However depending on the system the default may not apply.  E.g. when using slurm it is often preferred to use [srun to launch jobs ](http://slurm.schedmd.com/mpi_guide.html#intel_mpi).  Also starting MPI tests as part of the build will not work in many cases.  On many systems the frontend nodes are utilised to compile/link software while MPI running is only possible or permitted on the backend nodes.  The current behaviour is fine for a default.  However easy customisation is required to
- select a different mpi launcher - e.g. set an env-variable and have mpi_cmd_for listen to that, if set
- have the option in easybuild to split the task of build and testing into two tasks.  The first task builds and installs the second tasks performs the tests.  This would allow to executed task on on the frontend, while using a batch script to perform the second task.  
- document this in an easy to notice/access location 

When building software outside easybuild, I typically work along the above line, if the application provides a test-suite that requires MPI running.  Amber is an example. I installed that before supported in easybuild so I have not checked/reviewed the testing EasyBuild does.   In my experience the tests done by easybuild are typically simpler than running the full fledged testsuite.

I think the topic of testing MPI executables requires it's own discussion - this is not part of NAMD.  We should decide on the level of testing we require for NAMD.  If we engage MPI running in the current state of affairs I could not use the config and would resort to building it outside EB.

ocaisa:
I agree that we need an easy  way to customise the MPI launcher, currently for our slurm case we fake it with an alias. For this particular case, you could (configure and) use `--job` to do the build and run the tests. In fact there are very few cases that can't be built on the backend, we try to do this by default to ensure that packages from the OS are not being silently picked up when building on the front end (since they  may cause things to fail on the backend).

boegel:
@ocaisa's suggestion is indeed the best we can do for now, we also try to build&install all of our software on our workernodes (although for us it doesn't matter too much)

Adding support to be able to specify a custom MPI launcher is not terribly difficult though, it's a matter of defining a configuration option for it and making `mpi_cmd_for` aware of it.

Splitting of the test step to be run on the backend is something we have considered before as well, (e.g. via GC3Pie), but that's less trivial. Just building on the backend is then probably far easier.
Is there any reason that won't work in your setup @jhein32?

@jhein32: For this particular PR, we should enable the running of the test, and leave disabling it as a site customisation until we have a mechanism in place to customise the MPI launcher command to be used?

jhein32:
On 11 Aug 2016, at 12:08, Kenneth Hoste <notifications@github.com<mailto:notifications@github.com>> wrote:


@ocaisa<https://github.com/ocaisa>'s suggestion is indeed the best we can do for now, we also try to build&install all of our software on our workernodes (although for us it doesn't matter too much)

Adding support to be able to specify a custom MPI launcher is not terribly difficult though, it's a matter of defining a configuration option for it and making mpi_cmd_for aware of it.

Splitting of the test step to be run on the backend is something we have considered before as well, (e.g. via GC3Pie), but that's less trivial. Just building on the backend is then probably far easier.
Is there any reason that won't work in your setup @jhein32<https://github.com/jhein32>?


We currently do not have the yum installed dev-packages for stuff EB expects from the OS on the backend nodes.

I am actually surprised that this issue has not been raised before.  Most services I have supported since 2002 would require building on the frontend and testing of MPI jobs on the backend.


@jhein32<https://github.com/jhein32>: For this particular PR, we should enable the running of the test, and leave disabling it as a site customisation until we have a mechanism in place to customise the MPI launcher command to be used?


For the time being, I would go for a simpler test, as is common for EB.  Most packages I install out of easybuild use MPI, but very few seem to test the package by starting an MPI executable.  Otherwise I would have discussed test failures before ;)

In the long run, one wants to arrive at doing testing by running MPI executables.  However this can be a can of worms.  Full fledged test suites often require human intervention.  An example is the Amber test suite (competitor to NAMD).  Takes an eternity to run and one often needs to consult with an expert to understand whether a triggered “failure” is indeed a failure or the tolerances are just to tight.

Best wishes
  Joachim

jhein32:
@boegel 

We currently do not have the yum installed dev-packages for stuff EB expects from the OS on the backend nodes.

I am actually surprised that this issue has not been raised before.  Most services I have supported since 2002 would require building on the frontend and testing of MPI jobs on the backend.  

For the time being, I would go for a simpler test for NAMD, as is common for EB.  Most packages I install out of easybuild use MPI, but very few seem to test the package by starting an MPI executable.  Otherwise I would have discussed test failures before ;)

In the long run, one wants to arrive at doing testing by running MPI executables.  However this can be a can of worms.  Full fledged test suites often require human intervention.  An example is the Amber test suite (competitor to NAMD).  Takes an eternity to run and one often needs to consult with an expert to understand whether a triggered “failure” is indeed a failure or the tolerances are just to tight.

Best wishes
  Joachim

boegel:
@jhein32 in this case, I think it would make sense to simplify the quick test for NAMD, indeed (in the NAMD easyblock), since running the MPI-able NAMD executable is indeed hard to get right everywhere...

jhein32:
How would one do that?  As far as I can make out, the easyblocks live in a different repository.  Could I easily add this to this PR?

j.

boegel:
@jhein32 no, you'd need to make a separate PR to the easyblocks repository, and include a reference to it here (just copy-paste the PR URL)

jhein32:
@boegel This is still pending.  I understand 2.11 is still the "official" NAMD version, so no bump required.  I had a look, but struggle with the simplification of the test in the NAMD easyblock.  Not sure when you intend to roll out 2.9.0, but if we want to get NAMD in there quickly I was wondering whether someone could help with the easyblock.

boegel:
@jhein32 please test using `--mpi-cmd-template`, cfr. https://github.com/hpcugent/easybuild-framework/issues/1918

boegel:
@jhein32 ping on testing with `--mpi-cmd-template`?

boegel:
@jhein32 ping, any feedback on using `--mpi-cmd-template`?

Other than that, I think this PR is good to merge...

boegel:
Test report by @boegel
**SUCCESS**
node2532.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/c51ffe6ea0c4c1704c51b949d64ee9e0 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2114.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/4aa57b7e2d82bfdf7d9bdcbfad71263f for a full test report.

boegel:
Tested on top of https://github.com/hpcugent/easybuild-easyblocks/pull/1173, good to go.

PR #3203, opened 2016-06-13
also include vsc-install in list of sources, to support offline installation

boegel:
requires https://github.com/hpcugent/easybuild-easyblocks/pull/957 to actually be picked up


boegel:
*slaps @wpoely86 fingers*

Issue #3242, opened 2016-06-21
Update HMMER source_urls, possibly use github, etc etc


fgeorgatos:
Most EasyBuilders need no introduction to what `source_urls`'s churn rate is; 

HMMER was just added to the pile and now requires:
* wget http://eddylab.org/software/hmmer3/3.0/hmmer-3.0-linux-intel-x86_64.tar.gz

So, either the easyconfigs now require an update or, *there is a better way*:
* For each and every software that now lives on github, use the following construct:
  * source_urls = ['https://github.com/alexdobin/STAR/archive/'] # kudos for the one doing that
  * fi. https://github.com/EddyRivasLab/hmmer/tree/3.0 exists

Going that way is a marvel, because you could now feed a git commit ID at `--try-software-version`!

In fact, this might prove to be such a better approach that it might be worthy of its own templating:
https://github.com/hpcugent/easybuild-framework/blob/9a6df6828bd2e484318afe02d06eb6372f939bfa/easybuild/framework/easyconfig/templates.py#L91
(if so, this should become a framework issue; great advantage: no need to do that much checksums!)

@caylo: Looking for a worthy challenge? ;-)

p.s. 
Permit me stop the ranting here or, I will start on http://www.cython.org/release/ being down today ;-)

boegel:
@fgeorgatos for Cython, see https://github.com/hpcugent/easybuild-easyconfigs/pull/3212

That being said, we're always going to be at the mercy of upstream changing things (or screwing things up, cfr. https://bitbucket.org/pypa/pypi/issues/438/backwards-compatible-un-hashed-package)...

boegel:
fixed in #3246, follow-up there please

PR #3246, opened 2016-06-21
fix homepage & source_urls in HMMER easyconfigs

boegel:
fix for #3242

### DIFF
--- a/easybuild/easyconfigs/h/HMMER/HMMER-3.0-goalf-1.1.0-no-OFED.eb
+++ b/easybuild/easyconfigs/h/HMMER/HMMER-3.0-goalf-1.1.0-no-OFED.eb
@@ -15,7 +15,7 @@ easyblock = 'ConfigureMake'
 name = 'HMMER'
 version = '3.0'
 
-homepage = 'http://hmmer.janelia.org/'
+homepage = 'http://hmmer.org/'
 description = """HMMER is used for searching sequence databases for homologs of protein sequences,
  and for making protein sequence alignments. It implements methods using probabilistic models
  called profile hidden Markov models (profile HMMs).  Compared to BLAST, FASTA, and other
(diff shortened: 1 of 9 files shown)


easybuild/easyconfigs/h/HMMER/HMMER-3.0-goalf-1.1.0-no-OFED.eb
```diff
@@ -27,10 +27,8 @@ description = """HMMER is used for searching sequence databases for homologs of
 
 toolchain = {'name': 'goalf', 'version': '1.1.0-no-OFED'}
 
-# eg. http://selab.janelia.org/software/hmmer3/3.0/hmmer-3.0-linux-intel-x86_64.tar.gz
-versiondir = "hmmer%s" % ".".join(version.split('.')[:-1])
-sources = ['hmmer-%s-linux-intel-x86_64.tar.gz' % version]
-source_urls = ['http://selab.janelia.org/software/%s/%s/' % (versiondir, version)]
+source_urls = ['http://eddylab.org/software/hmmer%(version_major)s/%(version)s/']
+sources = ['hmmer-%(version)s-linux-intel-x86_64.tar.gz']
```
boegel: @fgeorgatos, @verdurin, @rjeschmi, @pescobar: any idea if this is what it's supposed to be?

This is the binary tarball, rather than the source tarball? Cfr. http://hmmer.org/download.html

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/cdc3c63dfb1fe508533ce07cf086a63c for a full test report.

easybuild/easyconfigs/h/HMMER/HMMER-3.0-goalf-1.1.0-no-OFED.eb
fgeorgatos: @boegel : indeed, I was mildly surprised to observe that; versionsuffix at least should be different.
It could be, that there was an assumed performance advantage in the prebuilt binary (I recall an individual mentioning it - don't ask me why it was said so, this would indeed be unusual).

If we change versionsuffix, then a couple of bundles may need to follow;
even better, provide the github-source built alternatives would be easy

fgeorgatos:
@boegel: ty, btw

easybuild/easyconfigs/h/HMMER/HMMER-3.0-goalf-1.1.0-no-OFED.eb
boegel: I'll tackle this in a follow-up issue, I want to include the fixed homepage/source URLs already...

boegel: cfr. #3313

PR #3281, opened 2016-07-01
Update of PSI4 to git version from June 30

tovrstra:
@SebWouters  @wpoely86
I've tested these on my HPC account. Everything compiles without error. (I did not test any further yet but that should be fine. It is a development version anyway.)

### DIFF
--- a/easybuild/easyconfigs/c/CheMPS2/CheMPS2-1.7.1-intel-2016a.eb
+++ b/easybuild/easyconfigs/c/CheMPS2/CheMPS2-1.7.1-intel-2016a.eb
@@ -1,0 +1,31 @@
+easyblock = 'CMakeMake'
+
+name = 'CheMPS2'
+version = '1.7.1'
+
+homepage = 'https://github.com/SebWouters/CheMPS2'
+description = """CheMPS2 is a scientific library which contains a spin-adapted implementation of the
+density matrix renormalization group (DMRG) for ab initio quantum chemistry."""
+
+toolchain = {'name': 'intel', 'version': '2016a'}
+
+sources = ['v%(version)s.tar.gz']
+source_urls = ['https://github.com/SebWouters/CheMPS2/archive/']
+
+builddependencies = [('CMake', '3.5.2')]
+
+dependencies = [
+    ('GSL', '2.1'),
+    ('HDF5', '1.8.17')
+]
+
+runtest = 'test'
+
+separate_build_dir = True
+
+sanity_check_paths = {
+    'files': ['bin/chemps2', 'lib64/libchemps2.%s' % SHLIB_EXT, 'lib64/libchemps2.a'],
+    'dirs': ['include/chemps2']
+}
+
+moduleclass = 'chem'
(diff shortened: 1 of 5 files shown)


wpoely86:
@tovrstra They tagged 1.0rc2 also yesterday?

tovrstra:
Oops. What a coincidence. Good point. I guess that also implies that we have to use "PSI4" as package name in future instead of PSI?

wpoely86:
yes, I tried to convince to keep using increasing version numbers but alas. We should from now on switch to using `PS4/1.0rc2`.

tovrstra:
OK. I'll make a new PR.

wpoely86:
@fgeorgatos The problem is that it's a nightmare to maintain. We currently have:
- PSI 4.0b1
- PSI4 1.0rc2

There is no sane way to do both with the same name. We should then change to version to something like `4.0-1.0rc` which might confuse users even more.

tovrstra:
@fgeorgatos I don't agree with you and I'd like to make a PR in which the new name will be `psi4`: lower case and with the four. I am a downstream user, and this is the way I would like it to be and how I guess most users would prefer it as well. 

In the community, e.g. conferences I attend, this program is always called `Psi4`, sometimes also `PSI4` and everything on the command line seems to be `psi4`, not `PSI4`, nor `Psi4`. EasyBuild should follow that convention and this is a nice opportunity to fix it.

Most users seem to have happily forgotten that there was once a Psi3. The 4 has just become a part of the name. If you like, we can always ask one of the Psi4 developers if they ever intend to make a Psi5, but that seems unlikely.

wpoely86:
@tovrstra I already asked them once: https://github.com/psi4/psi4/issues/213

tovrstra:
@wpoely86 That thread is just hilarious. So there may be a Psi5 one day. It will not be called a new version but rather a new generation. I'm glad I'm warned.

fgeorgatos:
I had a closer look and it is clear to me that @wpoely's argumentation is
strong:
* Not using PSI4 is problematic both due to ordering, as well as it is the
explicit naming by original developers
* The name is really PSI4 rather than Psi4 (per documentation). They simply
use small caps all over the place

On 4 July 2016 at 08:51, Toon Verstraelen <notifications@github.com> wrote:

> @wpoely86 <https://github.com/wpoely86> That thread is just hilarious. So
> there may be a Psi5 one day. It will not be called a new version but rather
> a new generation. I'm glad I'm warned.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/hpcugent/easybuild-easyconfigs/pull/3281#issuecomment-230227273>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABdgQJUJ7xtkawkZ26nxqchPIl10sgo_ks5qSLtmgaJpZM4JC62a>
> .
>



-- 
echo "sysadmin know better bash than english"|sed s/min/mins/ \
  | sed 's/better bash/bash better/' # signal detected in a CERN forum

PR #3332, opened 2016-07-13
ENT-v1.0.2

rtripath89:
Submitting an Easyconfig for ENT v1.0.2

### DIFF
--- a/easybuild/easyconfigs/e/ENT/ENT-1.0.2.eb
+++ b/easybuild/easyconfigs/e/ENT/ENT-1.0.2.eb
@@ -1,0 +1,37 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+# Author: Ravi Tripathi
+# Email: ravi89@uab.edu
+
+easyblock = "ConfigureMake"
+
+name = 'ENT'
+version = '1.0.2'
+
+homepage = "http://dna.engr.uconn.edu/?page_id=25"
+description = """ENT is a highly scalable genotype 
+phasing algorithm based on entropy minimization. 
+ENT is capable of phasing both unrelated and related 
+genotypes coming from complex pedigrees."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+source_urls = ['http://dna.engr.uconn.edu/~software/ent/']
+sources = ['ent_%(version)s.tar.gz']
+
+checksums = ['cbcc1923476443d2c80319802acc286d']
+
+buildopts = 'all'
+
+skipsteps = ['install']
+skipsteps = ['configure']
+
+dependencies = [
+     ('GCC', '4.7.2'),
+]
+sanity_check_paths = {
+    'files': ['ent'],
+    'dirs': [],
+}
+
+moduleclass = 'bio'
+
--- a/easybuild/easyconfigs/e/ENT/ENT-1.0.2.eb
+++ b/easybuild/easyconfigs/e/ENT/ENT-1.0.2.eb
@@ -2,7 +2,7 @@
 # Author: Ravi Tripathi
 # Email: ravi89@uab.edu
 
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/e/ENT/ENT-1.0.2.eb
```diff
@@ -0,0 +1,35 @@
...
+source_urls = ['http://dna.engr.uconn.edu/~software/ent/']
+sources = ['ent_%(version)s.tar.gz']
+
+install_cmd = "tar xvzf ent_%(version)s.tar.gz &&"
+install_cmd += "cd %(builddir)s/ent_%(version)s &&"
+install_cmd += "make all &&"
+install_cmd += "cp -a %(builddir)s/ent_%(version)s/* %(installdir)s"
+
+dependencies = [
+     ('GCC', '4.7.2'),
```
boegel: @rtripath89 why does this need `GCC` as a dependency, especially such an old version?

Wouldn't it be better to use GCC as a toolchain then rather than `dummy` to make this more explicit (e.g. in the module name)?

boegel:
Test report by @boegel
**FAILED**
See https://gist.github.com/18fcabd46cb1ebc88b86724dc348c11f for a full test report.

boegel:
@rtripath89 doesn't work for me, I'm running into problems like:

````
ain.cc:187: error: ‘strlen’ was not declared in this scope
```

rtripath89:
Strange...I tested it on an old system with older gcc and it worked fine, but when I tested it on newer GCC it's throwing up the error that you mentioned. One method, how I was able to get it to build with GCC/v4.7.2, was when I changed the main.cc (after untarring the source file) to include string.h instead of string and tar'd up the changed folder, and put it in the sources. I don't know how to automate this in an Easyconfig. Do you think adding comments for this would suffice, or if their is another method?

boegel:
@rtripath89 the proper way would be creating a patch file, something like this:

```
tar xfvz ent_1.0.2.tar.gz
cp -a ent-1.0.2 ent-1.0.2.orig
vim ent-1.0.2/main.cc  # change string to string.h
diff -ru ent-1.0.2.orig ent-1.0.2 > ent-1.0.2_fix-string-include.patch
```

and than add this to the easyconfig:

```python
patches = ['ent-%(version)s_fix-string-include.patch']
```

and include the patch file in this PR as well

you should also add a short description to the patch on top (any lines above the first `---` line are comments), and specify yourself as the author of the patch

easybuild/easyconfigs/e/ENT/ENT-1.0.2.eb
boegel: @rtripath89 This should be taken care of before we can merge this PR...

easybuild/easyconfigs/e/ENT/ent_1.0.2_fix-string-include.patch
```diff
@@ -0,0 +1,11 @@
+diff -ru ent_1.0.2.orig/main.cc ent_1.0.2/main.cc
```
boegel: @rtripath89 please include a brief description + authorship on top

easybuild/easyconfigs/e/ENT/ENT-1.0.2.eb
```diff
@@ -0,0 +1,37 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+# Author: Ravi Tripathi
+# Email: ravi89@uab.edu
+
+easyblock = "Binary"
```
boegel: @rtripath89 You'll want to use `BinariesTarball` instead, since `Binary` doesn't unpack the `sources`...

Did you test this easyconfig?

migueldiascosta:
@rtripath89, this PR is being closed for the following reason(s): no activity for > 1 year.

PR #3334, opened 2016-07-13
Add easyconfigs for ncurses when building with gcc 5.

akesandgren:
It also needs a patch due to changes in cpp.


boegel:
@akesandgren PRs should target `develop` rather than `master`, please open a new PR?

boegel:
replaced by #3335

PR #3390, opened 2016-07-27
Easyconfig for Spparks 

rtripath89:
I couldn't find a better way to create this Easyconfig. Let me know if there is one.

### DIFF
--- a/easybuild/easyconfigs/s/SPPARKS/SPPARKS-07192016.eb
+++ b/easybuild/easyconfigs/s/SPPARKS/SPPARKS-07192016.eb
@@ -1,0 +1,36 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+# Author: Ravi Tripathi
+# Email: ravi89@uab.edu
+
+easyblock = "Binary"
+
+name = 'SPPARKS'
+version = '07192016'
+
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/s/SPPARKS/SPPARKS-07192016.eb
```diff
@@ -0,0 +1,38 @@
...
+# You need to manually get the source file from http://www.sandia.gov/~sjplimp/download.html
+sources = ['spparks.tar.gz']
+
+checksums = ['e872813f3c4734b360d4ca9acab3e238']
+
+install_cmd = "tar xvf spparks.tar.gz &&"
+install_cmd += "cd %(builddir)s/spparks-19Jul16/src/ &&"
+install_cmd += "make linux &&"
+install_cmd += "mkdir ../bin && cp spk_linux ../bin &&"
+install_cmd += "cp -a %(builddir)s/spparks-19Jul16/* %(installdir)s"
```
boegel: @rtripath89 it looks like `MakeCp` would be a better option here, combined with use of `buildopts`

boegel: @rtripath89 ping?

PR #3440, opened 2016-08-19
{vis}[intel/2016b] X11 v20160819 (REVIEW)

boegel:
(created using `eb --new-pr`)
requires ~~https://github.com/hpcugent/easybuild-easyblocks/pull/979~~ https://github.com/hpcugent/easybuild-easyblocks/pull/980 and ~~#3438~~

PR #3446, opened 2016-08-22
{vis}[foss/2016a] PyBERT 1.7.9 w/Python-3.5.1 (WIP)

hajgato:


easybuild/easyconfigs/c/chaco/chaco-20160817-foss-2016a-Python-3.5.1.eb
```diff
@@ -0,0 +1,32 @@
...
+ writing plotting applications at all levels of complexity,
+ from simple scripts with hard-coded data to large plotting programs
+ with complex data interrelationships and a multitude of interactive tools.
+ While Chaco generates attractive static plots for publication and presentation,
+ it also works well for interactive data visualization and exploration."""
+
+toolchain = {'name': 'foss', 'version': '2016a'}
+
+source_urls = ['https://github.com/enthought/chaco/archive/']
+sources = ['%s.tar.gz' % commit_id]
```
boegel: @hajgato 4.5.0 is the latest official release, why not use that?

Once we start diverging from official releases, we'll need to stick to it forever, since `20160817` will always be newer than `4.5.0` or any other newer version...

easybuild/easyconfigs/e/enable/enable-20160817-foss-2016a-Python-3.5.1.eb
```diff
@@ -0,0 +1,65 @@
...
+ while enabling a high level of control over user interaction.
+ Enable is a supporting technology for the Chaco and BlockCanvas projects.
+ Kiva is a multi-platform DisplayPDF vector drawing engine that supports multiple output backends,
+ including Windows, GTK, and Macintosh native windowing systems,
+ a variety of raster image formats, PDF, and Postscript."""
+
+toolchain = {'name': 'foss', 'version': '2016a'}
+
+source_urls = ['https://github.com/enthought/enable/archive/']
+sources = ['%s.tar.gz' % commit_id]
```
boegel: @hajgato why not use latest official release (4.5.1)?

easybuild/easyconfigs/e/enable/enable-20160817_numpy.patch
```diff
@@ -0,0 +1,16 @@
...
+--- enable-168e222dababd4464062eb7a3ab92c4b591f7031/kiva/agg/setup.py.org	2016-08-17 17:12:22.000000000 +0200
++++ enable-168e222dababd4464062eb7a3ab92c4b591f7031/kiva/agg/setup.py	2016-08-25 11:31:29.272582528 +0200
+@@ -235,7 +235,9 @@
+         # Frequently, the 64-bit libraries are not in a known location and need
+         # manual configuration. From experience, this is usually not detected by
+         # the builder if we do not raise an exception.
+-        x11_info = get_info('x11', notfound_action=2)
++        x11_info = {'libraries': ['X11'],
++                    'library_dirs': [os.path.join(os.environ["EBROOTLIBX11"], "lib")],
++                    'include_dirs': [os.path.join(os.environ["EBROOTLIBX11"], "include")]}
```
boegel: hmm, relying on `$EBROOTLIBX11` is going to break if we use the components approach in #3440...

we may need to work around that by defining `$EBROOT*` variables for each of the components?

cc @wpoely86

easybuild/easyconfigs/e/enable/enable-20160817_kiva.patch
```diff
@@ -0,0 +1,29 @@
+#non-git version (without .git directory) build fails
```
boegel: how is this fixed here exactly?

easybuild/easyconfigs/e/enable/enable-20160817_kiva.patch
```diff
@@ -0,0 +1,29 @@
...
+#bundled kiva version is tied to enable
```
boegel: what does this mean, can you clarify? that we're including our own `kiva`?

easybuild/easyconfigs/k/kiwisolver/kiwisolver-20160309-foss-2016a-Python-3.5.1.eb
```diff
@@ -0,0 +1,27 @@
+easyblock = 'PythonPackage'
+
+name = 'kiwisolver'
+version = '20160309'
```
boegel: @hajgato same here, why not use latest release (0.1.3)?

easybuild/easyconfigs/p/Pillow/Pillow-3.2.0-foss-2016a-Python-3.5.1.eb
```diff
@@ -0,0 +1,27 @@
+easyblock = 'PythonPackage'
+
+name = 'Pillow'
+version = '3.2.0'
+versionsuffix = '-Python-%(pyver)s'
```
boegel: @hajgato why not use the existing `Pillow-3.2.0-intel-2016a-Python-2.7.11-freetype-2.6.3.eb,`?

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9-foss-2016a-Python-3.5.1.eb
```diff
@@ -0,0 +1,43 @@
...
+    ('Python', '3.5.1'),
+    ('matplotlib', '1.5.1', versionsuffix),
+    ('Pygments', '2.1.3', versionsuffix),
+    ('traits', '4.5.0', versionsuffix),
+    ('traitsui', '5.1.0', versionsuffix),
+    ('enable', '20160817', versionsuffix),
+    ('chaco', '20160817', versionsuffix),
+]
+
+use_pip = True
```
boegel: @hajgato any reason for preferring `pip` here?

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9_2to3.patch
```diff
@@ -0,0 +1,415 @@
+# Python 2to3
```
boegel: @hajgato how sure are you this is enough to make `PyBERT` Py3 compatible? seems like a long shot to me...

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9_sphinx.patch
```diff
@@ -0,0 +1,21 @@
+#Do not use Sphinx, as it opnly needs to make documantation.
```
boegel: typo: `opnly` -> `only`

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9-foss-2016a-Python-3.5.1.eb
```diff
@@ -0,0 +1,43 @@
...
+]
+
+dependencies = [
+    ('Python', '3.5.1'),
+    ('matplotlib', '1.5.1', versionsuffix),
+    ('Pygments', '2.1.3', versionsuffix),
+    ('traits', '4.5.0', versionsuffix),
+    ('traitsui', '5.1.0', versionsuffix),
+    ('enable', '20160817', versionsuffix),
+    ('chaco', '20160817', versionsuffix),
```
boegel: @hajgato would it make sense to make this a `Bundle` of Python packages (cfr. `matplotlib`), to avoid a bunch of modules/easyconfigs?

easybuild/easyconfigs/e/enable/enable-20160817-foss-2016a-Python-3.5.1.eb
```diff
@@ -0,0 +1,65 @@
...
+dependencies = [
+    ('Python', '3.5.1'),
+    ('PyQt', '4.11.4', versionsuffix),
+    ('Pillow', '3.2.0', versionsuffix),
+    ('traits', '4.5.0', versionsuffix),
+    ('traitsui', '5.1.0', versionsuffix),
+    ('ReportLab', '3.3.0', versionsuffix),
+    ('Pygments', '2.1.3', versionsuffix),
+    ('pyface', '5.1.0', versionsuffix),
+    ('kiwisolver', '20160309', versionsuffix),
```
boegel: @hajgato same here, a bundle would save us quite a bit of modules/easyconfigs?

easybuild/easyconfigs/c/chaco/chaco-20160817-foss-2016a-Python-3.5.1.eb
hajgato: @boegel I had two options. Try to do everything with `Python3`, as the user want to use `PyBERT` simultaneously with `pyGIMLi`. As far as I got from `pyGIMLi` homepage, its requires `Python3`. So I had to use `Python3` here as well. and the latest official release is not yet `Python3` compatible. I could use version`4.6.0beta.20160817` (or whatever is the version claimed by the code). 

Other possibility to make a `Python2/2.7.11-Python-3.5.1`, and use all the stuff here with that.
(`chaco/4.5.0-foss-2016a-Python-2.7.11-Python-3.5.1` or so). The option to use system `Python` is not a good way, since `PyBERT` requires `Python 2.7`

easybuild/easyconfigs/e/enable/enable-20160817-foss-2016a-Python-3.5.1.eb
hajgato: @boegel see comment for `chaco`

easybuild/easyconfigs/e/enable/enable-20160817_numpy.patch
hajgato: @boegel @wpoely86 I would break that. It's not anymore libX11, but a whole bunch of other stuff as well. If we define `$EBROOT` for all components that we defined individually, then we will have lots of `$EBROOT` variables.

easybuild/easyconfigs/e/enable/enable-20160817_kiva.patch
hajgato: @boegel Somewhere in the code it is buried to read version info from `.git` directory. 
I just made `kiva` use the version of `enable` (as far as I know, kiva is not distributed as standalone package, see next comment fo rthe next line). I took the inspiration from a 2 years old PR, which never been merged. 
If you use `git clone` instead of downloading the actual commit, you do not need the patch.

easybuild/easyconfigs/e/enable/enable-20160817_kiva.patch
hajgato: @boegel Yes a `kiva` is icluded as well as `agg` for `kiva`. As far as I know, `Enthought` does not distribute a standalone `kiva`, at the homepage of this package referred to have two parts, `enable` and `kiva`  (see http://code.enthought.com/projects/enable/)

easybuild/easyconfigs/k/kiwisolver/kiwisolver-20160309-foss-2016a-Python-3.5.1.eb
hajgato: @boegel `0.1.3` does not work with `Python3`, see commant @ chaco

easybuild/easyconfigs/p/Pillow/Pillow-3.2.0-foss-2016a-Python-3.5.1.eb
hajgato: @boegel for two reasons. This PR is based on `foss` toolcahin and not `intel`, and its needs `Python3` and not `Python2` ;)

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9-foss-2016a-Python-3.5.1.eb
hajgato: @boegel Its a Python package, and I think it should be installed using `pip`. Am I doing here something wrong?

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9_2to3.patch
hajgato: @boegel: it seems to work. There are some code that I changed by hand, not only `2to3`, although   I was not clear. The user is testing it....

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9_sphinx.patch
hajgato: fixed

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9-foss-2016a-Python-3.5.1.eb
hajgato: @boegel It makes sense, and I can use versions from github.

easybuild/easyconfigs/e/enable/enable-20160817-foss-2016a-Python-3.5.1.eb
hajgato: @boegel It makes sense, and I can use versions from github.

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9_2to3.patch
hajgato: @boegel: I mean the GUI starts, but I do not know anything about its functionality.

hajgato:
@boegel  I can confirm that PyBERT does not work (prblem with the matplotlib interface)

easybuild/easyconfigs/e/enable/enable-20160817_numpy.patch
wpoely86: :+1:

hajgato:
@boegel: Ok, I found the error, it needs an old version of `SWIG`

hajgato:
@boegel: I put this back as **`WIP`**, and I will make a bundle.

easybuild/easyconfigs/c/chaco/chaco-20160817-foss-2016a-Python-3.5.1.eb
boegel: according to http://www.pygimli.org/installation.html#optional-prerequisites, pyGIMLi is compatible with Python 2.7 as well?

boegel: also, I'm not sure what `Python2/2.7.11-Python-3.5.1` is? is that Python 2 and 3 combined? that's a nice recipe for disaster ;

easybuild/easyconfigs/e/enable/enable-20160817_numpy.patch
boegel: hmm, ok, good point @hajgato, it would be a lot of `$EBROOT`'s, indeed... and I guess it's fairly rare to have to grab the install prefix for one of the components

easybuild/easyconfigs/e/enable/enable-20160817_kiva.patch
boegel: @hajgato please include a reference to the PR?

easybuild/easyconfigs/e/enable/enable-20160817_kiva.patch
boegel: ok, please extend the comment a bit to reflect this better? e.g. include the URL, etc?

easybuild/easyconfigs/p/Pillow/Pillow-3.2.0-foss-2016a-Python-3.5.1.eb
boegel: ok, my bad, sorry

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9-foss-2016a-Python-3.5.1.eb
boegel: the `use_pip` is there for specific use cases, like installing Python packages that are only distributed in `.whl` format

for standard Python packages, there's no need to use `pip`

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9_2to3.patch
boegel: since `PyGIMLi` is also compatible with Python 2, there's probably no need for this?

it's hard to tell whether this broke anything, or whether this patch is complete...

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9-foss-2016a-Python-3.5.1.eb
boegel: @hajgato if you can stick to Python2, sticking to official releases should be fine?

hajgato:
Python2 version will be made.

easybuild/easyconfigs/c/chaco/chaco-20160817-foss-2016a-Python-3.5.1.eb
hajgato: Ok, for the future what is the recipe to use python2 and python3 together. Ok, in this project it is not needed, but if in the future this problem will pop up? If I have a linux distro, I can have both, and most of the cases its not a disaster.

boegel: I think the right approach would be to make a `Python2/2.7.11-*` module, and load Python (3) and Python2 together as deps

hajgato: How do you prevent users to load python (2) and python2 module together? I think that's also a recipe for disaster (for example which order did they loaded it)

easybuild/easyconfigs/p/PyBERT/PyBERT-1.7.9-foss-2016a-Python-3.5.1.eb
hajgato: Ok then.

easybuild/easyconfigs/c/chaco/chaco-20160817-foss-2016a-Python-3.5.1.eb
boegel: Nothing would break then, one module would just shadow the other?

PR #3421, opened 2016-08-17
{compiler}[dummy] GCC 4.9.4 w/ binutils 2.25

boegel:
(created using `eb --new-pr`)

PR #3462, opened 2016-08-25
Add a toolchain for foss/2016a + NVIDIA CUDA 7.5

eshelman:
These easyconfigs provide a build of the goolfc toolchain using the versions of compilers and libraries that are in the foss/2016a toolchain. In other words, it brings goolfc up to date.

Furthermore, these easyconfigs switch the version numbering to a date-based scheme.

### DIFF
--- a/easybuild/easyconfigs/f/FFTW/FFTW-3.3.4-gompic-2016.08.eb
+++ b/easybuild/easyconfigs/f/FFTW/FFTW-3.3.4-gompic-2016.08.eb
@@ -8,7 +8,7 @@ description = """FFTW is a C subroutine library for computing the discrete Fouri
  in one or more dimensions, of arbitrary input size, and of both real and complex data."""
 
 toolchain = {'name': 'gompic', 'version': '2016.08'}
-toolchainopts = {'optarch': True, 'pic': True}
+toolchainopts = {'pic': True}
 
 sources = [SOURCELOWER_TAR_GZ]
 source_urls = [homepage]
--- a/easybuild/easyconfigs/f/FFTW/FFTW-3.3.4-gompic-2016.08.eb
+++ b/easybuild/easyconfigs/f/FFTW/FFTW-3.3.4-gompic-2016.08.eb
@@ -1,0 +1,34 @@
+easyblock = 'ConfigureMake'
+
+name = 'FFTW'
+version = '3.3.4'
+
+homepage = 'http://www.fftw.org'
+description = """FFTW is a C subroutine library for computing the discrete Fourier transform (DFT)
+ in one or more dimensions, of arbitrary input size, and of both real and complex data."""
+
+toolchain = {'name': 'gompic', 'version': '2016.08'}
+toolchainopts = {'optarch': True, 'pic': True}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = [homepage]
+
+common_configopts = "--enable-threads --enable-openmp --with-pic"
+
+configopts = [
+    common_configopts + " --enable-single --enable-sse2 --enable-mpi",
+    common_configopts + " --enable-long-double --enable-mpi",
+    common_configopts + " --enable-quad-precision",
+    common_configopts + " --enable-sse2 --enable-mpi",  # default as last
+]
+
+sanity_check_paths = {
+    'files': ['bin/fftw%s' % x for x in ['-wisdom', '-wisdom-to-conf', 'f-wisdom', 'l-wisdom', 'q-wisdom']] +
+             ['include/fftw3%s' % x for x in ['-mpi.f03', '-mpi.h', '.f', '.f03',
+                                              '.h', 'l-mpi.f03', 'l.f03', 'q.f03']] +
+             ['lib/libfftw3%s%s.a' % (x, y) for x in ['', 'f', 'l'] for y in ['', '_mpi', '_omp', '_threads']] +
+             ['lib/libfftw3q.a', 'lib/libfftw3q_omp.a'],
+    'dirs': ['lib/pkgconfig'],
+}
+
+moduleclass = 'numlib'
--- a/easybuild/easyconfigs/g/gompic/gompic-2016.08.eb
+++ b/easybuild/easyconfigs/g/gompic/gompic-2016.08.eb
@@ -1,0 +1,23 @@
+easyblock = "Toolchain"
+
+name = 'gompic'
+version = '2016.08'
+
+homepage = '(none)'
+description = """GNU Compiler Collection (GCC) based compiler toolchain along with CUDA toolkit,
+ including OpenMPI for MPI support with CUDA features enabled."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+comp_name = 'GCC'
+comp_ver = '4.9.3-2.25'
+comp = (comp_name, comp_ver)
+
+# compiler toolchain dependencies
+dependencies = [
+    comp,  # part of gcccuda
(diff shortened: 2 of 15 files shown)


easybuild/easyconfigs/s/ScaLAPACK/ScaLAPACK-2.0.2-gompic-2016.08-OpenBLAS-0.2.15-LAPACK-3.6.0.eb
```diff
@@ -0,0 +1,25 @@
...
+source_urls = [homepage]
+sources = [SOURCELOWER_TGZ]
+
+blaslib = 'OpenBLAS'
+blasver = '0.2.15'
+blassuff = '-LAPACK-3.6.0'
+
+versionsuffix = "-%s-%s%s" % (blaslib, blasver, blassuff)
+
+dependencies = [(blaslib, blasver, blassuff, ('GCC', '4.9.3-2.25'))]
```
boegel: @eshelman tests fail because of this hardcoded toolchain spec for OpenBLAS

First, it's wrong, since the easyconfig for OpenBLAS you provide uses `gompic`; also, there's no need to hardcode a toolchain spec at all in this case...

Any particular reason you added it?

boegel:
@eshelman for toolchain PRs, we usually include an easyconfig for HPL with that particular toolchain, as a testcase for that toolchain... Can you add one for this new toolchain?

easybuild/easyconfigs/f/FFTW/FFTW-3.3.4-gompic-2016.08.eb
```diff
@@ -0,0 +1,34 @@
...
+
+name = 'FFTW'
+version = '3.3.4'
+
+homepage = 'http://www.fftw.org'
+description = """FFTW is a C subroutine library for computing the discrete Fourier transform (DFT)
+ in one or more dimensions, of arbitrary input size, and of both real and complex data."""
+
+toolchain = {'name': 'gompic', 'version': '2016.08'}
+toolchainopts = {'optarch': True, 'pic': True}
```
boegel: @eshelman please remove the `optarch` entry, since that's enabled by default:

```python
toolchainopts = {'pic': True}
```

easybuild/easyconfigs/g/gompic/gompic-2016.08.eb
```diff
@@ -0,0 +1,23 @@
...
+
+comp_name = 'GCC'
+comp_ver = '4.9.3-2.25'
+comp = (comp_name, comp_ver)
+
+# compiler toolchain dependencies
+dependencies = [
+    comp,  # part of gcccuda
+    ('CUDA', '7.5.18', '', comp),  # part of gcccuda
+    ('OpenMPI', '1.10.2', '', ('gcccuda', version)),
```
boegel: @eshelman any reason not to use `1.10.3`?

easybuild/easyconfigs/g/goolfc/goolfc-2016.08.eb
```diff
@@ -0,0 +1,38 @@
...
+version = '2016.08'
+
+homepage = '(none)'
+description = """GCC based compiler toolchain __with CUDA support__, and including
+ OpenMPI for MPI support, OpenBLAS (BLAS and LAPACK support), FFTW and ScaLAPACK."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+comp_name = 'GCC'
+comp_ver = '4.9.3-2.25'
```
boegel: @eshelman why not use `4.9.4-2.25` (GCC 4.9.4 was released Aug 3rd), cfr #3421

easybuild/easyconfigs/h/hwloc/hwloc-1.11.2-gcccuda-2016.08.eb
```diff
@@ -0,0 +1,19 @@
+easyblock = 'ConfigureMake'
+
+name = 'hwloc'
+version = "1.11.2"
```
boegel: there's a `1.11.3` of hwloc...

easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.2.15-gompic-2016.08-LAPACK-3.6.0.eb
```diff
@@ -0,0 +1,51 @@
+easyblock = 'ConfigureMake'
+
+name = 'OpenBLAS'
+version = '0.2.15'
```
boegel: @eshelman OpenBLAS is at 0.2.18

fgeorgatos:
also, this PR is against master, instead of develop!

boegel:
@fgeorgatos changed to `develop` (I can do this myself recently, thank you GitHub!)

eshelman:
Thanks for the tips and fixes on this! I'll revise.

The reason I'd specifically called out older versions of GCC/OpenMPI/hwloc/OpenBLAS was that I assumed it would be valuable to keep the CUDA toolchain in sync with the foss toolchain. In other words, I thought I would match the versions used by foss/2016a.

If people think it's better to use all the newest versions, I'm not opposed. We will have another new version quite soon though (based on CUDA 8.0).

boegel:
@eshelman as long as you stick to the same major version (GCC 4.x, OpenMPI 1.x), there should be no downside in using newer versions of the components compared to `foss/2016a`, right?
In fact, it may even help in fixing some problems that may be there with `foss/2016a` but would be fixed with a newer bugfix release of the compilers or MPI (I can't think of an example from the top of my head though, but you get the idea).

It'll also align better with the toolchain version `2016.08`, where the idea is that this is determined by the most recent component in the toolchain.

easybuild/easyconfigs/s/ScaLAPACK/ScaLAPACK-2.0.2-gompic-2016.08-OpenBLAS-0.2.15-LAPACK-3.6.0.eb
eshelman: I was trying really hard to match existing conventions here. See, for example, this comparison with the existing ScaLAPACK easybuild for the 2016a toolchain:

```
--- ScaLAPACK-2.0.2-gompi-2016a-OpenBLAS-0.2.15-LAPACK-3.6.0.eb 2016-08-21 12:51:24.231715518 -0400
+++ ScaLAPACK-2.0.2-gompic-2016.08-OpenBLAS-0.2.15-LAPACK-3.6.0.eb      2016-08-24 23:49:14.932432277 -0400
@@ -5,7 +5,7 @@
 description = """The ScaLAPACK (or Scalable LAPACK) library includes a subset of LAPACK routines
  redesigned for distributed memory MIMD parallel computers."""

-toolchain = {'name': 'gompi', 'version': '2016a'}
+toolchain = {'name': 'gompic', 'version': '2016.08'}
 toolchainopts = {'pic': True}

 source_urls = [homepage]
```

eshelman:
I'm working on getting this fixed up. I had tried to copy it as closely as I could from the older goolfc-2.6.10 toolchain. Either I made a mistake or things have changed since that toolchain was added.

eshelman:
Sorry, team. I borked the merge. If you'd rather see me try this on a fresh branch, I will do so.

Or any other suggestion? My knowledge of git is still developing...

eshelman:
This should all be clean and working now. Please let me know if anything else needs to be adjusted? :-)

boegel:
merged via #3796

PR #3463, opened 2016-08-25
{vis}[GCCcore/4.9.3] X.Org v2016-08-31

ocaisa:
This is a build of the (almost) complete X.Org stack using jhbuild.
It clones the git repos of all the software packages and so is latest master branches but not necessarily particular software releases. The version number is a date and that is passed in as the date stamp that you want for the repos. In total it builds 240 packages, all bar 2 are directly from X.Org the others are dependencies. The patches should give you an idea about how to hack the buildtool (jhbuild) and the moduleset it uses.

There is one small issue in that the latest MESA requires the Mako python module. When I added that as an EasyConfig dep (built on top of the system python), the configure step would pass but the build would fail saying it could not load the Mako module. If I use my own installation of Python and add the module directly to the installation, it builds fine. The other python module dependencies are `six` and `lxml`, I'm not sure if there are problems with adding those as deps as well. As it stands then this PR has system dependencies on these 3 python modules (but probably if I solved how to do one properly they would all be solved).

I also may not have all the builddeps and deps correctly separated correctly (or perhaps complete).

Also, even though this is a pretty short easyconfig, it may require an easyblock, as we will probably want to ensure we are getting the latest versions of the sources. The right way to grab them is by

```
git clone git://anongit.freedesktop.org/git/xorg/util/modular/ xorg/util/modular/
```

but doing that (and handling it properly for updates) is not supported by EB for sources yet.

I also have a working version of this PR for GCCcore 5.4.0

This is a competitor for the approach of #3440


ocaisa:
Word to the wise, this will take about a half a day to  build!

ocaisa:
You can take a look at the packages included at https://gist.github.com/ocaisa/13cff80455cf608e4606d17fd8ac6cb9

boegel:
@ocaisa do you mind fleshing out the easyconfigs that are just version/toolchain bumps into a separate PR (which should be trivial to test/review/merge), so we can see the forest through the trees here?

ocaisa:
Done in #3469, #3470, #3471, #3473, #3474, #3475, #3476, #3477

boegel:
@ocaisa please sync with `develop` & resolve conflicts, let's try and get this thoroughly tested & the known issues dealt with ASAP?

ocaisa:
@boegel Just testing this myself now before I pass the baton, will report back when I get a successful build

ocaisa:
The more-or-less complete list of X.Org packages included in the build can be found at https://gist.github.com/ocaisa/13cff80455cf608e4606d17fd8ac6cb9
These are the packages cloned from the respective repo, `libepoxy` for example is not included since that comes from a release tarball (and is a good example of how you include something like that in a moduleset).
I'll update the sanity check to include the mesa sanity check components.

ocaisa:
@boegel This builds successfully for me now, I'll upgrade the sanity checks a bit.

boegel:
Test report by @boegel
**FAILED**
node2494.golett.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/df18818bf148bcc04a5682452b88a4d5 for a full test report.

ocaisa:
You're not able to clone from the repos on the test machine, is it firewalled? I had to get them to allow it here at JSC.

boegel:
@ocaisa it's indeed possible this is blocked by our firewall, which highlights another issue with this approach: it will redownload the required sources every time, and this build can not be performed offline, next to the potential firewall issues... Is there any way around that with `jhbuild`?

ocaisa:
You can tell jhbuild where to put the sources (it's part of the current patch) and you can tell it to work offline with `--no-network` (`-n`).

This is not the only easyconfig install that cannot be performed offline, there are quite a few more!

boegel:
@ocaisa I'm not saying this is the only build that can't be performed offline, just that we should try and make sure it can work offline when we realise it won't in its current state.

Is there anything missing to make this work with `--no-network`? Maybe support for `%(sourcedir)s`?

Although you can already handle this via an easyblock... *wink* *wink*

ocaisa:
@boegel Do I have access to the build tools at the `fetch` step, otherwise (if I make an easyblock) I'll need to move the "real" fetch elsewhere

boegel:
@ocaisa: see https://github.com/hpcugent/easybuild-framework/issues/1376

migueldiascosta:
@ocaisa, this PR is being closed for the following reason(s): no activity for > 1 year.

PR #3485, opened 2016-08-26
Bugfix for python string format error in h5py

### LINKED ISSUE #3484: Bug in python string format of h5py package
Shuld be: `versionsuffix = '-Python-%(pyver)s-HDF5-%s' % hdf5ver` instead of `versionsuffix = '-Python-%%(pyver)s-HDF5-%s' % hdf5ver`

https://github.com/hpcugent/easybuild-easyconfigs/commit/623cc278c4037beb3540263441840a834b152116#diff-9f53eb89c91c407e3c44b7c2191322efR19

timeu:
This fixes #3484

### DIFF
--- a/easybuild/easyconfigs/h/h5py/h5py-2.5.0-foss-2016a-Python-2.7.11-HDF5-1.8.16-serial.eb
+++ b/easybuild/easyconfigs/h/h5py/h5py-2.5.0-foss-2016a-Python-2.7.11-HDF5-1.8.16-serial.eb
@@ -16,7 +16,7 @@ sources = [SOURCE_TAR_GZ]
 
 hdf5ver = '1.8.16'
 hdf5versuffix = '-serial'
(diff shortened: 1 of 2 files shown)


boegel:
@timeu as indicated by the failing tests, this doesn't do what you think it does, you're actually breaking the easyconfig file this way, see also https://github.com/hpcugent/easybuild-easyconfigs/issues/3484#issuecomment-242782456

Where/how did you run into a format error exactly that made you make this change?

timeu:
@boegel : Yes you are right. I think the issue was that we ran some easyconfigs with an old easybuild verison. I will close the pull request for now and in case I encounter the issue again, I will create a new issue.

boegel:
@timeu ah yes, that's possible, the `%(pyver)s` template is only supported since EasyBuild v2.7.0

PR #3510, opened 2016-09-01
{vis}[goolf/1.4.10] Blender 2.66a

AFCJamie:
An easyconfig for Blender 2.66a which uses the goolf 1.4.10 toolchain.

### DIFF
--- a/easybuild/easyconfigs/b/Blender/Blender-2.66a-goolf-1.4.10.eb
+++ b/easybuild/easyconfigs/b/Blender/Blender-2.66a-goolf-1.4.10.eb
@@ -1,0 +1,36 @@
+easyblock = 'CMakeMake'
+
+name = 'Blender'
+version = '2.66a'
+
+homepage = 'https://www.blender.org/'
+description = """Blender is the free and open source 3D creation suite. It supports
+ the entirety of the 3D pipeline-modeling, rigging, animation, simulation, rendering,
+ compositing and motion tracking, even video editing and game creation."""
+
+toolchain = {'name': 'goolf', 'version': '1.4.10'}
+
+sources = ['blender-2.66a.tar.gz']
+source_urls = ['http://download.blender.org/source/']
+
+# Remove and add configopts as desired
+configopts = '-DWITH_INSTALL_PORTABLE=OFF '
+configopts += '-DWITH_BUILDINFO=OFF '
+configopts += '-DWITH_GAMEENGINE=OFF '
+configopts += '-DWITH_CYCLES=OFF '
+configopts += '-DWITH_SYSTEM_GLEW=OFF '
+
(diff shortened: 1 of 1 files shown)


boegel:
@AFCJamie Blender is (high) on my personal TODO list to install with EasyBuild, so I'm happy to see someone else is working on it too, thanks for contributing this!

This looks *way* more basic than I was expecting though, especially taking into account the rather extensive list of dependencies of Blender (see the `Manual Depenency Installation` at https://wiki.blender.org/index.php/Dev:Doc/Building_Blender/Linux/dependencies).

Can you comment on that? Is this simply a significantly trimmed down version of Blender that doesn't require all those dependencies?
If so, it may be a good starting point, but getting a more fleshed out Blender build is probably going to require quite a bit more effort.

Also: any particular reason for using `goolf/1.4.10`, which is a very old toolchain? This won't work on sufficiently recent systems (due to the OpenBLAS version included in `goolf/1.4.10` not being aware of recent architectures like Haswell, for example).

boegel:
Test report by @boegel
**FAILED**
node2003.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/266b79c7a3248c219e8a1d3b6b5004e1 for a full test report.

AFCJamie:
Hi @boegel, I'm glad that I am helping to make some progress towards this.

After looking at that link, I imagine this would be a trimmed down version of Blender. A work colleague built this by hand a few years ago so I followed his notes on our internal wiki when constructing the easyconfig, noting that the process he used was a cmake, make, make install process and that he only loaded Python and Boost into the environment. He didn't mention any errors (which he would do) so I assumed this was all that was needed to get a working version.

(Rather erroneously) I decided to use goolf/1.4.10 because there were already easyconfigs for Python 3.3.2, Boost 1.51.0 and CMake 3.2.3 in the repo, so I was hoping that this toolchain would work too (as your test shows - it hasn't). In light on what you have said about OpenBLAS (I didn't realise this), I see now that it built for us because I used a custom version of foss (that's not on the repo) which uses OpenBLAS 0.2.18 (on a Haswell node too). Tomorrow, I will open a new pull request with the foss toolchain I used (and now as GCC 6 works with EasyBuild I can be happy to push it up because I initially built GCC 6 by hand into our EasyBuild tree to get it to work, pushing the foss toolchain back then would have failed any tests done on it) and I'll include all the extra dependencies required for it that used this toolchain.

I hope this helps justify my position, I am new to building applications from source (having only started back in June as part of a student role) so I am still getting used to the nooks and crannies that different software versions provide.

All the best.

AFCJamie:
I will close this PR and open a new one once I've made a PR for the version of foss I'm using.

PR #3518, opened 2016-09-02
{vis}[foss/2016.04] Blender 2.66a

AFCJamie:
Correction to #3510.

An easyconfig for Blender 2.66a using the foss 2016.04 toolchain.

@boegel  After some tinkering I found that I had some troubles using GCC 6 to build Blender, but it built okay using a toolchain that uses GCC 5 so this should be better. ~~I have named this 'Trimmed' to reflect the lesser dependencies that were needed to get it to build at a cost of reduced functionality - but those can be added in future easyconfigs when the other dependencies have easyconfigs of their own.~~

### DIFF
--- a/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-foss-2016.04.eb
+++ b/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-foss-2016.04.eb
@@ -1,0 +1,26 @@
+easyblock = 'ConfigureMake'
+
+name = 'Autoconf'
+version = '2.69'
+
+homepage = 'http://www.gnu.org/software/autoconf/'
+description = """Autoconf is an extensible package of M4 macros that produce shell scripts
+ to automatically configure software source code packages. These scripts can adapt the
+ packages to many kinds of UNIX-like systems without manual user intervention. Autoconf
+ creates a configuration script for a package from a template file that lists the
+ operating system features that the package can use, in the form of M4 macro calls."""
+
+toolchain = {'name': 'foss', 'version': '2016.04'}
+
+source_urls = [GNU_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [('M4', '1.4.17')]
+
+sanity_check_paths = {
+    'files': ["bin/%s" % x for x in ["autoconf", "autoheader", "autom4te", "autoreconf", "autoscan",
+                                     "autoupdate", "ifnames"]],
+    'dirs': [],
+}
+
+moduleclass = 'devel'
--- a/easybuild/easyconfigs/a/Automake/Automake-1.15-foss-2016.04.eb
+++ b/easybuild/easyconfigs/a/Automake/Automake-1.15-foss-2016.04.eb
@@ -1,0 +1,33 @@
+##
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+#
+# Copyright:: Copyright 2012-2014 Uni.Lu/LCSB, NTUA
+# Authors::   Fotis Georgatos <fotis@cern.ch>
+# License::   MIT/GPL
+# $Id$
+#
+# This work implements a part of the HPCBIOS project and is a component of the policy:
+# http://hpcbios.readthedocs.org/en/latest/
+##
+
+easyblock = 'ConfigureMake'
+
+name = 'Automake'
+version = "1.15"
+
+homepage = 'http://www.gnu.org/software/automake/automake.html'
+description = "Automake: GNU Standards-compliant Makefile generator"
+
+toolchain = {'name': 'foss', 'version': '2016.04'}
+
+source_urls = [GNU_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [('Autoconf', '2.69')]
+
+sanity_check_paths = {
+    'files': ['bin/automake', 'bin/aclocal'],
+    'dirs': []
+}
+
+moduleclass = 'devel'
--- a/easybuild/easyconfigs/a/Autotools/Autotools-20150215-foss-2016.04.eb
+++ b/easybuild/easyconfigs/a/Autotools/Autotools-20150215-foss-2016.04.eb
@@ -1,0 +1,17 @@
+easyblock = 'Bundle'
+
+name = 'Autotools'
+version = '20150215'  # date of the most recent change
+
+homepage = 'http://autotools.io'
+description = """This bundle collect the standard GNU build tools: Autoconf, Automake and libtool"""
+
+toolchain = {'name': 'foss', 'version': '2016.04'}
+
+dependencies = [
+    ('Autoconf', '2.69'),  # 20120424
+    ('Automake', '1.15'),  # 20150105
+    ('libtool', '2.4.6'),  # 20150215
+]
+
+moduleclass = 'devel'
--- a/easybuild/easyconfigs/b/Bison/Bison-3.0.4-GCCcore-6.1.0.eb
+++ b/easybuild/easyconfigs/b/Bison/Bison-3.0.4-GCCcore-6.1.0.eb
@@ -1,0 +1,26 @@
+easyblock = 'ConfigureMake'
+
+name = 'Bison'
+version = '3.0.4'
+
+homepage = 'http://www.gnu.org/software/bison'
+description = """Bison is a general-purpose parser generator that converts an annotated context-free grammar
+ into a deterministic LR or generalized LR (GLR) parser employing LALR(1) parser tables."""
+
+toolchain = {'name': 'GCCcore', 'version': '6.1.0'}
+
+sources = [SOURCELOWER_TAR_GZ]
+source_urls = [GNU_SOURCE]
+
+builddependencies = [
+    ('M4', '1.4.17'),
+    # use same binutils version that was used when building GCCcore toolchain
+    ('binutils', '2.27', '', True),
+]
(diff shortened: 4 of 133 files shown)


easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-trimmed.eb
```diff
@@ -0,0 +1,38 @@
+easyblock = 'CMakeMake'
+
+name = 'Blender'
+version = '2.66a'
+versionsuffix = '-trimmed'
```
boegel: @AFCJamie I would drop this suffix personally; I feel it's OK to initially have a slim build of Blender, and then extend it gradually by updating this easyconfig file by adding dependencies etc.

easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-trimmed.eb
```diff
@@ -0,0 +1,38 @@
...
+
+homepage = 'https://www.blender.org/'
+description = """Blender is the free and open source 3D creation suite. It supports
+ the entirety of the 3D pipeline-modeling, rigging, animation, simulation, rendering,
+ compositing and motion tracking, even video editing and game creation."""
+
+toolchain = {'name': 'foss', 'version': '2016.04'}
+
+sources = ['blender-2.66a.tar.gz']
```
boegel: @AFCJamie this is a common pattern for source filenames, so we have a constant for it `SOURCELOWER_TAR_GZ` (which expands to `%(namelower)s-%(version)s.tar.gz`)

it's better to avoid hardcoding the version here since otherwise you have to bump the version in multiple places if you're looking into upgrading Blender

easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-trimmed.eb
```diff
@@ -0,0 +1,38 @@
...
+
+# These are needed until extra dependencies are added to flesh it out 
+configopts = '-DWITH_INSTALL_PORTABLE=OFF '
+configopts += '-DWITH_BUILDINFO=OFF '
+configopts += '-DWITH_GAMEENGINE=OFF '
+configopts += '-DWITH_CYCLES=OFF '
+configopts += '-DWITH_SYSTEM_GLEW=OFF '
+
+dependencies = [
+    ('Python', '3.3.2'),
```
boegel: when Python is listed as a dependency, we generally label the easyconfig with a versionsuffix like `-Python-%(pyver)s` (where `%(pyver)s` is a template value that EasyBuild will fill in automagically)

this makes it clear that only modules using the same Python version will be compatible with this modules (just because Python is such a common dep)

easybuild/easyconfigs/c/CMake/CMake-3.4.3-foss-2016.04.eb
```diff
@@ -0,0 +1,32 @@
...
+sources = [SOURCELOWER_TAR_GZ]
+
+configopts = '--DCMAKE_USE_OPENSSL=1'
+
+dependencies = [
+    ('ncurses', '6.0'),
+    # OS dependency should be preferred (uncomment osdependencies line) if the os
+    # version is more recent then this version. It's nice to have an up to date openssl
+    # for security reasons
+    ('OpenSSL', '1.0.1p'),
```
boegel: please comment out this dep, and uncomment the `osdependencies` line below, cfr. comment above

easybuild/easyconfigs/n/ncurses/ncurses-6.0-foss-2016.04.eb
```diff
@@ -4,18 +4,17 @@ name = 'ncurses'
...
 
 toolchain = {'name': 'foss', 'version': '2016.04'}
-toolchainopts = {'optarch': True, 'pic': True}
+toolchainopts = {'pic': True}
 
 source_urls = [GNU_SOURCE]
 sources = [SOURCE_TAR_GZ]
 
-patches = ['ncurses-%(version)s_gcc-5.patch']
-
```
boegel: @AFCJamie this should not be undone, just use the ncurses easyconfig that we already have in `develop` for this version/toolchain combo?

easybuild/easyconfigs/o/OpenSSL/OpenSSL-1.0.1p-foss-2016.04.eb
```diff
@@ -0,0 +1,22 @@
...
+version = '1.0.1p'
+
+homepage = 'http://www.openssl.org/'
+description = """The OpenSSL Project is a collaborative effort to develop a robust,
+ commercial-grade, full-featured, and Open Source toolchain implementing the Secure
+ Sockets Layer (SSL v2/v3) and Transport Layer Security (TLS v1) protocols as well
+ as a full-strength general purpose cryptography library."""
+
+toolchain = {'name': 'foss', 'version': '2016.04'}
+toolchainopts = {'pic': True, 'opt': True, 'optarch': True}
```
boegel: @AFCJamie please drop the `opt` and `optarch` entries here, they're enabled by default

easybuild/easyconfigs/p/Python/Python-3.3.2-foss-2016.04.eb
```diff
@@ -0,0 +1,45 @@
+name = 'Python'
+version = '3.3.2'
+
+homepage = 'http://python.org/'
+description = "Python is a programming language that lets you work more quickly and integrate your systems more effectively."
+
+toolchain = {'name': 'foss', 'version': '2016.04'}
+toolchainopts = {'pic': True, 'opt': True, 'optarch': True}
```
boegel: @AFCJamie same here, please drop the opt and optarch entries here, they're enabled by default

boegel:
@AFCJamie we usually try to update dependencies to the latest version when adding new easyconfigs, especially for new software packages...

Older versions include `CMake` (3.6.1), `Python` (3.5.2), `libreadline` (6.3), `OpenSSL` (1.0.1t)

Would you mind bumping those to the latest version? For all (except maybe `OpenSSL`), easyconfigs for these are already available in `develop` with another toolchain.

One thing you may consider for the future is to jump on the "common toolchains" train, i.e. follow the series of `foss`/`intel` versions that are refreshed every 6 months (see `2016a` and `2016b` for the latest iterations).

Of course, you're welcome to contribute back easyconfigs with whatever toolchain you were using before, just something to think about...

AFCJamie:
Thanks for the feedback, I really appreciate it. I will get working on your suggestions and push back to this branch when they are done.

easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-trimmed.eb
AFCJamie: Okay, no worries.

easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-trimmed.eb
AFCJamie: Okay, I'll try to make sure that I'm more conscious of not hard coding things especially when I don't have an example to look up upon.

easybuild/easyconfigs/n/ncurses/ncurses-6.0-foss-2016.04.eb
AFCJamie: Sorry about this, I must have completely overlooked that easyconfig in the branch and not realised it was there. I've pushed it back up with exactly the same contents as the one that already exists in develop.

AFCJamie:
I have pushed up the minor changes to the easyconfigs, I will look at bumping the versions up of those applications and push them to here on Monday if I don't get chance over the weekend. I will drop another message to let you know when I think I've done everything for this PR.

boegel:
:+1:

AFCJamie:
Just to give you an update, I'm working on the dependencies for Python 3.5.2 at the moment (CMake, libreadline and OpenSSL have been updated) and if that builds okay then I'll push everything up to here.

boegel:
@AFCJamie please open a separate for Python 3.5.2, it'll be a lot easier to review & merge, and it'll avoid that this PR becomes too big too handle

Don't worry about dependencies in this PR, once what's required has been merged into `develop`, it's just a case of retriggering Travis for this PR, which is easy to do.

AFCJamie:
@boegel Okay, will do!

boegel:
@AFCJamie now that #3522 is merged, can you please update this PR with current `develop`?

I want to make sure that the easyconfigs you remove here and added via #3522 are still there...

AFCJamie:
Okay, will do.

AFCJamie:
It seems to have included a lot of extra files, I'm guessing this shouldn't have happened? Would it be easier to open a new PR with a fresh branch off develop?

boegel:
@AFCJamie something went wrong, indeed...

Did you do a `git rebase`? I've seen this happening before...

I usually do a `git merge develop`. That creates an extra merge commit, which is less 'clean' if you're a Git purist, but we're not. ;-)

Seems like a new branch/PR is going to be cleanest, indeed...

AFCJamie:
@boegel  Yes, I used rebase, I think something may have been out of sync before I pushed. No worries, I'll give that a try next time I need to do anything like this. I don't mind either way personally, I am never sure when is best to use either haha. I'll make a new PR now.

Issue #2261, opened 2016-01-05
Make failing for Tk-8.6.4-foss-2015b-no-X11.eb


nathanhaigh:
Trying to install this (a Python dependency) and I get errors from `make` like this:

``` bash
In file included from /mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tkPort.h:21:0,
                 from /mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tkInt.h:19,
                 from /mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tkStubLib.c:14:
/mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tk.h:96:25: fatal error: X11/Xlib.h: No such file or directory
 #   include <X11/Xlib.h>
                         ^
compilation terminated.
make: *** [tkStubLib.o] Error 1
make: *** Waiting for unfinished jobs....
In file included from /mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tkPort.h:21:0,
                 from /mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tkInt.h:19,
                 from /mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tk3d.c:14:
/mnt/bioinf-7/easybuild/build/Tk/8.6.4/foss-2015b-no-X11/tk8.6.4/unix/../generic/tk.h:96:25: fatal error: X11/Xlib.h: No such file or directory
 #   include <X11/Xlib.h>
                         ^
```

I assumed that since this is for `no-X11` then then I should get an error-free build. Am I missing something? For someone who know this stuff better than me, the following link may help: http://wiki.tcl.tk/1843.

Does it require a dependency adding for supplying these headers or should the ones supplied by Tk in the `xlib/X11/` subdirectory.

boegel:
The intention is indeed that this Tk build doesn't require X11 headers, which is why we specify `--without-x` as a configure option, but it apparently still those.

[...]

PR #3519, opened 2016-09-02
Add warning about Tk and X

wpoely86:
To build Tk, you need an X11 stack. All current python easyconfig need
X11 to build. Follow up in
https://github.com/hpcugent/easybuild-easyconfigs/issues/2261


boegel:
lgtm

backelj:
@wpoely86 @boegel : I have the same issue when building Tk-8.6.6-intel-2017a.eb, it needs X11, but there's no warning in those easyconfig files — what's the reasoning behind this? Should this not be a (os)-dependency?

wpoely86:
@backelj No particular reason other then we haven't done it yet :wink:
Add a dep to the X11 bundle in `Tk-8.6.6-intel-2017a.eb` and you should be fine.

I think X11 is a bit too much to drag in as a OS dep.

jrbosch:
builddependencies or dependencies??. boegel suggests builddependencies in issue #2261 . See [https://github.com/hpcugent/easybuild-easyconfigs/issues/2261#issuecomment-196709139]

PR #3522, opened 2016-09-05
{lang}[foss/2016.04] Python 3.5.2

AFCJamie:
An easyconfig for Python 3.5.2 as well as its (many) dependencies. The reason for foss 2016.04 is because this Python is needed for Blender 2.66a (#3518) and I used a foss toolchain that is similar to 2016.04 in order to build Blender.

PR #3530, opened 2016-09-06
{vis}[foss/2016.04] Blender 2.66a

AFCJamie:
Fixed PR for #3518.

An easyconfig for Blender 2.66a which was built with foss 2016.04 (includes dependencies Boost 1.60.0 and CMake 3.6.1 - Python 3.5.2 was merged in #3522).


easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-Python-3.5.2.eb
```diff
@@ -0,0 +1,40 @@
...
+
+# These are needed until extra dependencies are added for them to work 
+configopts = '-DWITH_INSTALL_PORTABLE=OFF '
+configopts += '-DWITH_BUILDINFO=OFF '
+configopts += '-DWITH_GAMEENGINE=OFF '
+configopts += '-DWITH_CYCLES=OFF '
+configopts += '-DWITH_SYSTEM_GLEW=OFF '
+
+dependencies = [
+    ('Python', pyver),
```
boegel: @AFCJamie no need to add a `pyver` local variable, just use `'3.5.2'` here, the EasyBuild framework will make sure that the `%(pyver)s` template is resolved (see http://easybuild.readthedocs.io/en/latest/version-specific/easyconfig_templates.html)

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 3 (3 easyconfigs in this PR)
node2689.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/7e011acd02c2b4711df144083709decf for a full test report.

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 3 (3 easyconfigs in this PR)
node2034.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/9b96935ec1860e941d493851e4134f78 for a full test report.

AFCJamie:
I'm investigating the error now.

boegel:
@AFCJamie looks like a missing dep on `libjpeg-turbo`?

AFCJamie:
Do you think that could be causing the Python errors? I'll try building it and adding it now but after adding the Python paths to CMake as configopts I get errors  like this at buildtime:

error: implicit declaration of function '_Py_atomic_load_relaxe
d' [-Werror=implicit-function-declaration]
 #define PYC_INTERPRETER_ACTIVE (((PyThreadState *)_Py_atomic_load_relaxed(&_PyThreadState_Current))
 != NULL)

...

cc1: some warnings being treated as errors

AFCJamie:
I've added libjpegturbo 1.5.0 but the Python errors still persist, I'll push up now so you can see.

AFCJamie:
Changing 
configopts += '-DPYTHON_LIBRARY=${EBROOTPYTHON}/lib '
to
configopts += '-DPYTHON_LIBRARY=${EBROOTPYTHON}/lib/libpython3.so '
didn't fix it either.

AFCJamie:
Is this relevant?
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=810065

AFCJamie:
To note as well, I don't think I had any build errors when building with Python 3.4.3.

boegel:
@AFCJamie sure, if this is of value to you, it makes sense to put together an easyconfig with dependency versions that work

I'm surprised the Python version matters at all though, in my experience bumping to latest is never a problem there.

Boost and CMake are different. Boost tends to change their API happily across releases, while I've seen problems with using a too recent CMake as well left and right..

boegel:
@AFCJamie https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=810065 seems to confirm that Python 3.5 isn't going to work out with this old Python version

Do you find any references w.r.t. the required Boost version? Because 1.51.0 is *very* old (and you may experience problems with getting it to work on a recent OS)...

AFCJamie:
@boegel  I figure it would be good to get at least one working easyconfig up on the repo as a starting point for a more developed Blender easyconfig that targets the more recent versions. It feels like a bit of a balancing act trying to find a combination that works, but I tried with Boost 1.60.0 (as well as Python 3.3.2 and CMake 3.4.3) and it does work, so that is good. However, the combination Boost 1.60.0, Python 3.4.3 and CMake 3.6.1 did not work, sadly. So I will remove the other files from this commit and push up the ones I think should work. Fortunately, Python 3.4.3 does not have as many dependencies as the dependencies needed for 3.5.2 so should I put everything in here?

boegel:
@AFCJamie if you can keep the amount of easyconfigs below 10, it's OK to include them all in this PR (especially since the ones for Python & co are easy to review via `eb --review-pr`)

FYI: I'm working on an easyconfig for the latest version of Blender on top of `intel/2016b` and Python 2.7.11 based on your easyconfig, PR coming soon.

AFCJamie:
@boegel Okay, fortunately bzip2, zlib and ncurses are already in develop from the PR for Python 3.5.2 so I only had to include libreadline with this PR for Python. I did try a build of Python 3.3.2 with libreadline 6.3 before pushing but that did not work, reverting back to 6.2 allowed the build to succeed so I've stuck with 6.2 here.

easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-Python-3.3.2.eb
```diff
@@ -21,17 +21,12 @@ configopts += '-DWITH_GAMEENGINE=OFF '
 configopts += '-DWITH_CYCLES=OFF '
 configopts += '-DWITH_SYSTEM_GLEW=OFF '
 
-# Python paths
-configopts += '-DPYTHON_LIBRARY=${EBROOTPYTHON}/lib '
-configopts += '-DPYTHON_INCLUDE_DIR=${EBROOTPYTHON}/include/python3.5m '
```
boegel: @AFCJamie why did you drop these? They look relevant, but may need some tweaking, e.g. you could replace the hardcoded `3.5` with `%(pyshortver)s`

boegel:
Test report by @boegel
**FAILED**
node2601.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/14f80038ec20419c132959a2444edbf9 for a full test report.

boegel:
Test report by @boegel
**FAILED**
node2011.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/d26add0e5250d7ec7eec85ab19ce3714 for a full test report.

boegel:
@AFCJamie last failed test looks like a missing `zlib` dep for `Blender`?

easybuild/easyconfigs/b/Blender/Blender-2.66a-foss-2016.04-Python-3.3.2.eb
AFCJamie: @boegel I didn't need them for Python 3.3.2 when I built it for some reason but I'll include them back in.

AFCJamie:
Hmm okay, I'm not sure if it used a system zlib here then when building without it specified in the dependencies and I didn't realise. I've added zlib and libjpeg-turbo back as dependencies so hopefully that helps.

boegel:
@AFCJamie an easyconfig for the latest version of Blender with the `intel/2016b` toolchain just got merged via #3553 

I had to make a couple of changes on top of your easyconfig to make it build/work:

* (include a patch to fix a compilation issue with the Intel compilers)
* Blender requires X11 libraries, so I had to include the `X11` bundle as a dependency
* Blender requires OpenGL support, so I had to include the `Mesa` bundle as a dependency
* include OpenImageIO (and OpenEXR on which it depends) to get the 'cycles' render engine working; I was forced to do this because the default render engine wasn't produced correct outputs for my test case (gray screen in an animated rendering); using the 'cycles' engine (via `blender -E CYCLES ...`) does work though
* I had to make sure that `$GALLIUM_DRIVER` is defined to `swr`; without this, `blender` segfaults at startup of the GUI...

I think the minimal that should be done here is to also include `X11` and `Mesa` as dependencies as well, since those are always required imho (you probably have these available on your test system)? Or, at the very least, include them as OS dependencies (via `osdependencies`)...

Thoughts on this?

AFCJamie:
@boegel That looks really good, it sounds like you have had to make a fair few additions to get it to work. Seeing as the latest version has been merged - do you feel this version is still relevant to try and merge given that it's considerably older? As I said before I only built this particular version due this version being the one we currently have on our system (from pre-EasyBuild times) so if 2.77a is working okay, 2.66a may not be necessary any longer?

boegel:
@AFCJamie unless you feel strongly about, I'm OK with leaving this as is...

AFCJamie:
I feel okay either way. I initially only made this PR because I noticed one didn't already exist on the repo but now, as there's a more recent one, I am happy to close this and try and look at different software/versions to try and push up. :-)

boegel:
@AFCJamie OK, in that case, please close this PR (we try to avoid closing PRs made by others ourselves)

AFCJamie:
Okay, will do.

Issue #3408, opened 2016-08-09
libxml2 depends on liblzma


akesandgren:
If the system has the development package for liblzma installed the build of libxlm2 will have dependencies on lzma_<functions>. The static libxml2.a will not have these resolved and thus building, for instance, GROMACS will fail with unresolved symbols.

I think libxml2 should either depend on lzma (better choice) or forcibly disable lzma (not as good).

boegel:
@akesandgren adding lzma (`XZ`) as a dependency for libxml2 is fine by me to avoid that it's being picked up from the OS

I also considered this recently, but I ran into a blocker: XZ (which provides `liblzma`) depends on `gettext`, which in turn depends on `libxml2`. So including `XZ` as a dep for `libxml2` would introduce a cyclic dep.

We're probably missing something here, like one of these actually only being a build dep or something like that...

akesandgren:
gettext is only a builddep for XZ, so this should be doable.

akesandgren:
xz doesn't need gettext to conf/build "correctly" (it will simply lack language support for its messages), so a bootstrap xz that doesn't have a builddep on gettext (may use the system gettext binaries if installed, which they almost always are) that can then be used to build libxml2 -> gettext -> xz

Another way would be to start with a bootstrap libxml2 without lzma support, and then take the chain forward.

And by bootstrap i of course mean using dummy toolchain.

akesandgren:
I currently have no libxml2, gettext, or XZ installed (by easybuild).
A bit of config writing gives me the following chain, does it look sane?

b-an03 [easybuild]$ eb libxml2-2.9.4-foss-2016b.eb --dry-run 
== temporary log file in case of crash /scratch/ake/tmp/eb-vzcjwj/easybuild-AXT1fx.log
Dry run: printing build status of easyconfigs and dependencies
 * [ ] /pfs/nobackup/home/a/ake/eb-tree/ebfiles_repo/l/libxml2/libxml2-2.9.4.eb (module: libxml2/2.9.4)
 * [x] /eb/opt/easybuild_repo/M4/M4-1.4.17.eb (module: M4/1.4.17)
 * [ ] /eb/common/software/EasyBuild/2.8.2/lib/python2.7/site-packages/easybuild_easyconfigs-2.8.2-py2.7.egg/easybuild/easyconfigs/a/Autoconf/Autoconf-2.69.eb (module: Autoconf/2.69)
[...]

PR #3568, opened 2016-09-14
Add missing dependency on XZ to libxml2.

akesandgren:
The original problem is that when building libxml2 on a system with
liblzma installed by the distro, it picks up dependencies on lzma
routines. Some of the versions of Gromacs are linking against libxml2
statically, causing link errors against lzma. And any other software that
does static linking against libxml2 will have the same problem.

There is a cyclic dependency between XZ, libxml2, and gettext.
Fix this by adding a dummy toolchain build of XZ without a dependency on
gettext. gettext is usually installed by the distribution, and if it
isn't XZ will still build but without language support for its messages.

libxml2 for the updated toolchains are then built against the
dummy-toolchain XZ. And then gettext and XZ can be built against that
libxml2.

This was originally discussed in issue #3408.


boegel:
@akesandgren this introduces a bunch of conflicts for stuff that depends on `libxml2`...

Can we make the `dummy` `XZ` a build-only dep for `libxml2`?

akesandgren:
Doh, example of conflict?

As stated in the original problem, libxml2 must have XZ as a dependency due to libxml2.a having dependencies on lzma functions. A build dep doesn't solve that.

akesandgren:
Ahh.... stuff that depends on XZ-<toolchain>.... (checked the details of the failed travis builds)
I'll check the built libxml2.a in more detail...

akesandgren:
Ok..... that's not going to fly at all then....

The minimum impact i can think of then is to build XZ-<toolchain> without gettext as a dependency and then build xml2 on that instead. But then XZ looses language support for messages. Which isn't that big a problem in my opinion.

The question then is if XZ should be built --without-gettext or always have gettext as an OS-dep, or just let it find it if it's there...?

boegel:
@akesandgren are you saying `libxml2.a` dynamically links to `XZ` libs? Can we statically link (and thus avoid a runtime dep on `XZ`?

akesandgren:
No, but using libxml2.a requires an explicit link with XZ libs (and zlib too actually).
And you can't statically link them together and you shouldn't. The libxml2.la contains the required link info but it's usually only used when autotools are in play. CMake ignores it.

boegel:
@akesandgren why couldn't (or shouldn't) you statically link the XZ lib in `libxml2.a`? That looks like a nice way to get out of this mess to me...

akesandgren:
because the libxml2.a will then contain the XZ functions and you can't link that libxml2.a with a different liblzma.a/so. And to do it you would need to extract all .o's from liblzma and add them to libxml2.a

This is bad behavior and something i strongly advise against.

akesandgren:
Pushed an amended commit which should work better and not cause conflicting dependencies.

akesandgren:
Also updated the original comment (at the top) to match the current PR version.

akesandgren:
Ok, so those changes to libxml2 requires more XZ .eb files.... fixing that too then.

easybuild/easyconfigs/x/XZ/XZ-5.2.2-GCCcore-4.9.3.eb
```diff
@@ -0,0 +1,27 @@
...
+description = "xz: XZ utilities"
+
+toolchain = {'name': 'GCCcore', 'version': '4.9.3'}
+
+sources = [SOURCELOWER_TAR_BZ2]
+source_urls = ['http://tukaani.org/xz/']
+
+builddependencies = [
+    ('Autotools', '20150215'),
+    ('gettext', '0.19.8', '', True),
```
boegel: @akesandgren please include `binutils` as a build dep too here, otherwise this fails with the following (from `config.log`):

```
collect2: fatal error: cannot find 'ld'  
``` 

```python
builddependencies = [
     ('binutils', '2.25'),
 ]
```

boegel: note: the `cannot find 'ld'` issue only happens on systems where `ld.gold` isn't available system-wide...

akesandgren: Fixed.

easybuild/easyconfigs/l/libxml2/libxml2-2.9.4-foss-2016a.eb
```diff
@@ -15,6 +15,9 @@ source_urls = [
 ]
 sources = [SOURCELOWER_TAR_GZ]
 
-dependencies = [('zlib', '1.2.8')]
+dependencies = [
+    ('zlib', '1.2.8'),
+    ('XZ', '5.2.2'),
+}
```
boegel: @akesandgren typo, should be `]`

akesandgren: Fixed with new commit

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 17 out of 17 (17 easyconfigs in this PR)
node2726.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/4a5c77648619fc265be184c2615b1180 for a full test report.

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 21 out of 24 (24 easyconfigs in this PR)
node2146.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/8aceaeef4ed8ef1e421261523b1146ce for a full test report.

boegel:
@akesandgren hmm, trouble in paradise with `gettext` 0.19.7:

```

libtool: link: gcc -std=gnu99 -I///usr/include/libxml2 -g -O2 -o .libs/cldr-plurals cldr_plurals-cldr-plural.o cldr_plurals-cldr-plural-exp.o cldr_plurals-cldr-plurals.o  ./.libs/libgettextsrc.so /tmp/vsc40023/easybuild_build/gettext/0.19.7/dummy-/gettext-0.19.7/gettext-tools/gnulib-lib/.libs/libgettextlib.so -lc ../gnulib-lib/.libs/libgettextlib.so -lm -lxml2 -lncurses -fopenmp -Wl,-rpath -Wl,/user/home/gent/vsc400/vsc40023/eb_phanpyscratch/SL6/sandybridge/software/gettext/0.19.7/lib
./.libs/libgettextsrc.so: undefined reference to `xmlOutputBufferGetContent'
./.libs/libgettextsrc.so: undefined reference to `xmlXPathSetContextNode'
collect2: ld returned 1 exit status
make[5]: *** [msgexec] Error 1
```

note the weird path passed to `-I`...

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 41 out of 45 (24 easyconfigs in this PR)
node2726.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/6870e1e16535f040df2401aaf22673eb for a full test report.

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 23 out of 27 (24 easyconfigs in this PR)
node2726.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/dc3bcabb1da0b548cf69b2a64a84c600 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 24 out of 24 (24 easyconfigs in this PR)
node2146.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/390656a09fe15a124e43405b537b88e2 for a full test report.

boegel:
@akesandgren ignore the failing test report on CentOS7, that's just because `goolf/1.4.10` isn't compatible with that hardware (OpenBLAS version is too old for Haswell)

boegel:
lgtm

boegel:

PR #3592, opened 2016-09-21
{chem}[intel/2016a] NWChem-6.6

RvDijk:


easybuild/easyconfigs/n/NWChem/NWChem-6.6.revision27746-foss-2016a-Python-2.7.11.eb
```diff
@@ -0,0 +1,40 @@
+name = 'NWChem'
+version = '6.6.revision27746'
+verdate = '2015-10-20'
+pyver = '2.7.11'
+versionsuffix = '-%s-Python-%s' % (verdate, pyver)
```
boegel: @RvDijk please make the easyconfig filename match this versionsuffix (`2015-10-20`  is missing)

easybuild/easyconfigs/n/NWChem/NWChem-6.6.revision27746-foss-2016a-Python-2.7.11.eb
```diff
@@ -0,0 +1,40 @@
+name = 'NWChem'
+version = '6.6.revision27746'
+verdate = '2015-10-20'
+pyver = '2.7.11'
```
boegel: please use `%(pyver)s` instead:

```python
versionsuffix = '-%s-Python-%%(pyver)s' % verdate
```

easybuild/easyconfigs/n/NWChem/NWChem-6.6.revision27746-foss-2016a-Python-2.7.11.eb
```diff
@@ -0,0 +1,40 @@
...
+
+patches = [
+    'NWChem_fix-date.patch',
+    'Ga_argv.patch',
+    'Raman_displ.patch',
+    'Ga_defs.patch',
+    'Cosmo_meminit.patch',
+    'Dplot_tolrho.patch',
+    'Sym_abelian.patch',
+    'Xccvs98.patch',
```
boegel: @RvDijk it's better to name these `NWChem-6.6.revision27746_foo.patch` to make it clear on top of which NWChem version these were created

easybuild/easyconfigs/n/NWChem/NWChem-6.6.revision27746-foss-2016a-Python-2.7.11.eb
```diff
@@ -0,0 +1,40 @@
...
+]
+
+dependencies = [
+    ('Python', '2.7.11')
+]
+
+modules = 'all python'
+
+# more test failures than normal?
+max_fail_ratio = 0.60
```
boegel: how many tests are failing, did you check?

easybuild/easyconfigs/n/NWChem/NWChem-6.6.revision27746-foss-2016a-Python-2.7.11.eb
boegel: also, where did you get these patches?

RvDijk: @boegel the patches are from their own website:
http://www.nwchem-sw.org/index.php/Download

wpoely86: Treat them like sources? As we do for the mpfr patches in GCC?

hajgato: +1

easybuild/easyconfigs/n/NWChem/NWChem-6.6.revision27746-foss-2016a-Python-2.7.11.eb
hajgato: Well, the version numbering is a bit problematic in case of `NWChem` but I am not able to find better way either. (For example how to indicate which patches are installed). `NWChem` is a bit rolling release, with snapshots (svn27746 = v6.6), and the patches are issued that affect old functionality, but patches are bringing new functionality will be released when a new snapshow will be released

easybuild/easyconfigs/n/NWChem/NWChem-6.6.revision27746-foss-2016a-Python-2.7.11.eb
RvDijk: @boegel too many.. I'm getting some strange errors about ARMCI Error():208 and ARMCI DASSERT fail.

boegel:
@RvDijk is this still WIP?

RvDijk:
@boegel I'm still on it... most of the times the installation works but the tests fail (armci error():208). Fixing this takes a lot of time since it takes a lot of time in order to get to the testing phase.
Also, has the easyblock to be adjusted s.t. it treats the patches as sources?

RvDijk:
@boegel we decided to stick with the intel version since the foss toolchain requires a complex compilation for which 32 bit integers are used.

boegel:
@RvDijk what's up with all the patches, where do they come from?

Can we collapse them all into a single large patch, and clearly mention how it was composed, and from where the patches come?

RvDijk:
@boegel the patches are provided by NWChem and are listed on http://www.nwchem-sw.org/index.php/Download right below the download link.  The last commit collapses them all into one single large patch.

RvDijk:
This PR has become outdated, deleting it.

Issue #3622, opened 2016-09-27
ncurses-6.0 fails to build on Ubuntu 16.04


riccardomurri:
Trying to build `ncurses-6.0.eb` on a Ubuntu 16.04 VM with the `foss-2016b`
toolchain installed ends with an error::

```
$ eb /opt/easybuild/software/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_easyconfigs-2.9.0-py2.7.egg/easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
== temporary log file in case of crash /tmp/eb-kxD7o6/easybuild-K6tscS.log
== resolving dependencies ...
== processing EasyBuild easyconfig /opt/easybuild/software/EasyBuild/2.9.0/lib/python2.7/site-packages/easybuild_easyconfigs-2.9.0-py2.7.egg/easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
== building and installing ncurses/6.0...
== fetching files...
== creating build dir, resetting environment...
== unpacking...
== patching...
== preparing...
== configuring...
== building...
== FAILED: Installation ended unsuccessfully (build directory: /tmp/ncurses/6.0/dummy-dummy): build failed (first 300 chars): cmd " make -j 1 " exited with exitcode 2 and output:
cd man && make DESTDIR="" RPATH_LIST="/opt/easybuild/software/ncurses/6.0/lib" all
make[1]: Entering directory '/tmp/ncurses/6.0/dummy-dummy/ncurses-6.0/man'
/bin/sh ./MKterminfo.sh ./terminfo.head ./../include/Caps ./terminfo.tail >terminfo.5
mak
== Results of the build can be found in the log file(s) /tmp/eb-kxD7o6/easybuild-ncurses-6.0-20160927.130156.iSdal.log
[...]

PR #3624, opened 2016-09-27
Make `ncurses` easyconfigs work on Debian/Ubuntu

riccardomurri:
Fixes issue #3622 when building on Debian/Ubuntu.

### DIFF
--- a/easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
+++ b/easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
@@ -13,10 +13,8 @@ toolchainopts = {'optarch': True, 'pic': True}
 
 source_urls = [GNU_SOURCE]
 sources = [SOURCE_TAR_GZ]
+patches = ['ncurses-6.0_mawk_compatibility.patch']
 
-# ncurses-6.0 builds fail with `mawk`, so force use of `gawk`
-# see http://lists.gnu.org/archive/html/bug-ncurses/2015-08/msg00013.html
-preconfigopts = "env AWK=/usr/bin/gawk "
 configopts = [
     # default build
     '--with-shared --enable-overwrite',
--- a/easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
+++ b/easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
@@ -14,6 +14,9 @@ toolchainopts = {'optarch': True, 'pic': True}
 source_urls = [GNU_SOURCE]
 sources = [SOURCE_TAR_GZ]
 
+# ncurses-6.0 builds fail with `mawk`, so force use of `gawk`
+# see http://lists.gnu.org/archive/html/bug-ncurses/2015-08/msg00013.html
+preconfigopts = "env AWK=/usr/bin/gawk "
 configopts = [
     # default build
     '--with-shared --enable-overwrite',
--- a/easybuild/easyconfigs/n/ncurses/ncurses-5.9.eb
+++ b/easybuild/easyconfigs/n/ncurses/ncurses-5.9.eb
@@ -14,7 +14,10 @@ toolchainopts = {'optarch': True, 'pic': True}
 source_urls = [GNU_SOURCE]
 sources = [SOURCE_TAR_GZ]
 
-patches = ['ncurses-%(version)s_configure_darwin.patch']
+patches = [
+        'ncurses-%(version)s_configure_darwin.patch',
+        'ncurses-5.9_gcc5x.patch',
+]
 
 configopts = [
     # default build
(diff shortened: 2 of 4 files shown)


easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
```diff
@@ -14,6 +14,9 @@ toolchainopts = {'optarch': True, 'pic': True}
 source_urls = [GNU_SOURCE]
 sources = [SOURCE_TAR_GZ]
 
+# ncurses-6.0 builds fail with `mawk`, so force use of `gawk`
+# see http://lists.gnu.org/archive/html/bug-ncurses/2015-08/msg00013.html
+preconfigopts = "env AWK=/usr/bin/gawk "
```
boegel: @riccardomurri what if `usr/bin/gawk` isn't available?

riccardomurri: What other `awk`'s are available on Linux? I only know of `mawk` and `gawk` -- and ncurse won't compile with `mawk`.

riccardomurri: I can change that to read `AWK=/usr/bin/awk` but then it relies on admins to do `update-alternatives --set awk /usr/bin/gawk` e.g. it might not work out of the box on Debian ...

boegel: @riccardomurri I'm worried about the case where *no* awk is available on the system... Or would the `ncurses` install then fail entirely? Should we add an `awk` build dep that is handled via EasyBuild?

boegel:
@riccardomurri can you update the PR according to your last update?

boegel:
@riccardomurri also, a test report on Ubuntu would be fantastic for this one... :-)

boegel:
Test report by @boegel
**SUCCESS**
node2564.golett.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/a3ca6d1273bb4465fe5ff6655c31bae4 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2009.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/e30b8019b4a645e80b297fe80e69af62 for a full test report.

easybuild/easyconfigs/n/ncurses/ncurses-6.0.eb
riccardomurri: **Outdated** as this part of the code has been superceded in later commits.

boegel:
Test report by @boegel
**SUCCESS**
daint101 - Linux SLES 11_SP3, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.9
See https://gist.github.com/852589a898a002b7cc23fb7331b665ac for a full test report.

riccardomurri:
> @riccardomurri can you update the PR according to your last update

Title updated.

riccardomurri:
> @riccardomurri also, a test report on Ubuntu would be fantastic for this one... :-)

Can you pls remind me how do I generate a test report?

boegel:
@riccardomurri by combining `--from-pr` and `--upload-test-report` (or `-u`), see http://easybuild.readthedocs.io/en/latest/Integration_with_GitHub.html#uploading-test-reports-upload-test-report

It requires a bit a setup (GitHub token mostly), you can verify with `eb --check-github` that everything is in place.

boegel:
cc @akesandgren, who mentioned to be suffering from this as well...

akesandgren:
Yes, that patch is the "good" one, but making a builddep on mawk 1.3.4 is probably better. See ncurses commit 53ea71aa108570a4a070e004484642a75f3492df.

I.e., there are more places which need an updated mawk compared to the one in Debian/Ubuntu

riccardomurri:
> making a builddep on mawk 1.3.4 is probably better

Well, but then ncurses > 6.0 would not need it, so we would end up with the strange situation that ncurses-6.0 depends on mawk>1.3.4 but ncurses-6.1 does not...

On the other hand, the patch would not apply to ncurses>6.0 so `--try-software-version=6.1` would be broken... so maybe depending on an `awk` with POSIX RE support is the way.  But then why don't we just use `gawk` (it's already packaged for EB)?

akesandgren:
mawk is also in EB.  The question then is if that awk script used in ncurses-6.0 is fully compatible with both of them or not. Including the other mawk scripts used in ncurses.

And ncurses > 6.0 will still need a working awk (mawk or gawk), it's not just this place that uses mawk scripts in ncurses.

boegel:
@riccardomurri I think #3792 is the proper fix here, I was sitting with someone who was running into the failing ncurses build with the patch in place, while it work fine with the patch...

boegel:
@riccardomurri Can you verify whether the fix added in #3792 fixes the problem you had, so this can be closed?

migueldiascosta:
@riccardomurri, this PR is being closed for the following reason(s): no activity for > 1 year.

PR #3629, opened 2016-09-27
Update GCCcore-5.4.0.eb

serge-p:
- added original source url for Integer Set Library (isl-0.15.tar.bz2)


boegel:
@serge-p should this same change be made in the other GCC(core) easyconfigs?

cc @wpoely86

wpoely86:
Yes, definitely. Weird we forgot that one. GCC usually has a copy too but the original is always better.

boegel:
@serge-p are you up for applying this same change to the other GCC(core) easyconfigs?

boegel:
merged via #3798

Issue #3635, opened 2016-09-30
Cython-0.22.tar.gz was not found (HTTP response code 404)


mictadlo:
Hi,
I used `eb BLAST+-2.2.31-foss-2015b-Python-2.7.10.eb --robot` and got `Download of http://www.python.org/ftp/python/2.7.10/Cython-0.22.tar.gz to /mnt/easybuild/sources/p/Python/extensions/Cython-0.22.tar.gz failed, done trying`.

How is it possible to put a new URL?

Thank you in advance, 

Best wishes,

Mic

boegel:
@mictadlo EasyBuild should also be trying `http://cython.org/release/Cython-0.22.tar.gz` (can you confirm that by checking the log?), which is supposed to be the correct URL for Cython, but it seems the sources are no longer available there...

You can download this tarball from PyPI instead: https://pypi.python.org/packages/f8/25/80f9ca7e31e2b68cc942ff1d6136588f33a7aef5e2d6abe3f2183cb9fad5/cython-0.22.tar.gz (obtained via https://pypi.python.org/pypi/Cython/0.22)

You can then place the downloaded tarball in the location where EasyBuild would download it, in the source path (cfr. http://easybuild.readthedocs.io/en/latest/Configuration.html#sourcepath). If EasyBuild finds the tarball there, it will not try to download it.

To figure out the location in your setup, run `eb --show-config`, look for `sourcepath`, and then copy the tarball to the `p/Python/extensions` subdirectory in that location.

Do note that you'll need to rename the tarball from `cython-0.22.tar.gz` to `Cython-0.22.tar.gz`; it seems they provide the tarball in a different name in PyPI... >_<

I'll look into fixing the download URLs for Cython, thanks for the report!

mictadlo:
Yes, it is correct. I downloaded from PyPi, but still had to rename it to `Cython-0.22.tar.gz`.

boegel:
fixed in #3636, thanks again for reporting @mictadlo!

PR #3636, opened 2016-09-30
fix source URLs for Cython

boegel:
fix for problem reported in #3635


boegel:
tested, downloads are fixed

PR #3700, opened 2016-10-22
Make OpenBLAS use the designated LAPACK version.

akesandgren:
The OpenBLAS Makefile doesn't unpack any lapack-x.y.z.tgz, it only deals
with timing.tgz and large.tgz.
To make OpenBLAS use the version of LAPACK specified in the easyconfig
we need to add a small patch to the OpenBLAS Makefile that unpacks it if
it exists.

This solves easyconfig issue# 1050.


akesandgren:
Pull request withdrawn, need to do this another way, The unpacked Lapack needs to be patchable before build starts,.

Issue #1050, opened 2014-08-28
OpenBLAS is not using provided LAPACK sources


boegel:
The OpenBLAS 0.2.8 and 0.2.9 easyconfigs we're using now are not using the provided LAPACK sources, but are resorting to the `lapack-netlib` included in the OpenBLAS sources.

For OpenBLAS 0.2.9 and more recent this is not a big problem, since the LAPACK included in the sources matching the one we're passing it (3.5.0), see https://github.com/xianyi/OpenBLAS/blob/develop/lapack-netlib/README .

akesandgren:
Still is still a problem for newer openblas/lapack combos, like openblas-0.2.18-lapack-3.6.1 where openblas contains lapack-3.6.0
I have a fix for this and will send a PR soon

akesandgren:
PR 3700 added

akesandgren:
Unfortunately the patch in PR 3700 is not good enough to make me add patches onto LAPACK after it has been unpacked, which is needed to fix some compile time bugs...

akesandgren:
I'll withdraw that PR. I really need to get this done the correct way to solve all build problems.

boegel:
@akesandgren thanks for looking into this!

akesandgren:
Made a new PR (3701) that handles this in a much better way, and doesn't include patches.

It also makes it possible to add patches on to of the unpacked lapack to deal with compiler problems (and lapack bugs).

I'll start generating more PR for those problems, if this one gets accepted.

boegel:
fixed by #3701 & #3795

PR #3701, opened 2016-10-22
Make OpenBLAS use the LAPACK version specified in the easyconfig.

akesandgren:
The OpenBLAS Makefile only handles large.tgz and timing.tgz and ignores
lapack-x.y.z.tgz. We need to manually unpack the lapack-x.y.z specified
in the easyconfig.

Doing it this way makes it possible to add patches on top of the
unpacked lapack version, which is neccesary to deal with various
compiler problems.

This solves easyconfig issue #1050.

### DIFF
--- a/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.2.18-GCC-5.4.0-2.26-LAPACK-3.6.1.eb
+++ b/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.2.18-GCC-5.4.0-2.26-LAPACK-3.6.1.eb
@@ -3,6 +3,8 @@ easyblock = 'ConfigureMake'
 name = 'OpenBLAS'
 version = '0.2.18'
 
+obv = '%s-%s' % (name, version) # To be used in the lapack-x.y.z unpack cmd
+
 lapackver = '3.6.1'
 versionsuffix = '-LAPACK-%s' % lapackver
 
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.2.18-GCC-5.4.0-2.26-LAPACK-3.6.1.eb
```diff
@@ -16,7 +18,7 @@ large_src = 'large.tgz'
 timing_src = 'timing.tgz'
 sources = [
     'v%(version)s.tar.gz',
-    lapack_src,
+    (lapack_src, 'mv %s/lapack-netlib %s/lapack-netlib.old; mkdir %s/lapack-netlib; tar -C %s/lapack-netlib --strip-components=1 -zxf %%s' % (obv, obv, obv, obv)),
```
boegel: @akesandgren I would try and shorten this a bit (lines should be < 120 chars long); also, I see no point in keeping `lapack-netlib.old` around...

```
lapack_unpack_cmd = "tar -C lapack-netlib --strip-components=1 -zxf %%s"
sources = [
    'v%(version)s.tar.gz',
    (lapack_src, "cd %%(name)s-%%(version)s; rm -rf lapack-netlib/*; %s; cd -" % lapack_unpack_cmd
    ...
```

akesandgren:
Dropped saving of old lapack-netlib and made a more easily readable lapack_unpack_cmd.

boegel:
merged via #3795

PR #3702, opened 2016-10-24
Update WRF-3.8.0_netCDF-Fortran_separate_path.patch

andreas-h:
I believe this is a typo preventing the `NETCDFFPATH` variable to be set correctly.


boegel:
@andreas-h hmm, sure looks like a typo indeed, but I'm a bit puzzled as to how we missed this up until now?

What part of WRF is broken without this fix?

andreas-h:
It's indeed very strange.  WRF3.8.1 builds fine without this fix on my installation, but during my ongoing work to compile WRF with Chemistry enabled (some extra environment variables set), some `Makefile` of the chemistry code started complaining that it couldn't find the netCDF-Fortran library ...  I'm awfully behind on this project; hopefully I can continue end of next week (and then also answer your question about the other PR).

boegel:
@andreas-h OK, thanks for the info. I'll make sure this doesn't break the WRF 3.8 installation on my end, but I don't expect it to. If it's fine, I'll merge this ASAP, thanks!

boegel:
tested, works, so going in, thanks @andreas-h!

PR #3703, opened 2016-10-25
use $CC,$CXX rather than $I_MPI_CC,$I_MPI_CXX in patch for OpenFOAM 4.0

boegel:
@wpoely86 just to ensure that this patch can be reused with other (non-Intel MPI) toolchains


boegel:
good to go, thanks for the review @wpoely86!

```
$ eb OpenFOAM-4.0-intel-2016b.eb  -x | egrep 'export CC=|export CXX=|export I_MPI_CC=|export I_MPI_CXX='
  export CC="icc"
  export CXX="icpc"
  export I_MPI_CC="icc"
  export I_MPI_CXX="icpc"

```

PR #3715, opened 2016-10-25
{math}[intel/2016b] Qhull 2015.2 (REVIEW)

hajgato:


boegel:
already merged via #3719, this easyconfig only includes an outdated/useless comment compared to the Qhull easyconfig from #3719

PR #3725, opened 2016-10-27
move easyconfigs for deprecated iqacml & iiqmpi toolchains to archive

boegel:
support for two software packages would be lost (without considering archived easyconfigs):

* CBLAS (fixed via ~~#3726~~)
* QLogicMPI (doesn't matter, it was just a very system-specific RPM)


boegel:

PR #3731, opened 2016-10-27
move easyconfigs for deprecated ictce toolchains to archive

boegel:
support lost for the following software packages (without considering archived easyconfigs):

* IronPython (fixed with ~~#3734~~)
* Mono (fixed with ~~#3734~~)
* MyMediaLite (fixed with ~~#3734~~)

(depends on ~~#3729~~ since `iomkl-4.6.13.eb` requires one of the OpenMPI easyconfigs being archived here)


boegel:

PR #1410, opened 2015-02-17
{toolchains} add [Clang/3.6.0] and deps, and toolchains above it, aligned with goolf/1.7.20 (REVIEW)

fgeorgatos:
This builds upon #1389 for the upcoming Clang/3.6.0 , 21 Feb. 2015, as per http://llvm.org
At moment of PR initiation, toolchains up to cgmvapich2 are included and more to be added asap.
It is aligned to recent toolchains #1294 & #1397 and features similar compatibility aspects.

PR #3735, opened 2016-10-28
move easyconfigs for deprecated Clang-based toolchains to archive

boegel:
This currently removes *all* existing Clang-based toolchains (`ClangGCC`, `cgmpich`, `cgmpolf`, `cgmvapich2`, `cgmvolf`, `cgompi`, `cgoolf`) to the archive, so even though it's still supported in the framework, no easyconfigs will be readily available as a part of EasyBuild v3.0 unless an update is contributed (e.g. #1410, but that got blocked because of compilation issues).


boegel:
@fgeorgatos I'm going to merge this as is; I know you'd like to see at least one Clang-based toolchain readily available, but that can be handled via a separate PR (either #1410, or an update of that).

PR #3747, opened 2016-11-01
new bio module prokka

fizwit:
update bioperl for foss-2016b; new easyconfig prokka


boegel:
@fizwit Travis tests fail because the easyconfig you added for `prokka` is not located in a subdirectory named `prokka` as it's expected to, can you please fix that?

boegel:
prokka contributed in #3755, so closing this

PR #3793, opened 2016-11-14
Eliminate dependency on mpi-mic-rt in ifort.

bartoldeman:
This dependency causes Intel mpirun/mpiexec to be added to $PATH
when the ifort module is loaded, which is not always desirable,
for instance when OpenMPI is already loaded.

Fixes hpcugent/easybuild-easyblocks#1032.

### DIFF
--- a/easybuild/easyconfigs/i/ifort/ifort-2016.0.109-GCC-4.9.3-2.25.eb
+++ b/easybuild/easyconfigs/i/ifort/ifort-2016.0.109-GCC-4.9.3-2.25.eb
@@ -10,6 +10,9 @@ toolchain = {'name': 'dummy', 'version': ''}
 
 sources = ['parallel_studio_xe_%(version_major)s_composer_edition_for_fortran.tgz']
 
+# remove dependency on intel-mpi-rt-mic
+patches = ['ifort_2016_no_mpi_mic_dependency.patch']
+
 checksums = ['bce7f6a71f7e44f67956197501d00b7c']
 
 gccver = '4.9.3'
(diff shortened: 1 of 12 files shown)


bartoldeman:
Note: I have tested the patch only on 2017.1, 2017.0 and 2016.2. Does your automatic testing test on the other 2016 varieties?

boegel:
@bartoldeman If I trigger a test report, all touched easyconfigs will be tested.

However, the required patch file `ifort_2016_no_mpi_mic_dependency.patch` is missing in this PR, can you please include that too?

Also, please also update `ifort-2016.0.109-GCC-4.9.3-2.25.eb` to use the patch, so we cover all `2016` and `2017` ifort versions?

Does this only apply to `ifort`, not `icc`?

bartoldeman:
my apologies for the missing files. There are there now.

icc is not affected at all.

for ifort I even see mpirt for 2015 versions, in
$EBROOTICC/mpirt/
but this is not picked up in the easyblock (easyblock looks for "mpi", not "mpirt").

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 11 out of 11 (10 easyconfigs in this PR)
node2059.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/54239399b29d4db5fac0d0016d560f76 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2407.golett.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/3eb909b87082cac7064c90f0c4eac7de for a full test report.

boegel:
Thanks for the update & info @bartoldeman!

I diffed the contents of the `ifort` install directory with and without this patch, and it looks as expected.

I see no reason to hold this back further, going in, thanks @bartoldeman!

PR #3798, opened 2016-11-14
include original download URL for ISL source tarball

boegel:
wrapper PR for #3629 which applies the fix consistently


boegel:
going in, thanks @serge-p!

PR #3801, opened 2016-11-15
disable patching of GCC to make it prefer /lib over /lib64 to avoid breaking the build

boegel:
In https://github.com/hpcugent/easybuild-easyblocks/pull/1030, patching of GCC was enabled to make it prefer `/lib` over `/lib64` directories.

This causes problems with older versions of GCC (4.7.2 and earlier, except for 4.6.4):

```
/usr/lib/crti.o: could not read symbols: File in wrong format
collect2: ld returned 1 exit status
```

with:

```
$ file /usr/lib/crti.o
/usr/lib/crti.o: ELF 32-bit LSB relocatable, Intel 80386, version 1 (SYSV), not stripped
```

This change simply disables the patching, to fix the build.
If needed, this can be revisited once the cause of the problem (and a fix) is figured out...

cc @sebth

### DIFF
--- a/easybuild/easyconfigs/g/GCC/GCC-4.2.4.eb
+++ b/easybuild/easyconfigs/g/GCC/GCC-4.2.4.eb
@@ -21,6 +21,8 @@ dependencies = [
 
 languages = ['c', 'c++', 'fortran']
 
+prefer_lib_subdir = False
+
 configopts = "--with-gmp=$EBROOTGMP --with-mpfr=$EBROOTMPFR"
 
 # building GCC sometimes fails if make parallelism is too high, so let's limit it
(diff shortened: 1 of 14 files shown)


boegel:
superseded by disabling `prefer_lib_subdir` by default in the `GCC` easyblock, see https://github.com/hpcugent/easybuild-easyblocks/issues/1035

boegel:
Test report by @boegel
**SUCCESS**
node2055.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/1d6d29bb79c1a91f59b153b69c9bf920 for a full test report.

PR #3806, opened 2016-11-16
add Octopus easyconfig for goolf toolchain

ssalama:


boegel:
@ssalama what happened here, did you delete your branch?

ssalama:
@boegel Yeah, but only for the mean time. The toolchain need another list of dependencies which is not there yet and it must be built with goolf-1.4.10. I am going to add this dependencies first, so it doesn't fail TravisCI build.

PR #3817, opened 2016-11-17
Fix incorrect descriptions for ifort.

akesandgren:
Some of the ifort .eb files had gotten an incorrect description,
claiming to be C/C++ instead of Fortran.


fgeorgatos:
Nice catch, Lgtm!

boegel:
:+1:

PR #3831, opened 2016-11-19
build both API62 and API80 in the libjpeg easyconfigs

pescobar:
Reasons for this PR are detailed in this issue: https://github.com/hpcugent/easybuild-easyconfigs/issues/3829

I think this won't break backward compatibility in any case because the latest `configure-make-makeinstall` cycle is still using the option `--with-jpeg8`


pescobar:
I fucked up something with git. closing this

Issue #3829, opened 2016-11-18
issues with current config in the libjpeg-turbo easyconfig


pescobar:
While creating new easyconfigs for some image analysis apps I noticed a really weird issue. See this output of `ldd`

```
tificc:
	linux-vdso.so.1 =>  (0x00007ffea04fc000)
	liblcms2.so.2 => /scicore/home/scicore/easybuild/test-installs/apps/LittleCMS/2.8-goolf-1.7.20/lib/liblcms2.so.2 (0x00007fbda7d7f000)
	libtiff.so.5 => /scicore/soft/apps/LibTIFF/4.0.4-goolf-1.7.20/lib/libtiff.so.5 (0x00007fbda7b0d000)
	libjpeg.so.8 => /scicore/soft/apps/libjpeg-turbo/1.4.1-goolf-1.7.20/lib/libjpeg.so.8 (0x00007fbda78b7000)
	libz.so.1 => /scicore/soft/apps/zlib/1.2.8-goolf-1.7.20/lib/libz.so.1 (0x00007fbda76a1000)
	libm.so.6 => /lib64/libm.so.6 (0x0000003823000000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x0000003822800000)
	libc.so.6 => /lib64/libc.so.6 (0x0000003822400000)
	libjpeg.so.62 => /usr/lib64/libjpeg.so.62 (0x00000032a5000000)
	/lib64/ld-linux-x86-64.so.2 (0x0000003821c00000)
```

Notice that this binary is using a `libjpeg.so.8` from the software stack and `libjpeg.so.62` from system libraries.  I have experienced the same exact issue with two different applications which had libjpeg-turbo as dependency. 

After debugging it a little bit I noticed that the libjpeg-easyconfig built with easybuild doesn't contain the library `libjpeg.so.62` because it's compiled with `configopts = "--with-jpeg8"` which removes support for the turbojpeg API 62. More details here:
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/BUILDING.md

After modifying my libjpeg easyconfig to remove the option `configopts = "--with-jpeg8` the missing library is generated and now my application correctly uses the libraries from my software stack instead of system libs. I couldn't find a way to force the application to don't use turbojpeg API 62

[...]

PR #3832, opened 2016-11-19
build both API62 and API80 for libjpeg

pescobar:
Reasons for this PR are detailed in this issue: #3829

I think this won't break backward compatibility in any case because the latest configure-make-makeinstall cycle is still using the option `--with-jpeg8` which was the default before this PR


pescobar:
More info can be found here: 
https://github.com/libjpeg-turbo/libjpeg-turbo#using-libjpeg-turbo

It seems that using `--with-jpeg8` only the TurboJPEG API is built. Applying this PR the libjpeg API is also built.

boegel:
@pescobar see https://github.com/hpcugent/easybuild-easyconfigs/issues/3829#issuecomment-261766069, this is not fully installing the most recent version of the API, since some files will get overwritten in the 2nd cycle.

As you have done it here, it's OK though, since we're not breaking backward-compatibility (the install cycle using `--with-jpeg8` is done last, so it still 'wins').

But I'm not sure this is actually what we want...
Thougths?

pescobar:
@boegel I think you are right and this is now what we want.

I have been looking at how other package managers handle this and I have seen that in ubuntu you can find two packages: `libjpeg-turbo8` and `libjpeg62`. Those two packages can be installed together. These are the provided libraries:

http://packages.ubuntu.com/xenial/amd64/libjpeg62/filelist
```
/usr/lib/x86_64-linux-gnu/libjpeg.so.62
/usr/lib/x86_64-linux-gnu/libjpeg.so.62.0.0
```

http://packages.ubuntu.com/xenial/amd64/libjpeg-turbo8/filelist
```
/usr/lib/x86_64-linux-gnu/libjpeg.so.8
/usr/lib/x86_64-linux-gnu/libjpeg.so.8.0.2
```

The problem appears with the devel packages. Both `libjpeg-turbo8-dev` and `libjpeg62-dev` provide a static lib `/usr/lib/x86_64-linux-gnu/libjpeg.a` so both are incompatible. You cannot install both packages together.

http://packages.ubuntu.com/xenial/amd64/libjpeg-turbo8-dev/filelist
http://packages.ubuntu.com/xenial/amd64/libjpeg62-dev/filelist

to be honest I don't know what would be the right approach to handle this :?

migueldiascosta:
@pescobar, this PR is being closed for the following reason(s): no activity for > 1 year.

PR #3871, opened 2016-11-30
{lang}[dummy,intel/2016b] Mono 4.6.2.7 (REVIEW)

backelj:
- Added Mono 4.6.2.7 with both dummy and intel-2016b toolchains
- No longer using mono.py easyblock, only ConfigureMake instead

### DIFF
--- a/easybuild/easyconfigs/m/Mono/Mono-4.6.2.7.eb
+++ b/easybuild/easyconfigs/m/Mono/Mono-4.6.2.7.eb
@@ -1,0 +1,17 @@
+easyblock = 'ConfigureMake'
+
+name = 'Mono'
+version = '4.6.2.7'
+
+homepage = 'http://mono-framework.com' 
+description = """An open source, cross-platform, implementation of C# and the CLR that is
+ binary compatible with Microsoft.NET."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+source_urls = ['http://download.mono-project.com/sources/mono/']
+sources = [SOURCELOWER_TAR_BZ2]
+
+osdependencies = ['gettext']
+
+moduleclass = 'lang'
--- a/easybuild/easyconfigs/m/Mono/Mono-4.6.2.7-intel-2016b.eb
+++ b/easybuild/easyconfigs/m/Mono/Mono-4.6.2.7-intel-2016b.eb
@@ -1,0 +1,20 @@
+easyblock = 'ConfigureMake'
+
+name = 'Mono'
+version = '4.6.2.7'
+
+homepage = 'http://mono-framework.com' 
+description = """An open source, cross-platform, implementation of C# and the CLR that is
+ binary compatible with Microsoft.NET."""
+
+toolchain = {'name': 'intel', 'version': '2016b'}
+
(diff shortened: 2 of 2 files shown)


easybuild/easyconfigs/m/Mono/Mono-4.6.2.7.eb
```diff
@@ -0,0 +1,17 @@
...
+homepage = 'http://mono-framework.com' 
+description = """An open source, cross-platform, implementation of C# and the CLR that is
+ binary compatible with Microsoft.NET."""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+source_urls = ['http://download.mono-project.com/sources/mono/']
+sources = [SOURCELOWER_TAR_BZ2]
+
+osdependencies = ['gettext']
```
boegel: @backelj why not use a proper dependency provided via EasyBuild?

easybuild/easyconfigs/m/Mono/Mono-4.6.2.7.eb
```diff
@@ -0,0 +1,17 @@
+easyblock = 'ConfigureMake'
```
boegel: it's great that they cleaned things up so you can now install Mono using just `configure`-`make`-`make install`

but since we *have* an easyblock for Mono, shouldn't we update it so it falls back to just `ConfigureMake` if the Mono version is recent enough?

easybuild/easyconfigs/m/Mono/Mono-4.6.2.7.eb
backelj: Well, I added it as an osdependency because it worked with the version of gettext that was installed as a package from the os, and I didn't want to add yet another module to our already overloaded module list. But that said, feel free to add a proper EasyBuild dependency for the sake of reproducability...

backelj: About the dependencies: where do you draw the line? The website http://www.mono-project.com/docs/compiling-mono/linux/ talks mentions " sudo apt-get install git autoconf libtool automake build-essential mono-devel gettext", so this would indicate we also have to add Autotools (which has autoconf, libtool and automake) in as well, right?

easybuild/easyconfigs/m/Mono/Mono-4.6.2.7.eb
boegel: My point is that we should avoid fragmentation where the existing software-specific easyblock is used in some easyconfigs, but not in others.

This is sort of a corner case though, since we can get away cleanly with using `ConfigureMake` if the Mono version is recent enough...

The changes that could/should be made to the Mono easyblock are basically to fall back to `ConfigureMake` for sufficiently recent versions of Mono.

boegel:
Test report by @boegel
**SUCCESS**
node2017.delcatty.os - Linux SL 6.8, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/3f00208d479c7f816416284e14c927a4 for a full test report.

boegel:
tested on top of https://github.com/hpcugent/easybuild-easyblocks/pull/1069, good to go, thanks @backelj!

PR #3907, opened 2016-12-07
fix source_urls in pkg-config easyconfigs

boegel:
Apparently this got moved (the old URL auto-directs to the new).


jhein32:
I had builds fail for not being able to download the sources.  So the URL change seems not reliable.  That is how I noticed.

boegel:
Thanks @jhein32!

PR #3928, opened 2016-12-12
fix broken links to infernal

SethosII:
the URL of infernals homepage and the download link changed, therefore
fetching files fails

### DIFF
--- a/easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
+++ b/easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
@@ -25,6 +25,8 @@ toolchainopts = {'optarch': True, 'pic': True}
 sources = [SOURCELOWER_TAR_GZ]
 source_urls = ['http://eddylab.org/infernal/%s' % name.lower()]
 
+checksums = ['300c208dd5550b422dd02d93c131958a']
+
 sanity_check_paths = {
     'files': ['bin/cm%s' % x for x in ['align', 'build', 'calibrate', 'convert', 'emit',
                                        'fetch', 'press', 'scan', 'search', 'stat']],
--- a/easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
+++ b/easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
@@ -15,7 +15,7 @@ easyblock = 'ConfigureMake'
 name = 'Infernal'
 version = '1.1rc1'
 
-homepage = 'http://infernal.janelia.org/'
+homepage = 'http://eddylab.org/infernal/'
 description = """Infernal ('INFERence of RNA ALignment') is for searching DNA sequence databases
  for RNA structure and sequence similarities."""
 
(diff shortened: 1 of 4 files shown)


easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
```diff
@@ -15,15 +15,15 @@ easyblock = 'ConfigureMake'
...
+homepage = 'http://eddylab.org/infernal/'
 description = """Infernal ('INFERence of RNA ALignment') is for searching DNA sequence databases
  for RNA structure and sequence similarities."""
 
 toolchain = {'name': 'ictce', 'version': '5.3.0'}
 toolchainopts = {'optarch': True, 'pic': True}
 
 sources = [SOURCELOWER_TAR_GZ]
-source_urls = ['ftp://selab.janelia.org/pub/software/%s' % name.lower()]
+source_urls = ['http://eddylab.org/infernal/%s' % name.lower()]
```
boegel: @SethosII the tabrall for `1.1rc` is not available here anymore?

boegel:
@SethosII can you please also include checksums for the tarballs, so we can make sure the downloaded source tarballs are the same as they were before?

easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
SethosII: yes, the URL is completely down

SethosII:
Hey @boegel, here are the checksums of my download:

```
> md5sum sources/i/Infernal/infernal-1.1rc1.tar.gz 
300c208dd5550b422dd02d93c131958a  sources/i/Infernal/infernal-1.1rc1.tar.gz
> sha1sum sources/i/Infernal/infernal-1.1rc1.tar.gz 
89077437c53d36fac37ba1e677ce05c112cd7588  sources/i/Infernal/infernal-1.1rc1.tar.gz
```

Where can I check them against the old ones?

easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
boegel: I mean, the tarball for `1.1rc` is not available in the new location...

I'd suggest to just remove `source_urls` instead then, and replace it with a comment, like so:

```python
# used to be at ftp://selab.janelia.org/pub/software/infernal, but not anymore (nor somwhere else)

**Comment by boegel** [comment]
@SethosII Here's what I have:

```
$ md5sum EasyBuild/sources/i/Infernal/*
300c208dd5550b422dd02d93c131958a  EasyBuild/sources/i/Infernal/infernal-1.1rc1.tar.gz
04bc8c9693cffba371b8559f023b6548  EasyBuild/sources/i/Infernal/infernal-1.1.tar.gz
```

Please include checksums like this in the easyconfig files themselves (right below `sources`/`source_urls`):

```python
checksums = ['04bc8c9693cffba371b8559f023b6548']
```

easybuild/easyconfigs/i/Infernal/Infernal-1.1rc1-ictce-5.3.0.eb
SethosII: The new location contains also the tarball for `1.1rc`.

SethosII:
The checksums from the new location match the old ones. I added them to the easyconfigs.

SethosII:
Oh sorry, it's fixed now.

boegel:
lgtm, thanks @SethosII!

PR #3932, opened 2016-12-14
fix test that verifies dumped easyconfig, take into account that dumped dependencies may include hardcoded dependency

boegel:
This change is required because of making `robot` fall back to using subtoolchains even without `--minimal-toolchains` in EasyBuild v3.0.0, cfr. https://github.com/hpcugent/easybuild-framework/pull/1973 .

An example PR that is blocked because of this is https://github.com/hpcugent/easybuild-easyconfigs/pull/3929, where the `CMake` dependency is being resolved via the available easyconfig using the `GCCcore` subtoolchain, leading to this error (without having the patch in this PR applied):

```
FAIL: Test for parsing of easyconfig libharu-2.3.0-foss-2016a.eb
  File "<string>", line 1, in innertest
  File "/home/travis/build/hpcugent/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 300, in template_easyconfig_test
    self.assertEqual(orig_val, dumped_val)
AssertionError: [('CMake', '3.6.1')] != [('CMake', '3.6.1', '', ('GCCcore', '4.9.3'))]
```

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -45,6 +45,7 @@ import easybuild.tools.options as eboptions
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.framework.easyblock import EasyBlock
 from easybuild.framework.easyconfig.default import DEFAULT_CONFIG
+from easybuild.framework.easyconfig.format.format import DEPENDENCY_PARAMETERS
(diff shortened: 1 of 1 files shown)


boegel:
@ocaisa please review?

boegel:
Thanks for the review @ocaisa!

PR #4010, opened 2017-01-12
Perl

valtandor:
New EasyConfigs for Perl 5.22.0 with the foss-2015a toolchain, namely single-threaded and multithreaded versions (either can be removed if not desired).


valtandor:
Whoops - wrongly directed to master! Closing.

boegel:
@valtandor GitHub now actually supports changing the target branch after opening the PR...

PR #4012, opened 2017-01-13
adding channel to perl-app-cpanminus

jerowe:
I'm adding the conda-forge channel to the perl-app-cpanminus config.

### DIFF
--- a/easybuild/easyconfigs/p/perl-app-cpanminus/perl-app-cpanminus-1.7039.eb
+++ b/easybuild/easyconfigs/p/perl-app-cpanminus/perl-app-cpanminus-1.7039.eb
@@ -22,7 +22,7 @@ description = """ cpanm - get, unpack build and install modules from CPAN """
 toolchain = {'name': 'dummy', 'version': ''}
 
 requirements = "%(name)s=%(version)s"
-channels = ['bioconda']
+channels = ['bioconda', 'conda-forge']
 
 builddependencies = [('Anaconda2', '4.0.0')]
 


boegel:
@jerowe the `perl-app-cpanminus` easyconfig still works fine, so why is this required?

Is there a similar fix for the `gencore_variant_detection` easyconfig, which is broken (cfr. https://github.com/hpcugent/easybuild-easyblocks/pull/1082#issuecomment-272259037).

jerowe:
If it works now it won't in the next day or so. Bioconda used to be a
standalone channel.  Now it requires some packages from conda-forge.

I think I fixed gencore_variant_detevtion,  could you try again?

Also,  I was not able to reproduce my error with a clean setup.  I am start
to think I broke something in my other environment,  but it's on my work
computer and I'm at home ;)

On Jan 13, 2017 1:06 PM, "Kenneth Hoste" <notifications@github.com> wrote:

@jerowe <https://github.com/jerowe> the perl-app-cpanminus easyconfig still
works fine, so why is this required?

Is there a similar fix for the gencore_variant_detection easyconfig, which
is broken (cfr. hpcugent/easybuild-easyblocks#1082 (comment)
<https://github.com/hpcugent/easybuild-easyblocks/pull/1082#issuecomment-272259037>
).

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/hpcugent/easybuild-easyconfigs/pull/4012#issuecomment-272394518>,
or mute the thread

boegel:
@jerowe how did you fix `gencore_variant_detection`, I don't see any changes to the easyconfig?

jerowe:
I changed the environment itself to include the conda_forge channel

boegel:
@jerowe please change the target branch to `develop`?

boegel:
Test report by @boegel
**SUCCESS**
node2623.swalot.os - Linux centos linux 7.2.1511, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/a79b1ccc3d219e24a5551fa4202cd3f6 for a full test report.

boegel:
lgtm

boegel:
Going in, thanks @jerowe!

boegel:
@jerowe installing `gencore_variant_detection` is still troublesome, error is now different though:

```
PaddingError: Placeholder of length '80' too short in package r::r-base-3.3.1-5.                                                                                                                               The package must be rebuilt with conda-build > 2.0.
```

jerowe:
@boegel , looking into this now.

jerowe:
@boegel , I pinned a few more versions. That might help. It builds on our docker image at least.

I think I've asked this before, but how could I go about testing things locally? A docker image that simulates travis would be really handy to have. I could do some testing locally to hopefully leave you with less testing. ;-)

boegel:
@jerowe still the same problem I'm afraid... I'm just using the existing easyconfig for `gencore_variant_detection` 1.0

jerowe:
@boegel , what is the prefix you are installing to? I know at some point I was having trouble with some modules, because the --prefix command supplied to conda was too long. I think this might have been the same error you are seeing.

boegel:
@jerowe Hah, seems like you're spot on. My default install prefix is rather long, and with a really short install prefix (`/tmp/$USER`), it installs just fine...

So, we should have a check in `conda.py` on the length of the install prefix or something like that?

jerowe:
That would be best, but I think the exact allowed prefix name depends upon the version of conda-build that was used to build the package. Packages are often build and rebuild, and there is no way of telling the version of conda-build used to build a package without going and hunting through the actual builds on travis.

Maybe we could catch any 'prefix blahblah' errors, and give a helpful warning that the prefix is probably too long?

boegel:
Yes, the latter would certainly be useful, since it wasn't clear quickly what the problem was even for us...

PR #4057, opened 2017-01-21
{lib}TensorFlow 0.12.1 with Foss 2016a and Python 2.7.11

RvDijk:
Work in progress: 
Tensorflow made pip installation available. Still need to resolve for the .whl extension in the python dependencies.

### DIFF
--- a/easybuild/easyconfigs/t/TensorFlow/TensorFlow-0.12.1-foss-2016a-Python-2.7.11.eb
+++ b/easybuild/easyconfigs/t/TensorFlow/TensorFlow-0.12.1-foss-2016a-Python-2.7.11.eb
@@ -1,4 +1,4 @@
-easyblock = 'Bundle'
+easyblock = 'PythonPackage'
 
 name = 'TensorFlow'
 version = '0.12.1'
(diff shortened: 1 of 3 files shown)


RvDijk:
@boegel there is no .tar.gz or .zip for tensorflow on pypi, so in this case I need to use "pip install" on the .whl extension. 
I tried to use .whl for each dependency in the exts_list to be consistent, however I can't find out how to disable the extracting step for each packages...

boegel:
@RvDijk have you tried using `use_pip = True` in the Tensorflow easyconfig?

boegel:
@Rv Ah, wait, you're doing that already...

easybuild/easyconfigs/t/TensorFlow/TensorFlow-0.12.1-foss-2016a-Python-2.7.11.eb
```diff
@@ -0,0 +1,66 @@
...
+engineers working on the Google Brain Team within Google's Machine Intelligence research 
+organization for the purposes of conducting machine learning and deep neural networks research,
+but the system is general enough to be applicable in a wide variety of other domains as well."""
+
+toolchain = {'name': 'foss', 'version': '2016a'}
+
+# this is a bundle of Python packages
+exts_defaultclass = 'PythonPackage'
+use_pip = True
+unpack_sources = False
```
boegel: @RvDijk the problem is that this is not being picked up for extensions, which is a bug we'll need to fix, probably in the `PythonPackage` easyblock...

boegel: @RvDijk please check whether this fixes the problem: https://github.com/hpcugent/easybuild-easyblocks/pull/1089

RvDijk:
@boegel I'm trying to verify with:
"eb TensorFlow-0.12.1... --include-easyblocks='$HOME/pythonpackage.py' "
However it does not seem to work, could it be the case that the "--include-easyblocks=XXX" is not used when a Bundle easyblock is used and XXX is the extension of that Bundle?

boegel:
@RvDijk you need to put a generic easyblock like `PythonPackage` in a subfolder named `generic`, e.g. `--include-easyblocks generic/pythonpackage.py`

This is an issue we need to resolve in framework at some point, see https://github.com/hpcugent/easybuild-framework/issues/1964

easybuild/easyconfigs/t/TensorFlow/TensorFlow-0.12.1-foss-2016a-Python-2.7.11.eb
boegel: @RvDijk Any luck?

RvDijk:
@boegel providing a subfolder seems to work for selecting the appropriate easyblock, however this still results in a error:

`== 2017-01-27 22:15:55,731 build_log.py:147 ERROR EasyBuild crashed with an error (at easybuild/software/EasyBuild/3.0.1/lib/python2.6/site-packages/vsc_base-2.5.5-py2.6.egg/vsc/utils/exceptions.py:124 in __init__): Unknown file type for file numpy-1.12.0-cp27-cp27mu-manylinux1_x86_64.whl (at easybuild/software/EasyBuild/3.0.1/lib/python2.6/site-packages/easybuild_framework-3.0.1-py2.6.egg/easybuild/tools/filetools.py:638 in extract_cmd)
`

boegel:
@RvDijk I just look into this myself... The problem is two-fold:

* the easyblock being used when parsing the easyconfig file is `Bundle`, which doesn't know about the `unpack_sources` or `use_pip` easyconfig parameters since those are specific to `PythonPackage`...
* not all easyconfig parameters can be specified on a per-extension basis (see also https://github.com/hpcugent/easybuild-framework/issues/991), so there's currently no way to specify `use_pip` for just `Tensorflow` itself

This is an issue we'll have to solve soon, we've hit this before and have always danced around it...

Note that in this case, we can probable still dance around it.

`numpy` is a part of the `Python` module already, and if you specifically want to use the latest and greatest `numpy`, you can include it as a separate dependency;
both the `mock` and `funcsigs` are also included in the `Python` installation already

So, rather than making this a `Bundle`, we can get away with just installing `Tensorflow` directly via the `PythonPackage` easyblock instead, where we can use `use_pip` without problems?

Of course, that doesn't mean the problem that occurred here shouldn't be resolved, it should.

boegel:
@RvDijk see #4084

RvDijk:
This PR has become outdated, closing...

PR #4071, opened 2017-01-26
{lang}[foss/2016b] Python 3.6.0 (w/ lots of extensions)

fizwit:
Python 3.6.0  This is a fat Python with lots of packages.

### DIFF
--- a/easybuild/easyconfigs/p/Python/Python-3.6.0-foss-2016b.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.6.0-foss-2016b.eb
@@ -1,0 +1,1026 @@
+# easybuild easycofig
+#
+# John Dey jfdey@fredhutch.org
+#
+# Fred Hutchinson Cancer Research Center
+
+name = 'Python'
+version = '3.6.0'
+
+homepage = 'http://python.org/'
+description = """Python is a programming language that lets you work more quickly and integrate your systems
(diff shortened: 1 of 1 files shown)


boegel:
@fizwit I changed the target branch to `develop`, otherwise this is impossible to review/merge :)

boegel:
@fizwit This is a lot fatter than what we usually go with in the central repo...
We usually try to limit the dependencies to those that are required for Python itself (not for particular extensions).

There are several extensions that are commented out, is there a good reason for that?

This looks more like a site-specific version of a Python easyconfig to me, since including this would be a switch in 'policy' for central Python easyconfigs...

Thoughts?

boegel:
@fizwit Also, do you mind changing the PR title a bit to better reflect the contents of the PR?

fizwit:
Sure Like Python3.6.0

fizwit:
What about Python-3.6.0-foss-2016b-bio for a new naming concept?  We have tried to incorporate all the module from anaconda into our local build.

boegel:
@fizwit Using a `-bio` versionsuffix may make sense, yes.
But it may also complicate things, e.g. if you start including this Python variant as a dependency in other easyconfigs...

In any case, the currently commented our extensions (including useful ones like `scikit-*`) should be resolved (i.e. either removed or decommented).

PR #4076, opened 2017-01-30
Easyconfig file for TASSEL 5.2.33

FokkeDijkstra:
Easyconfig file for TASSEL 5.2.33


easybuild/easyconfigs/t/TASSEL/TASSEL-5.2.33-Java-1.8.0_92.eb
```diff
@@ -0,0 +1,27 @@
...
+commit = 'dbad59cff68dd5753ff8bebd3546df694b5da3a4'
+versionsuffix = '-Java-%(javaver)s'
+
+homepage = 'http://www.maizegenetics.net/tassel'
+description = """TASSEL provides tools to investigate relationships between phenotypes and genotypes"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+source_urls = ['https://bitbucket.org/tasseladmin/tassel-5-standalone/get']
+sources = ['%s.zip' % commit]
```
boegel: @FokkeDijkstra We have a better way of dealing with this now, allowing to use a datestamp as version and renaming the downloading sources, see easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#alternative-formats-for-sources

FokkeDijkstra: I've read the instructions, but have no clue what to do. I also found out that Tassel currently can be downloaded using the version number. So I've used that instead.

PR #4117, opened 2017-02-09
disable_quad_on_power

d-rdgz:
disables attempts to build libquadmath versions of FFTW on Power


boegel:
see https://github.com/hpcugent/easybuild-easyblocks/pull/1102

PR #4142, opened 2017-02-15
platanus-1.2.4-foss-2017a

ysagon:


easybuild/easyconfigs/p/Platanus/Platanus-1.2.4-foss-2017a.eb
```diff
@@ -0,0 +1,25 @@
...
+name = 'Platanus'
+version = '1.2.4' 
+
+homepage = 'http://platanus.bio.titech.ac.jp/'
+description = """PLATform for Assembling NUcleotide Sequences"""
+
+toolchain = {'name': 'foss', 'version': '2017a'}
+toolchainopts = {'openmp': True}
+
+source_urls = ['http://platanus.bio.titech.ac.jp/?ddownload=150']
```
boegel: @ysagon does auto-downloading work with this for you?

easybuild/easyconfigs/p/Platanus/Platanus-1.2.4-foss-2017a.eb
```diff
@@ -0,0 +1,25 @@
...
+sources = ['%(name)s_v%(version)s.tar.gz']
+
+# fix Makefile to support newer gcc version
+patches = ['Platanus_new_gcc.patch']
```
boegel: @ysagon rather than using a patch (which may break with future versions of Platanus), how about something like

```python
buildopts = 'CXXFLAGS="$CXXFLAGS -fpermissive"`
```

This has as extra bonus that EasyBuild is in control of the compiler options.

You probably also need to include `'cstd': 'c++0x'` in `toolchainopts`, and you may also need to explicitly include some of the hardcoded flags like `-Dnullptr=0` and `-lm`

easybuild/easyconfigs/p/Platanus/Platanus-1.2.4-foss-2017a.eb
ysagon: @boegel  yes it does. But this is just by chance, as their webserver don't care of the end of the url:)
It could be a good idea to have a way to specify the sources, and an optional destination filename, because if I put ?download=150 in the sources, eb doesn't recognize the file type.

easybuild/easyconfigs/p/Platanus/Platanus-1.2.4-foss-2017a.eb
ysagon: Ok, seems better. A question anyway, in the Makefile there are other flags like:

    -funroll-loops -Wall -finline-limit-50000

Are they managed by eb in any way?

boegel: no, although I think `-funroll-loops` would be enabled if you use `-O3` (`'opt': True` in `toolchainopts`)

If you think the others should be included too, you'll need to add them in the `buildopts` line (there may be a good reason for it)

boegel:
Test report by @boegel
**SUCCESS**
node2605.swalot.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/7bf839e4b2ea73d13dcedd201cb649fc for a full test report.

ysagon:
Hi, are you waiting for an action on my side?

boegel:
@ysagon Not really, this one slipped through somehow, lgtm

boegel:
Test report by @boegel
**SUCCESS**
gligar02.gligar.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/ccf29f0af3f3925d90e1e75f9a4adf1a for a full test report.

boegel:
Going in, thanks @ysagon!

PR #4152, opened 2017-02-16
sprint_for_easyconfigs_using_2017a_common_toolchains

JackPerdue:
(created using `eb --new-pr`)
a different take on 2017a with minimal-toolchains and lots of examples


JackPerdue:
Again, I don't expect this to be accepted as is but hopefully gives some idea of where I think we could go.

JackPerdue:
Things I would do with more iterations:

1) consistly reorder depencies by osdependiencies/builddepbendencies/depenendcies
2) get rid of  my #EOF at the end of everything (sorry, habit)
3) verify that the needed "easyblock" is explicitly listed (my preference)
4) any pep8 cleanup
5) more cross-platform testing (RH6 vs RH7, x86_64 vs ppc64)
6) (I'll think of #6 after some more beer.... "you are #6" [bonus points if you know where that is from]).

JackPerdue:
BTW... I still think the GCCcore module for 2017a should have been named GCCcore/6.3.0-2.26 to provide for flexibility for other binutils.....

JackPerdue:
Oh.... and %(pyver)s and %(pyshortver)s don't work if Python is only a builddependency..... not sure if that is an ommission or by design... thus you'll see it being defined a lot in the above

boegel:
@JackPerdue still relevant?

JackPerdue:
#1 and #3 are still relevant but I can fight that battle in future versions. :)  Closing.

PR #4189, opened 2017-02-23
Update OpenMPI-1.10.2-GCC-4.9.3-2.25.eb

serge-p:
patches = ['pax_disable.patch']

PR #4194, opened 2017-02-23
Update OpenMPI-1.10.3-GCC-5.4.0-2.26.eb

serge-p:
pax patch for non-interactive, automated builds


boegel:
@serge-p Rather than spreading this change over multiple PRs (cfr. #4189), we should make the change once consistently across all OpenMPI easyconfigs.

Also, rather than using a patch file (which is quite prone to breaking), maybe we should do a 'runtime' patch of the configure script instead, for example via something like:

```python
# instruct pax not to prompt for a new volume, since that requires interaction
preconfigopts = "sed -i'' -e 's/pax -r/pax -r -O/g' configure && "
```

The benefit being that this should always work, while a patch file is prone to being broken for a future OpenMPI release...

boegel:
@serge-p Thoughts on the above?

migueldiascosta:
@serge-p, this PR is being closed for the following reason(s): no activity for > 6 months.

PR #4195, opened 2017-02-23
Easyconfig and patch file to install BLAST+ 2.6.0, tested at UAntwerp…

klust:
…en on hopper with 2016b toolchain

I adapted a easyconfig to install BLAST+ 2.6.0 and made a new patch file. The patches should also work for other toolchains: they correct an error in the root configure script (the developers accidentally forgot to remove the clean configure.orig script back to configure) and a problem with the install-toolkit target in the Makefile that caused an error message.

There are plenty of comments in the EasyConfig so that other users can easily adapt it to their own needs.

### DIFF
--- a/easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
+++ b/easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
@@ -1,0 +1,85 @@
+##
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+#
+# Copyright:: Copyright 2012-2014 Uni.Lu/LCSB, NTUA
+# Authors::   Fotis Georgatos <fotis@cern.ch>, Kenneth Hoste (UGent)
+# License::   MIT/GPL
+# $Id$
+#
+# This work implements a part of the HPCBIOS project and is a component of the policy:
+# http://hpcbios.readthedocs.org/en/latest/HPCBIOS_2012-94.html
+##
+
+easyblock = 'ConfigureMake'
+
+name = 'BLAST+'
+version = '2.6.0'
+versionsuffix = '-Python-2.7.12'
+
+homepage = 'http://blast.ncbi.nlm.nih.gov/'
+description = """Basic Local Alignment Search Tool, or BLAST, is an algorithm
+ for comparing primary biological sequence information, such as the amino-acid
+ sequences of different proteins or the nucleotides of DNA sequences."""
+
+toolchain = {'name': 'intel', 'version': '2016b'}
+
+sources = ['ncbi-blast-%(version)s+-src.tar.gz']
+source_urls = ['http://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/%(version)s/']
+
+# The patch contains two fixes:
+# - The authors forgot to reinstall the correct root configure script (which is 
+#   available as configure.orig).
+# - A correction to Makefile.in.top, install-toolkit target: The last install
+#   created an error message, so it has been replaced by two commands that 
+#   hopefully perform the action that the authors intended.
+patches = ['%(name)s-%(version)s_general-fixes.patch']
+
+# Loading this version of Boost will also automatically load Python, and the configure
+# of BLAST+ will discover it and enable Python support.
+#dependencies = [ 
+#	('Boost', '1.63.0', versionsuffix),
+#	('GMP', '6.1.1'),
+#	('libpng', '1.6.26'),
+#	('libjpeg-turbo', '1.5.0')
+#]
+dependencies = []
+
+# Select library type
+configopts  = '--with-static '   # Build failed when --with-dll was added.
+# Executables to generate
+configopts += '--with-static-exe --with-bin-release --without-debug --with-strip '
+# Use a couple of build-in libraries rather than their independent
+# counte3rpart, even if they are installed in seperate modules, to reduce dependencies.
+configopts += '--without-z --without-bz2 --without-pcre '
+# Enable support for SRA/VDB. Don't use a version that is already on the system,
+# but just get it from GitHub and compile into BLAST+. This again reduces dependencies.
+configopts += '--with-downloaded-vdb '
+
+# Uncomment the following lines to enable support for a Python module.
+# Make sure it is compatible with the Python used by Boost should you enable
+# support for a Boost module with Python support
+dependencies.append( ('Python', '2.7.12') )
+configopts += '--with-python=$EBROOTPYTHON '
+
+# Uncomment the following lines to enable Boost
+dependencies.append( ('Boost', '1.63.0', '-Python-2.7.12') )
+configopts += '--with-boost=$EBROOTBOOST '
+
+# Uncomment to enable the use of GMP
+dependencies.append( ('GMP', '6.1.1') )
+configopts += '--with-gmp=$EBROOTGMP '
+
+# Uncomment to enable PNG-support
+dependencies.append( ('libpng', '1.6.26') )
+configopts += '--with-png=$EBROOTLIBPNG '
+
+# Uncomment to enable JPEG-support
+dependencies.append( ('libjpeg-turbo', '1.5.0') )
+configopts += '--with-jpeg=$EBROOTLIBJPEGMINTURBO '
+
+sanity_check_paths = {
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+
+# Loading this version of Boost will also automatically load Python, and the configure
+# of BLAST+ will discover it and enable Python support.
+#dependencies = [ 
+#	('Boost', '1.63.0', versionsuffix),
+#	('GMP', '6.1.1'),
+#	('libpng', '1.6.26'),
+#	('libjpeg-turbo', '1.5.0')
+#]
+dependencies = []
```
boegel: @klust this doesn't look correct? Especially since you're using `$EBROOTPYTHON`, `$EBROOTBOOST` below?

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+
+sources = ['ncbi-blast-%(version)s+-src.tar.gz']
+source_urls = ['http://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/%(version)s/']
+
+# The patch contains two fixes:
+# - The authors forgot to reinstall the correct root configure script (which is 
+#   available as configure.orig).
+# - A correction to Makefile.in.top, install-toolkit target: The last install
+#   created an error message, so it has been replaced by two commands that 
+#   hopefully perform the action that the authors intended.
```
boegel: @klust We generally include a description of the patch in the patch itself, i.e. at the top of the patch file (any text above the first line that starts with `---` is comments).

Also, please mention the author of the patch, and mention references to external sources that were used to put it together, if any.

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+#	('libpng', '1.6.26'),
+#	('libjpeg-turbo', '1.5.0')
+#]
+dependencies = []
+
+# Select library type
+configopts  = '--with-static '   # Build failed when --with-dll was added.
+# Executables to generate
+configopts += '--with-static-exe --with-bin-release --without-debug --with-strip '
+# Use a couple of build-in libraries rather than their independent
```
boegel: `built-in`?

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+# counte3rpart, even if they are installed in seperate modules, to reduce dependencies.
```
boegel: `counterpart`?

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+configopts += '--without-z --without-bz2 --without-pcre '
```
boegel: What's the exact motivation for this?

The list of dependencies should be as complete as possible here. For sites that want to filter out stuff that is commonly provided via the OS, the support for `--filter-deps` is there (cfr. http://easybuild.readthedocs.io/en/latest/Manipulating_dependencies.html#filtering-out-dependencies-using-filter-deps).

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+# Enable support for SRA/VDB. Don't use a version that is already on the system,
+# but just get it from GitHub and compile into BLAST+. This again reduces dependencies.
+configopts += '--with-downloaded-vdb '
```
boegel: same here, strong preference for including `VDB` as a proper dependency

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+
+# Uncomment the following lines to enable support for a Python module.
```
boegel: they are uncommented already?

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,85 @@
...
+# Make sure it is compatible with the Python used by Boost should you enable
+# support for a Boost module with Python support
+dependencies.append( ('Python', '2.7.12') )
```
boegel: ah, that's why the list of `dependencies` is empty above...

This goes against the common style of keeping the list of `dependencies` nicely together, no other easyconfig that I know of uses `.append(...)`...

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
klust: The same comments were included in the patch also; I've omitted them from the easyconfig. I've also added me as an author to the patch. The patch is newly developed without using external sources; I wanted to make the changes as minimal as possible so didn't start from, e.g., previous patch files.

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
klust: Corrected the spelling mistake, and the one in the next mistake also.

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
klust: By doing it this way, I am sure that the versions that are compiled in are also tested by the authors of BLAST+ and known to be compatible. 

Moreover, the extensive loading from modules that is occurring when building software the traditional EasyBuild way is becoming a pain for users who are building workflows. In fact, the recompile started because a user needed software that we have installed in different toolchains that conflict. Though things seemed to work properly when unloading modules from older toolchains before loading packages that were compiled with a newer one, it was a pain due to the high number of modules involved.

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
klust: I've made the sentence more clear.

easybuild/easyconfigs/b/BLAST+/BLAST+-2.6.0-intel-2016b-Python-2.7.12.eb
klust: Keeping all dependencies together is a very good coding style if software really can't be compiled without, i.e., if these dependencies are non-optional. Packages as BLAST+ or, e.g., Petsc have so many options and can be configured in so many different ways that it is a pain in those cases to always see witch dependencies belong to which configopt switches and hence which dependencies can be safely omitted when a certain option is turned off.

Though I have to admit I've done it this way to also provoke some discussion about precisely this issue.

boegel:
An easyconfig for BLAST+ 2.6.0 with `intel/2017a` and Python 2.7.13 was merged as a part of  #4624.

The patch included here was helpful in fixing an installation issue, thanks for sharing it @klust!

I consider this PR obsolete now, so I'll go ahead and close it, also because it doesn't follow the usual 'coding' style we maintain in easyconfig files...

PR #4199, opened 2017-02-24
Switch to lowopt=True for libxc v2.2.* and v3.*

tovrstra:
I've noticed that some functionals in LibXC give significantly different results with too much compiler optimization. For example, the energy of a water molecule computed with the TPSS functional changes by 6.5 kJ/mol with lowopt=False. (The problem is in the TPSS correlation functional, somewhere in mgga_c_pkzb.c. I've ruled out numerical integration issues. Similar relative errors appear on every grid point of the tpss_c energy density.)

I can only test with libx-2.2.* and intel compilers at the moment. I'll make similar changes for other versions and compilers when the opportunity arises.

Can someone check if the combination `{'opt': False, 'lowopt': True}` is meaningful for `toolchainopts`? It seems to work for me.

### DIFF
--- a/easybuild/easyconfigs/l/libxc/libxc-2.2.0-intel-2014b.eb
+++ b/easybuild/easyconfigs/l/libxc/libxc-2.2.0-intel-2014b.eb
@@ -9,7 +9,7 @@ description = """Libxc is a library of exchange-correlation functionals for dens
 
 toolchain = {'name': 'intel', 'version': '2014b'}
 # Results for some functions (e.g. mgga_c_tpss) change with too aggressive optimization settings.
-toolchainopts = {'opt': False, 'lowopt': True}
+toolchainopts = {'lowopt': True}
 
 sources = [SOURCE_TAR_GZ]
 source_urls = ['http://www.tddft.org/programs/octopus/down.php?file=libxc/']
--- a/easybuild/easyconfigs/l/libxc/libxc-2.2.0-intel-2014b.eb
+++ b/easybuild/easyconfigs/l/libxc/libxc-2.2.0-intel-2014b.eb
@@ -8,6 +8,7 @@ description = """Libxc is a library of exchange-correlation functionals for dens
  The aim is to provide a portable, well tested and reliable set of exchange and correlation functionals."""
 
 toolchain = {'name': 'intel', 'version': '2014b'}
+# Results for some functions (e.g. mgga_c_tpss) change with too aggressive optimization settings.
 toolchainopts = {'opt': False, 'lowopt': True}
 
 sources = [SOURCE_TAR_GZ]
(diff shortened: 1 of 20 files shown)


boegel:
@tovrstra You say that the results are different, which is a clear indication of a problem, but do you know which of the results is *correct*?

In any case, if we make this change, it should probably be made consistent across all libxc 2.2.x easyconfigs, and maybe also for libxc 3.x?

easybuild/easyconfigs/l/libxc/libxc-2.2.0-intel-2014b.eb
```diff
@@ -8,7 +8,8 @@ description = """Libxc is a library of exchange-correlation functionals for dens
  The aim is to provide a portable, well tested and reliable set of exchange and correlation functionals."""
 
 toolchain = {'name': 'intel', 'version': '2014b'}
-toolchainopts = {'opt': True}
+# Results for some functions (e.g. mgga_c_tpss) change with too aggressive optimization settings.
+toolchainopts = {'opt': False, 'lowopt': True}
```
boegel: @tovrstra settign `opt` to `False` shouldn't be needed, it's `False` by default, see http://easybuild.readthedocs.io/en/latest/version-specific/toolchain_opts.html#intel

enabling `opt` would imply `-O3`, the default is to use `-O2`, while `lowopt` is equivalent with `-O1` (see also `eb -x libxc.eb | grep CFLAGS`

tovrstra: ok. is fixed.

tovrstra:
I'm reasonably sure that the with lowopt=True are correct. I noticed the following two problems with LibXC when trying to install a beta release of HORTON on delcatty:

1) This release of HORTON contains regression tests that compare results obtained previously (GNU compiler, -O1, stored in a script) with results after a new build. When HORTON is linked to LibXC with lowopt=True, the regression tests pass. If not, they fail. (When they pass, there are still minimal differences but these are several orders of magnitude smaller.)

2) Without lowopt=True, some consistency checks for the derivatives of the TPSS correlation functional begin to fail in the HORTON unit tests. (There may be other functionals with problems. We don't test them all in HORTON.)

The TPSS code has nearly not changed in the 3.0.0. (Only the reference to the paper is done differently.)  So, it seems right to make the same changes for libxc-3.0.0. I'm not sure about the other compilers. We always had consistent results with the GNU compilers, irrespective of optimization settings.

boegel:
@tovrstra Please also make the same change in the `libxc` easyconfigs for `foss` and `CrayGNU`; it may not be strictly needed there since those compilations are done with GCC, but let's keep things consistent to avoid re-introducing this problem?

tovrstra:
Good point. Done.

boegel:
Test report by @boegel
**SUCCESS**
node2097.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/94b9d06dd94497e62047826e22445598 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 11 out of 11 (11 easyconfigs in this PR)
node2304.phanpy.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/b5a7dcfec1aa269d924e4c4bb8edd6ca for a full test report.

boegel:
lgtm, tested with easyconfigs that still work on CentOS7 (others should be fine too)

boegel:
Going in, thanks @tovrstra!

PR #4204, opened 2017-02-24
{bio}[intel/2016b] ea-utils v1.04.807

boegel:
(created using `eb --new-pr`)

PR #3834, opened 2016-11-20
{bio}[foss/2016b] ea-utils v27a4809

verdurin:
(created using `eb --new-pr`)

PR #3634, opened 2016-09-29
{bio}[foss/2016a] ea-utils

RvDijk:

PR #4205, opened 2017-02-25
fix version (& homepage) in ea-utils easyconfigs

boegel:
when looking into #4204, I noticed that `ea-utils` now available with a proper version on GitHub, so we don't need to use the commit as a version anymore...

Originally contributed by @RvDijk (#3634), and updatd by @verdurin (#3834), so feedback from the both of you would be great since this is a 'breaking' change.

As far as I can tell `ea-utils` is not used as a dependency somewhere else, so changing the version like we do here is acceptable imho (since it won't break other easyconfigs).

### DIFF
--- a/easybuild/easyconfigs/e/ea-utils/ea-utils-1.04.807-foss-2016a.eb
+++ b/easybuild/easyconfigs/e/ea-utils/ea-utils-1.04.807-foss-2016a.eb
@@ -1,9 +1,9 @@
 easyblock = 'MakeCp'
 
 name = 'ea-utils'
-version = '27a4809'
+version = '1.04.807'
 
-homepage = 'https://code.google.com/p/ea-utils/'
+homepage = 'http://expressionanalysis.github.io/ea-utils/'
 description = """Command-line tools for processing biological sequencing data.
 Barcode demultiplexing, adapter trimming, etc.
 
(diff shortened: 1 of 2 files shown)


RvDijk:
@boegel looks good to me.

boegel:
Test report by @boegel
**SUCCESS**
node2304.phanpy.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/23d11508263d06ea044734f798af8e8e for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2097.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/fe9061f8d301b8af9e9fef65052929d3 for a full test report.

boegel:
@verdurin we usually don't bump dependencies in existing easyconfigs... Also, this is just a build dependency?

verdurin:
It is, yes.

That being the case, looks fine to me.

boegel:
Thanks for the feedback @RvDijk and @verdurin!

PR #4127, opened 2017-02-13
{lib}[foss/2016b]  MXNet 0.9.3 core library with Python-2.7.12 (REVIEW)

hajgato:

PR #4207, opened 2017-02-27
{lib}[intel/2016b] MXNet 0.9.3 core library with Python-2.7.12 (WIP)

hajgato:


hajgato:
compilation ends without problems, however when I try to import the `Python` module I get the following error:
```
>>> import mxnet as mx 
[14:25:33] /tmp/hajgato/build/MXNet/0.9.3/intel-2016b/mxnet/dmlc-core/include/dmlc/logging.h:300: [14:25:33] include/mxnet/operator.h:514: Check failed: this->name == type (BatchNorm vs. Activation) Register Name and TypeString mismatch, name="BatchNorm", but TypeString="Activation"
```

I am afraid to solve this problem is beyond my capabilities...

boegel:
@hajgato That's a pretty weird problem... This doesn't occur with `foss/2016b`? (#4127)?

hajgato:
@boegel with `GCC` is fine. It might be `c++std` difference. 
I have found something on `MXNet` github page, but this error is rather long and it is not corrected.
see:
https://github.com/dmlc/mxnet/issues/413
and around there

hajgato:
tried with `--std=c++14` and `--stc=gnu++11`, same result

hajgato:
better version:
https://github.com/hpcugent/easybuild-easyconfigs/pull/4346

PR #4226, opened 2017-02-28
fix check for Szip library in configure script for netCDF 4.1.3

boegel:
I have a use case for an old netCDF version, so I figured out why it didn't build anymore.

Not sure how this ever worked without the patch, maybe because an old Szip library that provides `-lszip` was installed system-wide?


boegel:
Test report by @boegel
**FAILED**
Build succeeded for 1 out of 5 (2 easyconfigs in this PR)
node2116.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/6397683dee6557fca320b8225c6a7c7c for a full test report.

PR #4234, opened 2017-03-02
{lang}[iomkl/2017A] R v3.3.2

JackPerdue:
(created using `eb --new-pr`)
Newest R/iomkl at TAMU HPRC


JackPerdue:
primarily submitted for review by https://github.com/hpcugent/easybuild-easyconfigs/issues/2742

JackPerdue:
See also https://github.com/hpcugent/easybuild-easyconfigs/issues/4235

boegel:
@JackPerdue PRs of this size are unlikely to get merged (quickly), since they require quite a bit of effort to review & test... If you intend to get these easyconfigs merged in, we should break this up into smaller PRs, starting with the toolchain & deps + HPL as a test case.

Also, the versioning scheme you use for `iomkl` is a concern, since it's too close to the common toolchain versioning scheme, see http://easybuild.readthedocs.io/en/latest/Common-toolchains.html#versioning-scheme-for-common-toolchains.
`2017A` and `2017a` are just too close to each other... 

We just merged an `iomkl/2017a` a couple of days ago, which is basically `intel/2017a` where Intel MPI is swapped out in favor of OpenMPI 2.0.2.

Having both `iomkl/2017a` and `iomkl/2017A` is going to be too confusing...

PR #4238, opened 2017-03-02
QUAST 4.4 foss-2016a

ysagon:


SethosII:
@ysagon I also try to create a config for `QUAST`. Why do you have `Boost` as a dependency? I couldn't find something in their documentation that it's required. One thing that's required is `Perl` but this is not in your config? And what is the line `options = {'modulename': 'site'}` for?

ysagon:
I don't remember.

If you have a look here http://quast.bioinf.spbau.ru/manual.html#sec1 it'w written that E-MEM quast submodule needs Boost. 

It's working for me. If it's not working for you, feel free to put another version or correct this one. 

If I remember correctly, the  options = {'modulename': 'site'} line si to prevent easybuild to try to load the quast python module.

easybuild/easyconfigs/q/QUAST/QUAST-4.4-foss-2016a-Python-2.7.11.eb
```diff
@@ -0,0 +1,31 @@
...
+source_urls = ['https://github.com/ablab/%(namelower)s/archive/']
+sources = ['%(namelower)s_%(version)s.tar.gz']
+
+dependencies = [
+    ('Python', '2.7.11'),
+    ('Boost', '1.61.0'),
+    ('matplotlib', '1.5.1', '-Python-%(pyver)s')
+]
+
+options = {'modulename': 'site'}
```
boegel: @ysagon Better clarify this with a comment, it's a bit of a hack since `import site` will always work (since `site` is a part of the Python standard library).

We should have a cleaner way of bypassing the `import` check...

migueldiascosta:
@ysagon, this PR is being closed for the following reason(s): no activity for > 6 months, obsoleted by more recent PRs.

easybuild/easyconfigs/q/QUAST/QUAST-4.4-foss-2016a-Python-2.7.11.eb
boegel: Just in case someone picks up on this: you can easily skip the default `import` check using `False` here:

```python
options = {'modulename': False}

PR #4225, opened 2017-02-28
{lang}[foss/2016b] Python v2.7.13, Tkinter v2.7.13

verdurin:
(created using `eb --new-pr`)

PR #4090, opened 2017-02-01
{perf}[dummy/dummy] ipp v2017.1.132

shahzebsiddiqui:
(created using `eb --new-pr`)

PR #4260, opened 2017-03-06
{lang}[intel/2017a] Python v2.7.13 (REVIEW)

boegel:
This Python easyconfig deliberately does not include `Tk` as a dependency (and hence doesn't support 'import Tkinter'), following the example in #4225 by @verdurin, to avoid requiring `X11` as an indirect dependency for `Python`.

I intend to also open a separate PR for a cleaned up version of the `Tkinter` easyconfig in #4225.

TODO:

* [ ] check whether the problem reported in #4049 still occurs (@vanzod: could you check that?)
* ~~[ ] strip out `Tkinter.py` to make it crystal clear that this Python installation doesn't provide it?~~
* ~~[ ] include a sanity check command to ensure that `Tkinter` support isn't added accidentally, e.g. by picking up `Tk` from the OS?~~

In any case, `_tkinter.so` is not a part of the Python installation I obtained using these easyconfigs.

edit: update on TODO: enhanced `Python` easyblock that makes sure that `Tkinter` is there when `Tk` is included as a dependency is done in https://github.com/hpcugent/easybuild-easyblocks/pull/1156; I'm deliberately not enforcing a lack of `Tkinter` when `Tk` is not listed as a dependency, since `Tk` may get picked up from the OS (although unlikely I think) when `--filter-deps=Tk` is used.

#4090 is mostly unrelated to this, so shouldn't be a blocker for this PR.

### DIFF
--- a/easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
+++ b/easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
@@ -15,8 +15,6 @@ toolchainopts = {'pic': True}
 sources = ['readline-%(version)s.tar.gz']
 source_urls = ['http://ftp.gnu.org/gnu/readline']
 
-patches = ['libreadline-%(version)s-bugfix.patch']
-
 builddependencies = [('binutils', '2.27')]
 dependencies = [('ncurses', '6.0')]
 
--- a/easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
+++ b/easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
@@ -1,0 +1,33 @@
+easyblock = 'ConfigureMake'
+
+name = 'libreadline'
+version = '7.0'
+
+homepage = 'http://cnswww.cns.cwru.edu/php/chet/readline/rltop.html'
+description = """The GNU Readline library provides a set of functions for use by applications that
+ allow users to edit command lines as they are typed in. Both Emacs and vi editing modes are available.
+ The Readline library includes additional functions to maintain a list of previously-entered command lines,
+ to recall and perhaps reedit those lines, and perform csh-like history expansion on previous commands."""
+
+toolchain = {'name': 'GCCcore', 'version': '6.3.0'}
+toolchainopts = {'pic': True}
+
+sources = ['readline-%(version)s.tar.gz']
+source_urls = ['http://ftp.gnu.org/gnu/readline']
+
+patches = ['libreadline-%(version)s-bugfix.patch']
+
+builddependencies = [('binutils', '2.27')]
+dependencies = [('ncurses', '6.0')]
+
+# for the termcap symbols, use EB ncurses
+preconfigopts = "env LDFLAGS='-lncurses'"
+
+sanity_check_paths = {
+    'files': ['lib/libreadline.a', 'lib/libhistory.a'] +
+             ['include/readline/%s' % x for x in ['chardefs.h', 'history.h', 'keymaps.h', 'readline.h', 'rlconf.h',
+                                                  'rlstdc.h', 'rltypedefs.h', 'tilde.h']],
+    'dirs': [],
+}
+
+moduleclass = 'lib'
--- a/easybuild/easyconfigs/p/Python/Python-2.7.13-intel-2017a.eb
+++ b/easybuild/easyconfigs/p/Python/Python-2.7.13-intel-2017a.eb
@@ -28,7 +28,7 @@ dependencies = [
 osdependencies = [('openssl-devel', 'libssl-dev', 'libopenssl-devel')]
 
 # order is important!
-# package versions updated March 6th 2017
+# package versions updated March 30th 2017
 exts_list = [
     # note: more recent versions of setuptools (v34.x) can not be installed from source anymore,
     # see https://github.com/pypa/setuptools/issues/980
@@ -42,13 +42,14 @@ exts_list = [
     ('nose', '1.3.7', {
         'source_urls': ['https://pypi.python.org/packages/source/n/nose/'],
     }),
-    ('numpy', '1.12.0', {
+    ('numpy', '1.12.1', {
         'source_urls': ['https://pypi.python.org/packages/source/n/numpy'],
         'source_tmpl': '%(name)s-%(version)s.zip',
         'patches': ['numpy-1.12.0-mkl.patch'],
     }),
-    ('scipy', '0.18.1', {
+    ('scipy', '0.19.0', {
         'source_urls': ['https://pypi.python.org/packages/source/s/scipy'],
+        'source_tmpl': '%(name)s-%(version)s.zip',
     }),
     ('blist', '1.3.6', {
         'source_urls': ['https://pypi.python.org/packages/source/b/blist/'],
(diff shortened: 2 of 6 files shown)


vanzod:
@boegel The problem in #4049 affects only Python 3 because of the presence of `XZ` in the dependencies and the resulting dependency loop with . Nothing to do with X11. Unless I missed something else...

boegel:
@vanzod I wasn't suggesting #4049 has something to do with X11, I was just bringing it up here since this PR includes a new easyconfig for `libreadline`.

You originally opened the issue for Python 2.7.12, see the description, so now I'm confused... :)

easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
```diff
@@ -0,0 +1,33 @@
...
+ The Readline library includes additional functions to maintain a list of previously-entered command lines,
+ to recall and perhaps reedit those lines, and perform csh-like history expansion on previous commands."""
+
+toolchain = {'name': 'GCCcore', 'version': '6.3.0'}
+toolchainopts = {'pic': True}
+
+sources = ['readline-%(version)s.tar.gz']
+source_urls = ['http://ftp.gnu.org/gnu/readline']
+
+patches = ['libreadline-%(version)s-bugfix.patch']
```
wpoely86: drop? Pretty sure it's upstream.

easybuild/easyconfigs/p/Python/Python-2.7.13-intel-2017a.eb
```diff
@@ -0,0 +1,133 @@
...
+source_urls = ['http://www.python.org/ftp/%(namelower)s/%(version)s/']
+sources = [SOURCE_TGZ]
+
+# python needs bzip2 to build the bz2 package
+dependencies = [
+    ('bzip2', '1.0.6'),
+    ('zlib', '1.2.11'),
+    ('libreadline', '7.0'),
+    ('ncurses', '6.0'),
+    ('SQLite', '3.17.0'),
```
wpoely86: you forgot this one.

verdurin:
@boegel agree with all the TODO items.

vanzod:
@boegel I am testing it for the `ncurses` linking issue but I am having issue with the `SQLite-3.17.0-intel-2017a` dependency. Since I could not find the easyconfig anywhere, I made one but I am hitting the following error:

```
sqlite3-sqlite3.o: In function `sqlite3VdbeExec':
sqlite3.c:(.text+0x2e9c4): undefined reference to `__builtin_add_overflow'
sqlite3.c:(.text+0x32364): undefined reference to `__builtin_mul_overflow'
sqlite3.c:(.text+0x323de): undefined reference to `__builtin_sub_overflow'
sqlite3.c:(.text+0x35661): undefined reference to `__builtin_add_overflow'
sqlite3-sqlite3.o: In function `sumStep':
sqlite3.c:(.text+0x8854e): undefined reference to `__builtin_add_overflow'
make: *** [sqlite3] Error 1
```

Could you add your `SQLite` easyconfig to this PR?

easybuild/easyconfigs/p/Python/Python-2.7.13-intel-2017a.eb
```diff
@@ -0,0 +1,133 @@
...
+    }),
+    ('nose', '1.3.7', {
+        'source_urls': ['https://pypi.python.org/packages/source/n/nose/'],
+    }),
+    ('numpy', '1.12.0', {
+        'source_urls': ['https://pypi.python.org/packages/source/n/numpy'],
+        'source_tmpl': '%(name)s-%(version)s.zip',
+        'patches': ['numpy-1.12.0-mkl.patch'],
+    }),
+    ('scipy', '0.18.1', {
```
boegel: should bump to 0.19.0...

easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
boegel: if it would be, patching would fail...

easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
```diff
@@ -0,0 +1,33 @@
...
+
+builddependencies = [('binutils', '2.27')]
+dependencies = [('ncurses', '6.0')]
+
+# for the termcap symbols, use EB ncurses
+preconfigopts = "env LDFLAGS='-lncurses'"
```
wpoely86: I'm a bit lost what we decided with ncurses but `-ltinfo` should suffice here.

easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
wpoely86: https://git.savannah.gnu.org/cgit/readline.git/commit/?h=devel&id=8ef852a5be72c75e17f2510bea52455f809b56ce

It's upstream, part of 7.0. Can be dropped.

easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
boegel: only if the `ncurses` we're using here provides a separate libtinfo, which it doesn't (only the one built with `dummy` does)

We discussed this during yesterday's conf call, @vanzod is going to come back to #4049, but it should only relate to the `dummy` ncurses (which was being picked up even though the `GCCcore` one should be used, for some reason)

wpoely86: why not for all `ncurses`? Many (including libreadline) only need `libtinfo`.

boegel: Well, because it's causing problems, see #4049 :)

The problem is mostly with `libncurses` having the tinfo symbols stripped out though, I think.

easybuild/easyconfigs/p/Python/Python-2.7.13-intel-2017a.eb
boegel: there now

easybuild/easyconfigs/l/libreadline/libreadline-7.0-GCCcore-6.3.0.eb
wpoely86: That's for 2016b. I think if we start cleanly with a new toolchain, most explicit linking to `libncurses` can be dropped and replaced to `libtinfo`.

boegel: The easyconfig for ncurses 6.0 with GCCcore, which is what we're using for the 2017a toolchains, is already there, so we can't easily split up libncurses and libtinfo there anymore?

boegel:
Test report by @boegel
**FAILED**
node2052.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/b63310440a6ae837d78d74a768ffc3b9 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2052.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/cac79e84a1336e2638e32b6a14cb2746 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 7 out of 7 (5 easyconfigs in this PR)
node2424.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/87001b2e80ca53fcd5c33f3ee85df3bc for a full test report.

boegel:
good to go, thanks for the feedback @wpoely86!

PR #4289, opened 2017-03-09
{bio}[foss/2016b] PileOMeth v0.1.11

robqiao:
New easyconfig file for PileOMeth added

### DIFF
--- a/easybuild/easyconfigs/h/HTSlib/HTSlib-1.2.1-foss-2016b.eb
+++ b/easybuild/easyconfigs/h/HTSlib/HTSlib-1.2.1-foss-2016b.eb
@@ -32,7 +32,7 @@ dependencies = [
 ]
 
 sanity_check_paths = {
-    'files': ["bin/bgzip", "bin/tabix", "lib/libhts.so"],
+    'files': ["bin/bgzip", "bin/tabix", "lib/libhts.%s" % SHLIB_EXT],
     'dirs': [],
 }
 
--- a/easybuild/easyconfigs/h/HTSlib/HTSlib-1.2.1-foss-2016b.eb
+++ b/easybuild/easyconfigs/h/HTSlib/HTSlib-1.2.1-foss-2016b.eb
@@ -1,0 +1,39 @@
+##
+# This is a contribution from Phoenix HPC Service, The University of Adelaide, Australia
+# Homepage: https://www.adelaide.edu.au/phoenix/
+#
+# Copyright::   adelaide.edu.au/phoenix
+# Authors::	Robert Qiao <robert.qiao@adelaide.edu.au>, Exe Escobedo <exequiel.sepulvedaescobedo@adelaide.edu.au>
+# License::
+#
+# Acknowledgement::  	
+# 		Original Author: Pablo Escobar Lopez
+# 		Swiss Institute of Bioinformatics
+# 		Biozentrum - University of Basel
+##
+
+easyblock = 'ConfigureMake'
+
+name = 'HTSlib'
+version = '1.2.1'
+
+homepage = "http://www.htslib.org/"
+description = """ A C library for reading/writing high-throughput sequencing data.
+ This package includes the utilities bgzip and tabix 
+"""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['https://github.com/samtools/%(namelower)s/releases/download/%(version)s/']
+sources = [SOURCELOWER_TAR_BZ2]
+
+dependencies = [
+    ('zlib', '1.2.8'),
+]
+
+sanity_check_paths = {
+    'files': ["bin/bgzip", "bin/tabix", "lib/libhts.so"],
+    'dirs': [],
+}
+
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/p/PileOMeth/PileOMeth-0.1.11.eb
```diff
@@ -0,0 +1,40 @@
...
+PileOMeth extracts per-base methylation metrics from them. 
+PileOMeth requires an indexed fasta file containing the reference genome as well.
+"""
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = ['%(version)s.zip']
+source_urls = ['https://github.com/dpryan79/PileOMeth/archive/']
+
+dependencies = [
+    ('HTSlib', '1.2.1', '', ('foss', '2017a'))
```
boegel: @robqiao You should include an easyconfig for this dependency in this pull request (see failing tests in Travis), and use `foss/2017a` as `toolchain` rather than specifying the toolchain for the dependency, i.e.:

```python
toolchain = {'name': 'foss', 'version': '2017a'}
...
dependencies = [
    ('HTSlib', '1.2.1'),
]
```

easybuild/easyconfigs/p/PileOMeth/PileOMeth-0.1.11.eb
```diff
@@ -0,0 +1,40 @@
...
+]
+
+files_to_copy = [
+    (["PileOMeth"], "bin"), "*.c", "*.h"
```
boegel: @robqiao Any particular reason you are also copying the `*.c` and `*.h` files?

easybuild/easyconfigs/p/PileOMeth/PileOMeth-0.1.11.eb
robqiao: @boegel thanks. Is that a convention to use a tool chain option if any dependencies use one, even when parent itself not use any?

easybuild/easyconfigs/p/PileOMeth/PileOMeth-0.1.11.eb
robqiao: @boegel some of our users use .h to build their own functions using existing functions, but I guess I should remove these for this purpose. :)

boegel: Well, than you should only include the `*.h` files, and copy them to the `include` subdirectory (so that the location gets added to `$CPATH`)?

easybuild/easyconfigs/p/PileOMeth/PileOMeth-0.1.11.eb
boegel: Since PileOMeth is also being compiled, it's a lot better to specify the `toolchain` to something that is not `dummy`, since then EasyBuild will also set up the build environment (stuff like `$CC`, `$CFLAGS`, etc.)

The dependencies by default will inherit the parent toolchain. You can hard specify to use a particular toolchain on a per-dependency basis, but then these should be a subtoolchain of the parent toolchain. In practice, we only do this when a particular dependency was installed with a `dummy` toolchain (e.g. a binary tool that doesn't require compiling).

See also http://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#dependencies

easybuild/easyconfigs/h/HTSlib/HTSlib-1.2.1-foss-2016b.eb
```diff
@@ -0,0 +1,39 @@
...
+
+source_urls = ['https://github.com/samtools/%(namelower)s/releases/download/%(version)s/']
+sources = [SOURCELOWER_TAR_BZ2]
+
+dependencies = [
+    ('zlib', '1.2.8'),
+]
+
+sanity_check_paths = {
+    'files': ["bin/bgzip", "bin/tabix", "lib/libhts.so"],
```
boegel: please use `'lib/libhts.%s' % SHLIB_EXT` rather than hardcoding the `.so`

robqiao: Thanks for the tip.

boegel:
Test report by @boegel
**SUCCESS**
node2054.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/90cb683d6a9daddcefdfde14873099d6 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2569.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/29c30e1e419749de34ff17833a24799c for a full test report.

boegel:
lgtm

boegel:
Going in, thanks @robqiao!

PR #4239, opened 2017-03-02
fix missing PYTHONPATH in ROOT v6.08.02 easyconfig

ysagon:
as ROOT compiles a ROOT.py module as well and put it in the lib folder, we need to set the PYTHONPATH as well.

PR #4320, opened 2017-03-10
correctly set $PYTHONPATH for ROOT v5.30.06

ysagon:
fix PYTHONPATH #4239


boegel:
@ysagon Because this (and the other PRs for the ROOT easyconfigs) include the commits from #4239, I can't test this via `--from-pr` anymore (which is maybe a bug in `--from-pr`, but fine).

Do you mind updating this (and the other ROOT easyconfig PRs) with current `develop`?

boegel:
@ysagon Nevermind, I've opened a wrapper PR that basically includes all your ROOT PRs on top of current `develop`, see #4331

I'll work on getting that one merged, which should also automatically mark all your ROOT PRs as merged, that's going to be less painful. :)

PR #4330, opened 2017-03-10
detect use of '$root', which is not compatible with module files in Lua syntax

boegel:


boegel:
@pescobar, @wpoely86 please review?

boegel:

PR #4357, opened 2017-03-16
{bio}[foss/2016b] Circos 0.69-4

SethosII:
Circos and GD with toolchain foss-2016b because `--try-toolchain=foss-2016b` on existing EasyConfigs fails.

### DIFF
--- a/easybuild/easyconfigs/c/Circos/Circos-0.69-4-foss-2016b-Perl-5.24.0.eb
+++ b/easybuild/easyconfigs/c/Circos/Circos-0.69-4-foss-2016b-Perl-5.24.0.eb
@@ -1,0 +1,37 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+
+easyblock = 'Tarball'
+
+name = 'Circos'
+version = '0.69-4'
+
+homepage = 'http://www.circos.ca/'
+description = """Circos is a software package for visualizing data and information.
+ It visualizes data in a circular layout - this makes Circos ideal for exploring
+ relationships between objects or positions."""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['http://circos.ca/distribution/']
+sources = [SOURCELOWER_TGZ]
+
+checksums = ['d84a23f03ebcba08671e0108182737b7']
+
+perl = 'Perl'
+perlver = '5.24.0'
+versionsuffix = '-Perl-%(perlver)s'
+dependencies = [
+    (perl, perlver),
+    ('GD', '2.52', versionsuffix),
+]
+
+sanity_check_paths = {
+    'files': ['bin/%(namelower)s'],
+    'dirs': ['lib/%(name)s'],
+}
+
+modextrapaths = {'PERL5LIB': 'lib'}
+
+sanity_check_commands = [('perl', '-e "use Circos"')]
+
+moduleclass = 'bio'
--- a/easybuild/easyconfigs/c/Circos/Circos-0.69-4-foss-2016b-Perl-5.24.0.eb
+++ b/easybuild/easyconfigs/c/Circos/Circos-0.69-4-foss-2016b-Perl-5.24.0.eb
@@ -4,6 +4,7 @@ easyblock = 'Tarball'
 
 name = 'Circos'
 version = '0.69-4'
+versionsuffix = '-Perl-%(perlver)s'
 
 homepage = 'http://www.circos.ca/'
 description = """Circos is a software package for visualizing data and information.
@@ -17,11 +18,8 @@ sources = [SOURCELOWER_TGZ]
 
 checksums = ['d84a23f03ebcba08671e0108182737b7']
 
-perl = 'Perl'
-perlver = '5.24.0'
-versionsuffix = '-Perl-%(perlver)s'
 dependencies = [
-    (perl, perlver),
+    ('Perl', '5.24.0'),
     ('GD', '2.52', versionsuffix),
 ]
 
--- a/easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
+++ b/easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
@@ -1,0 +1,29 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+
+easyblock = 'PerlModule'
+
+name = 'GD'
+version = '2.52'
+
+homepage = 'http://search.cpan.org/~lds/GD/'
+description = """GD.pm - Interface to Gd Graphics Library"""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['http://cpan.metacpan.org/authors/id/L/LD/LDS/']
+sources = [SOURCE_TAR_GZ]
+
+checksums = ['646a4b5ba587763ed791bad4f2f79e95']
+
+perl = 'Perl'
+perlver = '5.24.0'
+versionsuffix = '-Perl-%(perlver)s'
+
+dependencies = [
+    (perl, perlver),
+    ('libgd', '2.2.3'),
+    ('libpng', '1.6.26'),
+    ('libjpeg-turbo', '1.5.0'),
+]
+
+moduleclass = 'bio'
--- a/easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
+++ b/easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
@@ -4,8 +4,9 @@ easyblock = 'PerlModule'
 
 name = 'GD'
 version = '2.52'
+versionsuffix = '-Perl-%(perlver)s'
 
-homepage = 'http://search.cpan.org/~lds/GD/'
+homepage = 'http://search.cpan.org/~lds/%(name)s/'
 description = """GD.pm - Interface to Gd Graphics Library"""
 
 toolchain = {'name': 'foss', 'version': '2016b'}
@@ -15,15 +16,16 @@ sources = [SOURCE_TAR_GZ]
 
 checksums = ['646a4b5ba587763ed791bad4f2f79e95']
 
-perl = 'Perl'
-perlver = '5.24.0'
-versionsuffix = '-Perl-%(perlver)s'
-
 dependencies = [
-    (perl, perlver),
+    ('Perl', '5.24.0'),
     ('libgd', '2.2.3'),
     ('libpng', '1.6.26'),
     ('libjpeg-turbo', '1.5.0'),
 ]
+
+sanity_check_paths = {
+    'files': ['bin/bdf2gdfont.pl', 'lib/perl5/site_perl/%(perlver)s/x86_64-linux-thread-multi/%(name)s.pm'],
+    'dirs': ['lib/perl5/site_perl/%(perlver)s/x86_64-linux-thread-multi/%(name)s'],
+}
 
 moduleclass = 'bio'


easybuild/easyconfigs/c/Circos/Circos-0.69-4-foss-2016b-Perl-5.24.0.eb
```diff
@@ -0,0 +1,37 @@
...
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['http://circos.ca/distribution/']
+sources = [SOURCELOWER_TGZ]
+
+checksums = ['d84a23f03ebcba08671e0108182737b7']
+
+perl = 'Perl'
+perlver = '5.24.0'
+versionsuffix = '-Perl-%(perlver)s'
```
boegel: @SethosII please move this up right below `version`

easybuild/easyconfigs/c/Circos/Circos-0.69-4-foss-2016b-Perl-5.24.0.eb
```diff
@@ -0,0 +1,37 @@
...
+dependencies = [
+    (perl, perlver),
```
boegel: no need to use local variables, only used in one place now, so make this

```python
    ('Perl', '5.24.0'),
```

easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
```diff
@@ -0,0 +1,29 @@
...
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['http://cpan.metacpan.org/authors/id/L/LD/LDS/']
+sources = [SOURCE_TAR_GZ]
+
+checksums = ['646a4b5ba587763ed791bad4f2f79e95']
+
+perl = 'Perl'
+perlver = '5.24.0'
+versionsuffix = '-Perl-%(perlver)s'
```
boegel: please move up below `version`

easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
```diff
@@ -0,0 +1,29 @@
...
+
+dependencies = [
+    (perl, perlver),
```
boegel: no need for variables

easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
```diff
@@ -0,0 +1,29 @@
...
+    ('libgd', '2.2.3'),
+    ('libpng', '1.6.26'),
+    ('libjpeg-turbo', '1.5.0'),
+]
+
```
boegel: no custom `sanity_check_paths`?

SethosII:
@boegel Thanks for the notes. I added the requested changes.

easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
```diff
@@ -0,0 +1,31 @@
...
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['http://cpan.metacpan.org/authors/id/L/LD/LDS/']
+sources = [SOURCE_TAR_GZ]
+
+checksums = ['646a4b5ba587763ed791bad4f2f79e95']
+
+dependencies = [
+    ('Perl', '5.24.0'),
+    ('libgd', '2.2.3'),
```
boegel: the existing easyconfig for this depends on libpng 1.6.24, so including a dependency on libpng 1.6.26 below isn't going to work...

There's a libgd 2.2.4 release, maybe include an updated easyconfig for libgd, and use the latest libpng 1.6.27 both in libgd and here?

easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
```diff
@@ -0,0 +1,31 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+
+easyblock = 'PerlModule'
+
+name = 'GD'
+version = '2.52'
```
boegel: bump to latest 2.56?

easybuild/easyconfigs/g/GD/GD-2.52-foss-2016b-Perl-5.24.0.eb
SethosII: I get an error during the sanity check of libgd 2.2.4 (the log file contains no further information.):

```
== FAILED: Installation ended unsuccessfully (build directory: /dev/shm/libgd/2.2.4/foss-2016b): build failed (first 300 chars): Sanity check failed: no file of ('lib/libgd.%(SHLIB_EXT)s',) in /easybuild/test/software/libgd/2.2.4-foss-2016b
```

But the file is actually there:

```
$ ls /easybuild/test/software/libgd/2.2.4-foss-2016b/lib/
libgd.a  libgd.la  libgd.so  libgd.so.3  libgd.so.3.0.4  pkgconfig
```

Shouldn't `SHLIB_EXT` resolve to `so`? The sanity check is this:

```
sanity_check_paths = {
    'files': ['lib/libgd.a', 'lib/libgd.%(SHLIB_EXT)s'],
    'dirs': ['bin', 'include'],
}
```

boegel: @SethosII `SHLIB_EXT` is a constant, not a template (confusing, I know), so use:

```python
'lib/libgd.%s' % SHLIB_EXT
```

SethosII:
I did some testing with different dependencies:

Circos|GD  |libgd|libpng|works?
------|----|-----|------|-----
0.69-4|2.56|2.2.4|1.6.27|fails during install of GD: `Can't create bdf2gdfont.pl: Permission denied at bdf_scripts/bdf2gdfont.PLS line 11.`
0.69-4|2.52|2.2.4|1.6.27|fails during install of GD: `cmd "make test" exited with exitcode 2 and output: "/easybuild/16.04/software/Perl/5.24.0-foss-2016b/bin/perl" -MExtUtils::Command::MM -e 'cp_nonempty' -- GD.bs blib/arch/auto/GD/GD.bs 644`
0.69-4|2.56|2.2.3|1.6.27|fails during install of GD: `Can't create bdf2gdfont.pl: Permission denied at bdf_scripts/bdf2gdfont.PLS line 11.`
0.69-4|2.52|2.2.3|1.6.27|works (libgd-2.2.3 is a custom one with libpng-1.6.27 as dependency, the [current easyconfig](https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/l/libgd/libgd-2.2.3-foss-2016b.eb) uses libpng-1.6.24)

`bdf_scripts/bdf2gdfont.PLS` until line 11:

```
#!perl
use Config;
use File::Basename qw(&basename &dirname);
use Cwd;

$origdir = cwd;
chdir dirname($0);
$file = basename($0, '.PL','.PLS');
$file .= $^O eq 'VMS' ? '.com' : '.pl';

open OUT,">$file" or die "Can't create $file: $!";
```

So I looked into the packages for GD 2.52 and 2.56 and 2.56 contains already an bdf2gdfont.pl (starts with GD-2.55) with only read and execute rights set (555). I tried with deleting this file and adding write rights but now it fails during the installation at another point:

Circos|GD  |libgd|libpng|works?
------|----|-----|------|-----
0.69-4|2.56-patched|2.2.4|1.6.27|fails during install of GD: `cmd "make test" exited with exitcode 2 and output: PERL_DL_NONLAZY=1 "/mnt/fhgfs_ribdata/easybuild/16.04/software/Perl/5.24.0-foss-2016b/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t`
0.69-4|2.56-patched|2.2.3|1.6.27|fails during install of GD: `cmd "make test" exited with exitcode 2 and output: PERL_DL_NONLAZY=1 "/mnt/fhgfs_ribdata/easybuild/16.04/software/Perl/5.24.0-foss-2016b/bin/perl" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t`

So I don't really know what to try next? Maybe stick with the working version? Then the only problem is what to do with the libpng dependency: the current `libgd-2.2.3` uses `libpng-1.6.24` so also stick with this?

boegel:
@SethosII Thanks for the extensive report!

Do you have more details on the failing test for the patched GD 2.56, i.e. what is the exact problem there?
The `bdf2gdfont.pl` they include may be patched, so removing that is probably not a good idea. Or maybe it's a packaging problem, since it's GD itself that is trying to install a new `bdf2gdfont.pl`, right?

I'm fine with taking the easy way out and sticking to the old libpng, I unfortunately don't have time to dive into this one myself...

SethosII:
I also just realised that Circos fails with the older dependencies during execution. It's missing the modules `SVG` and `Statistics::Basic` (Perl version 5.24.0). These modules are also not included in Perl version 5.18.2 which is used by the current Circos EasyConfigs with the `ictce` toolchain. I don't know if they are working at all (I have no intel compiler to test)? The sanity check `sanity_check_commands = [('perl', '-e "use Circos"')]` doesn't fail on this because Circos still returns 0 although it detects missing modules.

Now back to the build errors. Here are the build logs for the patched version: [removed bdf2gdfont.pl](https://github.com/hpcugent/easybuild-easyconfigs/files/857489/easybuild-GD-2.56-removed-bdf2gdfont-20170321.092842.TNrYI.txt) and [writeable bdf2gdfont.pl](https://github.com/hpcugent/easybuild-easyconfigs/files/857488/easybuild-GD-2.56-writeable-bdf2gdfont-20170321.092842.TNrYI.txt). Both fail the tests with messages like this: `Failed test 'unable to generate comparison image for test 6: Can't locate object method "_new" via package "GD::Image" at /dev/shm/GD/2.56/foss-2016b-Perl-5.24.0/GD-2.56/t/../blib/lib/GD/Image.pm line 83.`. Here is the line with the context:

```
sub _make_filehandle {
  shift;  # get rid of class // first mention of shift, don't know what this comment should tell me?
  no strict 'refs';
  my $thing = shift;
  return $thing if defined(fileno $thing);

  # otherwise try qualifying it into caller's package
  my $fh;
  {   
    local $^W = 0;  # to avoid uninitialized variable warning from Symbol.pm
    $fh = qualify_to_ref($thing,caller(2));
  }
  return $fh if defined(fileno $fh);

  # otherwise treat it as a file to open
  $fh = gensym;
  open($fh,$thing) || return;
  return $fh;
}

sub new {
  my $pack = shift;
  if (@_ == 1) {
    if (my $type = _image_type($_[0])) {
      my $method = "newFrom${type}Data";
      return unless $pack->can($method);
      return $pack->$method($_[0]);
    }
    return unless my $fh = $pack->_make_filehandle($_[0]);
    my $magic;
    return unless read($fh,$magic,4);
    return unless my $type = _image_type($magic);
    seek($fh,0,0);
    my $method = "newFrom${type}";
    return $pack->$method($fh);
  }
  return $pack->_new(@_); // line 83
}
```

`_new` is indeed not in the file but I don't know much about Perl syntax for what the underscore means. So I can't really tell why the tests are failing.

boegel:
Unfortunately my Perl-fu is not sufficient to figure this one out either...Maybe @rjeschmi can help?

@SethosII W.r.t. the Circos requiring a sufficiently recent Perl version: is there an easy way to check this, i.e. a light-weight command that we could use as a sanity check command?

rjeschmi:
sure @boegel let me pull it and see.

SethosII:
@boegel Maybe i misunderstood you or I expressed my self weirdly: Circos doesn't require version 5.24.0 of Perl, I just used this version. But two of the needed modules (`SVG` and `Statistics::Basic`) aren't included in any Perl version from EasyBuild.

For the testing part: `circos -modules` reports the status of required modules but the exitcode is still 0 if stuff is missing, so it's not really helpful? An alternative which does real computation is the script located at `examples/run`. It just contains one command so one could also run this directly:

```
#!/bin/bash

echo "Creating image circos.png and writing report to run.out."

echo "Example takes ~45-60 seconds to generate."

../bin/circos -conf etc/circos.conf -debug_group summary,timer > run.out
```

The distribution already contains example output `circos.png` and `run.out` of this command.

I won't be able to respond in the next few days, just to let you know.

boegel:
@SethosII `circos -modules` may be sufficient, we can combine it with `grep` (or `grep -v`) to get a non-zero exit code when there's a problem

boegel:
@SethosII The problem seems to be a known issue, see http://www.perlmonks.org/?node_id=1133976 .

Are you also seeing the problem with `libgd.so.*`?

SethosII:
@boegel For testing: `circos -modules | grep "missing"` would be an option. There shouldn't be a line with `missing` in it (same is true for the currently used `perl -e "use Circos"`).

For GD: I don't have the exact same error but maybe something related. They have this:

```
#   Failed test 'use GD;'
#   at t/GD.t line 14.
#     Tried to use 'GD'.
#     Error:  Can't load  '/nv/vol80/hlilab/shailesh/benchmarking/modules/GD-2.56/t/../blib/arch/auto/GD/GD.so' for module GD: libgd.so.3: cannot open shared object file: No such file or directory at /usr/lib64/perl5/DynaLoader.pm line 200.
#  at (eval 17) line 2
# Compilation failed in require at (eval 17) line 2.
# BEGIN failed--compilation aborted at (eval 17) line 2.
```

and I got this:

```
#   Failed test 'use GD;'
#   at t/GD.t line 14.
#     Tried to use 'GD'.
#     Error:  Can't locate loadable object for module GD in @INC (@INC contains: /dev/shm/GD/2.56/foss-2016b-Perl-5.24.0/GD-2.56/t/../blib/lib /dev/shm/GD/2.56/foss-2016b-Perl-5.24.0/GD-2.56/t/../blib/arch /dev/shm/GD/2.56/foss-2016b-Perl-5.24.0/GD-2.56/t/../lib /dev/shm/GD/2.56/foss-2016b-Perl-5.24.0/GD-2.56/blib/lib /dev/shm/GD/2.56/foss-2016b-Perl-5.24.0/GD-2.56/blib/arch /easybuild/16.04/software/Perl/5.24.0-foss-2016b/lib/perl5/site_perl/5.24.0/x86_64-linux-thread-multi /easybuild/16.04/software/Perl/5.24.0-foss-2016b/lib/perl5/site_perl/5.24.0 /easybuild/16.04/software/Perl/5.24.0-foss-2016b/lib/perl5/5.24.0/x86_64-linux-thread-multi /easybuild/16.04/software/Perl/5.24.0-foss-2016b/lib/perl5/5.24.0 .) at t/GD.t line 14.
# Compilation failed in require at t/GD.t line 14.
# BEGIN failed--compilation aborted at t/GD.t line 14.
```

The paths indeed do not contain a `GD` file, just `GD.pm`. The line 14 is this: `use_ok('GD',':DEFAULT',':cmp');`.

boegel:
@SethosII `GD.pm` is the Perl module, so that looks right w.r.t. the `use`. Not being able to load the `.so` is probably also making the `use` fail.

@vanzod seems to be looking into `GD` too, cfr. #4394, you should join forces. ;-)

SethosII:
As of #4712 I can get a successfull build with the recent versions of Circos and GD/libgd, but it's not possible to run because the perl modules `Statistics::Basic` and `SVG` are missing. Can we just add them to the Perl module or is there a way to add the specific perl modules only for this module? And maybe we should close this one and I open another pull request?

boegel:
@SethosII Adding additional extensions in the Perl module is fine, as long as they don't imply additional dependences.
Alternatively, we can create either individual easyconfigs for these Perl modules, or a single separate easyconfig with a bundle of Perl modules.

SethosII:
@boegel Thanks, I will do this in a new pull request.

PR #4358, opened 2017-03-16
SAMtools 1.4 easyconfig + patch

klust:
2 additional dependencies (additional compression algorithms), but also needs a modified easyblock as errmod.h and kprobaln.h are no longer provided nor needed. The easyblock is provided in a pull request in that repository. As the Makefile has also changed from version 1.3.1, I also redid that patch.

edit: depends on https://github.com/hpcugent/easybuild-easyblocks/pull/1139


boegel:
@klust Did you check whether the patch is strictly required?

Since EasyBuild sets `$LIBRARY_PATH`, I would expect the injecting the `-L$EBROOTXXX/lib` is not required?

In fact, the one for zlib is wrong (missing `/lib`), which probably confirms my suspicion...

klust:
@boegel You are right, the patch is not needed. I took it as a given since there was a patch file for previous versions. I have recompiled 1.4 without the patch and checked the samtools executable with ldd: it loads the correct versions of the libraries. I've also removed the patch form the 1.3.1 easyconfig for intel/2016b with built-in htslib, and it turned out the patch was not needed either. So the patch may have been there for too long already. This also explains why another makefile that linked the other two compression libraries didn't need to be patched. I've included new versions of SAMtools-1.3.1-intel-2016b.eb and SAMtools-1.4-intel-2016b.eb and did a git rm on the patchfile for 1.4 which is clearly not needed.

boegel:
lgtm

boegel:
Test report by @boegel
**SUCCESS**
node2589.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/9d72c373ff1b9aea280a6e5090a2a0b0 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2128.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/e4e03de7d0159d0ba64f160d54e6ebba for a full test report.

boegel:
Going in, thanks @klust!

PR #4372, opened 2017-03-20
{bio}[intel/2016b] Bowtie 2.2.9

klust:
The Bowtie2 Makefile has the options for the GNU compiler hardcoded in the Makefile. Moreover, is forces sse2 rather than using -xHost to compile for the current architecture, and the Allinea blog (article link: https://www.allinea.com/blog/201501/tuning-bowtie2-better-performance) suggest that replacing this in the Makefile does produce slightly better code for newer Intel processors. The patch replaces the GNU compiler options with options that correspond to what EasyBuild uses. I have hardcoded them in the patch rather than using CXXFLAGS to have compatible debug options. 

I've also added LICENSE to the list of files to copy in the EascyConfig.


easybuild/easyconfigs/b/Bowtie2/Bowtie2-2.2.9_optimize-intel.patch
```diff
@@ -0,0 +1,22 @@
...
+   $(error bowtie2 compilation requires a 64-bit platform )
+ endif
+ 
+-SSE_FLAG=-msse2 
+-
+-DEBUG_FLAGS    = -O0 -g3 -m64 $(SSE_FLAG)
++DEBUG_FLAGS    = -std=c++11 -O0 -g3 -xHost -ftz -fp-speculation=safe -fp-model source
+ DEBUG_DEFS     = -DCOMPILER_OPTIONS="\"$(DEBUG_FLAGS) $(EXTRA_FLAGS)\""
+-RELEASE_FLAGS  = -O3 -m64 $(SSE_FLAG) -funroll-loops -g3
++RELEASE_FLAGS  = -std=c++11 -O2 -xHost -ftz -fp-speculation=safe -fp-model source
```
boegel: @klust You don't need a patch for this, it's easier/better to do this in the easyconfig file using:

```python
toolchainopts = {'cstd': 'c++11'}
...
buildopts = 'RELEASE_FLAGS="$CXXFLAGS"'
```

You could also redefine `DEBUG_FLAGS` the same way, but I doubt that is relevant since we're not doing a debug build (hopefully)?

klust:
Tested the procedure you suggested, submitted the modified easyconfig and removed the patch in the last commit. I thought the variable defined in the Makefile would overwrite the one in the environment, but this is indeed not the case.

boegel:
@klust arguments to the `make` command and environment variables are not the same thing

This:

```
CC=g++ make
```

is not the same as

```
make CC=g++
```

In the latter, you are telling make "use `g++` for the `CC` variable in the `Makefile`, regardless what the `Makefile` define it as".

easybuild/easyconfigs/b/Bowtie2/Bowtie2-2.2.9-intel-2016b.eb
```diff
@@ -0,0 +1,49 @@
...
+# Author: Pablo Escobar Lopez
+# Swiss Institute of Bioinformatics
+# Biozentrum - University of Basel
+# Modified by: Robert Schmidt <rjeschmi@gmail.com>
+# Ottawa Hospital Research Institute - Bioinformatics Team
+# Modified by: Adam Huffman <adam.huffman@crick.ac.uk>
+# The Francis Crick Institute
+# Modified by: Kurt Lust, UAntwerp: Patch for better optimization with the
+# Intel compilers, based on 
+# https://www.allinea.com/blog/201501/tuning-bowtie2-better-performance
```
boegel: no longer patched, and I'm not sure if advertising Allinea is required for something as simply as making sure the compilation flags provided by EasyBuild are being used?

boegel:
Test report by @boegel
**SUCCESS**
node2463.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/f7cda2d9c92aed30d86f6682029c6ea6 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2001.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/7c4c9911204f60ca57a57da8cbfed0df for a full test report.

klust:
OK, modified the comments; no further changes were made.

boegel:
lgtm

easybuild/easyconfigs/b/Bowtie2/Bowtie2-2.2.9-intel-2016b.eb
```diff
@@ -0,0 +1,48 @@
...
+toolchain = {'name': 'intel', 'version': '2016b'}
+
+sources = ['%(namelower)s-%(version)s-source.zip']
+source_urls = [('http://sourceforge.net/projects/bowtie-bio/files/%(namelower)s/%(version)s', 'download')]
+
+toolchainopts = {'cstd': 'c++11'}
+
+# Make sure that the Makefile uses options for intel rather than the ones for gcc coded
+# in the Makefile by setting RELEASE_FLAGS.
+buildopts = 'RELEASE_FLAGS="$CXXFLAGS"'
```
boegel: We should also apply this same fix to the other Bowtie2 easyconfigs (in a separate PR).

While looking into that, I noticed that we have a Bowtie2 easyblock that is only used by 3/18 Bowtie2 easyconfigs, we should clean that up, probably by enhancing the Bowtie2 easyblock to specify `RELEASE_FLAGS`, and to make sure it copies everything that the easyconfigs now copy...

I'll try and find some time for that.

boegel:
Going in, thanks @klust!

PR #4397, opened 2017-03-27
{chem}[intel/2017a] libxc v3.0.0, ABINIT v8.2.2

migueldiascosta:
(created using `eb --new-pr`)

PR #4400, opened 2017-03-27
{chem}[foss,2016b] GPAW 1.2.0

mikstr:
Latest releases of GPAW (v 1.2.0) and ASE (v 3.13.0) for the foss-2016b toolchain.

One thing to note is that to have GPAW use FFTW, FFTW should be compiled with shared libraries, otherwise GPAW uses numpy's FFTs.

### DIFF
--- a/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
@@ -9,23 +9,21 @@ in the Python scripting language."""
 
 toolchain = {'name': 'foss', 'version': '2016b'}
 
-# source_urls = ['https://wiki.fysik.dtu.dk/ase-files/']
-# sources = ['python-%s-%s.tar.gz' % (name.lower(), version)]
-source_urls = ['https://pypi.python.org/packages/84/a8/664c99fc94510163b5289c8e475660182f0f6ba098c549879bc5d36c17fd/']
-sources = ['%s-%s.tar.gz' % (name.lower(), version)]
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
 python = 'Python'
 pythonver = '3.5.2'
-pythonshortver = '.'.join(pythonver.split('.')[0:2])
 versionsuffix = '-%s-%s' % (python, pythonver)
 
 dependencies = [
    (python, pythonver),
-   ('matplotlib', '1.5.3', '-Python-' + pythonver)
+   ('matplotlib', '1.5.3', versionsuffix)
 ]
 
 sanity_check_paths = {
-    'files': ['bin/ase-gui'],
-    'dirs': ['lib/python%s/site-packages/%s' % (pythonshortver, name.lower())]
+    'files': ['bin/ase-build', 'bin/ase-db', 'bin/ase-gui', 'bin/ase-info', 'bin/ase-run'],
+    'dirs': ['lib/python%(pyshortver)s/site-packages/%(namelower)s'],
 }
 
 moduleclass = 'chem'
--- a/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
@@ -2,6 +2,7 @@ easyblock = "PythonPackage"
 
 name = 'ASE'
 version = '3.13.0'
+versionsuffix = '-Python-%(pyver)s'
 
 homepage = 'https://wiki.fysik.dtu.dk/ase/'
 description = """ASE is a python package providing an open source Atomic Simulation Environment
@@ -12,12 +13,8 @@ toolchain = {'name': 'foss', 'version': '2016b'}
 source_urls = [PYPI_LOWER_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
 
-python = 'Python'
-pythonver = '3.5.2'
-versionsuffix = '-%s-%s' % (python, pythonver)
-
 dependencies = [
-   (python, pythonver),
+   ('Python', '3.5.2'),
    ('matplotlib', '1.5.3', versionsuffix)
 ]
 
--- a/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
@@ -1,0 +1,31 @@
+easyblock = "PythonPackage"
+
+name = 'ASE'
+version = '3.13.0'
+
+homepage = 'https://wiki.fysik.dtu.dk/ase/'
+description = """ASE is a python package providing an open source Atomic Simulation Environment
+in the Python scripting language."""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+# source_urls = ['https://wiki.fysik.dtu.dk/ase-files/']
+# sources = ['python-%s-%s.tar.gz' % (name.lower(), version)]
+source_urls = ['https://pypi.python.org/packages/84/a8/664c99fc94510163b5289c8e475660182f0f6ba098c549879bc5d36c17fd/']
+sources = ['%s-%s.tar.gz' % (name.lower(), version)]
+python = 'Python'
+pythonver = '3.5.2'
+pythonshortver = '.'.join(pythonver.split('.')[0:2])
+versionsuffix = '-%s-%s' % (python, pythonver)
+
+dependencies = [
+   (python, pythonver),
+   ('matplotlib', '1.5.1', '-Python-' + pythonver)
+]
+
+sanity_check_paths = {
+    'files': ['bin/ase-gui'],
+    'dirs': ['lib/python%s/site-packages/%s' % (pythonshortver, name.lower())]
+}
+
+moduleclass = 'chem'
--- a/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
@@ -22,7 +22,7 @@ dependencies = [
 ]
 
 sanity_check_paths = {
-    'files': ['bin/ase-build', 'bin/ase-db', 'bin/ase-gui', 'bin/ase-info', 'bin/ase-run'],
+    'files': ['bin/ase-' + p for p in ['build', 'db', 'gui', 'info', 'run']],
     'dirs': ['lib/python%(pyshortver)s/site-packages/%(namelower)s'],
 }
 
--- a/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
@@ -20,7 +20,7 @@ versionsuffix = '-%s-%s' % (python, pythonver)
 
 dependencies = [
    (python, pythonver),
-   ('matplotlib', '1.5.1', '-Python-' + pythonver)
+   ('matplotlib', '1.5.3', '-Python-' + pythonver)
 ]
 
 sanity_check_paths = {
--- a/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
@@ -13,10 +13,7 @@ toolchain = {'name': 'foss', 'version': '2016b'}
 source_urls = [PYPI_LOWER_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
 
-dependencies = [
-   ('Python', '3.5.2'),
-   ('matplotlib', '1.5.3', versionsuffix)
-]
+dependencies = [('Python', '3.5.2'), ('matplotlib', '1.5.3', versionsuffix)]
 
 sanity_check_paths = {
     'files': ['bin/ase-' + p for p in ['build', 'db', 'gui', 'info', 'run']],
--- a/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
@@ -13,7 +13,10 @@ toolchain = {'name': 'foss', 'version': '2016b'}
 source_urls = [PYPI_LOWER_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
 
-dependencies = [('Python', '3.5.2'), ('matplotlib', '1.5.3', versionsuffix)]
+dependencies = [
+    ('Python', '3.5.2'),
+    ('matplotlib', '1.5.3', versionsuffix)
+]
 
 sanity_check_paths = {
     'files': ['bin/ase-' + p for p in ['build', 'db', 'gui', 'info', 'run']],
--- a/easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
+++ b/easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
@@ -1,18 +1,26 @@
---- customize.py	2017-02-24 15:10:13.635016617 +0100
-+++ /home/camp/mikst/Desktop/customize.py	2017-02-24 15:19:02.267012659 +0100
-@@ -32,15 +32,21 @@
+--- gpaw-1.2.0-orig/customize.py	2017-05-24 11:57:38.765096571 +0200
++++ gpaw-1.2.0/customize.py	    2017-05-24 11:58:43.189777767 +0200
+@@ -27,20 +27,28 @@
+ 
+     libraries += ['somelib', 'otherlib']
+ """
+-
++import os
+ # compiler = 'gcc'
  # mpicompiler = 'mpicc'  # use None if you don't want to build a gpaw-python
  # mpilinker = 'mpicc'
  # platform_id = ''
 -# scalapack = False
 +scalapack = True
++
 +libraries = ['readline',
 +             'gfortran',
-+             'scalapack',
 +             'openblas',
-+             'xc'] 
++             'xc']
  
-+extra_compile_args = ['-O2', '-std=c99', '-fPIC', '-Wall']
++scalapack_libs = [libfile[3:-2] for libfile in
++                  os.getenv('SCALAPACK_STATIC_LIBS').split(',')]
++mpi_libraries += scalapack_libs
  # Use ScaLAPACK:
  # Warning! At least scalapack 2.0.1 is required!
  # See https://trac.fysik.dtu.dk/projects/gpaw/ticket/230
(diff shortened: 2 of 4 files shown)


mikstr:
I'll fix the dependencies asap!

mikstr:
Would you prefer that I make a new branch for libxc-3.0.0-foss2016b?

boegel:
@mikstr If you're only doing a toolchain bump, feel free to include those dependencies in this PR.

mikstr:
@boegel thanks, I have added libxc-3.0.0-foss-2016b in this PR as it is simply a toolchain bump

easybuild/easyconfigs/l/libxc/libxc-3.0.0-foss-2016b.eb
```diff
@@ -0,0 +1,31 @@
...
+source_urls = ['http://www.tddft.org/programs/octopus/down.php?file=libxc/']
+
+configopts = 'FC="$F77" FCFLAGS="$FFLAGS" --enable-shared --enable-fortran'
+
+# From the libxc mailing list
+# To summarize: expect less tests to fail in libxc 2.0.2, but don't expect
+# a fully working testsuite soon (unless someone wants to volunteer to do
+# it, of course  ) In the meantime, unless the majority of the tests
+# fail, your build should be fine.
+# runtest = 'check'
```
migueldiascosta: see #4397

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,32 @@
+easyblock = "PythonPackage"
+
+name = 'GPAW'
+version = '1.2.0'
```
mikstr: thanks, fixed the title in the PR :-)

easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,31 @@
...
+pythonshortver = '.'.join(pythonver.split('.')[0:2])
+versionsuffix = '-%s-%s' % (python, pythonver)
+
+dependencies = [
+   (python, pythonver),
+   ('matplotlib', '1.5.3', '-Python-' + pythonver)
+]
+
+sanity_check_paths = {
+    'files': ['bin/ase-gui'],
```
mikstr: good one

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
```diff
@@ -0,0 +1,28 @@
...
+ if scalapack:
+-    libraries += ['scalapack-openmpi',
+-                  'blacsCinit-openmpi',
+-                  'blacs-openmpi']
++#    libraries += ['scalapack-openmpi',
++#                  'blacsCinit-openmpi',
++#                  'blacs-openmpi']
+     define_macros += [('GPAW_NO_UNDERSCORE_CBLACS', '1')]
+     define_macros += [('GPAW_NO_UNDERSCORE_CSCALAPACK', '1')]
+
```
mikstr: yes, _gpaw.so is for serial runs only, parallel runs are done using "gpaw-python", a custom python interpreter with MPI interface

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,32 @@
...
+
+homepage = 'https://wiki.fysik.dtu.dk/gpaw/'
+description = """GPAW is a density-functional theory (DFT) Python code based on the projector-augmented wave (PAW)
+method and the atomic simulation environment (ASE). It uses real-space uniform grids and multigrid methods or 
+atom-centered basis-functions."""
+toolchain = {'name': 'foss', 'version': '2016b'}
+source_urls = ['https://pypi.python.org/packages/3c/ed/c06fc0960c1ddc8bb5ae6a23d1164ffa78324758a3bfb50c677278bef14a']
```
mikstr: yes, thanks

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
```diff
@@ -0,0 +1,28 @@
...
+@@ -32,15 +32,21 @@
+ # mpicompiler = 'mpicc'  # use None if you don't want to build a gpaw-python
+ # mpilinker = 'mpicc'
+ # platform_id = ''
+-# scalapack = False
++scalapack = True
++libraries = ['readline',
++             'gfortran',
++             'scalapack',
++             'openblas',
```
mikstr: really new to easybuild, but is this the recommended way to go? (At least, I was not able to find another .eb file which use os.getenv('SCALAPACK_STATIC_LIBS'))

easybuild/easyconfigs/l/libxc/libxc-3.0.0-foss-2016b.eb
mikstr: @migueldiascosta should I just uncomment the line `# runtest = 'check'`?
Or is there more to it :-)?

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
migueldiascosta: @boegel, better way?

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
migueldiascosta: you're right, but for the intel toolchain, if `libraries` includes `blacs_intelmpi` and nothing else is done, `import _gpaw` fails with `undefined symbol: MPI_Finalize`. Setting `compiler=os.getenv('MPICC')` works around this...

Alternatively, and cleaner, one can use only lapack/blas in `libraries`, for building `_gpaw.so`, and scalapack/blacs in `mpi_libraries`, for building `gpaw-python`, that also works...

Again, there's probably a better way

mikstr: where and how do you see the blacs_intelmpi errors? I'm afraid I don't have access to intelmpi.

migueldiascosta: @mikstr it's not really about intelmpi, it's how mkl packages blacs (the same would happen with mkl_blacs_openmpi)

one solution is to include the MPI-interface in _gpaw.so, another is not to include scalapack/blacs when building _gpaw.so (as you said, it doesn't need it), only in gpaw-python 

The previous 0.9 easyconfig and patch did not have this problem because it used only lapack, not scalapack.

btw, you can test intel mkl/mpi if you want, they are now free (as in beer), https://software.intel.com/en-us/performance-libraries ("with a Named-User License per Intel’s standard end user licensing agreement")

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
```diff
@@ -0,0 +1,28 @@
+--- customize.py	2017-02-24 15:10:13.635016617 +0100
++++ /home/camp/mikst/Desktop/customize.py	2017-02-24 15:19:02.267012659 +0100
```
boegel: this location is wrong?

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
boegel: `$SCALAPACK_STATIC_LIBS` is set by the toolchain mechanism, and can be relied on sure

It's certainly better than hardcoding library names...

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
```diff
@@ -0,0 +1,28 @@
...
+ # platform_id = ''
+-# scalapack = False
++scalapack = True
++libraries = ['readline',
++             'gfortran',
++             'scalapack',
++             'openblas',
++             'xc'] 
+ 
++extra_compile_args = ['-O2', '-std=c99', '-fPIC', '-Wall']
```
boegel: these shouldn't be hardcoded either, use `$CFLAGS`

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,31 @@
...
+method and the atomic simulation environment (ASE). It uses real-space uniform grids and multigrid methods or 
+atom-centered basis-functions."""
+toolchain = {'name': 'foss', 'version': '2016b'}
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+patches = ['GPAW-1.2.0-customize.patch']
+python = 'Python'
+pythonver = '3.5.2'
+versionsuffix = '-%s-%s' % (python, pythonver)
```
boegel: please use the `%(pyver)s` template like we do in other easyconfigs, and move up the `versionsuffix` definition right under `version`

easybuild/easyconfigs/l/libxc/libxc-3.0.0-foss-2016b.eb
```diff
@@ -0,0 +1,31 @@
...
+source_urls = ['http://www.tddft.org/programs/octopus/down.php?file=libxc/']
+
+configopts = 'FC="$F77" FCFLAGS="$FFLAGS" --enable-shared --enable-fortran'
+
+# From the libxc mailing list
+# To summarize: expect less tests to fail in libxc 2.0.2, but don't expect
+# a fully working testsuite soon (unless someone wants to volunteer to do
+# it, of course  ) In the meantime, unless the majority of the tests
+# fail, your build should be fine.
+runtest = 'check'
```
boegel: So, all tests pass now?

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
mikstr: sorry, how do I change $CFLAGS in the easyconfig file?

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
mikstr: @migueldiascosta, thanks, went with the solution of not including scalapack in `libraries` but in `mpi_libraries`.

easybuild/easyconfigs/l/libxc/libxc-3.0.0-foss-2016b.eb
mikstr: it looks like it, but not and expert here. At least it says 
> PASS: xc-run_testsuite
in the logfile.

mikstr:
What more is needed to get this merged? 
The review comments have been adressed, anyting else missing.

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
```diff
@@ -0,0 +1,36 @@
+--- gpaw-1.2.0-orig/customize.py	2017-05-24 11:57:38.765096571 +0200
```
boegel: @mikstr Please include a short description for this patch at the top + also mention author (+ organisation) an/or include a reference to an upstream patch if relevant; anything above the first `---` line is considered comments

boegel:
@mikstr We've recently enabled automatic style checks, please fix these too to make Travis happy on the next commit:

```
$ eb --from-pr 4400 --check-style
== temporary log file in case of crash /tmp/eb-iI9h9n/easybuild-M9BB_K.log
Running style check on 3 easyconfig(s)...
/tmp/eb-iI9h9n/files_pr4400/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb:17:4: E121 continuation line under-indented for hanging indent
[FAIL] /tmp/eb-iI9h9n/files_pr4400/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
[PASS] /tmp/eb-iI9h9n/files_pr4400/g/GPAW/GPAW-1.2.0-foss-2016b-Python-3.5.2.eb
/tmp/eb-iI9h9n/files_pr4400/l/libxc/libxc-3.0.0-foss-2016b.eb:25:36: E231 missing whitespace after ','
[FAIL] /tmp/eb-iI9h9n/files_pr4400/l/libxc/libxc-3.0.0-foss-2016b.eb
ERROR: One or more style checks FAILED!
```

easybuild/easyconfigs/l/libxc/libxc-3.0.0-foss-2016b.eb
boegel: Tests still fail for me, see upcoming test report...

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 7 out of 9 (3 easyconfigs in this PR)
node2406.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/d8ebc62de08aef4c1aa2dbf9393ba31b for a full test report.

easybuild/easyconfigs/l/libxc/libxc-3.0.0-foss-2016b.eb
migueldiascosta: sorry, my comment "see #4397" was too laconic, let me fix that :) 

In #4397, we discuss the origin of those failures in Haswell only, and suggest disabling `optarch` for `libxc`, since the tests always pass without `optarch`. I'm of the opinion that it is more important for libxc to be reproducible than to be fast (and I doubt that libxc has much impact on the overwall execution times of the application using it, much less any eventual performance difference from the optimization flags of libxc alone, unless you're doing something very specific with the functionals themselves, in which case there could be a performance difference but it would still be more important to be reproducible), so I'd disable `optarch` and enable the tests for all libxc easyconfigs...

from the libxc homepage: "The aim is to provide a portable, well tested and reliable set of exchange and correlation functionals"

of course, one could argue that there is not exact reproducibility with floating point arithmetic, but I believe the libcx tests have already been relaxed to take that into account in version 2.1.0, this should be a different problem

too verbose? :)

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
```diff
@@ -0,0 +1,36 @@
...
+ # Use ScaLAPACK:
+ # Warning! At least scalapack 2.0.1 is required!
+ # See https://trac.fysik.dtu.dk/projects/gpaw/ticket/230
+ if scalapack:
+-    libraries += ['scalapack-openmpi',
+-                  'blacsCinit-openmpi',
+-                  'blacs-openmpi']
++#    libraries += ['scalapack-openmpi',
++#                  'blacsCinit-openmpi',
++#                  'blacs-openmpi']
```
migueldiascosta: this is specific to the foss toolchain... it would be better if the customize patch was independent of the toolchain (the existing one, for 0.9.0, is)

mikstr: not sure why that is, all lines with openmpi appear to be commented out, right?

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
mikstr: @boegel thanks for helping, added description, author and institution (see ae2977a88cb80bb388f17852e37ff11838037655)

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-customize.patch
mikstr: @migueldiascosta I think, it will probably be difficult to avoid a tool-chain specific patch until GPAW is made more clever when building it.

easybuild/easyconfigs/l/libxc/libxc-3.0.0-foss-2016b.eb
mikstr: @boegel @migueldiascosta I n oticed that there is now a [libxc-3.0.0-GCC-5.4.0-2.26](https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/l/libxc/libxc-3.0.0-GCC-5.4.0-2.26.eb) in master, which could perhaps be used for foss-2016b in general. In any case I borrowed the` toolchainopts = {'lowopt': True}`  from that (see a2ebc76dd1ea43e5821adc240ec9790ac9e2469b ) to hopefully pass the libxc tests 😅

mikstr:
@boegel @migueldiascosta I think it basically ready now, right?
And thanks for all the good suggestions.

easybuild/easyconfigs/a/ASE/ASE-3.13.0-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,23 @@
...
+homepage = 'https://wiki.fysik.dtu.dk/ase/'
+description = """ASE is a python package providing an open source Atomic Simulation Environment
+in the Python scripting language."""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [('Python', '3.5.2'), ('matplotlib', '1.5.3', versionsuffix)]
```
boegel: @mikstr We usually list one dependency per line:

```python
dependencies = [
    ('Python', '3.5.2'),
    ('matplotlib', '1.5.3', versionsuffix),
]

File: easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-foss-2016b-Python-3.5.2.eb [hunk 14]
```diff
@@ -0,0 +1,29 @@
...
+
+name = 'GPAW'
+version = '1.2.0'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://wiki.fysik.dtu.dk/gpaw/'
+description = """GPAW is a density-functional theory (DFT) Python code based on the projector-augmented wave (PAW)
+method and the atomic simulation environment (ASE). It uses real-space uniform grids and multigrid methods or 
+atom-centered basis-functions."""
+toolchain = {'name': 'foss', 'version': '2016b'}
```
boegel: @mikstr please add an empty line before/after the `toolchain =` line

easybuild/easyconfigs/g/GPAW/GPAW-1.2.0-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,29 @@
...
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+patches = ['GPAW-1.2.0-customize.patch']
```
boegel: @mikstr please avoid hardcoding the version here, use `%(version)s`:

```python
patches = ['GPAW-%(version)s-customize.patch']

boegel:
Test report by @boegel
**SUCCESS**
node2457.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/4d2ac9800f1caf9113c2106fd64aa107 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2088.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/6dc7841105ddd2dc28249c4a50fcca72 for a full test report.

boegel:
This should be checked against #4772, where the same GPAW version with the same toolchain is being contributed, incl. a (different) patch...

akesandgren:
Any progress on this one and #4772 getting "merged"?

boegel:
@akesandgren Someone should look into merging this with #4772 to get the best of both worlds.

We haven't had any installation requests for GPAW, so I currently can't spend much time on this, I'm afraid...

cc @pouillon

mikstr:
superseded by #6514

PR #4411, opened 2017-03-29
Add Open Knowledgebase of Interatomic Models (OpenKIM)

schiotz:
The Open Knowledgebase of Interatomic Models (openkim.org) is a library
of interatomic models that can be used by some simulation packages, including
LAMMPS and Asap. An easybuild of the latter will be contributed shortly.

~~This pull request depends on easybuild-easyblocks pull request https://github.com/hpcugent/easybuild-easyblocks/pull/1155~~

EDIT: Removed dependency on easyblock PR.


schiotz:
@boegel 
This easyconfig _no longer depends on the associated easyblock PR_.  Following the suggestions in that PR, we no longer download the list of KIM models at build time, instead the list is hardcoded into the .eb file.  The price of eliminating the custom easyblock is a for loop in the end of the .eb file.  Is that acceptable?

schiotz:
@boegel 
If the history of this PR is too messy, I can make a new one based on the tip of the develop branch.

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
```diff
@@ -0,0 +1,256 @@
...
+modextrapaths = {'KIM_HOME': '.'}
+
+moduleclass = 'chem'
+
+# Sources for all KIM items.  The list will be patched below.
+sources += [
+    #
+    # Model drivers
+    #
+    'MD_853402641673_000.tgz',  # EAM_CubicNaturalSpline
```
boegel: @schiotz I hope this list was generated somehow? If so, please mention how in comments?

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
```diff
@@ -0,0 +1,256 @@
...
+source_urls = ['https://openkim.org/download/', 'https://s3.openkim.org/kim-api/']
+
+parallel = 1
+
+modextrapaths = {'KIM_HOME': '.'}
+
+moduleclass = 'chem'
+
+# Sources for all KIM items.  The list will be patched below.
+sources += [
```
boegel: There's no reason to not collapse this with the above?

```python
sources = [
    'kim-api-v%(version)s.tgz',
    #
    # Model drivers
    #
    'MD_853402641673_000.tgz',  # EAM_CubicNaturalSpline
    ...

File: easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb [hunk 3]
```diff
@@ -0,0 +1,256 @@
...
+description = """Open Knowledgebase of Interatomic Models.
+
+OpenKIM is an API and a collection of interatomic models (potentials) for
+atomistic simulations.  It is a library that can be used by simulation programs
+to get access to the models in the OpenKIM database.
+"""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+sources = ['kim-api-v%s.tgz' % (version,)]
```
boegel: @schiotz please use `%(version)s` template

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
```diff
@@ -0,0 +1,256 @@
+easyblock = 'ConfigureMake'
+name = 'OpenKIM'
```
boegel: style nitpicking: missing empty line above the `name` line

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
```diff
@@ -0,0 +1,256 @@
...
+atomistic simulations.  It is a library that can be used by simulation programs
+to get access to the models in the OpenKIM database.
+"""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+sources = ['kim-api-v%s.tgz' % (version,)]
+
+# Models and drivers are downloaded from the first URL, the API files from the latter.
+source_urls = ['https://openkim.org/download/', 'https://s3.openkim.org/kim-api/']
```
boegel: please list these one per line:

```python
source_urls = [
    'https://openkim.org/download/',  # for model & drivers
    'https://s3.openkim.org/kim-api/',  # for API files
]

File: easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb [hunk 6]
```diff
@@ -0,0 +1,256 @@
...
+]
+
+
+# Now the sources list should be changed, so that models and model drivers
+# are unpacked in the right folders.
+
+kim_maindir = 'kim-api-v%s' % (version,)
+kim_subdir = {'MO': 'models',  'MD': 'model_drivers'}
+for i, x in enumerate(sources):
+    if x.startswith('MO_') or x.startswith('MD_'):
```
boegel: @schiotz I don't think this is something we want in an easyconfig file, anything that requires executing code should go into an easyblock...

Since the pattern is fairly obvious, how about doing something like this?

```python
mo_subdir = 'models'
md_subdir = 'model_drivers'
sources = [
    'kim-api-v%(version)s.tgz',
    #
    # Model drivers
    #
    {
        'filename': 'MD_853402641673_000.tgz',  # EAM_CubicNaturalSpline
        'extract_cmd': "tar xvz kim-api-v%%(version)s/src/%s -f %%s" % md_subdir,
    }
    ...
```

That would still be a nightmare to maintain though, it's a lot easier to just specify a 2 lists with `MD_*.tgz` and `MO_*.tgz` filenames, and then let a custom OpenKIM easyblock loop over them to extract them in the right place?

schiotz:
@boegel 
Thanks for your comments.  I think I have addressed them all.

I have moved the code into an EasyBlock.  It is no longer told the
exact folder name, instead it gets it from the original source file.
In that way it will still work if they change the way the source
folder is named.  The EasyBlock has been submitted as a brand-new pull
request, since it does not share any code with the previous drafts of
the easyblock: https://github.com/hpcugent/easybuild-easyblocks/pull/1204

There is one thing I have not done as suggested: You suggested
collapsing the sources list into one.  I still use `source += ...`
since that was I keep the autogenerated part of the easyconfig
separate from the part one might want to change manually in a future
version.

Finally, I include the script that autogenerates the sources list.  Is that appropriate?

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
```diff
@@ -0,0 +1,243 @@
...
+
+parallel = 1
+
+modextrapaths = {'KIM_HOME': '.'}
+
+moduleclass = 'chem'
+
+# Sources for all KIM items.  Autogenerated by make-kim-model-list.py
+# AUTOGENERATED ON: Fri Jun 30 08:44:55 2017 UTC
+sources += [
```
boegel: @schiotz OK to include like this, but do move it up below the `sources =` line

At some point, we will start for checking the order of easyconfig parameters as well in the automated style check

boegel:
@schiotz Including the script like you did makes perfect sense, thanks!

easybuild/easyconfigs/o/OpenKIM/make-kim-model-list.py
```diff
@@ -0,0 +1,32 @@
+from urllib.request import urlopen
+import json
+import time
+
+query = b'&fields={"kimcode":1,"kim-api-version":1,"driver":1}&database=obj&query={"type":"mo"}'
+#query = b'&database=obj&query={"type":"mo"}'
+url = 'https://query.openkim.org/api'
+timeout = 60        
+# We should maybe do some error handling here?
+url_fd = urlopen(url, data=query, timeout=timeout)
```
boegel: Maybe take a look at https://github.com/hpcugent/easybuild-framework/blob/master/easybuild/tools/filetools.py#L408 for inspiration w.r.t. error handling.

Also, it's better to use `urllib2` which is aware of proxies.

It's OK to include as-is though, up to you if you want to spend more time on this.

easybuild/easyconfigs/o/OpenKIM/make-kim-model-list.py
```diff
@@ -0,0 +1,32 @@
+from urllib.request import urlopen
```
boegel: @schiotz Please include license header + mention authorship (so people can contact you in case of questions/problems concerning this script)

schiotz:
@boegel 
I have moved the autogenerated part up under the `source = ...` line, although it bothers me a bit that autogenerated stuff then goes in the middle of human-maintainable lines.  Are you really going to enforce the order of things in EasyConfig files?  To me that seems to be needless consistency for consistency's own sake, rarely a good idea.  But I guess you know what you are doing :-)

I have added the requested headers to the supporting script, and fixed the import.  Very embarrassing, the script actually only worked with Python 3!

I hope everything is fine now, if not we will just do another iteration :-)

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
```diff
@@ -0,0 +1,243 @@
...
+    'MO_261274272789_001.tgz',  # model_Ta_PF_cubicsplines_derletnguyenmanhdudarev
+    'MO_683890323730_001.tgz',  # model_V_PF_cubicsplines_derletnguyenmanhdudarev
+    'MO_195478838873_001.tgz',  # model_W_PF_cubicsplines
+]
+# END OF AUTOGENERATED CODE.
+
+source_urls = [
+    'https://openkim.org/download/',    # For models and drivers
+    'https://s3.openkim.org/kim-api/',  # For API files
+]
```
boegel: we typically include `source_urls` above `sources`

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
```diff
@@ -0,0 +1,243 @@
...
+
+parallel = 1
+
+modextrapaths = {'KIM_HOME': '.'}
```
boegel: this should be handled by the `OpenKIM` easyblock, done in update to your easyblock via https://github.com/schiotz/easybuild-easyblocks/pull/1

easybuild/easyconfigs/o/OpenKIM/OpenKIM-1.8.2-foss-2016b.eb
schiotz: @boegel 
Something is badly wrong.  The checksum of at least one model driver file has changed.  I suspect it may be related to the release of a new version of the API, that may have triggered regeneration of the .tar.gz files (or perhaps they are always generated on the fly).

I am investigating!

schiotz:
@boegel 
We have a problem with the checksums: The OpenKIM.org servers generate the .tar.gz files on the fly.  Usually they are cached, but that cache may be cleared, causing the tar files to be regenerated.  This changes the checksum.  

Furthermore, metadata that does not affect the code may be updated without creating a new version number.  An example is the reference to the paper publishing the model, the models often first refer to a preprint (so the final paper can refer to the model), and then that reference is later updated.

They provide a database of SHA-1 checksums of all files *inside* the tar files, I do not think that this will help us.

I have asked the developers if they can provide autogenerated monthly archives of all models.  Then I could make an EB of the API, and a separate one with the latest monthly package.  It should work.

Another alternative may be to make an EB with the API alone, and then a script the user can run to install all models.  It looks like the latest API released last week has better support for this, but I have yet to look at it.

boegel:
@schiotz Ugh, that's indeed a poor match with our current approach of including checksums for downloaded source tarballs...

Keep us posted if you have any updates on the archived models!

schiotz:
@boegel 
I close this PR, and replace it with PR #5141 which only installs the API.  Version 1.9 of the API includes tools that makes it really easy for the users to maintain their own collection of models.  If the OpenKIM developers begin releasing snapshots of the model database (with stable checksums), then I will add a separate EasyConfig installing the models.

PR #4451, opened 2017-04-05
Fix pkgconfig patch for Qhull to only let CMake substitute @

bartoldeman:
see https://cmake.org/cmake/help/latest/command/configure_file.html
otherwise it replaces ${...} by empty strings which is not what we want and broke our Octave compilation.

### DIFF
--- a/easybuild/easyconfigs/q/Qhull/Qhull_pkgconfig.patch
+++ b/easybuild/easyconfigs/q/Qhull/Qhull_pkgconfig.patch
@@ -3,7 +3,7 @@ requires making CMake to fill in the gaps via 'configure_file(... @ONLY)'
 author: Kenneth Hoste (UGent)
 --- qhull/CMakeLists.txt.orig	2015-09-21 15:27:38.424110227 +0200
 +++ qhull/CMakeLists.txt	2015-09-21 15:27:06.403494984 +0200
-@@ -430,3 +430,9 @@
+@@ -430,3 +430,10 @@
  install(FILES html/rbox.man          DESTINATION ${MAN_INSTALL_DIR} RENAME rbox.1)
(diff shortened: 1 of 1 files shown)


boegel:
@bartoldeman How did it break you Octave build exactly?

note to self: stuff to retest that lists Qhull as a dep: `Octave`, `OpenMD`, `PCL`, `pyhull`, `scikit-image`

bartoldeman:
Somewhere in the Octave build procedure where using QHull it used
"-L -L/some/path". I submitted a PR for Octave 4.2.1 so perhaps you can see it for yourself.

boegel:
without the patch:

```
$ cat $EASYBUILD_PREFIX/software/Qhull/2015.2-intel-2016b/lib/pkgconfig/qhull.pc
prefix=/user/home/gent/vsc400/vsc40023/eb_phanpyscratch/CO7/haswell-ib/software/Qhull/2015.2-intel-2016b
exec_prefix=
includedir=/include
libdir=/lib

Name: qhull
Description: General dimension convex hull library
Version: 2015.2 2016/01/18
Requires:
Libs: -L -lqhull -lm
Cflags: -I
```

with the patch:

```
cat $EASYBUILD_PREFIX/software/Qhull/2015.2-intel-2016b/lib/pkgconfig/qhull.pc
prefix=/user/home/gent/vsc400/vsc40023/eb_phanpyscratch/CO7/haswell-ib/software/Qhull/2015.2-intel-2016b
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${prefix}/lib

Name: qhull
Description: General dimension convex hull library
Version: 2015.2 2016/01/18
Requires:
Libs: -L${libdir} -lqhull -lm
Cflags: -I${includedir}
```

boegel:
Thanks for the fix @bartoldeman!

PR #4452, opened 2017-04-06
Octave 4.2.1: add easyconfig for intel 2016b with updated deps incl Qt5

bartoldeman:

PR #4471, opened 2017-04-07
Fixed broken Qhull patch

damianam:
This got me scratching my head for a while...........

Octave depends on Qhull. Octave was failing with an obscure libtool error:

`require no space between -L and -L/one/path`

After much digging through the octave logs I found that `QHULL_LDFLAGS` was set to `-L`. That came from `pkg-config`. The issue is that in the patch applied to `Qhull`, `CMake` replaces both `@variable@` and `${variable}` kind of variables. So `${prefix}` and family are replaced with nothing, since they are not defined. See https://cmake.org/cmake/help/v3.0/command/configure_file.html The solution is to add `@ONLY` to the cmake command


boegel:
@damianam duplicate of #4451 (verified with #4452)? If you sync your branch with current `develop` this PR should become empty...

damianam:
Seems like @bartoldeman beat me by 1 day....... Next time I'll take a look at pending PR's

PR #4483, opened 2017-04-12
{bio}(foss/2015b) Swarm 2.1.13 (WIP)

nathanhaigh:
Swarm comes with a couple of Python helper scripts. I'm not sure how the dependencies in these should be specified in the easyconfig file.

### DIFF
--- a/easybuild/easyconfigs/s/Swarm/Swarm-2.1.13-foss-2015b-Python-2.7.10.eb
+++ b/easybuild/easyconfigs/s/Swarm/Swarm-2.1.13-foss-2015b-Python-2.7.10.eb
@@ -1,0 +1,38 @@
+easyblock = 'MakeCp'
+
+name = "Swarm"
+version = "2.1.13"
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://github.com/torognes/swarm'
+description = "A robust and fast clustering method for amplicon-based studies"
+
+toolchain = {'name': 'foss', 'version': '2015b'}
+
+source_urls = ['https://github.com/torognes/swarm/archive']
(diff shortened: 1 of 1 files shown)


wpoely86:
@nathanhaigh New dependencies that are not in the `Python` easyconfig yet?

nathanhaigh:
@wpoely86 That's right, one of the scripts requires `igraph` and `optparse`:

```
from igraph import Graph, plot
from optparse import OptionParser
```

wpoely86:
Than you can either add them in this easyconfig as PythonExtensions or add a dependency on them?

boegel:
@nathanhaigh +1 on listing the required non-standard Python packages as additional extensions, see for example https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/m/matplotlib/matplotlib-2.0.0-intel-2016b-Python-2.7.12.eb (which `easyblock` is used for the parent installation is irrelevant).

nathanhaigh:
Thanks @boegel. Turns out that `python-igraph` requires the `igraph` C library to be installed. There is an easyconfig for the latter. Should i create a separate easyconfig "bundle" for `python-igraph` which depends on Python and igraph, then add that as a dependency here?

boegel:
@nathanhaigh either that, or just also include `igraph` as a dependency to Swarm, and include `python-igraph` as an extension

the downside of including `python-igraph` as a dependency here is that you'll end up with another module for `python-igraph`

boegel:
@nathanhaigh Is this still WIP?

migueldiascosta:
@nathanhaigh, this PR is being closed for the following reason(s): no activity for > 6 months.

PR #4484, opened 2017-04-13
{chem}[intel/2017a] atomate v0.4.4 and dependencies

migueldiascosta:
(created using `eb --new-pr`)


migueldiascosta:
**SUCCESS**
Build succeeded for 19 out of 19 (19 easyconfigs in this PR)
nscc04 - Linux RHEL 6.8, Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz, Python 2.7.12
See https://gist.github.com/98069be639020df380af39006ffa39e6 for a full test report.

boegel:
@migueldiascosta I realise this has taken a lot of effort, but wouldn't it be better to install `atomate` as a single bundle of Python packages, rather than installing each Python package in a separate module?

The downside of the current approach is that you'll end up with a bunch of additional modules for Python packages that may only be relevant as dependencies to `atomate`, and that Python startup performance with several Python packages that are spread out across different locations on a shared filesystem is significantly slower...

See for example how `IPython` is installed: https://github.com/hpcugent/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/i/IPython/IPython-5.2.2-intel-2016b-Python-2.7.12.eb.

Some Python packages are still listed as dependencies, but generally only those that have 3rd-party (non-Python) dependencies. `PyYAML` would be one of those here...

migueldiascosta:
@boegel no worries. I've never tried using a bundle but originally I simply let the package's setup install the dependencies implicitly, which works but I guess is also not recommended. What are the advantages of the bundle over this, besides being explicit? 

Also, I'm not sure I completely understand the concept of bundles (of modules, right?) and extensions (in this case, PythonPackages, installed in a central location for that bundle?), maybe the writing easyconfigs documentation should mention them?

Regarding what should be an extension or a separate module, the thing is that the dependencies are much more generic than atomate (in particular pymatgen, which is much more widely used, not to mention matplotlib :) ). I suppose someone else will create the matplotlib 2.0.0 easyconfig for intel-2017a but regarding pymatgen, shall I include it as an extension in an atomate bundle and additionally open a separate PR for it or leave it as a separate module?

migueldiascosta:
**FAILED**
Build succeeded for 5 out of 8 (8 easyconfigs in this PR)
nscc04 - Linux RHEL 6.8, Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz, Python 2.7.13
See https://gist.github.com/3453dd1eccc8f33a1beb02a95bf53795 for a full test report.

boegel:
@migueldiascosta The advantage of using a bundle is mainly two-fold: the required packaging are installed in a single location (so no unnecessary explosion of modules), and also performance (spreading Python packages across a lot of locations has an impact on how many entries are in `$PYTHONPATH`, which can slow down Python startup time when it goes hunting for modules on `import`).

With a bundle I mean an easyconfig file using the `Bundle` easyblock, which doesn't install anything by itself, but it can serve as a way to combine multiple modules together (a toolchain is a special kind of bundle), and can be (ab) used to install a bunch of extensions together in a single location.

Covering this better in the documentation is a good idea, any suggestions for a location or how to explain this exactly?

For stuff like `pymatgen` that is also useful for things other than `atomate` (or maybe stand-alone),  providing a dedicated easyconfig file for it (like we already do for `matplotlib`) makes sense.

boegel:
W.r.t. the failing test: it seems like you need to enable the `pic` toolchain option for `pymatgen` (or a dependency thereof)?

easybuild/easyconfigs/c/custodian/custodian-1.1.0-intel-2017a-Python-2.7.13.eb
```diff
@@ -0,0 +1,39 @@
...
+]
+
+# this is a bundle of Python packages
+exts_defaultclass = 'PythonPackage'
+
+exts_list = [
+    ('custodian', version, {
+        'source_urls': ['https://pypi.python.org/packages/source/c/custodian/'],
+    }),
+    ('monty', '0.9.6', {
```
boegel: `monty` a dependency for `custodian`, so it should be listed first (to avoid auto-installing it)

easybuild/easyconfigs/f/FireWorks/FireWorks-1.4.2-intel-2017a-Python-2.7.13.eb
```diff
@@ -0,0 +1,60 @@
...
+]
+
+# this is a bundle of Python packages
+exts_defaultclass = 'PythonPackage'
+
+exts_list = [
+    ('FireWorks', version, {
+        'source_urls': ['https://pypi.python.org/packages/source/F/FireWorks/'],
+    }),
+    ('Flask', '0.12', {
```
boegel: same here, dependencies of `FireWorks` should be listed first

easybuild/easyconfigs/p/pymatgen-db/pymatgen-db-0.6.5-intel-2017a-Python-2.7.13.eb
```diff
@@ -0,0 +1,41 @@
...
+
+# this is a bundle of Python packages
+exts_defaultclass = 'PythonPackage'
+
+exts_list = [
+    ('pymatgen-db', '0.6.5', {
+        'modulename': 'matgendb',
+        'source_urls': ['https://pypi.python.org/packages/source/p/pymatgen-db/'],
+    }),
+    ('smoqe', '0.1.3', {
```
boegel: move up above `pymatgen-db`?

easybuild/easyconfigs/p/pymatgen/pymatgen-4.7.3-intel-2017a-Python-2.7.13.eb
```diff
@@ -0,0 +1,62 @@
...
+    ('monty', '0.9.6', {
+        'source_urls': ['https://pypi.python.org/packages/source/m/monty/'],
+    }),
+    ('mpmath', '0.19', {
+        'source_urls': ['https://pypi.python.org/packages/source/m/mpmath/'],
+    }),
+    ('pymatgen', version, {
+        'source_urls': ['https://pypi.python.org/packages/source/p/pymatgen/'],
+    }),
+    ('requests', '2.13.0', {
```
boegel: this is already included in the Python module?

easybuild/easyconfigs/p/pymatgen/pymatgen-4.7.3-intel-2017a-Python-2.7.13.eb
```diff
@@ -0,0 +1,62 @@
...
+        'source_urls': ['https://pypi.python.org/packages/source/r/requests/'],
+    }),
+    ('sympy', '1.0', {
```
boegel: should be moved up?

migueldiascosta:
@boegel still regarding the use of the `Bundle` easyblock for python packages, in terms of packaging and performance, how is this different than simply letting the `PythonPackage` easyblock install the requirements (when they are listed in `setup.py`'s `install_requires`), besides being more explicit? The end result seems to be the same (just trying to understand, being more explicit is reason enough for me)

migueldiascosta:
**SUCCESS**
Build succeeded for 8 out of 8 (8 easyconfigs in this PR)
nscc04 - Linux RHEL 6.8, Intel(R) Xeon(R) CPU E5-2690 v3 @ 2.60GHz, Python 2.7.13
See https://gist.github.com/24d4f9eeec5d22de76041822b93f6e5e for a full test report.

boegel:
@migueldiascosta The difference is that we're in control over all versions, just like we are when specifying regular dependencies where we also fix versions.

When required packages get pulled in via `setup.py` automatically, it's possible that when the most recent version gets pulled in that stuff doesn't work anymore if the version requirements were not specified strictly enough. E.g. `mpi4py` is now at version 2.0.0, which is not backwards-compatible with v1.x.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 8 out of 8 (8 easyconfigs in this PR)
node2066.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/b8b134c61367419e35a35f3f41e21039 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 8 out of 8 (8 easyconfigs in this PR)
node2571.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/001be5e51f9a41d6eca9ab3e2adba6f8 for a full test report.

boegel:
lgtm

boegel:

PR #4505, opened 2017-04-20
Change source_urls of pycrypto to encrypted https://pypi.python.org/...

haribert:
There are at least three reasons for this:
- A crypto library should be downloaded via http**s**:// (it is remarkable that pycrypto is almost the only library which was downloaded unencrypted via http://)
- python.org is the more reliable source than dlitz.net
- dlitz.net is at the moment (20.4.2017) not online.


boegel:
Thanks for the consistent change @haribert, there's no question it's better to download `pycrypto` from PyPI via `https`.

I'm getting the same MD5sum for the new download (`55a61a054aa66812daf5161a0d5d7eda`), so going in!

PR #4606, opened 2017-05-16
use 'git diff --name-only' and $TRAVIS_COMMIT_RANGE in Travis config to get list of changed files

boegel:
For now, this will just print a list of changed files to the Travis log.

The idea is to see if we can use this to get a list of changes files for a PR, so we can do a partial style check, or check whether `SHA256` checksums are available for only the files touched by a PR rather than everything at once...

The `sed` is required because of https://github.com/travis-ci/travis-ci/issues/4596.


boegel:

PR #4619, opened 2017-05-19
run 'git diff --name-only' from target branch to current HEAD in Travis config to get relevant list of changed files

boegel:
We should be comparing with `HEAD` to include the merge commit that Travis creates when testing PRs, otherwise the list of changes files is not what we expect to be if the branch used for the PR is not in sync with `develop`.

follow-up for #4606


boegel:
grr, for some reason Travis isn't picking up new commits, need to close/open the PR over and over to make it realise something changed...

boegel:
I think this does what we want, see https://travis-ci.org/hpcugent/easybuild-easyconfigs/jobs/234268643#L856.

http://stackoverflow.com/a/41159710/1467322 and http://stackoverflow.com/questions/7251477/what-are-the-differences-between-double-dot-and-triple-dot-in-git-dif/7256391#7256391 were a big help...

@wpoely86 I'd like to merge this in, if it does what we want for incoming PRs, we can look into enhancing the tests to pick up the list of changed files using this command to start enforcing style checks and SHA256 checksums for files being touched in PRs...

wpoely86:
Looks good but this doesn't do anything other the printing the list of files?

boegel:
@wpoely86 Not yet, I'm still evaluating whether this does what I hope it does, we can only do that in actual PRs once this is included...

PR #4633, opened 2017-05-24
{vis}[intel/2017a] Ferret 7.1 (WNIP)

hajgato:
Work Not In Progress
It compiles without any error, but when I try to run ferret it gives:
```
**ERROR Ferret crash; signal = 11
```

debugger:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000631da6 in define_special_grids_ ()
```


hajgato:
`-DG77_SIGNAL` should be removed from macro definitions for `Intel` compilers. Newer version is in #5731

PR #4681, opened 2017-06-07
Feature/pymatgen 4.3.2

ingridbr:


boegel:
@ingridbr Can you clarify what is going on here, this looks a bit messy?

Did you intend to contribute back a single easyconfig, or everything in one go?

If the latter, this is going to be a nightmare to get merged, especially since you're changing dependency versions in existing easyconfig files (which we don't do, unless there's something very wrong) and also some site-specific tweaks seems to have crept in...

ingridbr:
Hi Kenneth,

Sorry, I did the pull reqyuest to Gent by mistake instead of to our repository. You can forget about it.

Best regards,

Ingrid

PR #4697, opened 2017-06-09
{bio}{dummy} SEED added

robqiao:
SEED easyconfig added

### DIFF
--- a/easybuild/easyconfigs/s/SEED/SEED-2015.eb
+++ b/easybuild/easyconfigs/s/SEED/SEED-2015.eb
@@ -1,0 +1,36 @@
+##
+# This is a contribution from Phoenix HPC Service, The University of Adelaide, Australia
+# Homepage: https://www.adelaide.edu.au/phoenix/
+#
+# Copyright::   Copyright 2014-2017 adelaide.edu.au/phoenix
+# Authors::	Robert Qiao <robert.qiao@adelaide.edu.au>, Exequiel Sepulveda <exequiel.sepulveda@gmail.com>
+# License::     Artistic-v2.0
+#
+# Notes::
+##
+
+easyblock = 'CmdCp'
+
+name = 'SEED'
+version = '2015'
+
+homepage = 'http://www.theseed.org/wiki/Main_Page'
+description = """ SEED is a software for clustering large sets of Next Generation Sequences (NGS) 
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/s/SEED/SEED-2015.eb
```diff
@@ -0,0 +1,36 @@
...
+# Authors::	Robert Qiao <robert.qiao@adelaide.edu.au>, Exequiel Sepulveda <exequiel.sepulveda@gmail.com>
+# License::     Artistic-v2.0
+#
+# Notes::
+##
+
+easyblock = 'CmdCp'
+
+name = 'SEED'
+version = '2015'
```
boegel: @robqiao please use `20150725` (i.e. the full datestamp of the commit) as version, and tie it to the commit ID:

```python
version = '20160725'
commit = 'f9784b0'
```

and then use the commit ID in the `sources` spec:

```python
sources = ['%s.tar.gz' % commit]

File: easybuild/easyconfigs/s/SEED/SEED-2015.eb [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+
+homepage = 'http://www.theseed.org/wiki/Main_Page'
+description = """ SEED is a software for clustering large sets of Next Generation Sequences (NGS) 
+with hundreds of millions of reads in a time and memory efficient manner. 
+Its algorithm joins highly similar sequences into clusters that can differ 
+by up to three mismatches and three overhanging residues. 
+"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
```
boegel: @robqiao since the source is available, shouldn't we compile it ourselves with a proper toolchain?

easybuild/easyconfigs/s/SEED/SEED-2015.eb
fgeorgatos: this is elegant in style, yet I'm afraid there is no `--try-*` escape path in case someone wishes to advance a version,
since `--try-amend=commit=bla` goes at the end of a modified easyconfig and it won't be picked up correctly.

any feedback on that?

boegel:
@robqiao You'll need to issue a new pull request for this... GitHub is reporting it as merged, but that's because the commit here was also included in #4703, where you later removed the `SEED` easyconfig...

PR #4713, opened 2017-06-15
{bio}{foss/2016b}GenomeStrip

robqiao:
GenomeStrip easyconfig


easybuild/easyconfigs/g/GenomeSTRiP/GenomeStrip-2.00.1747-foss-2016b.eb
```diff
@@ -0,0 +1,69 @@
...
+homepage = 'http://software.broadinstitute.org/software/genomestrip/'
+description = """Genome STRiP (Genome STRucture In Populations) is a suite of tools 
+for discovering and genotyping structural variations using sequencing data. 
+The methods are designed to detect shared variation using data from multiple individuals.
+"""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+toolchainopts = {'usempi': True}
+
+sources = ['svtoolkit_2.00.1747.tar']
```
robqiao: Download Tarball has to be downloaded manually, due to rego requirement

boegel: @robqiao Please mention the location where you can download from in a comment above this `sources` line

Also, don't hardcode the version here, use `%(version)s`:

```python
sources = ['svtoolkit_%(version)s.tar']
```

easybuild/easyconfigs/g/GenomeSTRiP/GenomeStrip-2.00.1747-foss-2016b.eb
```diff
@@ -0,0 +1,69 @@
...
+sanity_check_paths = {
+    'files': ["lib/SVToolkit.jar"],
+    'dirs': ["lib/gatk", "lib/depend"],
+}
+
+modextravars = {
+    'SV_DIR': '$EBROOTGENOMESTRIP',
+}
+
+modloadmsg = {
```
boegel: hmm, this is a pretty large load message...

easybuild/easyconfigs/g/GenomeSTRiP/GenomeStrip-2.00.1747-foss-2016b.eb
```diff
@@ -0,0 +1,69 @@
...
+    ('R', '3.4.0'),
+]
+
+files_to_copy = [ 
+    'lib', 'bwa', 'conf', 'qscript', 'R', 'scripts', 'wdl', 'cookbook', 'installtest', 'misc', 'LICENSE.txt',
+]
+
+sanity_check_paths = {
+    'files': ["lib/SVToolkit.jar"],
+    'dirs': ["lib/gatk", "lib/depend"],
```
boegel: you're listing way more in `files_to_copy`, so extend this a bit (no need to list all, focus on the key ones)?

robqiao: @boegel only thing I can think to ignore maybe "bwa". This is really a bioinformatics pipeline, which contains many small tools and "cookbook" contains examples I would imagine to be useful. I can give a go to minimize and see what you reckon.

easybuild/easyconfigs/g/GenomeSTRiP/GenomeStrip-2.00.1747-foss-2016b.eb
robqiao: @boegel Yeh, this is a pipeline, so some instruction on how to use it properly would be useful I would think. Message is concatenated into a single line, I would rather them to have some kind of formatting, is there something we can do?

easybuild/easyconfigs/g/GenomeSTRiP/GenomeStrip-2.00.1747-foss-2016b.eb
boegel: I'm not suggesting to *install* less, just to be a bit more specific in the sanity check, e.g. by also listing `scripts` and a couple of others in `sanity_check_paths`

easybuild/easyconfigs/g/GenomeSTRiP/GenomeStrip-2.00.1747-foss-2016b.eb
boegel: The formatting isn't touched by EasyBuild, you're avoiding the use of newlines yourself by using `\` at the end of each line.
Also, you don't need to wrap the `"""..."""` string `{ ... }`.

So, something like this:

```python
modloadmsg = """First line
second line
third line
"""

It would try to avoid a load message that prints more than 5 lines.

migueldiascosta:
@robqiao, this PR is being closed for the following reason(s): no activity for > 1 year.

PR #4727, opened 2017-06-16
{devel}[GCCcore/5.4.0] CMake v3.7.1

vanzod:
(created using `eb --new-pr`)


boegel:
@vanzod What's this about? We don't modify existing easyconfigs, unless there's a strong technical reason for it...

This looks like a case for `--minimal-toolchains` with `--use-existing-modules` and `--add-dummy-to-minimal-toolchains`?

vanzod:
For some reason my local installation did not see the already present easyconfig and I pushed it without double checking. Friday evening mistakes. Sorry for the unnecessary noise.
Closing

PR #4749, opened 2017-06-20
fix source URLs for HDF5

boegel:
following the issue opened by @reedts on the download issues for HDF5 1.8.18, and contacting HDF5 support on this, I figured out that we're actually using the wrong source URLs for HDF5...

In https://support.hdfgroup.org/ftp/HDF5/releases, there are subdirectories for the 1.8.x and 1.10.x versions of HDF5, which in turn have a subdirectory per version.

For some/most of these there is also a symlink directly in https://support.hdfgroup.org/ftp/HDF5/releases, but this is apparently something they are considering to remove...

This PR fixes the HDF5 source URLs to use the subdirectories specific to 1.8.x and 1.10.x, and also aligns the homepage while at it...

### DIFF
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.10.0-patch1-foss-2016b.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.10.0-patch1-foss-2016b.eb
@@ -1,7 +1,7 @@
 name = 'HDF5'
 version = '1.10.0-patch1'
 
-homepage = 'http://www.hdfgroup.org/HDF5/'
+homepage = 'https://support.hdfgroup.org/HDF5/'
 description = """HDF5 is a unique technology suite that makes possible the management of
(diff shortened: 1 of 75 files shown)


vanzod:
lgtm

Tested `HDF5-1.8.17-foss-2016b.eb` and it correctly download source archive

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 75 out of 75 (75 easyconfigs in this PR)
node2137.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/639aae36a51b9386d9a0e9a154d786b2 for a full test report.

boegel:
Test report shows that downloads still work for all easyconfigs.

Thanks for the review @vanzod, going in.

PR #2506, opened 2016-02-17
enable automatic style checks in easyconfig tests

wpoely86:
Needs ~~hpcugent/easybuild-framework#1618~~

PR #4762, opened 2017-06-21
avoid use of import in ANSYS & MATLAB easyconfigs

boegel:
requires https://github.com/hpcugent/easybuild-easyblocks/pull/1195 to retain backward compatibility

(required to make #2506 eligible for merging)


boegel:
Tested on top of https://github.com/hpcugent/easybuild-easyblocks/pull/1195, works as intended.

PR #4763, opened 2017-06-21
set $OMP_NUM_THREADS during CP2K test step via dedicated easyconfig parameter

boegel:
requires https://github.com/hpcugent/easybuild-easyblocks/pull/1196

(required to make #2506 eligible for merging)


boegel:
Tested on top of https://github.com/hpcugent/easybuild-easyblocks/pull/1196, going in.

PR #4767, opened 2017-06-22
{devel} [foss/2017a] [intel/2017a] LikWid 4.2.0

Falkor:
Updated reciPY for Likwid on top of foss and intel 2017a
Signed-off-by: Sebastien Varrette <Sebastien.Varrette@uni.lu>

### DIFF
--- a/easybuild/easyconfigs/l/likwid/likwid-4.2.0-foss-2017a.eb
+++ b/easybuild/easyconfigs/l/likwid/likwid-4.2.0-foss-2017a.eb
@@ -1,0 +1,50 @@
+##
+# -*- mode: python; -*-
+# EasyBuild reciPY for LikWid -- see https://github.com/RRZE-HPC/likwid
+# as per https://github.com/hpcugent/easybuild
+#
+# Copyright:: Copyright 2017 UL HPC -- hpc.uni.lu
+# Authors::   UL HPC Team <hpc-sysadmins@uni.lu>
+# License::   MIT/GPL
+#
+#
+easyblock = 'ConfigureMake'
+
+name = 'likwid'
+version = '4.2.0'
+
+homepage = 'https://github.com/RRZE-HPC/likwid'
+description = """Likwid stands for 'Like I knew what I am doing'. Likwid is a
+simple to install and use toolsuite of command line applications for performance
+oriented programmers. It works for Intel and AMD processors on the Linux
+operating system.
+"""
+
+toolchain = {'name': 'foss', 'version': '2017a'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://github.com/RRZE-HPC/likwid/archive/']
+sources = ['%(version)s.tar.gz']
+
+patches = ['likwid-%(version)s-foss-config-mk.patch']
+
+builddependencies = [('binutils', '2.27')]
+
+skipsteps = ['configure']
+buildopts = 'CC="$CC" CFLAGS="$CFLAGS -std=c99" PREFIX=%(installdir)s'
+installopts = 'PREFIX=%(installdir)s'
+
+sanity_check_paths = {
+    'files': ["bin/likwid-memsweeper", "bin/likwid-mpirun", "bin/likwid-perfctr",
+              "bin/likwid-perfscope", "bin/likwid-pin", "bin/likwid-powermeter", "bin/likwid-topology",
+              "lib/liblikwidpin.%s" % SHLIB_EXT, "lib/liblikwid.%s" % SHLIB_EXT],
+    'dirs': ["man/man1"]
+}
+
+postinstallcmds = [
+    'echo "================================================================================\nIMPORTANT: you will have to manually apply the following changes **as root**:\n    sudo chown root:root \$EBROOTLIKWID/sbin/likwid-accessD\n    sudo chmod u+s \$EBROOTLIKWID/sbin/likwid-accessD\n ================================================================================"'
+]
+
+maxparallel = 1
+
+moduleclass = 'devel'
--- a/easybuild/easyconfigs/l/likwid/likwid-4.2.0-intel-2017a.eb
+++ b/easybuild/easyconfigs/l/likwid/likwid-4.2.0-intel-2017a.eb
@@ -1,0 +1,49 @@
+##
+# -*- mode: python; -*-
+# EasyBuild reciPY for LikWid -- see https://github.com/RRZE-HPC/likwid
+# as per https://github.com/hpcugent/easybuild
+#
+# Copyright:: Copyright 2017 UL HPC -- hpc.uni.lu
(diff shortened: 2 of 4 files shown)


easybuild/easyconfigs/l/likwid/likwid-4.2.0-foss-2017a.eb
```diff
@@ -0,0 +1,53 @@
...
+              "bin/likwid-perfscope", "bin/likwid-pin", "bin/likwid-powermeter", "bin/likwid-topology",
+              "lib/liblikwidpin.%s" % SHLIB_EXT, "lib/liblikwid.%s" % SHLIB_EXT],
+    'dirs': ["man/man1"]
+}
+
+postinstallcmds = [
+    'echo "================================================================================"',
+    'echo "IMPORTANT: you will have to manually apply the following changes **as root**:"',
+    'echo "   sudo chown root:root \$EBROOTLIKWID/sbin/likwid-accessD"',
+    'echo "   sudo chmod u+s \$EBROOTLIKWID/sbin/likwid-accessD"'
```
boegel: @Falkor should we put his on hold until we have proper support for https://github.com/hpcugent/easybuild-framework/issues/542?

easybuild/easyconfigs/l/likwid/likwid-4.2.0-intel-config-mk.patch
```diff
@@ -0,0 +1,49 @@
+--- likwid-likwid-4.2.orig/config.mk	2016-12-22 13:22:23.000000000 +0100
++++ likwid-likwid-4.2/config.mk	2017-06-22 10:14:28.585696000 +0200
+@@ -1,7 +1,7 @@
+ # Please have a look in INSTALL and the WIKI for details on
+ # configuration options setup steps.
+ # supported: GCC, CLANG, ICC, MIC (ICC), GCCX86 (for 32bit systems)
+-COMPILER = GCC#NO SPACE
++COMPILER = ICC#NO SPACE
```
boegel: hmm, do we really need a patch for this?

isn't using `COMPILER=ICC` in `buildopts` sufficient?

easybuild/easyconfigs/l/likwid/likwid-4.2.0-foss-2017a.eb
Falkor: as you wish but as soon as hpcugent/easybuild-framework#542 is solved, I'll update the recipes accordingly in all cases, so it could be the context of a second pool request...

easybuild/easyconfigs/l/likwid/likwid-4.2.0-intel-config-mk.patch
Falkor: That's the advised guidelines (from [INSTALL](https://github.com/RRZE-HPC/likwid/blob/master/INSTALL) as the `COMPILER` setting in `config.mk` permits to  `include_[GCC|CLANG|ICC|MIC].mk` which are required for correct compilation. 
I wouldn't risk to diverge from this approach. Unless you mean there is a way to reuse the foss patch and provide the appropriate `sed` command to be performed in addition to the patch yet _before_ the build ? That would be confusing for me.

boegel: It's a common misconception that you need to patch `Makefile`s (or the files they include) to make `make` use the right values.

Just running `make COMPILER=ICC` (which is equivalent to using `buildopts = "COMPILER=ICC"` in the easyconfig file) has the exact same effect of changing `GCC` to `ICC` like you're doing here...

If we can avoid maintaining patches while still doing the right thing, we should, it makes things a lot simpler (e.g. `--try-software-version` could be used reliably).

easybuild/easyconfigs/l/likwid/likwid-4.2.0-foss-2017a.eb
boegel: OK, I guess that shouldn't be a blocker then.

easybuild/easyconfigs/l/likwid/likwid-4.2.0-intel-config-mk.patch
Falkor: you're fully right, I updated and remove the patch files accordingly

easybuild/easyconfigs/l/likwid/likwid-4.2.0-intel-2017a.eb
```diff
@@ -0,0 +1,52 @@
...
+
+toolchain = {'name': 'intel', 'version': '2017a'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://github.com/RRZE-HPC/likwid/archive/']
+sources = ['%(version)s.tar.gz']
+
+patches = ['likwid-%(version)s-config-mk.patch']
+
+builddependencies = [('binutils', '2.27')]
```
boegel: this is not required when using the `intel` toolchain, since it will be loaded already

easybuild/easyconfigs/l/likwid/likwid-4.2.0-foss-2017a.eb
```diff
@@ -0,0 +1,53 @@
...
+
+toolchain = {'name': 'foss', 'version': '2017a'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://github.com/RRZE-HPC/likwid/archive/']
+sources = ['%(version)s.tar.gz']
+
+patches = ['likwid-%(version)s-config-mk.patch']
+
+builddependencies = [('binutils', '2.27')]
```
boegel: this is not required when using the `intel` toolchain, since it will be loaded already

boegel:
Test report by @boegel
**SUCCESS**
node2404.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/67c1cc8338297750f785057ab6e11b80 for a full test report.

boegel:
Test report by @boegel
**FAILED**
gligar01.gligar.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/cb56c9f9ccda0b6342540737a7633cf6 for a full test report.

boegel:
@Falkor ignore the failed test report, our Intel license server was briefly down, I'll submit a new report

lgtm

boegel:
Test report by @boegel
**SUCCESS**
node843.raichu.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/4a5b00aed310eeb1db1e83657dace86e for a full test report.

boegel:
Going in, thanks @Falkor!

PR #4774, opened 2017-06-23
stop using deprecated 2-element tuple format in sources, use equivalent dict format instead

boegel:
cfr. http://easybuild.readthedocs.io/en/develop/Deprecated-functionality.html#depr-sources-2-element-tuple

requires ~~https://github.com/hpcugent/easybuild-framework/pull/2223~~

### DIFF
--- a/easybuild/easyconfigs/v/VarScan/VarScan-2.3.6-Java-1.7.0_80.eb
+++ b/easybuild/easyconfigs/v/VarScan/VarScan-2.3.6-Java-1.7.0_80.eb
@@ -13,7 +13,7 @@ description = """Variant calling and somatic mutation/CNV detection for next-gen
 
 toolchain = {'name': 'dummy', 'version': ''}
 
-sources = [('%(name)s.v%(version)s.jar', 'download')]
+sources = ['%(name)s.v%(version)s.jar']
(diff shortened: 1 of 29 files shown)


boegel:
@vanzod please review?

easybuild/easyconfigs/v/VarScan/VarScan-2.3.6-Java-1.7.0_80.eb
```diff
@@ -13,7 +13,7 @@ description = """Variant calling and somatic mutation/CNV detection for next-gen
 
 toolchain = {'name': 'dummy', 'version': ''}
 
-sources = [('%(name)s.v%(version)s.jar', 'download')]
+sources = ['%(name)s.v%(version)s.jar']
```
wpoely86: what was this supposed to do?

boegel: Good question, I was wondering the exact same thing when I ran into it...

It's a bug, this syntax was used to specify a custom extract command, but since the JAR easyblock just skips the extract step this went unnoticed.

boegel:

PR #4802, opened 2017-07-03
New EasyConfig for GLib 2.52.0 with the gimkl-2017a toolchain (plus dependency)

valtandor:


boegel:
@valtandor Is the missing easyconfig for `Python` provide via another PR?

valtandor:
Ah...On our system this is built against a "bare" Python, but I recalled what you said about not wanting specially decorated versions in the dependencies or builddependencies lists. So I took out the "-bare" suffix, and forgot that the resulting Python doesn't exist.

I will supply a Python easyconfig for gimkl-2017a, but haven't got to it quite yet.

migueldiascosta:
@valtandor is this still relevant?

boegel:
I'm closing this, since it's no longer worth the effort to get this merged given the old toolchains.

@valtandor My apologies for not following up on this closer at the time...

PR #4814, opened 2017-07-05
new OpenFOAM-1612, OpenFOAM-1706 versions

olesenm:
edit: requires ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1205~~

### DIFF
--- a/easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-cleanup.patch
+++ b/easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-cleanup.patch
@@ -1,0 +1,409 @@
+# - Replace the OpenFOAM third-party libraries with EASYBUILD variables.
+# - Set install dir, compiler, etc using EASYBUILD
+# <Mark.Olesen@esi-group.com>
+--- OpenFOAM-v1706.orig/applications/utilities/mesh/manipulation/setSet/Allwmake	2017-06-30 10:50:14.000000000 +0200
++++ OpenFOAM-v1706/applications/utilities/mesh/manipulation/setSet/Allwmake	2017-07-05 13:36:32.826823741 +0200
+@@ -9,11 +9,11 @@
+ #
+ # use readline if available
+ #
+-if [ -f /usr/include/readline/readline.h ]
++if [ -f $EBROOTLIBREADLINE/include/readline/readline.h ]
+ then
+     echo "    found <readline/readline.h>  --  enabling readline support."
+     export COMP_FLAGS="-DHAS_READLINE"
+-    export LINK_FLAGS="-lreadline"
++    export LINK_FLAGS="-L$EBROOTLIBREADLINE/lib -lreadline -L$EBROOTNCURSES -lncurses"
+ fi
+ 
+ wmake $targetType
+--- OpenFOAM-v1706.orig/src/parallel/decompose/Allwmake	2017-06-27 15:04:26.213523686 +0200
++++ OpenFOAM-v1706/src/parallel/decompose/Allwmake	2017-07-11 11:42:41.467097531 +0200
+@@ -39,7 +39,9 @@
+ 
+     # Library
+     [ -r $FOAM_EXT_LIBBIN/libmetis.so ] || \
++    [ -r $METIS_ARCH_PATH/lib/libmetis.a ] || \
+     [ -r $METIS_ARCH_PATH/lib/libmetis.so ] || \
++    [ -r $METIS_ARCH_PATH/lib$WM_COMPILER_LIB_ARCH/libmetis.a ] || \
+     [ -r $METIS_ARCH_PATH/lib$WM_COMPILER_LIB_ARCH/libmetis.so ] || \
+     [ "${METIS_ARCH_PATH##*-}" = system ] || {
+         echo "$warning (missing library)"
+@@ -94,7 +96,9 @@
+ 
+     # Library
+     [ -r $FOAM_EXT_LIBBIN/libscotch.so ] || \
++    [ -r $SCOTCH_ARCH_PATH/lib/libscotch.a ] || \
+     [ -r $SCOTCH_ARCH_PATH/lib/libscotch.so ] || \
++    [ -r $SCOTCH_ARCH_PATH/lib$WM_COMPILER_LIB_ARCH/libscotch.a ] || \
+     [ -r $SCOTCH_ARCH_PATH/lib$WM_COMPILER_LIB_ARCH/libscotch.so ] || \
+     [ "${SCOTCH_ARCH_PATH##*-}" = system ] || {
+         echo "$warning (missing library)"
+--- OpenFOAM-v1706.orig/src/parallel/decompose/ptscotchDecomp/Make/options	2017-06-30 10:50:14.000000000 +0200
++++ OpenFOAM-v1706/src/parallel/decompose/ptscotchDecomp/Make/options	2017-07-05 12:23:10.494399931 +0200
+@@ -5,7 +5,6 @@
+     $(PFLAGS) $(PINC) \
+     -I$(SCOTCH_ARCH_PATH)/include/$(FOAM_MPI) \
+     -I$(SCOTCH_ARCH_PATH)/include \
+-    -I/usr/include/scotch \
+     -I../decompositionMethods/lnInclude
+ 
+ /*
+--- OpenFOAM-v1706.orig/src/parallel/decompose/scotchDecomp/Make/options	2017-06-30 10:50:14.000000000 +0200
++++ OpenFOAM-v1706/src/parallel/decompose/scotchDecomp/Make/options	2017-07-05 12:24:29.590283915 +0200
+@@ -8,7 +8,6 @@
+ EXE_INC = \
+     $(PFLAGS) $(PINC) \
+     -I$(SCOTCH_ARCH_PATH)/include \
+-    -I/usr/include/scotch \
+     -I../decompositionMethods/lnInclude
+ 
+ /*
+--- OpenFOAM-v1706.orig/etc/bashrc	2017-06-30 10:50:14.000000000 +0200
++++ OpenFOAM-v1706/etc/bashrc	2017-07-11 09:17:28.319350909 +0200
+@@ -43,9 +43,10 @@
+ # - If this does not produce the desired result, please set one of the fallback
+ #   values to an appropriate path.
+ #
+-rc="${BASH_SOURCE:-${ZSH_NAME:+$0}}"
+-[ -n "$rc" ] && FOAM_INST_DIR=$(\cd $(dirname $rc)/../.. && \pwd -L) || \
+-FOAM_INST_DIR=$HOME/$WM_PROJECT
++##EasyBuild## module sets FOAM_INST_DIR
++# rc="${BASH_SOURCE:-${ZSH_NAME:+$0}}"
++# [ -n "$rc" ] && FOAM_INST_DIR=$(\cd $(dirname $rc)/../.. && \pwd -L) || \
++# FOAM_INST_DIR=$HOME/$WM_PROJECT
+ # FOAM_INST_DIR=/opt/$WM_PROJECT
+ # FOAM_INST_DIR=/usr/local/$WM_PROJECT
+ #
+@@ -90,7 +91,7 @@
+ #- MPI implementation:
+ #    WM_MPLIB = SYSTEMOPENMPI | OPENMPI | SYSTEMMPI | MPICH | MPICH-GM | HPMPI
+ #               | MPI | FJMPI | QSMPI | SGIMPI | INTELMPI | USERMPI
+-export WM_MPLIB=SYSTEMOPENMPI
++export WM_MPLIB=EASYBUILDMPI
+ 
+ #- Operating System:
+ #    WM_OSTYPE = POSIX
+@@ -119,7 +120,7 @@
+ 
+ # Location of third-party software
+ # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+-export WM_THIRD_PARTY_DIR=$WM_PROJECT_INST_DIR/ThirdParty-$WM_PROJECT_VERSION
++export WM_THIRD_PARTY_DIR=$WM_PROJECT_DIR/ThirdParty  ##Easybuild## none
+ 
+ # Location of site-specific templates etc
+ # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+--- OpenFOAM-v1706.orig/etc/config.sh/CGAL	2017-06-30 10:50:14.000000000 +0200
++++ OpenFOAM-v1706/etc/config.sh/CGAL	2017-07-05 12:43:36.924551892 +0200
+@@ -50,41 +50,9 @@
+ #------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


olesenm:
Now adjusted package names too.

olesenm:
Hello @easybuilders/easybuild-easyconfigs-maintainers - review requested

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1612+-foss-2016a.eb
```diff
@@ -0,0 +1,46 @@
...
+]
+
+sources = [ 'OpenFOAM-%(version)s.tgz' ]
+checksums = ['ca02c491369150ab127cbb88ec60fbdf']
+
+patches = [ 'OpenFOAM-%(version)s-cleanup.patch' ]
+
+dependencies = [
+    # > causes bash issue? ('libreadline', '6.3'),
+    # > causes bash issue? ('ncurses', '6.0'),
```
wpoely86: please explain?

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-cleanup.patch
```diff
@@ -0,0 +1,406 @@
+--- OpenFOAM-v1706.orig/applications/utilities/mesh/manipulation/setSet/Allwmake	2017-06-30 10:50:14.000000000 +0200
```
wpoely86: it's good practice to claim authorship of the patches and explain what they are supposed to do.

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016a.eb
```diff
@@ -0,0 +1,44 @@
...
+version = 'v1706'
+
+homepage = 'http://www.openfoam.com/'
+description = """OpenFOAM is a GPL-opensource C++ CFD-toolbox.
+This offering is not approved or endorsed by OpenCFD Ltd,
+producer and distributor of the OpenFOAM software via www.openfoam.com,
+and owner of the OPENFOAM trademark.
+"""
+
+toolchain = {'name': 'foss', 'version': '2016a'}
```
wpoely86: any reason why you add them for both 2016a and 2016b? Is there any difference?

olesenm: Honestly, I have no idea. I'm very new to easybuild but I understood that the configs need to specify a toolchain. The 2016b toolchang uses updated gcc etc.

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-cleanup.patch
olesenm: Where do the explanation go? I've seen some patches with `#` comments, but this seems to be fairly non-standard.

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1612+-foss-2016a.eb
olesenm: I experienced an issue where the LD_LIBRARY_PATH pointed to the easybuild libreadline, which provoked symbol conflicts for my bash.

olesenm:
What is still needed to get this restarted? @easybuilders/easybuild-easyconfigs-maintainers

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1612+-cleanup.patch
```diff
@@ -0,0 +1,479 @@
+--- OpenFOAM-v1612+.orig/applications/utilities/mesh/manipulation/setSet/Allwmake	2016-12-23 15:22:59.000000000 +0100
```
boegel: please include a brief description of the patch above this line and mention the author

any lines above this line are comments

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-cleanup.patch
boegel: see my comment in the other patch

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016a.eb
boegel: There's little point in providing easyconfigs for an older toolchain like `foss/2016a` imho.

If people really need to install OpenFOAM with it, they can use `--try-toolchain`.

Generally speaking, it's better to stick to more recent toolchains, so I'd prefer just having easyconfigs for `foss/2016b` here.

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1612+-foss-2016a.eb
boegel: Can you elaborate on that a bit, exact error messages, etc.?

These dependencies should not be commented out, since otherwise we're at the mercy of whatever the OS provides (or doesn't), which is something we're trying to avoid.

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
```diff
@@ -0,0 +1,44 @@
...
+toolchain = {'name': 'foss', 'version': '2016b'}
+toolchainopts = {'cstd': 'c++11'}
+
+source_urls = [
+    ('https://sourceforge.net/projects/openfoamplus/files/%(version)s',
+     'download')
+]
+
+sources = [ 'OpenFOAM-%(version)s.tgz' ]
+checksums = ['630d30770f7b54d6809efbf94b7d7c8f']
```
boegel: please use `SHA256` checksums (see also https://github.com/easybuilders/easybuild-easyconfigs/pull/5005)

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
```diff
@@ -0,0 +1,44 @@
...
+    # > causes bash issue? ('libreadline', '6.3'),
+    # > causes bash issue? ('ncurses', '6.0'),
+    # OpenFOAM requires 64 bit METIS using 32 bit indexes (array indexes)
+    ('METIS', '5.1.0'),
+    ('SCOTCH', '6.0.4'),
+    ('Boost', '1.61.0'),
+    # Could build with CGAL, but this adds an extra python dependency
+    # that we don't want/need
+    # ('CGAL', '1.60.0'),
+    # ('ParaView', '5.4.0', '-mpi'),
```
boegel: We usually install OpenFOAM with CGAL & ParaView included as dependencies, so I'd prefer to stick to that.

What's the problem with the extra (indirect) Python dependency?

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
```diff
@@ -0,0 +1,44 @@
...
+and owner of the OPENFOAM trademark.
+"""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+toolchainopts = {'cstd': 'c++11'}
+
+source_urls = [
+    ('https://sourceforge.net/projects/openfoamplus/files/%(version)s',
+     'download')
+]
```
boegel: style nitpicking: let's just make this a single line? I don't see the need to spread this over 4 lines...

```python
source_urls = [('https://sourceforge.net/projects/openfoamplus/files/%(version)s', 'download')]
```

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
```diff
@@ -0,0 +1,44 @@
...
+
+patches = [ 'OpenFOAM-%(version)s-cleanup.patch' ]
+
+dependencies = [
+    # > causes bash issue? ('libreadline', '6.3'),
+    # > causes bash issue? ('ncurses', '6.0'),
+    # OpenFOAM requires 64 bit METIS using 32 bit indexes (array indexes)
+    ('METIS', '5.1.0'),
+    ('SCOTCH', '6.0.4'),
+    ('Boost', '1.61.0'),
```
boegel: shouldn't this be bumped to a more recent version of Boost (e.g. 1.64.0)?

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
```diff
@@ -0,0 +1,44 @@
...
+    # Could build with CGAL, but this adds an extra python dependency
+    # that we don't want/need
+    # ('CGAL', '1.60.0'),
+    # ('ParaView', '5.4.0', '-mpi'),
+    # Part of foss-2016b -> ('FFTW', '3.3.5'),
+]
+
+builddependencies = [
+    ('CMake', '3.6.1'),
```
boegel: this can be bumped to 3.8.2, there's an easyconfig available already for it

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
```diff
@@ -0,0 +1,44 @@
...
+    ('flex', '2.6.0'),
```
boegel: this can be bumped to 2.6.4

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
olesenm: OK. On spack they run flake8 tests that complain when lines exceed 80 chars.

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
olesenm: If I just want OpenFOAM, why a huge slew of python things? With paraview it would likely be good to distinguish between an off-screen and hardware mode too. However, I didn't want to get into this. All I wanted was to get a new version of OpenFOAM building with easybuild.

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
olesenm: probably a holdover from before. I think that 2.6.1-2.6.3 were broken (weird namespaces if I remember).

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
boegel: We have automated style checks in place too now, but allow lines up to 120 chars; too much silly splits (like this), we're not in the 80s anymore so 120 chars makes more sense imho ;)

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
boegel: It's OK if you're not up for looking into this, it's just that our previous OpenFOAM easyconfigs do include CGAL & ParaView, so this is going to surprise people potentially...

easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-v1706-foss-2016b.eb
boegel: OK, but I haven't seen any issues with 2.6.4 as far as I can recall, so let's bump it?

akesandgren:
There is a bit of problem with OpenFOAM since there are (at least) 2 different distributions, openfoam.com and openfoam.org. And the version numbering differs.

Does anyone have any good insights on this subject?

Will the easyblock work correctly with both distributions?

boegel:
@akesandgren Some updates are done in https://github.com/easybuilders/easybuild-easyblocks/pull/1205, but for now it seems like the OpenFOAM easyblock is able to copy with both flavours (+ OpenFOAM-Extend as well)...

boegel:
@olesenm Are you up for following up on this, or do you prefer that we take it from here?

olesenm:
I'm fine with whatever works best for you (ie, please take over if possible). In a longer term I think that the various flavour/fork check should really only been done once, but that is just a style issue. At least for openfoam, I think that the changes to its config files would be better to do directly (if possible) rather that using patching to modify things. I've used this with reasonably good success in spack. When building openfoam-1706 with spack, I only use patch to adjust two entries in a single file and the rest of the configuration is simply emitted directly.

olesenm:
I've picked this up again (would be nice to get it past the line before the 1712 release). However, I'm currently going crazy with all sorts of foss-2016b checksum errors. The tarfiles in question (Autoconf, libtool, M4, numactl) all have sha256sums that match what their easybuild files require, but I always get errors. Must be some flags somewhere I guess.

boegel:
@olesenm Can you share which exact errors you are getting, and how you're triggering them? Can you also share your EasyBuild configuration (output of `eb --show-config` should suffice).

olesenm:
Here is a rebuild log of one of the offenders (m4):
[easybuild-oIFDJq.log](https://github.com/easybuilders/easybuild-easyconfigs/files/1554488/easybuild-oIFDJq.log)

A manual verification of the check sums from the command line

| checksum | result
|---------------|------------
| md5sum      | a5e9954b1dae036762f7b13673a2cf76   /home/mol/.local/easybuild/sources/m/M4/m4-1.4.17.tar.gz
| sha256sum   | 3ce725133ee552b8b4baca7837fb772940b25e81b2a9dc92537aeaf733538c9e  /home/mol/.local/easybuild/sources/m/M4/m4-1.4.17.tar.gz

So the sha256sum looks correct, but it would appear that verify_checksum still doesn't like it.

The easybuild config (`eb --show-config`):

    #
    # Current EasyBuild configuration
    # (C: command line argument, D: default value, E: environment variable, F: configuration file)
    #
    buildpath      (D) = /home/mol/.local/easybuild/build
    installpath    (D) = /home/mol/.local/easybuild
    module-syntax  (E) = Tcl
    modules-tool   (E) = EnvironmentModulesC
    repositorypath (D) = /home/mol/.local/easybuild/ebfiles_repo
    robot-paths    (D) = /home/mol/Software/easybuild/easybuild-easyconfigs/easybuild/easyconfigs
    sourcepath     (D) = /home/mol/.local/easybuild/sources


The easyconfig is develop at 4a6192723f378ca27a4f5d7c338596ecf07e6762
The easyblocks is develop at 96efeb6eda2fd6c1abcdb596fdcc75d833ab5c4e.

olesenm:
I think that this content has probably been merged, so I'll close it and see it that is true.

boegel:
@olesenm We have only merged easyconfigs for OpenFOAM v1712, not for the older versions you were adding here, but that's good enough going forward?

olesenm:
Sounds good to me.

PR #4817, opened 2017-07-05
adding Anaconda2 with GCC-5.4.0

shahzebsiddiqui:


vanzod:
@shahzebsiddiqui Since Anaconda is distributed as binaries, what is the reason to have it at the `GCC` toolchain level since it is already available at the `dummy` level?

shahzebsiddiqui:
Sorry this shouldn't have been committed, this is my custom GCC-5.4.0-2.27. I agree with @vanzod should be build with dummy toolchain. Initially thought Anaconda2 should be treated like Python but I presume that is not the case.

PR #4826, opened 2017-07-06
Eigen  3.2.8, 3.3.3 for GCCcore-5.4.0

shahzebsiddiqui:


boegel:
@shahzebsiddiqui `Eigen` is only header files, so it doesn't make sense to use a non-dummy toolchain for this, nothing is getting compiled?

PR #4839, opened 2017-07-09
Revert "notify EasyBuild maintainers of failing tests"

boegel:
This reverts commit 953a508d0f013c50dc8b9eebcadcfa0e65fcab95.

Turns out this also results in mails for every branch that gets created, which is definitely *not* what we want...

I don't see a better alternative now, I don't think we can limit this to only the central `develop` and `master` branches...


boegel:
Seems like we're not the only ones who want this, but it's not supported by Travis yet, see https://github.com/travis-ci/travis-ci/issues/1405...

Maybe I can cook something up via @boegelbot sometime...

PR #2265, opened 2016-01-06
{bio}[foss/2015b] libff1 3.2.1, gettext 0.19.7, GLib 2.46.2, Python 2.7.9, Exonerate 2.4.0 (REVIEW)

nathanhaigh:
Working on Easyconfigs for Exonerate-2.4.0-foss-2015b

PR #4840, opened 2017-07-09
{lib}[GCC/6.4.0-2.28,GCCcore/6.4.0] libsigsegv v2.11, libsigsegv v2.11

JackPerdue:
(created using `eb --new-pr`)
ligsigsegv - GCC or GCCcore if I want to use with intel/2017b??


JackPerdue:
In the past, there seems there was an issue with --minimal-toolchains using GCC (vs GCCcore) when building with the "intel" toolchain.  I stuck with GCCcore in our last 2017A (based on 2017a) toolchains.  I don't care either way (though the GCC one is easier since I don't have to include the binutils).  As long as it will be found by intel/iomkl./foss toolchains (I don't need a version for each).

Note, we have a flat module naming system here (not hierarchical) so am not sure of the overall impact.

What is the preference between the two that allow me to reuse across all toolchains?

boegel:
@JackPerdue There's a known issue when easyconfigs with `GCC` & `GCCcore` are being mixed, see https://github.com/easybuilders/easybuild-framework/issues/2265.

`GCC` is not a subtoolchain for `intel`, but `GCCcore` is, so if you want to share installations between `foss` & `intel`, you should go with `GCCcore`.

I do agree that having to include `binutils` as a build dep is a nuisance though, so we can consider changing this (in a backwards-compatible way)...

JackPerdue:
<from Slack>

boegel 3:03 PM hiya siliconslick!

siliconslick 3:09 PM howdy!   Any guidance to offer on EC PR#4840?   i.e. GCC vs GCCcore for use across toolchains

boegel 3:26 PM @siliconslick Haven't had time to look into that one yet. Maybe related to the issue @vanzod opened though: https://github.com/easybuilders/easybuild-framework/issues/2265
GitHub
Robot picks same dependency from different builds · Issue #2265 · easybuilders/easybuild-framework
When minimal toolchains are not enforced, it is possible that robot picks two different builds of the same dependency from two different toolchain levels and Lmod complains as in [1]. In this speci...
siliconslick 3:40 PM OK.... just wanted to pick one of the other before going too much further (I got more for 6.4.0....)

boegel 4:06 PM @siliconslick reviewed your PR #4840

vanzod 4:09 PM @siliconslick I always operate under the paradigm of building with the lowest possible toolchain. The only exception to this rule is when building math libraries (e.g. NLopt). In such cases, although I could build them under GCCcore, I prefer to build them with the GCC and iccifort toolchains in order to take advantage of the different level of optimization that a compiler can provide wrt the other.
4:10 This approach may be overkill and if others disagree, I gladly accept comments.

siliconslick 4:59 PMI guess thats for the sake of intel to make sure it finds the iccifort one and not reach all the way down to GCCcore... I would think the optimizations of GCC and GCCcore (no intel) would be same

vanzod 5:31 PM @siliconslick Well, if you use --minimal-toolchain if you have both iccifort and GCCcore, it will pick the latter. So it is important to have only the GCC and iccfort available

siliconslick 5:32 PM That complicates things... instead of a nice simple "easy" heuristic like "use GCCcore unless you need MPI or linear algebra(MKL)" tosomething like...."well, depending whether the package would benefit from Intel optimiazations you might want to use GCC instead (to cooperate with intel toolchain)"..... then there is the question of what packages to go which way with

5:34 If I submit an easyconfig using GCCcore do you have a way later to ignore the easyconfigs from EB in your robot path?   e.g. only consider built tools on the system so far (and not EB)
siliconslick

5:40 PM e.g. will --use-existing-modules help  in those cases if there is a GCCcore version in the EB repo?   I know I've been confused by that option in the past so am not sure.

5:41 e.g. I commit GCCcore for now but if you want to come along later and add the GCC ver and then only build that one on your system (???)

</end Slack>

boegel:
@JackPerdue With `--use-existing-modules` enabled, the modules you have installed will determine which subtoolchain is used. If no modules are available, `eb` will fall back to picking a subtoolchain based on the available easyconfig files. See also http://easybuild.readthedocs.io/en/latest/Manipulating_dependencies.html?#using-minimal-toolchains-for-dependencies

@vanzod's point of sometimes preferring `iccifort` & `GCC` over `GCCcore` for performance reasons makes sense, the Intel compilers are known to provide significantly better optimised binaries...

vanzod:
@JackPerdue If you delete one of the two files, I can proceed with the merging process.

easybuild/easyconfigs/l/libsigsegv/libsigsegv-2.11-GCCcore-6.4.0.eb
```diff
@@ -0,0 +1,28 @@
...
+easyblock = 'ConfigureMake'
+
+name = 'libsigsegv'
+version = '2.11'
+
+homepage = 'https://www.gnu.org/software/libsigsegv/'
+
+description = """
+ GNU libsigsegv is a library for handling page faults in user mode.
+"""
```
vanzod: Please merge description in a single line.

easybuild/easyconfigs/l/libsigsegv/libsigsegv-2.11-GCCcore-6.4.0.eb
```diff
@@ -0,0 +1,28 @@
...
+
+toolchain = {'name': 'GCCcore', 'version': '6.4.0'}
+
+source_urls = [GNU_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
```
vanzod: Please add checksum

vanzod:
Test report by @vanzod
**SUCCESS**
cermis - Linux debian 8.8, Intel(R) Xeon(R) CPU           E5620  @ 2.40GHz, Python 2.7.9
See https://gist.github.com/6f26442f888901a55c21298a23e43972 for a full test report.

vanzod:
Going in, thanks @JackPerdue!

PR #4412, opened 2017-03-29
{devel}[foss/2016b] TensorFlow v1.0.1, SWIG v3.0.10, wheel v0.30.0a0, ...

gppezzi:
(created using `eb --new-pr`)

PR #4871, opened 2017-07-12
{devel}[foss/2016b] Tensorflow v1.2.1, protobuf-python v3.3.0, protobuf v3.3.0

schiotz:
(created using `eb --new-pr`)


schiotz:
This updates to the latest Tensorflow, but also uses the FOSS toolchain instead of the Intel toolchain that the other Tensorflow ebuilds use (since that is what I need).

@boegel Two questions:

1. The toolchain is not really used, it appears to be specified just to select the right version of the prerequisites (Python and protobuf).  It this the right way to do it?

2.  I am also making an easyconfig that installs tensorflow with GPU support.  In pip, these are two different packages (tensorflow and tensorflow-gpu), although both packages install a Python module named tensorflow.  Should I make this a separate packages (i.e. Tensorflow-gpu), or should it be easyconfigs in the same package but with different suffix (and in that case, what suffix?  -CUDA-version?).

/Jakob

verdurin:
Test report by @verdurin
**SUCCESS**
ca024.camp.thecrick.org - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2640 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/f39dc92168dbb58754db6b652eea5245 for a full test report.

boegel:
@schiotz 

1. It makes sense to use `foss` as toolchain. For now, it indeed only really matters for the dependencies, but we have plans to switch to building Tensorflow from source in the future (see comments in #4412)

2. I think a simple `-gpu` as version suffix suffices. Also mentioning the CUDA version would needlessly confuse users imho. We can include the CUDA version to discriminate other Tensorflow installations with the same version/toolchain on top of a another CUDA version later, if the needs arises.

easybuild/easyconfigs/p/protobuf-python/protobuf-python-3.3.0-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,29 @@
...
+version = '3.3.0'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://github.com/google/protobuf/'
+description = """Python Protocol Buffers runtime library."""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['https://github.com/google/protobuf/archive/v%(version)s']
+sources = [SOURCE_TAR_GZ]
```
verdurin: Please add an SHA256 checksum

easybuild/easyconfigs/p/protobuf/protobuf-3.3.0-foss-2016b.eb
```diff
@@ -0,0 +1,23 @@
...
+name = 'protobuf'
+version = '3.3.0'
+
+homepage = 'https://github.com/google/protobuf/'
+description = """Google Protocol Buffers"""
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+source_urls = ['https://github.com/google/protobuf/archive/v%(version)s/']
+sources = [SOURCE_TAR_GZ]
```
verdurin: Please add a checksum here too

easybuild/easyconfigs/t/Tensorflow/Tensorflow-1.2.1-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,57 @@
...
+
+dependencies = [
+    ('Python', pyver),
+    ('protobuf-python', '3.3.0', versionsuffix),
+]
+
+exts_list = [
+    ('backports.weakref', '1.0rc1', {
+        'modulename': 'backports.weakref',
+        'source_urls': ['https://pypi.python.org/packages/source/b/backports.weakref'],
```
verdurin: Please add checksums for these extensions, too. Thanks.

verdurin:
Test report by @verdurin
**SUCCESS**
ca004.camp.thecrick.org - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2640 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/6c5e35a01a9e30bcf210fd01316219cf for a full test report.

schiotz:
@verdurin I have added the requested checksums and will also add them to PR #4904 (where relevant).

easybuild/easyconfigs/t/Tensorflow/Tensorflow-1.2.1-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,61 @@
...
+toolchain = {'name': 'foss', 'version': '2016b'}
+
+# this is a bundle of Python packages
+exts_defaultclass = 'PythonPackage'
+
+pyver = '3.5.2'
+pyshortver = '35'
+
+dependencies = [
+    ('Python', pyver),
```
boegel: `pyver` is only used in one place, so you can replace this with

```python
('Python', '3.5.2'),
```

and drop the `pyver = ...` line

you'll need to keep the `pyshortver` though, which is why I took a different approach in existing Tensorflow easyconfigs, since now you're risking to forget syncing the `35` with the `3.5.2`

That risk is avoided entirely with an approach like this:

```python
pymaj = '3'
pymin = '6'
pyshortver = pymaj + pymin

dependencies = [
    ('Python', '%s.%s.1' % (pymaj, pymin)),
```

See https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/t/Tensorflow/Tensorflow-1.2.0-intel-2017a-Python-3.6.1.eb

schiotz: @boegel 
Sorry!  Was away most of August, and am currently playing catch-up with a deadline Friday.  I will return to this and my other pull requests next week.

schiotz: @boegel 
In this case I am pretty sure I would forget to update the third digit in the Python version.  In fact, I already did when I copied from the 3.6.1 easybuild. :-)

What about this solution:
```
pyver = '3.5.2'
pyshortver = ''.join(pyver.split('.')[:2])
```

Feel free to veto it if it is too obscure!

boegel: @schiotz makes total sense, let's go with this!

schiotz:
Hi @boegel and @verdurin ,

I was just about to upload the updated Tensorflow EasyBuild, when I did a last test-build.  For unrelated reasons, I had removed the folder where eb keeps the sources, so they were downloaded again.  The protobuf sources have a new checksum!  All the files in the tar archive are identical, but the tar file itself differs!  I can only assume that it is being autogenerated and that a cache has been flushed.  How do we handle something like this?  I could just update the SHA256, but I have to assume this will happen again.

schiotz:
Hi @boegel and @verdurin ,

A follow-up:  I found this discussion of the problem https://github.com/google/protobuf/issues/3619

The really worrying quotes:

> I am seeing this as well. It appears that github's archives' tgz headers changed in a way that leaves the unpacked files identical, but the the archive block headers have changed. Something in the tar part, not the gz part, related to how >100-character file paths are represented.

> This is pervasive for many tagged source releases, not just 3.3.x etc., and is pervasive for other projects hosted on github as well.

and

> FYI, it's not specific to this particular repository. It's Github who is making this change.

Ouch! So this is going to be a general problem for many of your EasyBuilds.  It also means that it looks like a change occurring only once, so it makes sense that I update the SHA256 and update this PR.

boegel:
@schiotz Ugh, that sucks indeed...

I guess the only thing we can do is to update the SHA256 checksum though, and hope this doesn't happen again.

We should probably report this to the EasyBuild mailing list as well, before people go blame EasyBuild that the checksums it includes are wrong.

Bad @github, bad...

schiotz:
@boegel 
Apparently github does not guarantee stable SHA checksums on tgz files.  It looks like google is going to begin hosting the tgz files themselves to avoid this problem.

schiotz:
@boegel 
Updated!

If the changes are hitting many .eb files, might it be a good idea to release a 3.4.1?

boegel:
@schiotz It's only a problem for the generated tarballs, as far as I can tell.

For protobuf specifically, we can probably use https://github.com/google/protobuf/releases/download/v3.4.1/protobuf-cpp-3.4.1.tar.gz instead.

That's not a generated tarball, but an actual packaged release, so that shouldn't change over time, and it's what we need here I think.

In the long run, we should avoid downloading the generated tarball from GitHub for any projects, since those aren't guaranteed to have the same checksum over time. That's going to be a problem for some project though, where no 'packaged' tarball is available.

Basically anything that uses `github.com/<account>/<project>/archive` as source URL is going to be affected by this...

boegel:
@schiotz I was planning to release EasyBuild v3.4.1 already, but fixing this for everything is going to be a problem short term:

```
$ grep -R "github.com/.*/archive" easybuild/easyconfigs | grep -v __archive__ | wc -l
     578
```

boegel:
@schiotz I've opened a dedicated issue on this: https://github.com/easybuilders/easybuild-easyconfigs/issues/5151

boegel:
@schiotz Do you think using `https://github.com/google/protobuf/releases/download/v3.4.1/protobuf-cpp-3.4.1.tar.gz` instead makes sense for `protobuf`?

schiotz:
@boegel I am not sure.  The contents is different (even comparing 3.3.0 versions to each other), the -cpp version contains fewer files, and a handful of source files are different.

akesandgren:
I took a look at the diff between the source v3.3.0.tar.gz and the protobuf-cpp-3.3.0.tar.gz.

The only difference i can see is that protobuf-cpp-3.3.0.tar.gz contains a post-autoconfigure set of files
And that the pure source stuff contains some objectiveC things.

So my guess is the the protobuf-cpp-<version>.tar.gz is usable as a replacement of the source v3.3.0.tar.gz

akesandgren:
PS.

I need Tensorflow 1.3 asap... so i'd really like this to get merged.

Do you want me to test anything (apart from the obvious --from-pr)

akesandgren:
Well the --from-pr fails due to bad checksum in protobuf i think it was. (Log files already gone)

schiotz:
@akesandgren If you have cached the old .tgz file, then it will fail with the new easyconfig file, since you still have a file with the old checksum.  Try deleting `$EASYBUILD_PREFIX/sources/p/protobuf*.tgz`

akesandgren:
Crap. I was 100% sure i didn't have that version and didn't check... *RED FACE*

akesandgren:
Tensorflow 1.3.0 GPU version build based on this PR worked nicely.
Though they seem to have moved the URL for tensorflow.

boegel:
Test report by @boegel
**SUCCESS**
node2026.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/66a5f5b0070b11cba7f338f9d97d24a4 for a full test report.

boegel:
I suggest we reconsider switching to the 'stable' `protobuf-cpp-*.tar.gz` sources in a separate issue/PR, maybe after consulting the protobuf developers (or documentation).

Are you up for looking into that @schiotz?

boegel:
Going in, thanks @schiotz!

PR #4873, opened 2017-07-12
{tools}[intel/2017a] cadaver v0.23.3

boegel:
(created using `eb --new-pr`)

PR #4874, opened 2017-07-12
Revert "{tools}[intel/2017a] cadaver v0.23.3"

verdurin:
Reverts easybuilders/easybuild-easyconfigs#4873

Doesn't yet include checksum, which I should have spotted.


boegel:
@verdurin I wouldn't revert the whole thing, but just add the checksum?

PR #4919, opened 2017-07-20
Update PLINK urls

pol:
The old PLINK urls were no longer active, updated to new URLs.


pol:
Why did I not scroll down prior to submitting the PR... ok, third time's the charm.  Sorry.

PR #5019, opened 2017-08-16
{lib}[GCC-5.4.0-2.26] lftp 

robqiao:
{lib}[GCC-5.4.0] lftp config


easybuild/easyconfigs/l/lftp/lftp-4.7.8-GCC-5.4.0-2.26.eb
```diff
@@ -0,0 +1,50 @@
...
+    'http://lftp.yar.ru/ftp/',
+    'http://lftp.yar.ru/ftp/old/',
+]
+
+
+dependencies = [
+    ('GnuTLS', '3.5.14', '', ('GCCcore', '5.4.0')),
+    ('gettext', '0.19.8'),
+    ('Autoconf', '2.69', '', True), 
+    ('libreadline', '7.0', '', ('GCCcore', '5.4.0')),
```
boegel: I think you'll need to stick with `6.3` here, due to other dependencies already using that

There's no need to hard specify which toolchain to use here, `GCCcore/5.4.0` will be considered automatically by EasyBuild v3 if no matching easyconfig using `GCC/5.4.0-2.26` is found

robqiao: @boegel do we have a flag that can update easyconfig file to a new toolchain and correct file name correspondingly?

robqiao:
@boegel plz review this pr?

easybuild/easyconfigs/a/Autotools/Autotools-20170619-GCCcore-6.3.0.eb
```diff
@@ -0,0 +1,35 @@
+##
+# This is a contribution from Phoenix HPC Service, The University of Adelaide, Australia
+# Homepage: https://www.adelaide.edu.au/phoenix/
+#
+# Copyright::   adelaide.edu.au/phoenix
+# Authors::     Robert Qiao <robert.qiao@adelaide.edu.au>, Exe Escobedo <exequiel.sepulvedaescobedo@adelaide.edu.au>
+# License::     GNU GPL-v3.0
```
fgeorgatos: if i recall right, EB under GPLv2 so this is potentially a redistribution issue...

easybuild/easyconfigs/l/lftp/lftp-4.7.8-GCC-5.4.0-2.26.eb
boegel: I think you're looking for `--try-toolchain`?

easybuild/easyconfigs/a/Autotools/Autotools-20170619-GCCcore-6.3.0.eb
boegel: Yes, indeed, we ship `easybuild-easyconfigs` packages under GPLv2.

On the other hand, there's very little creativity involved with composing easyconfig files, so I'm not sure whether this is really an issue...

@robqiao Can you please reconsider either changing the license to something compatible with GPLv2, or simply dropping the license line?

boegel: To clarify: the license you specify here is assumed to be the license for the easyconfig file, not the license for the software it is installing...

robqiao:
@boegel Since lftp-4.7.8-GCC-6.3.0-2.28.eb is now in the repo, I will now close this PR.

boegel:
Thanks for the follow-up @robqiao!

PR #5024, opened 2017-08-16
iompi: OpenMPI dependency must come after icc/ifort

mkbosmans:


boegel:
@mkbosmans Can you clarify why you're making this change? I can't think of a scenario where this is strictly required?

mkbosmans:
In the generated module, for iompi, the OpenMPI module gets loaded before icc and ifort, which fails, because OpenMPI is not available for GCCcore.

boegel:
Thanks for the fix @mkbosmans!

PR #5050, opened 2017-08-19
Add Clang 4.0.1 based on GCC 6.4.0

mkbosmans:


easybuild/easyconfigs/c/Clang/Clang-4.0.1-GCC-6.4.0-2.28.eb
```diff
@@ -9,34 +9,41 @@
 ##
 
 name = "Clang"
-version = "3.8.1"
+version = "4.0.1"
```
verdurin: @mkbosmans Normally we just add a new `easyconfig`, rather than altering and/or renaming an existing one. That way we don't affect reproducibility.

boegel:
Since this has not been updated in several months, and other PRs for more recent versions of `Clang` have been opened recently (cfr. #5532), I'll go ahead and close this one...

easybuild/easyconfigs/c/Clang/Clang-4.0.1-GCC-6.4.0-2.28.eb
mkbosmans: This is an artifact of git squashing the diff of the two separate commits into one.

As mentioned in the first commit message, I think it is bad style to have Clang depend on the foss toolchain instead of the GCC compiler. That's why I removed Clang-3.7.1-foss-2016a.eb

Is it never OK to remove any eb file from easybuilds-easyconfigs that has been in a release?

boegel: @mkbosmans We indeed never remove easyconfigs, except when we archive old and obsolete easyconfigs with a new major release of EasyBuild (see e.g. #3732).

I agree with you though that using `foss` for Clang is a bit strange, which is why I used `GCC` in #5532.

PR #5067, opened 2017-08-28
Update ncurses to not build a separate libtinfo but provide a soft link instead. Force linking to ncurses in libreadline

ocaisa:
(created using `eb --new-pr`)

PR #5074, opened 2017-08-30
Linking to ncurses was not done correctly...

ocaisa:
Use `buildopts` instead, see http://www.linuxfromscratch.org/lfs/view/8.0/chapter06/readline.html


fgeorgatos:
@ocaisa: ugh, who else gets affected by this? (the old ghosts are waking up now)

ocaisa:
I don't really know, I'm surprised no-one has caught it until now. I only caught it because of the #5067 where the incorrectly linked libreadline was causing the system Lua (and therefore Lmod) to die. Appears to depend on the OS if this makes a difference (I was using Ubuntu 16.10)

boegel:
@ocaisa We should make this change consistent across all `libreadline` easyconfigs?

ocaisa:
I was afraid you might say that...in this case I just focussed on the latest version of `libreadline`, the previous one probably also needs a patch (see http://www.linuxfromscratch.org/lfs/view/7.10/chapter06/readline.html )

boegel:
Test report by @boegel
**SUCCESS**
node2049.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/c12661eb56f1d2899a8b30efce91501a for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2443.golett.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/784a414d4b9ac2a84ba0cbf9c520e0c8 for a full test report.

boegel:
Going in, thanks @ocaisa!

boegel:
@ocaisa To get back to applying the same fix consistently to the `libreadline` 6.3 easyconfigs: the patching mentioned in http://www.linuxfromscratch.org/lfs/view/7.10/chapter06/readline.html is only to avoid renaming the existing library files in case of an in-place update, which is not relevant in  the EasyBuild context (since each libreadline installation goes to its own installation prefix).

So probably we can/should also apply the `SHLIB_EXT` change to older `libreadline` easyconfigs, like we did for `libreadline-6.3.eb` in #5067?

Not sure it's worth the effort though, since nobody has complained about problems there for a very long time...

PR #4966, opened 2017-08-01
{lang}[GCCcore/6.4.0] Cython v0.26

JackPerdue:
(created using `eb --new-pr`)

PR #4917, opened 2017-07-20
{devel}[GCCcore/6.4.0] six v1.10.0

JackPerdue:
(created using `eb --new-pr`)
depends on PR #4914

PR #4916, opened 2017-07-20
{devel}[GCCcore/6.4.0] setuptools v36.2.0

JackPerdue:
(created using `eb --new-pr`)
depends on PR #4914

PR #4915, opened 2017-07-20
{lib}[GCCcore/6.4.0] PyYAML v3.12

JackPerdue:
(created using `eb --new-pr`)
depends on PR #4914

PR #5075, opened 2017-08-30
remove bare Python easyconfig using GCCcore/6.4.0 toolchain + easyconfigs that require it

boegel:
Following the discussion in #4962 and #5072, I propose we remove the bare Python easyconfig built with `GCCcore/6.4.0` (merged in #4962) and the handful of easyconfigs that require it (merged in #4915, #4916, #4917, #4966).

We should take a step back and figure out a long-term sustainable approach to having a `Python` installation at `GCCcore`, preferably with feedback from @damianam who has some experience with this too at JSC...


boegel:
@JackPerdue Please let us know if you're OK with this.

JackPerdue:
I totally approve of backing out any Python packages targeted for 2017b until we figure out what we want to do.

PR #5085, opened 2017-08-31
Texlive 20170524

fizwit:
texlive is very ugly.  The easyconfig is not self-contained and requires a small config file for the texlive installer script.  The config file if performing the equivalent of --prefix.  This will break with Jenkins.

### DIFF
--- a/easybuild/easyconfigs/t/texlive/texlive-20170524-foss-2016b.eb
+++ b/easybuild/easyconfigs/t/texlive/texlive-20170524-foss-2016b.eb
@@ -1,0 +1,71 @@
+# Easybuild easyconfig
+#
+# John Dey jfdey@fredhutch.org
+#
+# TexLive 2017
+# Net Install for texlive requires a configuration file named: texlive.profile
+# texlive.profile needs to be located in the current working directory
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/t/texlive/texlive-20170524-foss-2016b.eb
```diff
@@ -0,0 +1,71 @@
...
+    ('HarfBuzz', '1.3.1'), 
+    ('poppler', '0.54.0'),
+    ('cairo', '1.14.6'),
+    ('fontconfig', '2.12.1'),
+    ('zlib', '1.2.8'),
+    ('graphite2', '1.3.10'),
+]
+
+postinstallcmds = ['%(builddir)s/install-tl-%(version)s/install-tl \
+                   -profile /app/easybuild/fh_easyconfigs/texlive.profile']
```
boegel: Is the `texlive.profile` file only needed for this post-install command?

If so, we can probably embed it in the easyconfig, and put it in the TexLive installation directory itself, no?

We can use the `%(installdir)s` to inject the install location?

boegel:
@fizwit Can you clarify why this was closed? It seems like this was relatively close to being good to merge...

PR #5115, opened 2017-09-13
{math}[foss/2015b] NEST v2.12.0

jdonners:
(created using `eb --new-pr`)


easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b.eb
```diff
@@ -0,0 +1,35 @@
...
+toolchainopts = {'optarch': True, 'pic': True}
+
+sources = ['nest-%(version)s.tar.gz']
+source_urls = ['https://github.com/nest/nest-simulator/releases/download/v%(version)s/']
+
+dependencies = [
+    ('Python', '2.7.9'),
+    ('libreadline', '6.3'),
+    ('GSL', '2.1'),
+    ('libtool', '2.4.6', '-GNU-4.9.3-2.25', True),
```
boegel: @jdonners There should be no need to hardcode the toolchain anymore, since EasyBuild v3 subtoolchains are considered automatically.

Also, isn't `libtool` a build dependency only?

And also, rather than using `libtool` as (build) dependency, we usually use the `Autotools` bundle as a build dep instead:

```python
builddependencies = [('Autotools', '20150215')]

File: easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b.eb [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+
+configopts = '-DREADLINE_ROOT_DIR=$EBROOTLIBREADLINE'
+
+builddependencies = [
+    ('CMake', '3.4.1'),
+    ('Doxygen', '1.8.10'),
+]
+
+modextrapaths = {
+    'PYTHONPATH': ['lib/python2.7/site-packages'],
```
boegel: please don't hardcode the Python version, use the `%(pyshortver)s` template instead (see http://easybuild.readthedocs.io/en/latest/version-specific/easyconfig_templates.html)

easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b.eb
```diff
@@ -0,0 +1,35 @@
...
+ than on the exact morphology of individual neurons."""
+
+toolchain = {'name': 'foss', 'version': '2015b'}
+toolchainopts = {'optarch': True, 'pic': True}
+
+sources = ['nest-%(version)s.tar.gz']
+source_urls = ['https://github.com/nest/nest-simulator/releases/download/v%(version)s/']
+
+dependencies = [
+    ('Python', '2.7.9'),
```
boegel: if `Python` is included as a dependency, we usually use a versionsuffix, so please include this (right below the `version =` line):

```python
versionsuffix = '-Python-%(pyver)s`

File: easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b.eb [hunk 4]
```diff
@@ -0,0 +1,35 @@
...
+name = 'NEST'
+version = '2.12.0'
+
+homepage = 'http://www.nest-simulator.org/'
+description = """NEST is a simulator for spiking neural network models
+ that focuses on the dynamics, size and structure of neural systems rather
+ than on the exact morphology of individual neurons."""
+
+toolchain = {'name': 'foss', 'version': '2015b'}
+toolchainopts = {'optarch': True, 'pic': True}
```
boegel: `optarch` is enabled by default, no need to specify it (but keep the `pic` obviously)

easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b.eb
```diff
@@ -0,0 +1,35 @@
...
+
+sources = ['nest-%(version)s.tar.gz']
+source_urls = ['https://github.com/nest/nest-simulator/releases/download/v%(version)s/']
```
boegel: Please include a `SHA256` checksum.

With EasyBuild v3.4.0, `eb` can do that for you, using `eb NEST-2.12.0-foss-2015b.eb --inject-checksums`, see also http://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#adding-or-replacing-checksums-using-inject-checksums

jdonners:
hello @boegel,

thanks for the remarks, I updated the eb accordingly. I kept libtool as dependency, since NEST uses libltdl. CMake is used for configuration, not the Autotools. Added checksum. Added extra builddependency pkg-config and dependency ncurses. Added flag for MPI, although it is disabled (which is the default).

boegel:
@jdonners Please remove the easyconfig that doesn't include the `versionsuffix`

This should do the trick:
```
eb --update-pr 5115 :NEST-2.12.0-foss-2015b.eb
```

easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b-Python-2.7.9.eb
```diff
@@ -0,0 +1,40 @@
...
+]
+
+
+configopts = '-DREADLINE_ROOT_DIR=$EBROOTLIBREADLINE -Dwith-mpi=OFF'
+
+builddependencies = [
+    ('CMake', '3.4.1'),
+    ('Doxygen', '1.8.10'),
+    ('pkg-config', '0.27.1'),
+]
```
boegel: @jdonners Last style-related change request: please move this up, right above `dependencies =` (same block)

Also, please include a custom `sanity_check_paths`? Usually this suffices for Python packages (unless it also includes a `/bin`):

```python
sanity_check_paths = {
    'files': [],
    'dirs': ['lib/python%(pyshortver)s/site-packages'],
}
```

boegel:
@jdonners 
```
$ eb --check-style --from-pr 5115
== temporary log file in case of crash /tmp/eb-Cd3vLH/easybuild-SwcPUZ.log

*** WARNING: Using easyconfigs from unstable (pending/failed tests or merge conflict) PR #5115 ***

Running style check on 1 easyconfig(s)...
/tmp/eb-Cd3vLH/files_pr5115/n/NEST/NEST-2.12.0-foss-2015b-Python-2.7.9.eb:40:25: E231 missing whitespace after ','
/tmp/eb-Cd3vLH/files_pr5115/n/NEST/NEST-2.12.0-foss-2015b-Python-2.7.9.eb:40:42: E231 missing whitespace after ','
[FAIL] /tmp/eb-Cd3vLH/files_pr5115/n/NEST/NEST-2.12.0-foss-2015b-Python-2.7.9.eb
ERROR: One or more style checks FAILED!
```

easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b-Python-2.7.9.eb
```diff
@@ -0,0 +1,44 @@
...
+ than on the exact morphology of individual neurons."""
+
+toolchain = {'name': 'foss', 'version': '2015b'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://github.com/nest/nest-simulator/releases/download/v%(version)s/']
+sources = ['nest-%(version)s.tar.gz']
+checksums = ['bac578f38bb0621618ee9d5f2f1febfee60cddc000ff32e51a5f5470bb3df40d']
+
+configopts = '-DREADLINE_ROOT_DIR=$EBROOTLIBREADLINE -DCMAKE_INSTALL_LIBDIR=lib -Dwith-mpi=OFF'
```
boegel: @jdonners Can you clarify the need for `-DCMAKE_INSTALL_LIBDIR=lib`, what happens if you don't specify that?

easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b-Python-2.7.9.eb
```diff
@@ -0,0 +1,44 @@
...
+    ('Doxygen', '1.8.10'),
+    ('pkg-config', '0.27.1'),
+]
+
+modextrapaths = {
+    'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages'],
+}
+
+sanity_check_paths = {
+    'files': ['bin/nest', 'lib/libnest.so', 'lib/libmodels.so'],
```
boegel: @jdonners We try to avoid hardcoding `.so` (since that's incorrect on macOS), please use `SHLIB_EXT` instead:

```python
'files': ['bin/nest', 'lib/libnest.%s' % SHLIB_EXT, 'lib/libmodels.%s' % SHLIB_EXT],
```

boegel:
Test report by @boegel
**SUCCESS**
node2038.delcatty.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/231e1f7be5cd8bc24efe80560622c83c for a full test report.

jdonners:
The class is problably more bio rather than math. Are there any policies about categorization?

boegel:
@jdonners We're not very strict about this, since it's sometimes difficult to assign a single class for some software packages... I agree with you that `bio` seems more applicable here though.

easybuild/easyconfigs/n/NEST/NEST-2.12.0-foss-2015b-Python-2.7.9.eb
jdonners: I installed it both on Debian 7 and RedHat 7. On Debian it would install in 'lib', but on RedHat it would use lib64 by default. The sanity check then fails on RedHat 7.

boegel: OK, that could also be handled by using `( 'lib/libnest.%s' % SHLIB_EXT,  'lib64/libnest.%s' % SHLIB_EXT)` in the sanity check (where `(..., ...)` specifies alternatives, one of those must be available), but this works too.

Thanks for testing on different OSs!

boegel:
Test report by @boegel
**SUCCESS**
gligar03.gligar.os - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/75c326c8edda4bbb1b25d62c9b180f6a for a full test report.

boegel:
Going in, thanks @jdonners!

PR #5005, opened 2017-08-10
add test to enforce SHA256 checksums in touched files in pull requests to develop (REVIEW)

boegel:
Merging this will make Travis enforce having SHA256 checksums for sources & patches in all easyconfigs touched by future PRs, including for extensions in Python, R, Perl easyconfigs...

Before we merge this, we should probably make it a bit less painful by:

* [x] making `eb --check-style` also complain about missing SHA256 checksums
   * done via `--check-contrib`, see ~~https://github.com/easybuilders/easybuild-framework/pull/2551~~
* [x] support something like `eb --inject-checksums` to avoid the need to manually add/adjust checksums in easyconfigs
    * done, see ~~https://github.com/easybuilders/easybuild-framework/pull/2286~~

(proof-of-concept in ~~#5001~~ https://github.com/boegel/easybuild-easyconfigs/pull/40)

edit: now requires ~~https://github.com/easybuilders/easybuild-framework/pull/2551~~

PR #5152, opened 2017-09-21
{lang} [intel/2016b] PyCUDA/2016.1.2 (REVIEW)

smoors:
This version does not work:
PyCUDA-2016.1.2-intel-2017a-Python-2.7.13.eb

### DIFF
--- a/easybuild/easyconfigs/p/PyCUDA/PyCUDA-2016.1.2-intel-2016b-Python-2.7.12.eb
+++ b/easybuild/easyconfigs/p/PyCUDA/PyCUDA-2016.1.2-intel-2016b-Python-2.7.12.eb
@@ -1,0 +1,31 @@
+easyblock = 'PythonPackage'
+
+name = 'PyCUDA'
+version = '2016.1.2'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'http://mathema.tician.de/software/pycuda'
+description = """PyCUDA lets you access Nvidia’s CUDA parallel computation API from Python."""
+
+toolchain = {'name': 'intel', 'version': '2016b'}
+
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [
+    ('Python', '2.7.12'),
+    ('CUDA', '8.0.61', '', True),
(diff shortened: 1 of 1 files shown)


wpoely86:
Test report by @wpoely86
**SUCCESS**
nic66 - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz, Python 2.7.5
See https://gist.github.com/6cea524e38a404d30cf12a105304a7d8 for a full test report.

easybuild/easyconfigs/p/PyCUDA/PyCUDA-2016.1.2-intel-2016b-Python-2.7.12.eb
```diff
@@ -0,0 +1,31 @@
...
+version = '2016.1.2'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'http://mathema.tician.de/software/pycuda'
+description = """PyCUDA lets you access Nvidia’s CUDA parallel computation API from Python."""
+
+toolchain = {'name': 'intel', 'version': '2016b'}
+
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
```
boegel: @wpoely86 please keep an eye on the presence of checksums, we should start forcing that in Travis soon (cfr. #5005)

boegel:
@smoors: for future reference, how does the existing `PyCUDA-2016.1.2-intel-2017a-Python-2.7.13.eb` not work?

smoors:
CUDA/8.0.61 needs a GCC version <= 5, but intel/2017a uses GCC 6

boegel:
@smoors Do you mind copy-pasting the error message here, so people can track it down if they run into it.

Does it make sense to retract the previous easyconfig, since it just doesn't work?
Can we add a sanity check command that avoids this from happening again?

Also, please submit a follow-up PR that adds a SHA256 checksum for this easyconfig.

wpoely86:
This is just general CUDA stuff. It doesn't work with recent GCC, it will complain loudly and the error is already all over the internet.

It would be good to retrack the other easyconfig as it can never work with the current versions of CUDA.

PR #5317, opened 2017-11-09
{chem,math,lib}[intel/2017b] NAMD 2.12, MDTraj 1.9.1, molmod 1.4.3, yaff 1.4.2

tovrstra:
I have some questions about the NAMD easyconfig:
- The download from the NAMD website (requires login) is a `tar` file and not a `tar.gz` file. Still, the `.gz` extension is present. I've added a comment to reduce the confusion. I hope this is OK.
- Is there a way to change the easyconfig such that it uses `Tcl/8.6.7-GCCcore-6.4.0` instead of `Tcl/8.6.6-intel-2017b`. The former is loaded with Python/3.6.3-intel-2017b and may cause conflicts when loading NAMD afterwards. How should we deal with this?


boegel:
@tovrstra Which comment are you referring to for NAMD, I don't see it in the easyconfig file?

If the source filename is incorrectly named, you probably need to specify the correct extraction command yourself, since EasyBuild relies on the file extension being correct. See http://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#common-easyconfig-param-sources-alt.

To fix the Tcl issue, why not bump the Tcl version for NAMD to 8.6.7? Note that `Tcl-8.6.7-GCCcore-6.4.0.eb` suffices to resolve that dependency; since EasyBuild v3.x, `eb` will also consider subtoolchains if an easyconfig with the 'parent' toolchain (`intel` here) is not found.

easybuild/easyconfigs/m/MDTraj/MDTraj-1.9.1-intel-2017b-Python-3.6.3.eb
```diff
@@ -0,0 +1,29 @@
...
+    ('Python', '3.6.3'),
+    ('zlib', '1.2.11'),
+]
+
+# The unit tests of MDTraj are a pain to get to work: they require
+# a massive number of extra dependencies. See
+# https://github.com/mdtraj/mdtraj/blob/master/devtools/conda-recipe/meta.yaml
+sanity_check_paths = {
+    'files': [],
+    'dirs': ['lib/python%(pyshortver)s/site-packages/%(namelower)s-%(version)s-py%(pyshortver)s-linux-x86_64.egg/%(namelower)s'],
```
boegel: This line is too long, so it should be trimmed somehow (that's also why the tests are failing in Travis, we're stricter w.r.t. style now in easyconfigs).

You can just check for `lib/python%(pyshortver)s/site-packages` here, since EasyBuild already does an `import mdtraj` check.

easybuild/easyconfigs/n/NAMD/NAMD-2.12-intel-2017b-mpi.eb
```diff
@@ -0,0 +1,24 @@
...
+versionsuffix = '-mpi'
+
+homepage = 'http://www.ks.uiuc.edu/Research/namd/'
+description = """NAMD is a parallel molecular dynamics code designed for high-performance simulation of
+ large biomolecular systems."""
+
+toolchain = {'name': 'intel', 'version': '2017b'}
+toolchainopts = {'pic': True, 'usempi': True}
+
+sources = ['NAMD_%(version)s_Source.tar.gz']
```
boegel: @tovrstra Please include a SHA256 checksum here too (you can use `--inject-checksums` for that).

easybuild/easyconfigs/y/yaff/yaff-1.4.2-intel-2017b-Python-3.6.3.eb
```diff
@@ -0,0 +1,32 @@
...
+
+homepage = 'http://molmod.github.io/yaff/'
+description = """Yaff stands for 'Yet another force field'. It is a pythonic force-field code."""
+
+toolchain = {'name': 'intel', 'version': '2017b'}
+toolchainopts = {'lowopt': True}
+
+source_urls = ['https://github.com/molmod/yaff/releases/download/%(version)s']
+sources = [SOURCE_TAR_GZ]
+patches = ['yaff-1.4.2-add-missing-include.patch']
```
boegel: Please include `SHA256` checksums for both `sources` and `patches` (pro tip: use `--inject-checksums`).

Also avoid hardcoding the `yaff` version in `patches`, use `%(version)s` instead.

easybuild/easyconfigs/m/MDTraj/MDTraj-1.9.1-intel-2017b-Python-3.6.3.eb
tovrstra: Good. How does easybuild now that it should try `import mdtraj`? The module name is not necessarily related to the package name.

boegel: By default is uses `name.lower()`, you can override it with:

```python
options = {'modulename': 'foo'}
```

easybuild/easyconfigs/n/NAMD/NAMD-2.12-foss-2017a-mpi.eb
```diff
@@ -9,7 +9,11 @@ description = """NAMD is a parallel molecular dynamics code designed for high-pe
 toolchain = {'name': 'foss', 'version': '2017a'}
 toolchainopts = {'usempi': True, 'pic': True}
 
-sources = ['NAMD_%(version)s_Source.tar.gz']
+sources = [{
+    'filename': 'NAMD_%(version)s_Source.tar.gz',
+    'extract_cmd': "tar xfv %s",  # source file is actually not gzipped
```
boegel: Hmm, seems like I have been modifying the downloaded sources myself (i.e. by zipping it again) up until now, because the source file I have *is* gzipped. It's not when I redownloaded it though.

Does this ring a bell for you @akesandgren?

tovrstra: I just subscribed to the NAMD mailinglist and posted a question for this. I'll post the replies I'll get here.

easybuild/easyconfigs/m/MDTraj/MDTraj-1.9.1-intel-2017b-Python-3.6.3.eb
tovrstra: Allright! Thanks. In that case, I can remove this line from the yaff eb file.

tovrstra:
It seems that `Tcl-8.6.7` cannot be used as a dependency in `NAMD-2.12-intel-2017a-mpi.eb`? Why doesn't it use `Tcl-8.6.7-GCCcore-6.4.0` as a dependency in that case?

easybuild/easyconfigs/n/NAMD/NAMD-2.12-intel-2017a-mpi.eb
```diff
@@ -9,10 +9,14 @@ description = """NAMD is a parallel molecular dynamics code designed for high-pe
...
-sources = ['NAMD_%(version)s_Source.tar.gz']
+sources = [{
+    'filename': 'NAMD_%(version)s_Source.tar.gz',
+    'extract_cmd': "tar xfv %s",  # source file is actually not gzipped
+}]
+checksums = ['3be34f04c64a1e4225c04560c676ca8ff9bb57aa8830e42a46bf9ac6b16eb403']
 
 dependencies = [
-    ('Tcl', '8.6.6'),
+    ('Tcl', '8.6.7'),
```
boegel: @tovrstra Don't bump dependency versions in existing easyconfigs, since that may cause problems elsewhere (e.g. version conflicts on Tcl when this NAMD is reinstalled and then combined with existing modules).

Also, there's no easyconfig file for Tcl 8.6.7 that is compatible with `inte/2017a`.

easybuild/easyconfigs/y/yaff/yaff-1.4.2-intel-2017b-Python-2.7.14.eb
```diff
@@ -11,15 +11,18 @@ toolchain = {'name': 'intel', 'version': '2017b'}
 
 source_urls = ['https://github.com/molmod/yaff/releases/download/%(version)s']
 sources = [SOURCE_TAR_GZ]
-checksums = ['f61a85dcbf375acb21d7540ad7a80850996315d5eca608cfe2d57867f6120710']
+patches = ['yaff-%(version)s-add-missing-include.patch']
```
boegel: Hmm, this worked fine without this patch though?

boegel:
@tovrstra For `intel/2017a`, `GCCcore/6.3.0` is a subtoolchain, so `Tcl-8.6.7-GCCcore-6.4.0.eb` wouldn't satisfy that dependency.

But you shouldn't be updating dep versions in existing easyconfigs anyway... :)

tovrstra:
ok. I'll fix that. Sorry...

tovrstra:
I've removed the dependency change. It should be OK now.

easybuild/easyconfigs/y/yaff/yaff-1.4.2-intel-2017b-Python-2.7.14.eb
tovrstra: I really have no idea. I also don't understand why it passed the unit tests without setting `lowopt=True`. I'll check.

easybuild/easyconfigs/n/NAMD/NAMD-2.12-foss-2017a-mpi.eb
akesandgren: The file I have on disk is NAMD_2.12_Source.tar.gz
But i don't remember having done anything to it.
And the older files i have are also tar.gz

tovrstra: Yes, the extension is tar.gz, but can you run the file command on it? When I do this, I get

```
$ file NAMD_2.12_Source.tar.gz 
NAMD_2.12_Source.tar.gz: POSIX tar archive (GNU)
```

So not gzipped, despite the extension.

easybuild/easyconfigs/y/yaff/yaff-1.4.2-intel-2017b-Python-2.7.14.eb
tovrstra: This is weird! Indeed, the patch is not needed for `yaff-1.4.2-intel-2017b-Python-2.7.14.eb` but it is still needed to let the tests pass for `yaff-1.4.2-intel-2017b-Python-3.6.3.eb`. Even stranger, the tests run fine on Delcatty without `lowopt=True`, while on Golett `lowopt=True` is needed. I'll revert the changes in `yaff-1.4.2-intel-2017b-Python-2.7.14.eb` but I can't make much sense out of the current behavior.

tovrstra:
Changes in `yaff-1.4.2-intel-2017b-Python-2.7.14.eb` were reverted.

easybuild/easyconfigs/y/yaff/yaff-1.4.2-intel-2017b-Python-2.7.14.eb
boegel: @tovrstra Delcatty is Intel Sandy Bridge, Golett is Intel Haswell, which may explain why `lowopt=True` is only needed for the latter for `yaff`.

easybuild/easyconfigs/n/NAMD/NAMD-2.12-foss-2017a-mpi.eb
boegel: @akesandgren Indeed, the downloaded file has `.tar.gz` extension, but it's really just a (non-gzipped) tarball.

I see the same thing when re-downloading it, so I suspect I just fixed that myself manually by gzipping it after downloading.

This approach here is better, but then we should also make it consistent (i.e. apply the same change, including a checksum, in the existing NAMD 2.12 easyconfigs).

boegel: ... which is exactly what @tovrstra did, thanks. :)

easybuild/easyconfigs/t/TAMkin/TAMkin-1.2.4-intel-2016a-Python-2.7.11.eb
```diff
@@ -0,0 +1,33 @@
...
+name = 'TAMkin'
+version = '1.2.4'
+versionsuffix = "-Python-%(pyver)s"
+
+homepage = 'http://molmod.github.io/tamkin/'
+description = """TAMkin is a post-processing toolkit for normal mode analysis,
+ thermochemistry and reaction kinetics. It uses a Hessian computation from a
+ standard computational chemistry program as its input."""
+
+toolchain = {'name': 'intel', 'version': '2016a'}
```
boegel: Hmm, this is the only one using `intel/2016a` rather than `intel/2017b`, does it belong to the same batch of software?

tovrstra: No, this is an older one, in fact unrelated to the other ones. I haven't converted TAMkin to work with Python 3 yet.

easybuild/easyconfigs/n/NAMD/NAMD-2.12-foss-2017a-mpi.eb
tovrstra: You can follow up my message on the mailing list here: http://www.ks.uiuc.edu/Research/namd/mailing_list/namd-l.2016-2017/2442.html
No replies yet. Nobody seems to care, so don't expect this to change anytime soon.

I've looked around a bit: you can also get the source code here apparently: https://github.com/divijverma/NAMD/releases I'm not sure how legal and reliable that is. At least you'll get a .tar.gz when you're told so :)

tovrstra:
Are there any other changes needed. I think all is fixed now?

boegel:
@tovrstra The `TAMkin` easyconfig is proving to be a bit of a PITA, since the tests seem to be hanging for some reason.

Maybe it's better to remove the TAMkin easyconfig here, and open a separate pull request for it to avoid it blocking this PR?

easybuild/easyconfigs/n/NAMD/NAMD-2.12-foss-2017a-mpi.eb
boegel: We should stick to the official sources, and just deal with the faulty naming, it's an easy/clean fix now in the easyconfig.

tovrstra:
The TAMkin easyconfig was moved to separate PR: #5333

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 7 (7 easyconfigs in this PR)
node2101.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/8c73c26172a9cdcb4fadbed812039b1c for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2432.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/a6a14de5be0d025ed4d0726f6ccf715f for a full test report.

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 6 out of 7 (7 easyconfigs in this PR)
See https://gist.github.com/786bfce2190739b571707909bf65cdb2 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2101.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/6f8a482ae6d353d2735a2ba872c09656 for a full test report.

boegel:
Going in, thanks @tovrstra!

PR #5390, opened 2017-11-22
{phys}[dummy] COMSOL v5.2a

Micket:
Depends on easybuilders/easybuild-easyblocks#1296

### DIFF
--- a/easybuild/easyconfigs/c/COMSOL/COMSOL-5.2a.eb
+++ b/easybuild/easyconfigs/c/COMSOL/COMSOL-5.2a.eb
@@ -1,0 +1,17 @@
+name = 'COMSOL'
+version = '5.2a'
+
+homepage = 'https://www.comsol.se/'
+description = """COMSOL Multiphysics - Simulation Tool for Electrical, Mechanical, Fluid Flow, and Chemical
+Applications"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = [SOURCELOWER_TAR_GZ]
+
+# License is required (EB_COMSOL_LICENSE), e.g:
+# EB_COMSOL_LICENSE="FFFFFFFFF-BBB-123456-123456"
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/c/COMSOL/COMSOL-5.2a.eb
```diff
@@ -0,0 +1,17 @@
...
+Applications"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = [SOURCELOWER_TAR_GZ]
+
+# License is required (EB_COMSOL_LICENSE), e.g:
+# EB_COMSOL_LICENSE="FFFFFFFFF-BBB-123456-123456"
+# EB_COMSOL_LICENSE="1718@licsvr1 1718@licsvr2"
+# EB_COMSOL_LICENSE="/path/to/license.dat"
```
boegel: This should be updated according to the changes in https://github.com/easybuilders/easybuild-easyblocks/pull/1317?

easybuild/easyconfigs/c/COMSOL/COMSOL-5.2a.eb
```diff
@@ -0,0 +1,17 @@
+name = 'COMSOL'
+version = '5.2a'
+
+homepage = 'https://www.comsol.se/'
+description = """COMSOL Multiphysics - Simulation Tool for Electrical, Mechanical, Fluid Flow, and Chemical
+Applications"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = [SOURCELOWER_TAR_GZ]
```
boegel: I think it's worth including a comment on how to obtain this tarball?

I think COMSOL typically comes as a `.iso`? Is it clear which tarball to grab from that, or should we specify it more clearly?

Also, please include a SHA256 checksum?

migueldiascosta:
@Micket ping?

Micket:
This is outdated by several new comsol releases already, so probably not very interesting anymore.

PR #5404, opened 2017-11-24
{bio}[foss/2016b] FastQ_Screen 0.11.3 and 0.11.4

SethosII:
This adds FastQ_Screen.

It needs either of Bowtie, Bowtie2 or BWA to work. I mentioned this in the `modloadmsg`. Is this sufficient?

### DIFF
--- a/easybuild/easyconfigs/f/FastQ_Screen/FastQ_Screen-0.11.3-foss-2016b-Perl-5.24.0.eb
+++ b/easybuild/easyconfigs/f/FastQ_Screen/FastQ_Screen-0.11.3-foss-2016b-Perl-5.24.0.eb
@@ -24,11 +24,11 @@ dependencies = [
 ]
 
 sanity_check_paths = {
-    'files': ['fastq_screen'],
+    'files': ['%(namelower)s'],
     'dirs': [],
 }
 
-sanity_check_commands = [('fastq_screen -v')]
+sanity_check_commands = [('%(namelower)s -v')]
 
 modloadmsg = "You will also need to load either of Bowtie, Bowtie2 or BWA for FastQ Screen\n"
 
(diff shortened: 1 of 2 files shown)


SethosII:
Also one question regarding the easyblock. Is `PackedBinary` the correct one here? There is also `Tarball` which seems to be very similar?

SethosII:
Version 0.11.4 was released a few days ago, so also adding an easyconfig for it.

boegel:
@SethosII I would say `Tarball` is more natural here, but in practice there's no difference here with `PackedBinary`.

Does `FastQ Screen` automatically pick `Bowtie`, `Bowtie2` or `BWA`, or do you need to specify what it should use? Would it make sense to specify all three as dependencies?

SethosII:
@boegel I tried `Tarball` and the sanity check fails because it can't find the binary, so it seems Tarball doesn't set `PATH` (or at least not the same way as `PackedBinary`), should I switch to `Tarball` and `modextrapaths = {'Path': ''}`?

You have to specify what aligner should be used (`-aligner bowtie|bowtie2|bwa`). `FastQ_Screen` will complain if you miss the option. I have no opinion on whether we should include all aligners as dependencies.

boegel:
@SethosII That's right, when using `Tarball` you'll need to specify yourself that the programs are located in the top-level installation directory, using:

```python
modextrapaths = {'PATH': ''}
```

Since you have to specify the aligner explicitly, I think it makes sense to include all three possible aligners as dependencies to `FastQ_Screen` and give the user full freedom?

SethosII:
@boegel I changed it to the `Tarball` easyblock and added all aligners as dependencies.

boegel:
Test report by @boegel
**SUCCESS**
node2066.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/1f5da8f7d19e0f8a5f38a803a4cf0923 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 6 out of 6 (2 easyconfigs in this PR)
node2406.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/dcb7eb200bd4f0bae70ea4b2de7c7eb4 for a full test report.

boegel:
Going in, thanks @SethosII!

PR #5422, opened 2017-11-28
OpenFoam versions 4 and above are not on openfoam.com but on openfoam…

mboisson:
….org


boegel:
Hmm, not sure why this never made it in, it's clearly good to go. Thanks @mboisson!

PR #5452, opened 2017-12-01
{bio}[intel/2017b] OpenMM 7.1.1

tovrstra:
I'm trying to make an EasyConfig for the latest OpenMM version with recent Intel compilers.

I run into a rather silly issue. Easybuild complains that it cannot find `CMake/3.9.5-intel-2017b` but it should actually find `CMake/3.9.5-GCCcore-6.4.0`. There is something going wrong with the dependency resolution but I don't see how I can fix this. Any ideas?

### DIFF
--- a/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
+++ b/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
@@ -1,0 +1,50 @@
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+# Author: Pablo Escobar Lopez
+# sciCORE - University of Basel
+# SIB Swiss Institute of Bioinformatics 
+
+easyblock = "CMakeMake"
+
+name = "OpenMM"
+version = "7.1.1"
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://simtk.org/home/openmm'
+description = """OpenMM is a toolkit for molecular simulation."""
+
+toolchain = {'name': 'intel', 'version': '2017b'}
+
+source_urls = ['https://github.com/pandegroup/openmm/archive/']
+sources = ['%(version)s.tar.gz']
+
+builddependencies = [
+    ('CMake', '3.9.5'),
+]
+
+separate_build_dir = True
+
+dependencies = [
+    ('Python', '3.6.3'),
+    ('SWIG', '3.0.12', '-Python-%(pyver)s'),
+]
+
+runtest = ' test'
+
+preinstallopts = ' export OPENMM_INCLUDE_PATH=%(installdir)s/include && '
+preinstallopts += ' export OPENMM_LIB_PATH=%(installdir)s/lib && '
+
+# required to install the python API
+installopts = ' && cd python && python setup.py build && python setup.py install --prefix=%(installdir)s'
+
+sanity_check_paths = {
+    'files': ["lib/libOpenMM.%s" % SHLIB_EXT, "lib/python%(pyshortver)s/site-packages/simtk/openmm/openmm.py"],
+    'dirs': []
+}
+
+modextrapaths = {
+    'PYTHONPATH': 'lib/python%(pyshortver)s/site-packages',
+    'OPENMM_INCLUDE_PATH': 'include',
+    'OPENMM_LIB_PATH': 'lib',
+}
+
+moduleclass = 'bio'
--- a/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
+++ b/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
@@ -13,13 +13,14 @@ homepage = 'https://simtk.org/home/openmm'
 description = """OpenMM is a toolkit for molecular simulation."""
 
 toolchain = {'name': 'intel', 'version': '2017b'}
+toolchainopts = {'opt': True}
 
 source_urls = ['https://github.com/pandegroup/openmm/archive/']
 sources = ['%(version)s.tar.gz']
 checksums = ['ba915d1bc2ea8643c699e48fa4386646dc799a7b4f165350ff1317d8c71a1493']
 
 builddependencies = [
-    ('CMake', '3.9.5'),
+    ('CMake', '3.9.1'),
 ]
 
 separate_build_dir = True
--- a/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
+++ b/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
@@ -20,7 +20,7 @@ sources = ['%(version)s.tar.gz']
 checksums = ['ba915d1bc2ea8643c699e48fa4386646dc799a7b4f165350ff1317d8c71a1493']
 
 builddependencies = [
-    ('CMake', '3.9.1'),
+    ('CMake', '3.9.5'),
 ]
 
 separate_build_dir = True
@@ -30,7 +30,7 @@ dependencies = [
     ('SWIG', '3.0.12', '-Python-%(pyver)s'),
 ]
 
-runtest = ' test'
+runtest = 'test -e ARGS="-E \'(Integrator)|(Thermostat)|(Barostat)|(Rpmd)\'"'
 
 preinstallopts = ' export OPENMM_INCLUDE_PATH=%(installdir)s/include && '
 preinstallopts += ' export OPENMM_LIB_PATH=%(installdir)s/lib && '
--- a/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
+++ b/easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
@@ -30,7 +30,7 @@ dependencies = [
     ('SWIG', '3.0.12', '-Python-%(pyver)s'),
 ]
 
-runtest = 'test -e ARGS="-E \'(Integrator)|(Thermostat)|(Barostat)|(Rpmd)\'"'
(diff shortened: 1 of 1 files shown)


tovrstra:
For the sake of completeness, this is the standard error of EasyBuild:

```ERROR: Build of /vscmnt/gent_kyukon_data/_kyukon_data_gent/.../OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb failed (err: 'build failed (first 300 chars): Missing modules for one or more dependencies: CMake/3.9.5-intel-2017b')```

File: easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+version = "7.1.1"
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://simtk.org/home/openmm'
+description = """OpenMM is a toolkit for molecular simulation."""
+
+toolchain = {'name': 'intel', 'version': '2017b'}
+
+source_urls = ['https://github.com/pandegroup/openmm/archive/']
+sources = ['%(version)s.tar.gz']
```
verdurin: @tovrstra please add an `sha256` ckecksum. You can use the option `--inject-checksums`.

verdurin:
Test report by @verdurin
**FAILED**
Build succeeded for 1 out of 2 (1 easyconfigs in this PR)
rescomp2 - Linux centos 6.8, Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz, Python 2.7.14
See https://gist.github.com/33b83c99b5263426ae54e73247701cf0 for a full test report.

verdurin:
@tovrstra in my local build two of the tests failed:
```
98% tests passed, 2 tests failed out of 105

Total Test time (real) = 156.26 sec

The following tests FAILED:
         66 - TestReferenceRpmd (Failed)
         68 - TestReferenceDrudeLangevinIntegrator (Failed)
Errors while running CTest
```

easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
tovrstra: Done.

tovrstra:
@verdurin I wish I could get it working, so I could reproduce your tests. Did you build with `CMake/3.9.5-GCCcore-6.4.0` or with `CMake/3.9.5-intel-2017b`? It should work with the former, right? I'm using Easybuild 3.4.0.

tovrstra:
I see in the logs, your build was done with `CMake/3.9.5-GCCcore-6.4.0`.

boegel:
@tovrstra The `CMake` issue you're running into is indeed a bit confusing...

For EasyBuild to consider `CMake-3.9.5-GCCcore-6.4.0.eb`, it needs to be able to find that easyconfig via the robot search path (see http://easybuild.readthedocs.io/en/latest/Using_the_EasyBuild_command_line.html#robot-search-path).

Just having the module available is not enough, it must have the easyconfig available (and not on the command, must be via robot search path)...

boegel:
Test report by @boegel
**FAILED**
node2067.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/214c0cbc8fe9feef2d32d5efb9c3a7f2 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2402.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/e0dcb84000e1ece68ea13074803eeb21 for a full test report.

boegel:
Hmm, 1 test failed on Intel Sandy Bridge, works fine on Intel Haswell... :-/

tovrstra:
@boegel Thanks for explaining. Adding `toolchainpots = {'opt': True}` or `toolchainpots = {'precise': True}` fixes the test. Would that be ok? I don't think the former one has a large performance impact.

boegel:
Using `'opt': True` in `toolchainopts` basically means compiling with `-O2` rather than `-O3`, so if anything, it should actually help with performance...

tovrstra:
OK, but the default (without toolchainopts, causing the failing test) is `'optarch': True`, which is more aggressive. So, it makes sense that `toolchainpots = {'opt': True}` can fix the test, because it makes the compiler optimization less aggressive right?

boegel:
@tovrstra To disable the default w.r.t. `optarch`, you need to specify `'optarch': False`.

Only using `'opt': True` results in only changing `-O2 ... -xHost` to `-O3 ... -xHost`.

tovrstra:
@boegel Did you mean `'opt': True` results in only changing `-O3 ... -xHost` to `-O2 ... -xHost` instead? 

Also, I was probably not understanding the toolchain options properly. So, the optimization flags in toolchainopts are not mutually exclusive? I.e. setting one to True does (not) imply all other ones are set to False?

tovrstra:
@boegel I'm sorry for getting completely confused about `toolchainopts`. I tried to figure out from the documentation and even source code how the flags in toolchainopts translate into compiler flags but I couldn't figure it out. Is there a way to get to that information?

The only thing I found is this table here, which contains the default values: http://easybuild.readthedocs.io/en/latest/version-specific/toolchain_opts.html#intel Are these correct?

Are there combinations that should be avoided or that have no effect? A Rosetta stone would be very convenient, i.e. something that just says for every flag in toolchainopts what the corresponding changes are in terms of compiler flags.

tovrstra:
Sorry, one more question: when `toolchainopts` is specified in an easyconfig, does this replace the dictionary with default `toolchainopts`, or is the default updated with `toolchainopts` from the easyconfig?

boegel:
@tovrstra When specifying `toolchainopts`, only the values listed in there are updated, the others are kept at the default.
So unless `'optarch': False` is used, the default value (`True`) for `optarch` remains active, etc.

`opt` only has an effect on the optimisation level being used: `-O3` rather than `-O2`, it does not affect the use of `-xHost` (which is controlled via the `optarch` toolchain option), similar for `noopt` (`-O0`), `lowopt` (`-O1`), etc.

The precision-related compiler options that are added to `$CFLAGS` & co are controlled by the toolchain options related to precision (`strict`, `loose`, `precise`).

This has grown historically, and we could probably do a better job of organising/documenting this...

tovrstra:
@boegel OK Got it. Thanks! In that case, I understand that passing tests with `'opt`: True` is an odd coincidence here. My best guess explanation is that the reference results were obtained with binaries compiled with `-O3`, but even then it would be a coincidence that they used the same compiler etc.

boegel:
@tovrstra I agree, so enabling `precise` instead seems like a better approach to me? Have you tried with the slightly looser `strict`?

boegel:
Ugh, scratch that @tovrstra, `strict` is stricter w.r.t. precision than `precise`, I had them mixed up.

tovrstra:
No, I'll try. hang on.

tovrstra:
It turns out that some tests are not deterministic. E.g., I tried twice with `'strict': True` and otherwise also exactly the same easyconfig. The first attempt failed on a different test and the second attempt passed all tests.

Some tests involve integration algorithms for stochastic PDEs. The time averages from such simulations eventually converge to some reproducible number in the limit of infinitely long simulations. Obviously, the tests are not long enough or the thresholds on the tests are too tight, or yet something else. This is hard to say without digging deeper. I just disable tests that could have this behavior. `'opt': True` seems to ok.

tovrstra:
Not sure why the tests fail. It works fine for me. Probably related to quotes the line

```
runtest = 'test -e ARGS="-E \'(Integrator)|(Thermostat)|(Barostat)|(Rpmd)\'"'
```

easybuild/easyconfigs/o/OpenMM/OpenMM-7.1.1-intel-2017b-Python-3.6.3.eb
```diff
@@ -0,0 +1,52 @@
...
+]
+
+separate_build_dir = True
+
+dependencies = [
+    ('Python', '3.6.3'),
+    ('SWIG', '3.0.12', '-Python-%(pyver)s'),
+]
+
+runtest = 'test -e ARGS="-E \'(Integrator)|(Thermostat)|(Barostat)|(Rpmd)\'"'
```
boegel: @tovrstra Can you try wrapping this in triple quotes, and maybe with a space at the end of the string to keep it readable?

```python
runtest = """test -e ARGS="-E \'(Integrator)|(Thermostat)|(Barostat)|(Rpmd)\"' """

boegel:
@tovrstra Seems like triple quotes did the trick :+1:

boegel:
Test report by @boegel
**SUCCESS**
node2404.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/bc0c07eed7692c3338736f934d911ad4 for a full test report.

tovrstra:
Good. Then I have no more changes planned for the moment.

boegel:
Test report by @boegel
**SUCCESS**
node2404.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/c7e12da05763d19c68930f9193f02971 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2029.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/13b20fda560edc5f16b7e0338baf178b for a full test report.

boegel:
@verdurin Can you try submitting a test report again?

verdurin:
Test report by @verdurin
**SUCCESS**
rescomp2 - Linux centos 6.8, Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz, Python 2.7.14
See https://gist.github.com/873c0fcfd0575081fb5d867adffad803 for a full test report.

boegel:
Going in, thanks @tovrstra!

PR #5456, opened 2017-12-04
{devel} [foss/2017a] [intel/2017a] CMake 3.9.4

Falkor:


Falkor:
Hum, I don't catch the error -- looks to me as normal that "No conflicts detected" ?

boegel:
@Falkor The assertion that there are no conflicts failed, which means conflicts were found, see the Travis log in https://travis-ci.org/easybuilders/easybuild-easyconfigs/jobs/311257026; you can also check using `eb --check-conflicts --from-pr 5456`.

Two questions though: why CMake with `intel`, we usually just use the `GCCcore` subtoolchain, and why not the most recent 3.9.5 or 3.10.0?

easybuild/easyconfigs/c/CMake/CMake-3.9.4-foss-2017a.eb
```diff
@@ -0,0 +1,41 @@
...
+"""
+
+toolchain = {'name': 'foss', 'version': '2017a'}
+
+source_urls = ['http://www.cmake.org/files/v%(version_major_minor)s']
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['b5d86f12ae0072db520fdbdad67405f799eb728b610ed66043c20a92b4906ca1']
+
+builddependencies = [
+    ('binutils', '2.27'),
```
boegel: this makes no sense, a full toolchain like `foss` already includes `binutils`...

boegel:
@Falkor Ping on my question on why to add an easyconfig for `CMake` built with `intel` while the existing easyconfigs using `GCCcore` can be used just fine for building other software with `intel` or `foss`?

Falkor:
@boegel concerning your two questions:

- I'm not up-to-date with the `GCCcore` precised version is supported and as we do offer support for `{foss,intel}-2017a` toolchains, it was natural to offer it for these two flavors. 
- as regards the proposed CMake version:  indeed I came to this pool request when building Boost 1.65.1 -- it happens that FindBoost fails to find imported target for Boost 1.65.1  with Cmake < 3.9.4. I didn't checked carefully enough that this was not he last up-to-date version. 

If you prefer, I can close this pull request and restart a new proposal for 3.10.0, over GCCcore.

boegel:
@Falkor Since EasyBuild v3.0, `eb` will also consider easyconfigs using subtoolchains when resolving dependencies.

So, when you're building something using `foss/2017a` that requires `CMake` 3.9.1 (for example), it search for these easyconfigs to resolve that dependency (in this order):

* `CMake-3.9.1-foss-2017a.eb`
* `CMake-3.9.1-gompi-2017a.eb`
* `CMake-3.9.1-GCC-6.3.0-2.27.eb`
* `CMake-3.9.1-GCCcore-6.3.0.eb`

We tend to only provide easyconfigs for `CMake` recently that are installed with `GCCcore`, since those can also be used when installing software with `intel/2017a` (we use the same `GCCcore` version in both `foss` and `intel` of the same version for that exact reason).

Since you're using the previous common toolchain `2017a`, I would just stick to using CMake 3.9.1, for which an easyconfig is already available, unless you have a technical reason why 3.9.1 does not work. It seems like you do with Boost 1.65.1, so in that case, a new PR for CMake 3.10.0 with `GCCcore` 6.3.0 makes sense.

Falkor:
@boegel Excellent, thanks for the clear and concise explanation. 
I'll indeed close the pull request.  I may wait for the 2018a toolchain to propose a new PR for CMake 3.10.x with GCCcore 6.4.0 unless it's already provided.

PR #5443, opened 2017-12-01
{vis}[foss/2017b] Tk v8.6.7

Helios07:
(created using `eb --new-pr`)

PR #5360, opened 2017-11-17
{lang}[intel/2017b] R 3.4.3 (REVIEW)

boegel:
requires ~~https://github.com/easybuilders/easybuild-framework/pull/2345~~ and ~~#5328, #5334, #5336 , #5341, #5337, #5343, #5346~~

now also requires updates to R easyblock: ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1292, https://github.com/easybuilders/easybuild-easyblocks/pull/1297, https://github.com/easybuilders/easybuild-easyblocks/pull/1300~~

PR #5457, opened 2017-12-04
{lang}[foss/2017b] R v3.4.3

Helios07:
(created using `eb --new-pr`)
'Depends on ~~#5443~~


verdurin:
Test report by @verdurin
**FAILED**
Build succeeded for 22 out of 28 (1 easyconfigs in this PR)
dell.lan - Linux fedora 27, Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz, Python 2.7.14
See https://gist.github.com/f8515b014b629aca9874e47c26eff735 for a full test report.

boegel:
@Helios07 I just updated #5360 last night to R 3.4.3 as well.

Do you mind syncing up your easyconfig with that ones, in terms of dependencies & (more importantly) extensions?

Helios07:
I tried to synchronize the dependencies as far as possible and added checksums to the extensions. I did not compare them one by one, but the versions should be up to date.

easybuild/easyconfigs/r/R/R-3.4.3-foss-2017b-X11-20171023.eb
```diff
@@ -0,0 +1,3459 @@
...
+    ('Tcl', '8.6.7'),  # for tcltk
+    ('Tk', '8.6.7'),  # for tcltk
+    ('cURL', '7.55.1'),  # for RCurl
+    ('libxml2', '2.9.4'),  # for XML
+    ('GDAL', '2.2.3', '-Python-2.7.14'),  # for rgdal
+    ('PROJ', '4.9.3'),  # for rgdal
+    ('GMP', '6.1.2'),  # for igraph
+    ('NLopt', '2.4.2'),  # for nloptr
+    ('FFTW', '3.3.6'),  # for fftw
+    ('libsndfile', '1.0.28'), # for seewave
```
boegel: double space required before inline comment

easybuild/easyconfigs/r/R/R-3.4.3-foss-2017b-X11-20171023.eb
```diff
@@ -0,0 +1,3459 @@
...
+    # OS dependency should be preferred if the os version is more recent then this version,
+    # it's nice to have an up to date openssl for security reasons
+    # ('OpenSSL', '1.0.2h'),
+]
+
+osdependencies = [('openssl-devel', 'libssl-dev', 'libopenssl-devel')]
+
+name_tmpl = '%(name)s_%(version)s.tar.gz'
+ext_options = {
```
boegel: Please use the new `exts_default_options` construct added in https://github.com/easybuilders/easybuild-framework/pull/2345, see also easyconfig for R 3.4.3 that just got merged.

The easiest way to tackle this is probably to copy the contents from https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/r/R/R-3.4.3-intel-2017b-X11-20171023.eb, and change the toolchain to `foss`. By preference, use the same dependency versions (it sort of useful to keep easyconfigs consistent between the `foss` and `intel` common toolchains).

Helios07: I tried to copy your easyconfig file for intel2017b, but this is not building because it tries do download the extensions from http://cran.us.r-project.org/src/base/ instead of http://cran.r-project.org/src/contrib/. I have no idea, why.

boegel: @Helios07 If you're not using current `develop` of EasyBuild (which will become v3.5.0 very soon), `exts_default_options` will be ignored because it's not a known easyconfig parameter yet in EasyBuild v3.4.1 & earlier, and so extensions will be downloaded via the 'parent' `source_urls`.

Helios07: Installing the development version did not work in time. I will close this PR and try again after 3.5.0 is released.

PR #5464, opened 2017-12-05
save Theano's compile dir to /tmp-$USER to make it arch-specific

smoors:
Include architecture name in compiledir to fix problems with architecture-specific compilation when running the software on a heterogeneous compute cluster.

### DIFF
--- a/easybuild/easyconfigs/t/Theano/Theano-1.0.0-intel-2017b-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/t/Theano/Theano-1.0.0-intel-2017b-Python-2.7.14.eb
@@ -19,10 +19,10 @@ dependencies = [('Python', '2.7.14')]
 # Include architecture name in compiledir to fix problems with architecture-specific compilation
 # when running the software on a heterogeneous compute cluster.
 # Change the environment variable "VSC_ARCH_LOCAL" to match your cluster setup.
-# import os
-# architecture = os.environ['VSC_ARCH_LOCAL']
-# modextravars = {'THEANO_FLAGS': 'compiledir_format=compiledir_%s-%%(short_platform)s-%%(processor)s-'
-#                  '%%(python_version)s-%%(python_bitwidth)s' % architecture}
+import os
+architecture = os.environ['VSC_ARCH_LOCAL']
+modextravars = {'THEANO_FLAGS': 'compiledir_format=compiledir_%s-%%(short_platform)s-%%(processor)s-'
+                 '%%(python_version)s-%%(python_bitwidth)s' % architecture}
 
 sanity_check_paths = {
     'files': ['bin/theano-cache', 'bin/theano-nose'],
--- a/easybuild/easyconfigs/t/Theano/Theano-1.0.0-intel-2017b-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/t/Theano/Theano-1.0.0-intel-2017b-Python-2.7.14.eb
@@ -16,6 +16,14 @@ checksums = ['7bc82e4fcaac023cac23c390b30d5df81bc09d33568812df75b0bd2b86d2fa8f']
 
 dependencies = [('Python', '2.7.14')]
 
+# Include architecture name in compiledir to fix problems with architecture-specific compilation
+# when running the software on a heterogeneous compute cluster.
+# Change the environment variable "VSC_ARCH_LOCAL" to match your cluster setup.
+# import os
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/t/Theano/Theano-1.0.0-intel-2017b-Python-2.7.14.eb
```diff
@@ -16,6 +16,14 @@ checksums = ['7bc82e4fcaac023cac23c390b30d5df81bc09d33568812df75b0bd2b86d2fa8f']
 
 dependencies = [('Python', '2.7.14')]
 
+# Include architecture name in compiledir to fix problems with architecture-specific compilation
+# when running the software on a heterogeneous compute cluster.
+# Change the environment variable "VSC_ARCH_LOCAL" to match your cluster setup.
+# import os
+# architecture = os.environ['VSC_ARCH_LOCAL']
+# modextravars = {'THEANO_FLAGS': 'compiledir_format=compiledir_%s-%%(short_platform)s-%%(processor)s-'
+#                  '%%(python_version)s-%%(python_bitwidth)s' % architecture}
```
boegel: @smoors Please make the PR subject reflect what you're changing here?

I'm not sure if we should drag this along in the Theano easyconfig(s) though...

Also, the `architecture` value isn't used at all?

smoors: I leave it up to you to decide that, but I suspect many HPC clusters may run into this issue. At least we did at VUB. Also for VSC this is highly relevant: for example, when a user from UGent tries to use Theano at VUB or vice versa.
Feel free to change the architecture name to something else.

boegel: What kind of problems pop up without this?

smoors: When you use the module for the first time on a node (import theano in python), an architecture-specific shared object is created at runtime in ~/.teano/compiledir_xxx. When you later try to use it on a node with an incompatible architecture, it will try to use the existing .so and fail.

boegel: I'd say it makes more sense to somehow avoid that Theano uses `$HOME/.theano`, and make it use something like `/tmp/theano-$USER/`?

That's still sort of site-specific, but way less so than your approach...

smoors: That should work too! Would you like to have this the default behavior, or commented out?

boegel: I guess it makes sense to make that the default, since it would work in most cases imho.

For sites where this doesn't work, they can customise it?

It's certainly a more sensible default than using `$HOME` in an HPC context imho...

smoors: Yes, users can always change the env var after loading, or change the default in a .theanorc file.

easybuild/easyconfigs/t/Theano/Theano-1.0.0-intel-2017b-Python-2.7.14.eb
```diff
@@ -16,6 +16,12 @@ checksums = ['7bc82e4fcaac023cac23c390b30d5df81bc09d33568812df75b0bd2b86d2fa8f']
 
 dependencies = [('Python', '2.7.14')]
 
+# Save the compiledir on the local disk to avoid issues,
+# when running the software on a heterogeneous compute cluster.
+# Change temporary directory "/tmp" and  environment variable "USER" to match your cluster setup.
+import os
```
boegel: @smoors Hmm, not too happy with the `import os` here...

Also, it doesn't make sense, since this will make `Theano` use the the `$USER` value of the installation user, rather than the value of `$USER` that is *using* Theano.

migueldiascosta:
@smoors ping?

smoors:
closing, see follow-up here:
https://github.com/easybuilders/easybuild-easyconfigs/pull/6378
https://github.com/easybuilders/easybuild-easyblocks/pull/1353

PR #5505, opened 2017-12-11
Fix typo in old URL assert.

akesandgren:
The assert incorrectly prints "AssertionError: Old URL 'github.com/hpcugent/easybuild' not found in ..."
when it finds an old URL. Note the "not found".


boegel:
@akesandgren The assertion is that the URL is *not* found, and that fails, hence the assertion error...

Many people interpret this the other way, but consistently use this approach in all `self.assert*` statements.

akesandgren:
Ah yes, you're correct of course. Forgot that it was an assertion error...

It is slightly confusing though. especially when you're tired.

PR #5538, opened 2017-12-18
{bio}[foss/2016b] BamTools 2.5.0

SethosII:
This adds version 2.5.0 of BamTools `--try-software-version` doesn't work because they changed some paths.

### DIFF
--- a/easybuild/easyconfigs/b/BamTools/BamTools-2.5.0-foss-2016b.eb
+++ b/easybuild/easyconfigs/b/BamTools/BamTools-2.5.0-foss-2016b.eb
@@ -1,0 +1,19 @@
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+
+name = 'BamTools'
+version = '2.5.0'
+
+homepage = 'https://github.com/pezmaster31/%(namelower)s'
+description = "BamTools provides both a programmer's API and an end-user's toolkit for handling BAM files."
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://github.com/pezmaster31/%(namelower)s/archive']
+sources = ['v%(version)s.tar.gz']
+
+checksums = ['85e02e04998a67cbda7ab68cdab36cee133db024e814b34e06bb617b627caf9c']
+
+builddependencies = [('CMake', '3.7.2')]
+
+moduleclass = 'bio'


easybuild/easyconfigs/b/BamTools/BamTools-2.5.0-foss-2016b.eb
```diff
@@ -0,0 +1,28 @@
...
+
+toolchain = {'name': 'foss', 'version': '2016b'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://github.com/pezmaster31/%(namelower)s/archive']
+sources = ['v%(version)s.tar.gz']
+
+checksums = ['85e02e04998a67cbda7ab68cdab36cee133db024e814b34e06bb617b627caf9c']
+
+builddependencies =[('CMake', '3.7.2')]
```
boegel: @SethosII style check fails because of missing space after `=` here...

easybuild/easyconfigs/b/BamTools/BamTools-2.5.0-foss-2016b.eb
```diff
@@ -0,0 +1,28 @@
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+
+easyblock = 'CMakeMake'
```
boegel: @SethosII There's a dedicated easyblock for `BamTools`, that should be updated for recent `BamTools` versions instead of using the generic `CMakeMake` easyblock?

SethosII: @boegel I think a dedicated easyblock isn't needed anymore as of [version 2.4.2](https://github.com/pezmaster31/bamtools/releases/tag/v2.4.2):

> CMake has been sanitized and made more modern, idiomatic and standardized. Out-of-source builds are guaranteed now. Instead of picking binaries/libraries from the build tree, users should stage and/or install BamTools instead.

Whats do you think?

boegel: @SethosII When we have an existing software-specific easyblock, we try to keep it updated and used in all easyconfig for that software, since it allows trimming the easyconfigs quite a bit (e.g. `sanity_check_paths`, etc.).

Are you up for updating the `BamTools` easyblock? If not, can you let me know what would need to change?

SethosII: @boegel Sorry, my understanding of easyblocks is to limited right now so I won't be able to figure it out on my own. The changes needed from 2.4.2 (based on the release notes) onwards are these:

* `MakeCp` part is superflous, normal install will work
* only the static library is build, so no *.so files anymore (you can pass `-DBUILD_SHARED_LIBS=ON` but it will only build shared object files then, no direct way for building static and shared)
* the headers have an additional subdirectory `bamtools`
* make install doesn't copy docs

So the `custom_paths` variable should be altered like this:

```
custom_paths['files'].remove("lib/libbamtools.%s" % sharedlib_ext)
custom_paths['files'].remove("include/shared/bamtools_global.h")
custom_paths['files'].append("include/bamtools/shared/bamtools_global.h")
custom_paths['dirs'].remove("include/api")
custom_paths['dirs'].append("include/bamtools/api")
custom_paths['dirs'].remove("docs")
```

boegel: @SethosII Updated `BamTools` easyblock is available in https://github.com/easybuilders/easybuild-easyblocks/pull/1332, PR for BamTools 2.5.0 (with `intel/2017b`) in #5576

easybuild/easyconfigs/b/BamTools/BamTools-2.5.0-foss-2016b.eb
SethosII: @boegel I changed this (with a force push).

boegel:
@SethosII Please update the easyconfig to use the updated `BamTools` easyblock from https://github.com/easybuilders/easybuild-easyblocks/pull/1332 (see also #5576)

SethosII:
@boegel I updated the easyconfig and used the updated EasyBlock but the sanity check fails because the libraries are in the lib directory and not in the lib64 directory:

```
Sanity check failed: no file of ('lib64/libbamtools.a',) in /easybuild/test/software/BamTools/2.5.0-foss-2016b, no (non-empty) directory of ('lib64/pkgconfig',) in /easybuild/test/software/BamTools/2.5.0-foss-2016b")
$ ls /easybuild/test/software/BamTools/2.5.0-foss-2016b/lib/
libbamtools.a  pkgconfig/
```

boegel:
@SethosII Please try again with the updated `BamTools` easyblock from https://github.com/easybuilders/easybuild-easyblocks/pull/1337

SethosII:
@boegel It's still failing because of a typo. See https://github.com/easybuilders/easybuild-easyblocks/pull/1337/files#r160088175.

boegel:
@SethosII Please re-test with https://github.com/easybuilders/easybuild-easyblocks/pull/1337?

SethosII:
@boegel It's working now.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/d688eae9025a40fc4f242889230eace3 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2412.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/fb979a4863b91157d47afcdc14056684 for a full test report.

boegel:
Going in, thanks @SethosII!

PR #5624, opened 2018-01-11
Update FSLeyes to 0.20.0 (w dep modules)

tutufan:
Update FSLeyes to the latest version.  Prior Bundle easyblock changed to
PythonPackage, with extensions changed to regular dependencies (some
explicit).

The rationale for reshuffling is that the setup.py for this version of FSLeyes
seems to be installing some of its dependencies into the build tree (instead
of the install tree), causing import failures at runtime.  This doesn't fix
the root problem, but works around it.

Also, set `$FSLDIR`, which the app expects.

---
This is more for feedback than immediate merge (probably).  There are several possible issues:

1. The prior easyconfig was adding a lot of things as extensions, whereas this uses explicit and implicit module deps.  The impetus for this is that for reasons unknown, during build some Python modules are being _installed_ into the _build_ tree, resulting in disaster.  I think this is probably due to funny business in `setup.py`, but I've given up trying to diagnose it, for the moment.  This version works around the issue by adding explicit dependencies--this seems to prevent the problem from happening.
2.  Zapped a number of deps that were present in 0.15 easyconfig.  Seems to work fine, but maybe some of these were important in some subtle way?
3.  Some of these Python modules might be Python-only.  Is there some way to not lock them to `intel-2017a`?
4.  It doesn't show here, but `sphinx_rtd_theme` won't build if `$LANG=C`.  Not sure how much that matters here, but annoying.

### DIFF
--- a/easybuild/easyconfigs/f/FSLeyes/FSLeyes-0.20.0-intel-2017a-Python-2.7.13.eb
+++ b/easybuild/easyconfigs/f/FSLeyes/FSLeyes-0.20.0-intel-2017a-Python-2.7.13.eb
@@ -1,0 +1,48 @@
+easyblock = 'PythonPackage'
+
+name = 'FSLeyes'
+version = '0.20.0'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://git.fmrib.ox.ac.uk/fsl/fsleyes/fsleyes'
+description = "FSLeyes is the FSL image viewer."
+
+toolchain = {'name': 'intel', 'version': '2017a'}
+
+source_urls = [PYPI_LOWER_SOURCE]
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [
+    ('Python', '2.7.13'),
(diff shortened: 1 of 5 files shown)


tutufan:
Is there an easy way to run locally the checks (or at least some of them) that Travis will run?  It seems silly to go through a whole 15m push/CI cycle just to get bounced because one of my comments starts with `##`.  :-/

tutufan:
Oops, didn't mean to push that last WIP.  The prior push basically seemed to work, except for GUI redraw bug on Windows 10 (probably Xming) that makes the app unusable.  Am going to try to upgrade to the latest wxPython, but that's looking like a long slog, as it uses GTK3, the build for which is very different than GTK2.

tutufan:
This could be productively added in its current state, I think.  The Windows 10 bug mentioned above is being discussed on the FSL list, but it doesn't seem to affect non-Windows platforms.

tutufan:
This PR is out of date, and unfortunately I don't currently have time to update.

PR #5655, opened 2018-01-17
add vowpal_wabbit-8.5.0, Boost-1.66.0-foss-2017b

michaelkarlcoleman:
Seems to be working.  A couple of its tests fail (out of dozens); doubt these
are significant/correct.

This version of Boost adds "all" of its libraries (except python).
vowpal_wabbit needs program_options, which wasn't being built.


easybuild/easyconfigs/b/Boost/Boost-1.66.0-foss-2017b.eb
```diff
@@ -0,0 +1,26 @@
...
+checksums = ['bd0df411efd9a585e5a2212275f8762079fed8842264954675a4fddc46cfcf60']
+
+dependencies = [
+    ('bzip2', '1.0.6'),
+    ('zlib', '1.2.11'),
+]
+
+# Beware: There are bugs in the handling of --with-libraries vs
+# --without-libraries.
+configopts = '--with-libraries=all --without-libraries=python'
```
migueldiascosta: using `-with-libraries=all` probably warrants a version suffix, to distinguish from all the other Boost easyconfigs (?)

boegel:
@michaelkarlcoleman Is it worth spending time on getting this merged, given the old toolchain?

michaelkarlcoleman:
@boegel Not for us.  Thanks.

PR #5671, opened 2018-01-19
{bio}[foss/2015b]RStudio v1.1.414(WIP)

nathanhaigh:


nathanhaigh:
@boegel locally I've attempted more than I've added here but it didn't get me much further. As it stands, the build seems to be complaining about Qt, which shouldn't be needed for the server version being compiled.

It also says that `getpeereid` is not found but I can't seem to resolve that.

FYI: A normal build of RStudio involves running the `dependencies/linux/install-dependencies-debian` script. That script installs a bunch of system packages and then calls `dependencies/common/install-common`. `install-common` runs several other scripts for downloading and unpacking several tools into the build directory. The ones I played with a bit were: `./install-dictionaries`, `./install-mathjax` and `./install-pandoc `.

migueldiascosta:
@nathanhaigh were you able to make any further progress on this?

nathanhaigh:
In a word - No :)

nathanhaigh:
FYI: RStudio Server is now at `v1.1.463`: https://www.rstudio.com/products/rstudio/download-server/

boegel:
I don't see this ever getting merged due to the ancient (and no longer supported in EasyBuild 4.x) toolchain.

@nathanhaigh Any objections to closing this? It could still serve as a starting point in case someone else wants to take another stab at this, I guess...

nathanhaigh:
No objections to closing.

PR #5680, opened 2018-01-19
{chem}[intel/2017a,intel/2018a] VASP 5.4.4 + BEEF 0.1.x + Wannier90

eta000:
VASP 5.4.4 config if interesting.

### DIFF
--- a/easybuild/easyconfigs/b/BEEF/BEEF-0.1.1-r16-intel-2017a.eb
+++ b/easybuild/easyconfigs/b/BEEF/BEEF-0.1.1-r16-intel-2017a.eb
@@ -1,0 +1,32 @@
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+# Author: Benjamin P. Roberts
+# New Zealand eScience Infrastructure
+# The University of Auckland, Auckland, New Zealand
+
+easyblock = 'ConfigureMake'
+
+name = 'BEEF'
+version = '0.1.1-r16'
+
+homepage = 'http://suncat.stanford.edu/facility/software/functional'
+description = """BEEF is a library-based implementation of the Bayesian
+Error Estimation Functional, suitable for linking against by Fortran-
+or C-based DFT codes. A description of BEEF can be found at
+http://dx.doi.org/10.1103/PhysRevB.85.235149."""
+
+toolchain = {'name': 'intel', 'version': '2017a'}
+toolchainopts = {'openmp': False, 'usempi': False}
+
+# To make a source tarball from the SVN repository:
+# 1. svn co -r 16 svn://suncatls1.slac.stanford.edu/beef
+# 2. tar --exclude \.svn -cjvf beef-0.1.1-r16.tar.bz2 beef/trunk
+sources = [SOURCELOWER_TAR_BZ2]
+
+configopts = 'CC="$CC"'
+
+sanity_check_paths = {
+    'files': ['bin/bee', 'lib/libbeef.a'],
+    'dirs': [],
+}
+
+moduleclass = 'chem'
--- a/easybuild/easyconfigs/b/BEEF/BEEF-0.1.3-2822afe-intel-2017a.eb
+++ b/easybuild/easyconfigs/b/BEEF/BEEF-0.1.3-2822afe-intel-2017a.eb
@@ -1,0 +1,30 @@
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+# Author: Espen Tangen
+# Norwegian Academic High Performance Compution collaboration
+# UiT Norwegian Arctic University
+
+easyblock = 'ConfigureMake'
+
+name = 'BEEF'
+version = '0.1.3-2822afe'
+
+homepage = 'https://github.com/vossjo/libbeef'
+description = """libbeef is a package providing the class of BEEF-vdW functionals and Bayesian error estimates to Fortran- or C-based DFT codes."""
+
+toolchain = {'name': 'intel', 'version': '2017a'}
+toolchainopts = {'openmp': False, 'usempi': False}
+
+# To make a source tarball from the git repository with unique version:
+# git clone https://github.com/vossjo/libbeef.git
+# cd libbeef && git reset --hard <commit_SHA> && cd -
+# commit_SHA = 2822afe3da255a57b078779201f2fd6582b9a5e7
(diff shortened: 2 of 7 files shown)


boegel:
@eta000 Please target the `develop` branch when making pull requests.

Can you take a look at the reported style issues? See also the output of `eb --from-pr 5680 --check-style`

migueldiascosta:
IMHO, It may be preferable to have one base easyconfig for VASP and separately add optional support for other stuff (wannier90, functionals, tools, etc.), using separate easyconfigs with a different `versionsuffix`. 

Most of them are independent and activated via separate input flags so I suppose could be bundled together, but for instance, we have separate builds with Wannier90 v1 and v2 support, because support for v2 is incomplete (the VASP wiki still says it's not supported), and last time I checked the only thing that the define `VASP2WANNIER90v2` added on top of `VASP2WANNIER90` was add an argument to `wannier_setup` so that the function signature matches...

eta000:
I must admit that I have a certain conditional disagreement with the last statement. I agree that a single base easyconfig for VASP is the better way, but I strongly disagree that we should have different versionsuffixes for the install. I do, however, agree that we should use flags - preferably in the config file - and bundle all versions together. VASP 5.4.4 do support Wan90 v2. This has been tested at our site.

I will look carefully into the 5.4.4 config and update according to feedback.

Regards

Et.

Fra: Miguel Dias Costa <notifications@github.com>
Svar til: easybuilders/easybuild-easyconfigs <reply@reply.github.com>
Dato: tirsdag 23. januar 2018 02:36
Til: easybuilders/easybuild-easyconfigs <easybuild-easyconfigs@noreply.github.com>
Kopi: Espen Tangen <espen.tangen@uit.no>, Mention <mention@noreply.github.com>
Emne: Re: [easybuilders/easybuild-easyconfigs] {chem}[intel/2017a,intel/2018a] VASP 5.4.4 + BEEF 0.1.x + Wannier90 (#5680)


IMHO, It may be preferable to have one base easyconfig for VASP and separately add optional support for other stuff (wannier90, functionals, tools, etc.), using separate easyconfigs with a different versionsuffix.

Most of them are independent and activated via separate input flags so I suppose could be bundled together, but for instance, we have separate builds with Wannier90 v1 and v2 support, because support for v2 is incomplete (the VASP wiki still says it's not supported), and last time I checked the only thing that the define VASP2WANNIER90v2 added on top of VASP2WANNIER90 was add an argument to wannier_setup so that the function signature matches...

migueldiascosta:
@eta000 to be clear, I don't have a very strong position about what should be included in the base easyconfig, mostly because I will need to locally create different versions anyway (e.g. for things like VASP TST, other functionals, GPU support, our own local modifications...), so it would be better to hear the thoughts of other people about what exactly should be in the base easyconfig. 

I'd admit that one reason for bundling as much as possible in is that it's then easier for people to comment out stuff than add it. But even better would be a vasp easyblock with those options - if a direction is agreed, I'd be willing start one.

P.S.1 I would add though, FWIW, that one should take into account the fact that VASP is a very different case than, for example, an open source integrated suite such as QuantumESPRESSO. The development process of VASP is opaque, the documentation is chronically outdated and their idea of release notes is "List of changes: We will not attempt to explicitly list changes, there are simply too many", so I would prefer to err on the conservative side.

P.S.2 Regarding Wannier90, I didn't say that VASP support for Wannier90 v2 is inexistent but that it is incomplete. It does work but it doesn't support the v2 features (e.g. spinors) and some of our users report that it fails in cases where v1 works (admittedly, the users could be doing something wrong...), hence our need for a separate build with Wannier90 v1. This requires separate easyconfigs because naturally Wannier90 modules for different versions conflict with each other. But if it's just us, no need to reflect this on the contributed easyconfig(s)/block.

eta000:
@Miguel Dias Costa<mailto:notifications@github.com>

I agree with you. Seems that we have rather similar experience with VASP and their developers. I have also been told that v6 is just around the corner, and involved developers admits that lib support and manipulation to source code should be flags in input, not at compile stage  - as it is for quite a few other material science/band, and should be.

I have also been considering an easyblock for VASP.

P.S: We have more or less decided to go for a stripped basis version of VASP as central install and let users compiler their own versions on top of that. Currently we have been asked to maintain 61 binaries for VASP 5.4.1 - unanagable for an install team; I do however have access to a makefile that does quite a lot of this, and also a wrapper that generates VASPsol, vTST and occupation matrix manipulation support. D.S.


Fra: Miguel Dias Costa <notifications@github.com>
Svar til: easybuilders/easybuild-easyconfigs <reply@reply.github.com>
Dato: torsdag 25. januar 2018 06:59
Til: easybuilders/easybuild-easyconfigs <easybuild-easyconfigs@noreply.github.com>
Kopi: Espen Tangen <espen.tangen@uit.no>, Mention <mention@noreply.github.com>
Emne: Re: [easybuilders/easybuild-easyconfigs] {chem}[intel/2017a,intel/2018a] VASP 5.4.4 + BEEF 0.1.x + Wannier90 (#5680)


@eta000<https://github.com/eta000> to be clear, I don't have a very strong position about what should be included in the base easyconfig, mostly because I will need to locally create different versions anyway (e.g. for things like VASP TST, other functionals, GPU support, our own local modifications...), so it would be better to hear the thoughts of other people about what exactly should be in the base easyconfig.

I'd admit that one reason for bundling as much as possible in is that it's then easier for people to comment out stuff than add it. But even better would be a vasp easyblock with those options - if a direction is agreed, I'd be willing start one.

P.S.1 I would add though, FWIW, that one should take into account the fact that VASP is a very different case than, for example, an open source integrated suite such as QuantumESPRESSO. The development process of VASP is opaque, the documentation is chronically outdated and their idea of release notes is "List of changes: We will not attempt to explicitly list changes, there are simply too many", so I would prefer to err on the conservative side.

P.S.2 Regarding Wannier90, I didn't say that VASP support for Wannier90 v2 is inexistent but that it is incomplete. It does work but it doesn't support the v2 features (e.g. spinors) and some of our users report that it fails in cases where v1 works (admittedly, the users could be doing something wrong...), hence our need for a separate build with Wannier90 v1. This requires separate easyconfigs because naturally Wannier90 modules for different versions conflict with each other. But if it's just us, no need to reflect this on the contributed easyconfig(s)/block.

migueldiascosta:
@eta000, this PR is being closed for the following reason(s): no activity for > 6 months.

PR #5509, opened 2017-12-12
{cae}[foss/2017b] OpenFOAM v5.0 + ParaView 5.4.1 (+ deps)

boegel:
(created using `eb --new-pr`)

~~edit: WIP because of #5344~~

PR #5697, opened 2018-01-22
{math}[foss/2017b] SCOTCH v6.0.4

smoors:


boegel:
@smoors Empty diff? This easyconfig was already added to `develop` recently via #5509

PR #5698, opened 2018-01-22
{math}[intel/2017a] Octave: add ext packages, fix mkoctfile bug

michaelkarlcoleman:
Add several commonly used Octave Forge packages to
Octave-4.2.1-intel-2017a.eb (and -mt version).  Also add PostgreSQL and Perl
dependencies these packages require.

Fix a bug in mkoctfile that prevents shared library extensions (e.g., 'image')
from compiling with intel.

(At the moment, the -mt and non-mt version segfault in Qt when X11 used.  This
is true for develop as well, so this PR doesn't appear to be the cause.)

### DIFF
--- a/easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
+++ b/easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
@@ -6,14 +6,15 @@ homepage = 'http://www.gnu.org/software/octave/'
 description = """GNU Octave is a high-level interpreted language, primarily intended for numerical computations."""
 
 toolchain = {'name': 'intel', 'version': '2017a'}
-toolchainopts = {'cstd': 'c++11'}
+toolchainopts = {'cstd': 'c++11', 'pic': True}
 
 source_urls = [GNU_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
-patches = ['%(name)s-%(version)s_intel.patch']
+patches = ['%(name)s-%(version)s_intel.patch', '%(name)s-%(version)s_mkoctfile.patch']
 checksums = [
     '80c28f6398576b50faca0e602defb9598d6f7308b0903724442c2a35a605333b',  # octave-4.2.1.tar.gz
     'f35b6e28f1ceaf7e67c935544ad2547931a30cd03ab0a1587715c6bd94521de9',  # Octave-4.2.1_intel.patch
+    'ed014514cbaf0f65642b24ae2fe2f7f29adebc0dad7b25859ed9345bca161a9c',  # Octave-4.2.1_mkoctfile.patch
 ]
 
 builddependencies = [
@@ -48,6 +49,8 @@ dependencies = [
     ('SuiteSparse', '4.5.5', '-METIS-5.1.0'),
     ('GraphicsMagick', '1.3.25'),
     ('FFTW', '3.3.6'),
+    ('GSL', '2.3'),             # for gsl extension
+    ('PostgreSQL', '9.6.2', '-Python-2.7.13'),  # for io extension
 ]
 
 # link with multi-threaded BLAS/LAPACK library
@@ -60,9 +63,31 @@ configopts += '--enable-fortran-calling-convention=gfortran'
 pkg_url = 'https://downloads.sourceforge.net/project/octave/Octave%20Forge%20Packages/Individual%20Package%20Releases/'
 exts_default_options = {'source_urls': [pkg_url]}
 
+# order matters (deps first)
 exts_list = [
     ('control', '3.0.0', {
         'checksums': ['41e9a5627565444140cf8bb8cf978239245355a6fc7d61ff122327d8f822eca6'],
+    }),
+    ('general', '2.0.0', {
+        'checksums': ['80a6921bea3437ea233873ca4382bdc7c31c9e61bd1bcb9059da18db39df7194'],
+    }),
+    ('gsl', '2.1.0', {
+        'checksums': ['940142b7d655d9196e1ec5197b4176388c69671e03ba8da717782d9f040fcedb'],
+    }),
+    ('image', '2.6.2', {
+        'checksums': ['fe01e0683dd55de985662bd86eefcf4cd0dc46c53050fd15e65d17370b2833d9'],
+    }),
+    ('io', '2.4.9', {
+        'checksums': ['519493c9635d8461fec1ecf6240ce364d6ead2906743c612fe5a33a5dc6f8f7c'],
+    }),
+    ('struct', '1.0.14', {      # required by optim
+        'checksums': ['ad4e17687bc24650f032757271b20b70fe32c35513d4dd9ab1e549919df36b47'],
+    }),
+    ('optim', '1.5.2', {
+        'checksums': ['7b36033c5581559dc3e7616f97d402bc44dde0dfd74c0e3afdf47d452a76dddf'],
+    }),
+    ('parallel', '3.1.1', {
+        'checksums': ['4cd8f4619557c91c046882b056322da8de7f6047bfce6f102f40565bd4b4b37f'],
     }),
     ('signal', '1.3.2', {
         'patches': ['signal-1.3.2_fix-formula.patch'],
@@ -70,6 +95,9 @@ exts_list = [
             'a277f12e9f47cc66a8c56ab26284ac4839b223f32de13e35e710401fd0539b60',  # signal-1.3.2.tar.gz
             'e4971524077a19e655c7c70c11d9a27ca1a3aa2638150c5779e72ebce9ec38b5',  # signal-1.3.2_fix-formula.patch
         ],
+    }),
+    ('splines', '1.3.2', {
+        'checksums': ['61cb4c8420e23e7a9febd014ecfc4314d8977635976571edaae7749f651d2b1f'],
     }),
 ]
 
--- a/easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
+++ b/easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
@@ -32,7 +32,7 @@ dependencies = [
     ('X11', '20170314'),
     ('PCRE', '8.40'),
     ('ncurses', '6.0'),
-    ('libreadline', '6.3'),
+    ('libreadline', '7.0'),
     ('arpack-ng', '3.5.0'),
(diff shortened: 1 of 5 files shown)


tutufan:
An issue for the segfault mentioned above is here: https://github.com/easybuilders/easybuild-easyconfigs/issues/5699

tutufan:
The Travis check is failing with a libreadline dep graph conflict.  Why, though, is this building without error on my setup?  (I'm using EasyBuild 3.5.1.)

Best theories:
1. Something has changed in the EasyBuild framework since 3.5.1 was released, and Travis is failing because of that.
2. Somehow a conflict was introduced into my built modules since I built them.  My 'eb' isn't spotting the conflict (because I'm only rebuilding the Octave module), but the Travis eb is (?) rebuilding everything from scratch and does spot it?

Not sure how to proceed.

tutufan:
Arguably, the assert message for the condition of having dep conflicts should not be `No conflicts detected`.  :-)

easybuild/easyconfigs/p/PostgreSQL/PostgreSQL-9.6.2-intel-2017a-Python-2.7.13.eb
```diff
@@ -0,0 +1,46 @@
...
+ sounds, or video. It has native programming interfaces for C/C++, Java,
+ .Net, Perl, Python, Ruby, Tcl, ODBC, among others, and exceptional documentation."""
+
+toolchain = {'name': 'intel', 'version': '2017a'}
+
+source_urls = ['http://ftp.postgresql.org/pub/source/v%(version)s/']
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [
+    ('libreadline', '6.3'),
```
boegel: @tutufan The problem with the conflict seems to be here: you're including `libreadline` 6.3 as dependency for `PostgreSQL`, while `Python` 2.7.13 depends on `libreadline` 7.0 (see https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/p/Python/Python-2.7.13-intel-2017a.eb#L18).

I'm not sure how this is not an issue for you... Doesn't your `Python/2.7.13-intel-2017a` module depend on `libreadline` 7.0?

The `Python-2.7.13-intel-2017a.eb` in the central repository has always depended on `libreadline` 7.0, see #4260

easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
```diff
@@ -6,14 +6,15 @@ homepage = 'http://www.gnu.org/software/octave/'
 description = """GNU Octave is a high-level interpreted language, primarily intended for numerical computations."""
 
 toolchain = {'name': 'intel', 'version': '2017a'}
-toolchainopts = {'cstd': 'c++11'}
+toolchainopts = {'cstd': 'c++11', 'pic': True}
```
boegel: Can you clarify the need to enable `-fPIC`?

easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
```diff
@@ -6,14 +6,15 @@ homepage = 'http://www.gnu.org/software/octave/'
...
 
 source_urls = [GNU_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
-patches = ['%(name)s-%(version)s_intel.patch']
+patches = ['%(name)s-%(version)s_intel.patch', '%(name)s-%(version)s_mkoctfile.patch']
```
boegel: include one per line please:

```python
patches = [
    '%(name)s-%(version)s_intel.patch',
    '%(name)s-%(version)s_mkoctfile.patch',
]

File: easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a.eb [hunk 4]
```diff
@@ -5,14 +5,15 @@ homepage = 'http://www.gnu.org/software/octave/'
 description = """GNU Octave is a high-level interpreted language, primarily intended for numerical computations."""
 
 toolchain = {'name': 'intel', 'version': '2017a'}
-toolchainopts = {'cstd': 'c++11'}
+toolchainopts = {'cstd': 'c++11', 'pic': True}
 
 source_urls = [GNU_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
-patches = ['%(name)s-%(version)s_intel.patch']
+patches = ['%(name)s-%(version)s_intel.patch', '%(name)s-%(version)s_mkoctfile.patch']
```
boegel: same here, one patch per line please

easybuild/easyconfigs/p/PostgreSQL/PostgreSQL-9.6.2-intel-2017a-Python-2.7.13.eb
tutufan: Hmm.  Not sure how this is supposed to work.  Octave depends on PostgreSQL, which in turn depends on Python.  In addition, all three directly depend on libreadline.  Octave depends on 6.3, PostgreSQL depends on 6.3, and Python depends on 7.0.  (I'm reading these deps from the lua files.)

And just to be extra annoying, `lmod` is happily sailing through this without a complaint.  It loads only 7.0.

One wrinkle that might matter is that I probably built all three of these modules in separate eb runs.  That is, I don't think that they weren't built as a result of a `--robot` dep.

How exactly does `eb` come up with the dep graph that it checks for consistency?  I'd like it to use the deps from the already-built modules that are being used in the build (together with deps for any robot modules it plans to build).  But it could also be using the deps from the easyconfig tree, and theoretically those could be somewhat out of sync with the built modules, since I have been occasionally switching from branch to branch (master, develop, and my topic branches).

I'll try to attach a log from the Octave build.  It's certainly possible that I've confused myself somehow, but not sure why neither `eb` nor `lmod` are scolding me.
[easybuild-Octave-4.2.1-20180122.132842.log.txt](https://github.com/easybuilders/easybuild-easyconfigs/files/1658035/easybuild-Octave-4.2.1-20180122.132842.log.txt)

easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
tutufan: I believe this occurs because the `image` extension package has a `.so` component.  Without this flag, I got an `ld` error.  In other words, I think `pic` is required here for the same reason that (say) `Python` requires it.

easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
tutufan: Done.

easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a.eb
tutufan: Done.

tutufan:
I went ahead and updated all of the libreadline versions to 7.0, which seems to solve the issue.

boegel:
@tutufan This PR needs a sync with `develop` to resolve merge conflicts in the Octave easyconfigs (because extensions have been added through other PRs, cfr. https://github.com/easybuilders/easybuild-easyconfigs/pull/5798).

I fixed those conflicts myself, please merge https://github.com/tutufan/easybuild-easyconfigs/pull/1 to update this PR.

easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
```diff
@@ -28,7 +32,7 @@ dependencies = [
     ('X11', '20170314'),
     ('PCRE', '8.40'),
     ('ncurses', '6.0'),
-    ('libreadline', '6.3'),
+    ('libreadline', '7.0'),
```
boegel: @tutufan We usually don't fiddle with existing dependencies because that may result in introducing conflicts elsewhere.
The reason that this conflict popped up is because of the `Python` dependency of `PostgreSQL`.
We can avoid this by using a `PostgreSQL` dependency that doesn't have `Python` bindings (and thus no `Python` dependency).

Is `PostgreSQL` strictly needed for the `io` extension btw? It built fine for me without including `PostgreSQL`, but maybe that's because we have `postgresql-libs` installed in our system...

tutufan: It's been a while, but I believe `PostgreSQL` is required.  I almost certainly chose this `PostgreSQL` because it's the only reasonably up-to-date one.  It looks to me like the only non-Python one is several years old.  Are you thinking we should create a new one without Python, or am I missing something?

boegel: Yes, I think adding a recent Python-less `PostgreSQL` and using that as a dependency here is the way forward.

easybuild/easyconfigs/o/Octave/Octave-4.2.1-intel-2017a-mt.eb
```diff
@@ -60,17 +66,38 @@ configopts += '--enable-fortran-calling-convention=gfortran'
...
+    }),
     ('signal', '1.3.2', {
         'patches': ['signal-1.3.2_fix-formula.patch'],
         'checksums': [
             'a277f12e9f47cc66a8c56ab26284ac4839b223f32de13e35e710401fd0539b60',  # signal-1.3.2.tar.gz
             'e4971524077a19e655c7c70c11d9a27ca1a3aa2638150c5779e72ebce9ec38b5',  # signal-1.3.2_fix-formula.patch
         ],
     }),
+    ('splines', '1.3.2', {
+        'checksums': ['61cb4c8420e23e7a9febd014ecfc4314d8977635976571edaae7749f651d2b1f'],
```
boegel: @tutufan missing closing `}` here, I introduced that probably, my apologies :(

boegel: @tutufan fixed in https://github.com/tutufan/easybuild-easyconfigs/pull/2

boegel:
I don't think it's worth the effort to get this merged due to the old toolchain, so closing...

Sorry for not staying on top of this @michaelkarlcoleman.

michaelkarlcoleman:
@boegel No worries.  Life gets busy.  Be well.

PR #5748, opened 2018-01-29
{bio}[foss/2018a] add MEGAHIT-1.1.2

tutufan:
New module.  This version is cpu-only.  It's also possible to compile
this with CUDA, but it seems useful to have two separate
versions.  (Still working on getting a CUDA version to compile.)

### DIFF
--- a/easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a-Python-2.7.14.eb
@@ -1,0 +1,42 @@
+easyblock = 'MakeCp'
+
+name = 'MEGAHIT'
+version = '1.1.2'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://github.com/voutcn/megahit'
+description = """An ultra-fast single-node solution for large and complex
+ metagenomics assembly via succinct de Bruijn graph"""
+
+toolchain = {'name': 'foss', 'version': '2018a'}
+
+source_urls = ['https://github.com/voutcn/%(namelower)s/archive/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['d0d3965dd49c6fdaea958ef66146cb6b30b7d51acbbfe94194c437f15a424cb5']
+
+dependencies = [
+    ('Python', '2.7.14'),
+]
+
+# This is the CPU-only version.
+#
+# We're specifying version because otherwise the Makefile will try to
+# guess it using 'git describe --tag', which seems less reliable.
+buildopts = 'version=v%(version)s verbose=1 use_gpu=0'
+
+files_to_copy = [
+    (['%(namelower)s', '%(namelower)s_asm_core', '%(namelower)s_sdbg_build', '%(namelower)s_toolkit'], 'bin'),
+    'LICENSE',
+    'README.md',
+]
+
+runtest = 'test'
+
+sanity_check_paths = {
+    'files': ['bin/%(namelower)s', 'bin/%(namelower)s_asm_core', 'bin/%(namelower)s_sdbg_build', 'bin/%(namelower)s_toolkit'],
+    'dirs': []
+}
+
+sanity_check_commands = ['%(namelower)s -v']
+
+moduleclass = 'bio'
--- a/easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a-Python-2.7.14.eb
@@ -20,7 +20,7 @@ dependencies = [
 
 # This is the CPU-only version.
 #
-# We're specifying version because otherwise the Makefile will try to
+# We're specifying 'version' because otherwise the Makefile will try to
 # guess it using 'git describe --tag', which seems less reliable.
 buildopts = 'version=v%(version)s verbose=1 use_gpu=0'
 
--- a/easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a-Python-2.7.14.eb
@@ -33,7 +33,7 @@ files_to_copy = [
 runtest = 'test'
 
 sanity_check_paths = {
-    'files': ['bin/%(namelower)s', 'bin/%(namelower)s_asm_core', 'bin/%(namelower)s_sdbg_build', 'bin/%(namelower)s_toolkit'],
+    'files': ['bin/%(namelower)s', 'bin/%(namelower)s_asm_core', 'bin/%(namelower)s_sdbg_build'],
     'dirs': []
 }
 


tutufan:
A prior Travis check bombed with
```
AssertionError: Filename '/home/travis/build/easybuilders/easybuild-easyconfigs/easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-Python-2.7.4-foss-2018a.eb' of parsed easyconfig matches expected filename 'MEGAHIT-1.1.2-foss-2018a.eb'
```

There are potentially two things wrong here.  First, a typo--actually Python-2.7.14 is being used.

The other might be a complaint about using the `-Python-2.7.4` string in the filename on stylistic grounds.  I had considered leaving it out, but it wasn't clear to me that that was a better option.  (Roughly, if that string is not present, users of this module are locked into that Python version, which might be a problem if it conflicts with a different version that they need.  Putting the string in the filename leaves the door open for supporting several popular versions.)

Trying it out after fixing the first only, it still fails, so presumably the `-Python` is not allowed.  What criterion is being used there, though?  (That is, how can I tell myself, rather than being told by Travis?)

verdurin:
Test report by @verdurin
**SUCCESS**
dell.lan - Linux fedora 27, Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz, Python 2.7.14
See https://gist.github.com/069cc6b080cf8329ea12eb3a04bbbb53 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2430.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/b487ea6554086c3a38b6915148be18a6 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2040.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/39c8927e66eaac515a11d5ba549af153 for a full test report.

verdurin:
@tutufan - normally you would use `versionsuffix` to indicate a specific language runtime dependency, such as `Python`, which would need to be reflected in the filename.

Does that clarify?

tutufan:
@verdurin Ah, derp.  I get it.  The calculated suffix is just coming from name, version, and versionsuffix, and the easyconfig filename has to match.

So, that mistake notwithstanding, I could have gone either way here.  In this case, the package absolutely depends on Python 2, but probably doesn't care that much about the specific version.  Should this be changed?

boegel:
@tutufan The easyconfig filename must be `<name>-<version>-<toolchain>-<versionsuffix>.eb` (or without the `-<versionsuffix>` if `versionsuffix` is not specified), because that's how `--robot` hunts for easyconfig files to resolve dependencies with (which it has to, since the other option would be to open & parse *every* file it runs in to...).

easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a.eb
```diff
@@ -0,0 +1,47 @@
+easyblock = 'MakeCp'
+
+name = 'MEGAHIT'
+version = '1.1.2'
```
boegel: If `Python` is a dependency, we usually make that clear via the `versionsuffix` (since `Python` is a very common dependency)

```python
versionsuffix = '-Python-%(pyver)s`
```

Note that including this implies that you need to rename the easyconfig file to end with `-Python-2.7.14.eb`...

easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a.eb
```diff
@@ -0,0 +1,47 @@
...
+ metagenomics assembly via succinct de Bruijn graph"""
+
+toolchain = {'name': 'foss', 'version': '2018a'}
+
+source_urls = ['https://github.com/voutcn/%(namelower)s/archive/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['d0d3965dd49c6fdaea958ef66146cb6b30b7d51acbbfe94194c437f15a424cb5']
+
+# This version is CPU-only.  It's also possible to compile this with
+# CUDA, but that isn't working yet (and requires libcuda.so!).
```
boegel: @tutufan Hmm, not sure if it's worth keeping this comment in?

Not having CUDA as dependency, and the `gpu=0` make it fairly clear this is a non-GPU build?

easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a.eb
```diff
@@ -0,0 +1,47 @@
...
+
+dependencies = [
+    ('Python', '2.7.14'),
+]
+
+# for GPU, 'use_gpu=1 abi=1' at least, plus CUDA dep above
+buildopts = 'version=v%(version)s verbose=1 use_gpu=0'
```
boegel: This looks a bit strange, what would happen if you don't include the `version=v%(version)s` part?
Maybe it's worth clarifying that with a comment above this line?

easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a.eb
```diff
@@ -0,0 +1,47 @@
...
+     'bin'),
+    'LICENSE',
+    'README.md',
+]
+
+# for GPU, 'test_gpu' and must use buildopts
+runtest = 'test'
+
+sanity_check_paths = {
+    'files': ['bin/%(namelower)s'],
```
boegel: please check all binaries here

easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a.eb
```diff
@@ -0,0 +1,47 @@
...
+# for GPU, 'use_gpu=1 abi=1' at least, plus CUDA dep above
+buildopts = 'version=v%(version)s verbose=1 use_gpu=0'
+
+# for GPU, more binaries?
+files_to_copy = [
+    (['%(namelower)s',
+      '%(namelower)s_asm_core',
+      '%(namelower)s_sdbg_build',
+      '%(namelower)s_toolkit'],
+     'bin'),
```
boegel: minor style suggestion:

```python
files_to_copy = [
    (['%(namelower)s', '%(namelower)s_asm_core', '%(namelower)s_sdbg_build', '%(namelower)s_toolkit'], 'bin'),
    'LICENSE',
    'README.md',
]

tutufan: I considered this, but had the idea that multi-element lists were to be one-per-line.  (In fact, you requested that in https://github.com/easybuilders/easybuild-easyconfigs/pull/5698.)  Is there an easy rule-of-thumb to remember?

easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a.eb
tutufan: In the `Makefile`, `version` will be set to `git describe --tag`.  I'm not sure that even works for release tarballs.  Even if it does, though, it potentially fails badly if the commit that has a release tag also has other tags.  Since we *know* the version (since we're using it to pull the same tag from github), it seems safest to specify it explicitly.  Will add a comment.

tutufan:
@boegel This latest version addresses your comments, I believe.  Some of the comments were breadcrumbs to me or to whomever follows, pointing out a bit of how to proceed for a CUDA built.  Anyway, those are mostly zapped.

I tried the "all on one line" approach in `sanity_check_paths`, and that's now failing the style check because the line is too long.  Not sure exactly how to proceed.  I can just fill it to 80 or 120 chars, but it seems like in some cases you want elements to be one per line?

verdurin:
@tutufan - the simple workaround here would be to remove one of the sanity check entries. It's acceptable not to check for every single binary, as long as a meaningful number are covered.

tutufan:
@verdurin That would be my thought as well, but @boegel requested above that all binaries be checked.

My personal thought would be that list formatting is mostly a matter of taste, though I'd prefer respecting PEP8 with respect to the 80-column limit.  I don't really care, though, as long as the rule is spelled out in enough detail that it's possible to hit the target without iterating.

tutufan:
Updated with @verdurin 's suggestion, in hopes of getting this in.

tutufan:
Kicking Travis, which failed on one of eight runs due to a timeout/etc during download of the source.  (It would be nice if the downloader would try a bit harder before giving up.)

easybuild/easyconfigs/m/MEGAHIT/MEGAHIT-1.1.2-foss-2018a.eb
boegel: The format I suggested makes it clearer that there's basically 3 things being copied: binaries, LICENSE & README.

There's no clear rule for this, it's mostly intuition (and potential a bike-shedding opportunity).

boegel:
@tutufan If the line becomes too long, just wrap to the next line. Dropping one if fine too, since the binaries are being copied it actually can't go wrong undetected in this case, either the copying works and the binaries are there, or the copying fails and then it never gets to the sanity check (copying won't fail silently).

Looks good as is now, thanks for your persistence. We try to be consistent (in places where Travis isn't on our behalf), but we're only human. ;-)

boegel:
Test report by @boegel
**SUCCESS**
node2504.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/d5d0dff9844a36922725f113d915e420 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2005.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/e017512f547a4c465e5317b7a6764d68 for a full test report.

boegel:
Thanks @tutufan!

PR #5754, opened 2018-01-31
update R-bundle-Bioconductor Java version, etc

tutufan:
Update Java version because the existing (1.8.0_152) is no longer
viable (i.e., downloadable).  Updated a few R packages for which the
old version no longer existed.  Also, updated the checksums for a few
modules.  Not sure why they would have changed.


verdurin:
@tutufan - actually, you can still download `1.8.0_152`, in the "Java archive" part of the Oracle site e.g. http://www.oracle.com/technetwork/java/javase/downloads/java-archive-javase8-2177648.html

Not that we shouldn't update it.

tutufan:
Thanks @verdurin !  No idea why I wasn't seeing that...

tutufan:
Outdated.

PR #5768, opened 2018-02-02
{toolchain}[dummy/dummy] goolfc v2017b

akesandgren:
(created using `eb --new-pr`)
New version of goolfc toolchain. Uses same versions of dependencies as foss/2017b but adds CUDA 9.0

PR #5769, opened 2018-02-02
{lib}[goolfc/2017b] TensorFlow 1.5.0 w/ Python 3.6.3 (built from source)

akesandgren:
(created using `eb --new-pr`)
TensorFlow 1.5.0 built from source. Uses goolfc/2017b (~~#5768~~).


akesandgren:
Don't merge this yet, I think one of the patches can be removed. Will try this soon.

easybuild/easyconfigs/t/TensorFlow/TensorFlow-1.5.0-goolfc-2017b-Python-3.6.3.eb
```diff
@@ -0,0 +1,38 @@
...
+builddependencies = [
+    ('Bazel', '0.7.0'),
+    ('wheel', '0.30.0', versionsuffix),
+]
+dependencies = [
+    ('Python', '3.6.3'),
+    ('cuDNN', '7.0.5.15'),
+]
+
+with_mkl_dnn = False
```
boegel: @akesandgren Isn't this something that could/should be handled by the `TensorFlow` easyblock? Basically `if cuDNN then disable mkl_dnn`?

akesandgren: Yes, it probably should be done in the easyblock. (As I think i said in the easyblock PR?)
But since it wasn't done I added it here for the time being.

boegel: I updated the `TensorFlow` to do this automagically now, see https://github.com/easybuilders/easybuild-easyblocks/pull/1287/commits/960e97c889c14b78a90c4b8ab384eeeeefa4d52e, so you can drop this line.

easybuild/easyconfigs/t/TensorFlow/TensorFlow-1.5.0-goolfc-2017b-Python-3.6.3.eb
```diff
@@ -0,0 +1,38 @@
...
+
+cuda_compute_capabilities = ['3.0', '3.2', '3.5', '3.7']
```
boegel: @akesandgren How did you pick this exactly? I'm not sure what the best default would be here...

akesandgren: That is what is currently useful for us. The main problem is that CUDA 9 doesn't support 2.0 which is in the default list for tensorflow 1.5 and it must be removed.
One could pick up the existing list from tensorflow, remove 2.0 when CUDA is >= 9 and be done with it.
Or just set it to the current list which i think should be 3.0, 3.2, 3.5, 3.7, 5.0 6.0, 7.0 but i didn't check the actual cuda 9 list. (And 7.0 is not supported in cuda 8)

boegel: At the very least, we should include a pointer to https://developer.nvidia.com/cuda-gpus .

As for the default value, I think it makes sense to also include 5.2, 6.0, 6.1 and 7.0 (based on the list of Tesla GPUs). Or would that cause problems when building TF on a system that doesn't support 5.x & up?

Not sure why you need 3.2?

akesandgren: 3.2 is for Jetson TK1

If the list is kept in this recipy then the full list excluding 2.0 can be used since it is based on CUDA 9

If done in the easyblock then it needs to be created based on CUDA version which is fairly easy, but needs to be updated every time a new CUDA comes out...

akesandgren: So, should i change the list in the recipy or should we move this to the easyblock?

boegel: It makes sense to me to keep this in the easyconfig file, since it's closely tied to the CUDA version being used.

We should just try and make sure we also target 'modern' cards by default, to get proper performance.

Here's what is used by default for the TF binaries: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docker/Dockerfile.devel-gpu#L88 .

See also https://github.com/tensorflow/tensorflow/issues/6914 .

Not including `7.0` is a mistake imho, I consider performance a bigger factor than code size or compilation time...

boegel:
@akesandgren lgtm now, can you look into submitting a test report?

akesandgren:
Can try to start something after coffee...

akesandgren:
Test report by @akesandgren
**SUCCESS**
b-an04.hpc2n.umu.se - Linux ubuntu 16.04, Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz, Python 2.7.12
See https://gist.github.com/72156a2fd6928a803934f60f159cd08d for a full test report.

boegel:
@wpoely86 Can you look into submitting a test report on one of your GPU systems, with the latest `TensorFlow` easyblock from https://github.com/easybuilders/easybuild-easyblocks/pull/1287?

wpoely86:
Test report by @wpoely86
**FAILED**
Build succeeded for 0 out of 11 (4 easyconfigs in this PR)
nic170 - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2683 v4 @ 2.10GHz, Python 2.7.5
See https://gist.github.com/2f87f04c1acce91e09ce3d49b4e8423a for a full test report.

boegel:
@wpoely86 `--from-pr` should merge with `develop` already when testing a PR...

akesandgren:
@wpoely86 You need to build --from-pr=5768 first (that's the compiler toolchain it uses which is new)

wpoely86:
Test report by @wpoely86
**FAILED**
Build succeeded for 2 out of 4 (4 easyconfigs in this PR)
nic170 - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2683 v4 @ 2.10GHz, Python 2.7.5
See https://gist.github.com/b9f77611d27d8440575ebe7862c5c772 for a full test report.

boegel:
@wpoely86 You need to manually download the `cuDNN` "sources".

wpoely86:
Test report by @wpoely86
**SUCCESS**
nic170 - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2683 v4 @ 2.10GHz, Python 2.7.5
See https://gist.github.com/5817bf56d49ce51245be81b03146bf9d for a full test report.

boegel:

PR #5785, opened 2018-02-06
config file for https://pre-commit.com/

pescobar:
This PR includes a config file for https://pre-commit.com/ which runs all these checks for any commit:

* Runs autopep8 over python source (also for every .eb file)
* check whether the files parse as valid python
* trims trailing whitespace
* Ensures that a file is either empty, or ends with one newline
* Check for files that contain merge conflict strings
* Check that the url to the old easybuild repo is not present

Many of these checks are also done by Travis but travis usually takes longer to run the checks and trigger the error so this could help to reduce Travis waiting time for common mistakes and also avoid commit messages like "typo fix".."style fix"...and similars one


.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+        types: [file]
+        files: \.eb$
+      - id: autopep8-wrapper
+      - id: check-ast
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-merge-conflict
+
+  - repo: https://github.com/pescobar/pre-commit-hooks-easybuild
+    sha: master
```
asottile: Sorry about the drive-by review -- was just doing some googling and came across this.

Note that branch names aren't going to be repeatable and so [they're discouraged](https://pre-commit.com/#using-the-latest-sha-for-a-repository)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
+# To use these pre-commit hooks you should first install http://pre-commit.com
+# and then run "pre-commit install" in the root folder of your local clone
+# so the hooks are deployed to .git/hooks
+#
+# See http://pre-commit.com for more information
+# See http://pre-commit.com/hooks.html for more hooks
```
asottile: just a note, we recently switched to http*s* :)

.pre-commit-config.yaml
boegel: @asottile Your feedback is much appreciated!

boegel:
> check whether the files parse as valid python

@pescobar Is that also done for easyconfig files? If so, that's not a good idea, since EasyBuild injects constants and stuff like that in the environment in which the easyconfig files are parsed, so they may not parse as Python code, but that doesn't necessarily mean they're broken...

pescobar:
@asottile thanks for the feedback! 

I have created a tag `v0.1` in https://github.com/pescobar/pre-commit-hooks-easybuild and pointed the pre-commit config to it. I also updated all the urls to use https

@boegel I have updated the config so the `check-ast` hook (python syntax check) doesn't apply to .eb files. For .eb files only pep8 is checked

asottile:
I think `check-ast` is probably a fine check @boegel ?  It doesn't actually evaluate them -- just checks that they are valid *syntax*.

All of the currently checked in `.eb` files pass this:

```console
$ git ls-files -- '*.eb' | xargs check-ast 
$ echo $?
0
$ git ls-files -- '*.eb' | wc -l
9493

```

pescobar:
@boegel I think this one is ready to merge. Please review?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+        name: 'check-ast (.eb)'
+        types: [file]
+        files: \.eb$
+      - id: autopep8-wrapper
+      - id: check-ast
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-merge-conflict
+
+  - repo: https://github.com/pescobar/pre-commit-hooks-easybuild
```
boegel: @pescobar Wouldn't it be better to host this under `easybuilders`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+    sha: v0.1
```
boegel: v0.1 is not a SHA?

asottile: It's an unfortunate mistake, this is correct though (and tags are preferred here).

When I initially made the framework it only worked with sha1 references. It's confusing, but difficult to change (the oldest open issue in our tracker is to migrate this)

pescobar:
@boegel what do you think about merging this one?

.pre-commit-config.yaml
pescobar: @boegel fine for me

.pre-commit-config.yaml
asottile: An update on this btw, the latest version of pre-commit supports `rev` as well as `sha` (originally part of [v1.7.0](https://github.com/pre-commit/pre-commit/releases/tag/v1.7.0))

.pre-commit-config.yaml
boegel: OK, then you should transfer ownership of that repo to the `easybuilders` organisation, via https://github.com/pescobar/pre-commit-hooks-easybuild/settings

pescobar: @boegel I have transferred the repo to the easybuilders organization

I have granted access permissions to all the maintainers in easybuilders. Please let me know if you think the permissions should be adjusted

pescobar:
@asottile right now to define a custom pep8 config I am including an extra file `.pep8` , see here https://github.com/pescobar/easybuild-easyconfigs/blob/a3136e8eef1d096149ff8e7eb9ddf1396408f349/.pep8

Is this the recommended approach or is there any way to include the custom pep8 config in `.pre-commit-config.yaml` ?

thanks in advance for your help!

asottile:
@pescobar I usually keep [them separate](https://github.com/pre-commit/pre-commit/blob/834ed0f229a39c986b241374f6d338632e003b5f/tox.ini#L20-L21) -- though any pep8 configuration could be set as `args: [...]` for the particular hooks.

The benefit to keeping them separate is the pre-commit hook will still read that configuration, but any other tool (IDE plugin, etc.) can also read it :)

pescobar:
@asottile thanks for the feedback! I will keep it in a separate file as you recommend

boegel:
@pescobar: You'll need to sync with `develop` first...

The PRs that got merged recently (#6045, #6047) in which you removed `.pre-commit-config.yaml` are causing conflicts between this PR and current `develop`.

pescobar:
@boegel I have synced the repo with develop

PR #5790, opened 2018-02-07
{devel}[foss/2018a] intltool 0.51.0

RvDijk:


RvDijk:
@boegel this probably should have been GCCcore, it is now included in #5810 .. Should I close this PR?

boegel:
Yes, let's close this and tackle this one via #5810.

PR #5743, opened 2018-01-29
{vis}[foss/2018a] matplotlib 2.1.2 w/ Python 2.7.14 + 3.6.4

RvDijk:

PR #5791, opened 2018-02-07
{bio}[foss/2018a] QUAST 4.6.3

RvDijk:
Depending on ~~#5743~~


easybuild/easyconfigs/q/QUAST/QUAST-4.6.3-foss-2018a-Python-3.6.4.eb
```diff
@@ -0,0 +1,42 @@
...
+checksums = ['d7f5e670563d17d683f6df057086f7b816b6a088266c6270f7114a1406aaab63']
+
+install_target = 'install_full'
+
+dependencies = [
+    ('Python', '3.6.4'),
+    ('Perl', '5.26.1'),
+    ('matplotlib', '2.1.2', versionsuffix),
+    ('Java', '1.8.0_162', '', True),
+    ('Boost', '1.66.0', versionsuffix),
```
boegel: @RvDijk Can you check whether `QUAST` really needs Boost with Python bindings?

We are trying to limit the number of dependency variants to just 1 per toolchain generation (cfr. #5970), and we already have `2018a` easyconfigs using Boost 1.66.0 without Python bindings used as a dependency...

RvDijk: @boegel not using python bindings seems fine.

easybuild/easyconfigs/q/QUAST/QUAST-4.6.3-foss-2018a-Python-3.6.4.eb
```diff
@@ -0,0 +1,42 @@
...
+
+libraries = ['bedtools', 'bwa', 'ca_utils', 'gage', 'genemark', 'genemark-es', 'glimmer', 'gnuplot', 'html_saver',
+             'kronatools', 'manta0.29.6', 'MUMmer', 'ra_utils', 'sambamba', 'silva']
+sanity_check_paths = {
+    'files': ['bin/%(namelower)s.py', 'bin/meta%(namelower)s.py'],
+    'dirs': [('lib/python%%(pyshortver)s/site-packages/%%(namelower)s-%%(version)s-py%%(pyshortver)s.egg/'
+              '%%(namelower)s_libs/%s') % x for x in libraries]
+}
+
+sanity_check_commands = ['%(namelower)s.py --version']
```
boegel: @RvDijk This was changed from `--test` to `--version`, probably to dance around this:

```
You are probably running QUAST installed via pip, which does not include test data.
This is fine, just start using QUAST on your own data!

If you still want to run tests, please download and unpack test data to CWD:
wget quast.sf.net/test_data.tar.gz && tar xzf test_data.tar.gz
```

Shouldn't we download that test data and run the tests instead?

RvDijk: probably 😀

boegel:
Going in, thanks @RvDijk!

PR #5882, opened 2018-02-22
{phys}[dummy/dummy] COMSOL v5.3a

boegel:
(created using `eb --new-pr`)

edit: requires ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1317~~


easybuild/easyconfigs/c/COMSOL/COMSOL-5.3a.eb
```diff
@@ -0,0 +1,14 @@
+name = 'COMSOL'
+version = '5.3a'
+
+homepage = 'https://www.comsol.se/'
+description = """COMSOL Multiphysics - Simulation Tool for Electrical, Mechanical, Fluid Flow, and Chemical
+Applications"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+sources = [SOURCELOWER_TAR_GZ]
```
migueldiascosta: missing checksum (also, a comment on where to find this file?)

migueldiascosta:
@boegel, this PR is being closed for the following reason(s): closing and reopening to trigger tests.

easybuild/easyconfigs/c/COMSOL/COMSOL-5.3a.eb
boegel: The file has to be created manually from the `COMSOL53a_dvd.iso` (so including a checksum is useless).

But, if we fix https://github.com/easybuilders/easybuild-framework/issues/1376, we can consider using the ISO file as a source, and unpacking that instead?

boegel: I've updated the PR to use `None` as checksum, which is a supported escape hatch that is acceptable in this case imho...

boegel:
I can't submit a test report for this one, because I don't have access to the license myself.

I did use this easyconfig to install COMSOL for researchers who do have access to a license however, so it's good to go...

boegel:
Obsolete, see #11513

PR #5689, opened 2018-01-21
Add Tkinter sanity check to matplotlib (foss toolchains)

schiotz:
(created using `eb --new-pr`)

PR #5896, opened 2018-02-22
add sanity check command in matplotlib easyconfigs built with intel toolchain to ensure Tkinter is available

boegel:
follow-up for ~~#5689~~ (cc @schiotz)

I was forced to also fiddle with the `freetype` dependency in some `matplotlib` easyconfig to avoid introducing conflicts (since `Tkinter` indirectly depends on `freetype`).
We usually don't do that, but here that was the more sensible option, especially since it doesn't introduce any conflicts elsewhere (where `matplotlib` is used as a dependency).


boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 19 out of 19 (18 easyconfigs in this PR)
node2469.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/bac92550a7c42a9fddca731abd9b7f1c for a full test report.

PR #5938, opened 2018-03-01
fix installation of TensorFlow 1.3 via binary wheel after introducing TensorFlow easyblock

boegel:
Workaround for framework issue https://github.com/easybuilders/easybuild-framework/issues/2417, this only started occurring after the custom easyblock for `TensorFlow` was added in https://github.com/easybuilders/easybuild-easyblocks/pull/1287.

Changing the extension name to `tensorflow` rather than `TensorFlow` makes the `import` of the  class from `tensorflow.py` fail (because there's no `EB_tensorflow` class, only `EB_TensorFlow`).

Since `tensorflow` is also used as extension name in older `TensorFlow` easyconfigs (v1.2.x) I feel this  is an acceptable workaround for now...

### DIFF
--- a/easybuild/easyconfigs/p/PGI/PGI-15.10-GCC-4.9.3-2.25.eb
+++ b/easybuild/easyconfigs/p/PGI/PGI-15.10-GCC-4.9.3-2.25.eb
@@ -6,6 +6,7 @@ description = "C, C++ and Fortran compilers from The Portland Group - PGI"
 
 toolchain = {'name': 'dummy', 'version': ''}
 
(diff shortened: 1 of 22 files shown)


boegel:
Test report by @boegel
**SUCCESS**
gligar03.gligar.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/ea0b0e4d1b203717a22f046e1510d9f1 for a full test report.

PR #5946, opened 2018-03-03
Fix "unescaped left brace" in Automake 1.15 so that Perl >5.22 is supported

froland:


easybuild/easyconfigs/a/Automake/Automake-1.15-fix-brace.patch
```diff
@@ -0,0 +1,10 @@
+--- automake-1.15/bin/automake.in.orig	2015-01-06 03:25:55.000000000 +0800
```
boegel: @froland Did you create this patch yourself, or did you get it somewhere else?

Please mention the author or upstream reference above this line (anything above this line are comments ignored by `patch`).

boegel:
@froland We should apply this patch consistently in all `Automake` 1.15 easyconfigs? Is it still relevant for `Automake` 1.15.1?

Can you mention the exact error you ran into that this patch fixes in the PR description (so people will be pointed here when searching the error)?

froland:
Hi @boegel,

I added some details and the original author of the patch.

There is no need to apply this patch to newer version of Automake as it was included upstream.

Regards

boegel:
@froland OK, but we should apply the patch consistently in all Automake 1.15 easyconfigs?

boegel:
(merged via #6358)

PR #5952, opened 2018-03-06
Builds of bgen and QCTOOL using hg hash from bitbucket as best approx…

edmondac:
…imation to the versions

### DIFF
--- a/easybuild/easyconfigs/b/bgen/bgen-1.0.2-iomkl-2018a.eb
+++ b/easybuild/easyconfigs/b/bgen/bgen-1.0.2-iomkl-2018a.eb
@@ -4,7 +4,10 @@ name = 'bgen'
 version = '1.0.2'
 
 homepage = "https://bitbucket.org/gavinband/bgen"
-description = """A reference implementation of the BGEN format in C++, originally sourced from the qctool implementation. In addition three utilities - bgenix, which provides indexed access to BGEN files, cat-bgen which efficiently concatenates BGEN files, and edit-bgen which is used to manipulate BGEN metadata - are provided. See the wiki for documentation on these programs."""
+description = """A reference implementation of the BGEN format in C++, originally sourced from
+ the qctool implementation. In addition three utilities - bgenix, which provides indexed access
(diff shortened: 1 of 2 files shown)


akesandgren:
Rekicking travis due to changes in EasyBuild

akesandgren:
Travis kicked

easybuild/easyconfigs/b/bgen/bgen-1.0.2-iomkl-2018a.eb
```diff
@@ -0,0 +1,27 @@
...
+homepage = "https://bitbucket.org/gavinband/bgen"
+description = """A reference implementation of the BGEN format in C++, originally sourced from
+ the qctool implementation. In addition three utilities - bgenix, which provides indexed access
+ to BGEN files, cat-bgen which efficiently concatenates BGEN files, and edit-bgen which is used
+ to manipulate BGEN metadata - are provided. See the wiki for documentation on these programs."""
+
+toolchain = {'name': 'iomkl', 'version': '2018a'}
+
+sources = ['32abb11ef9dd.tar.gz']
+source_urls = ['https://bitbucket.org/gavinband/bgen/get/']
```
akesandgren: It might be preferable to make use of the filename/download_filename syntax for sources to give the downloaded file a name that matches the version in the easyconfig file. See https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html?highlight=sources%20specification#alternative-formats-for-sources

And the same for qctool

akesandgren: And add the missing checksums.

zao:
A note for the future - Bitbucket plans to to [remove](https://bitbucket.org/blog/sunsetting-mercurial-support-in-bitbucket) all Mercurial-based repositories by 2020-06-01. Both `bgen` and `QCTOOL` are Mercurial repositories, so these sources may be lost by then.

edmondac:
This is quite old, and not worth completing IMO.

PR #5967, opened 2018-03-07
sync with master after EasyBuild v3.5.3 release

boegel:


vanzod:
Going in, thanks @boegel!

PR #6015, opened 2018-03-14
{bio}[foss/2018a] BLAST+ v2.7.1 + Boost v1.63

Serpens:


Serpens:
I can deal with NASM, libjpeg-turbo and libpng variants, but what should I do about Boost? Should I find a way to compile BLAST+ with 1.66, or use some workaround?

boegel:
@Serpens Hmm, good question... What kind of problems are you running into with Boost 1.66.0 as dependency for BLAST+?

If there's no other way, we can whitelist this as an exception, but then we need a good way to deal with exceptions...

Serpens:
With Boost 1.66, compillation fails:
https://gist.github.com/Serpens/04b2653d7da9796584f5cd59668ca2a5#file-gistfile1-txt

It's possible to configure it --without-boost, but I don't think it's a good idea.

boegel:
@Serpens Can you check whether the `Boost` libraries can be linked statically in `BLAST+`?

We will seen need to look into some form of whitelisting for the check, but that way we can at least avoid that the `BLAST+` module we obtain is incompatible with other easyconfigs that require Boost `1.66.0`...

Serpens:
@boegel Ok, I'll try to do it.

It looks to me that only parts that are incompatible with `Boost 1.66.0` are some unit tests. It's risky, but it might be safe to remove them.

boegel:
@Serpens That's worth considering, if we have other good ways to test `BLAST+`.

Serpens:
`BLAST+` doesn't give an option to statically link `Boost` and keep dynamic linking for other libraries. I changed a script in configuration stage to do it. I checked executables and shared libraries with `ldd` and `objdump`, they don't need `Boost`.

I also checked `BLAST+/2.6.0-foss-2016b-Python-2.7.12`, it also doesn't reference `Boost`, so I'm not sure this change is needed.

Serpens:
@boegel What should I do to get this pr accepted?

easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1_static_boost.patch
```diff
@@ -0,0 +1,105 @@
+diff -Nru ncbi-blast-2.7.1+-src.orig/c++/src/build-system/configure.ac ncbi-blast-2.7.1+-src/c++/src/build-system/configure.ac
```
boegel: @Serpens Please include some comments above this line, to explain what this patch does (and why) + mention author or reference to source of this patch.

easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1-foss-2018a-Python-2.7.14.eb
```diff
@@ -0,0 +1,59 @@
...
+toolchain = {'name': 'foss', 'version': '2018a'}
+toolchainopts = {'usempi': True}
+
+source_urls = ['http://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/%(version)s/']
+sources = ['ncbi-blast-%(version)s+-src.tar.gz']
+checksums = [
+    '10a78d3007413a6d4c983d2acbf03ef84b622b82bd9a59c6bd9fbdde9d0298ca',  # ncbi-blast-2.7.1+-src.tar.gz
+]
+
+patches = ['%(name)s-%(version)s_static_boost.patch']
```
boegel: @Serpens Please move this up above `checksums`, and also include a SHA256 checksum for the patch file (after you include the requested comments).

`eb --inject-checksums` is your friend here

easybuild/easyconfigs/b/Boost/Boost-1.63.0-foss-2018a-Python-2.7.14.eb
```diff
@@ -0,0 +1,23 @@
...
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'http://www.boost.org/'
+description = "Boost provides free peer-reviewed portable C++ source libraries."
+
+toolchain = {'name': 'foss', 'version': '2018a'}
+toolchainopts = {'usempi': True, 'pic': True}
+
+sources = ['%%(namelower)s_%s.tar.gz' % '_'.join(version.split('.'))]
+source_urls = ['https://sourceforge.net/projects/%(namelower)s/files/%(namelower)s/%(version)s']
```
boegel: @Serpens Please include a SH256 checksum here too.

Serpens:
I found another way. BLAST+ 2.8 (currently alpha) supports Boost 1.66. I replaced 2 files that caused errors with their 2.8 versions and it works.

easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1_boost_backport.patch
```diff
@@ -0,0 +1,157 @@
+backporting Boost 1.66 support from BLAST+ 2.8 alpha
```
boegel: @Serpens Do you have an upstream reference (URL) for this patch? Please include it...

easybuild/easyconfigs/b/Boost/Boost-1.63.0-foss-2018a-Python-2.7.14.eb
```diff
@@ -0,0 +1,23 @@
+name = 'Boost'
+version = '1.63.0'
```
boegel: @Serpens This easyconfig file can be removed now?

easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1_boost_backport.patch
Serpens: @boegel I got the idea from this discussion https://aur.archlinux.org/packages/blast%2B/ but I did it on my own, so no upstream reference. Should I find who and when added Boost 1.66 support in NCBI's SVN?

boegel: If the patch is based on an idea and you created it yourself, just including yourself as an author is sufficient imho.

So, let's just get rid of the easyconfig for Boost 1.63.0?

easybuild/easyconfigs/b/Boost/Boost-1.63.0-foss-2018a-Python-2.7.14.eb
Serpens: Yes

boegel:
Test report by @boegel
**SUCCESS**
node2111.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/bf64194e48251c07162267173d3f7ee4 for a full test report.

boegel:
Going in, thanks @Serpens!

PR #5658, opened 2018-01-17
{foss/2017b} add Tkinter as dependency for ASE & matplotlib using Python 3.6.3

schiotz:
(created using `eb --new-pr`)

PR #6024, opened 2018-03-15
use non-interactive plotting backend by default for matplotlib 2.1.2

boegel:
Without this, you run into this error (ever since including `Tkinter` as dependency to restore interactive plotting capabilities, cfr. #5658, #5689) when trying to make plots in an environment where X11 forwarding is not enabled.

```
>>> import numpy as np
>>> import matplotlib.pyplot as pl
>>> pl.plot(np.random.random(3), np.random.random(3))
  File "<stdin>", line 1, in <module>
  File "/apps/gent/CO7/skylake-ib-PILOT/software/matplotlib/2.1.2-intel-2018a-Python-2.7.14/lib/python2.7/site-packages/matplotlib-2.1.2-py2.7-linux-x86_64.egg/matplotlib/pyplot.py", line 3250, in plot
    ax = gca()
  File "/apps/gent/CO7/skylake-ib-PILOT/software/matplotlib/2.1.2-intel-2018a-Python-2.7.14/lib/python2.7/site-packages/matplotlib-2.1.2-py2.7-linux-x86_64.egg/matplotlib/pyplot.py", line 962, in gca
    return gcf().gca(**kwargs)
  File "/apps/gent/CO7/skylake-ib-PILOT/software/matplotlib/2.1.2-intel-2018a-Python-2.7.14/lib/python2.7/site-packages/matplotlib-2.1.2-py2.7-linux-x86_64.egg/matplotlib/pyplot.py", line 592, in gcf
    return figure()
  File "/apps/gent/CO7/skylake-ib-PILOT/software/matplotlib/2.1.2-intel-2018a-Python-2.7.14/lib/python2.7/site-packages/matplotlib-2.1.2-py2.7-linux-x86_64.egg/matplotlib/pyplot.py", line 539, in figure
    **kwargs)
  File "/apps/gent/CO7/skylake-ib-PILOT/software/matplotlib/2.1.2-intel-2018a-Python-2.7.14/lib/python2.7/site-packages/matplotlib-2.1.2-py2.7-linux-x86_64.egg/matplotlib/backend_bases.py", line 171, in new_figure_manager
    return cls.new_figure_manager_given_figure(num, fig)
  File "/apps/gent/CO7/skylake-ib-PILOT/software/matplotlib/2.1.2-intel-2018a-Python-2.7.14/lib/python2.7/site-packages/matplotlib-2.1.2-py2.7-linux-x86_64.egg/matplotlib/backends/backend_tkagg.py", line 1049, in new_figure_manager_given_figure
    window = Tk.Tk(className="matplotlib")
  File "/apps/gent/CO7/skylake-ib-PILOT/software/Python/2.7.14-intel-2018a/lib/python2.7/lib-tk/Tkinter.py", line 1819, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: couldn't connect to display "localhost:21.0"
```

I consider it more common to use`matplotlib` in a non-interactive environment, so the default should be to be able to create plots non-interactively imho...

@schiotz: thoughts?

### DIFF
--- a/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-foss-2018a-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-foss-2018a-Python-2.7.14.eb
@@ -52,4 +52,8 @@ sanity_check_commands = ["""python -c 'import matplotlib; matplotlib.use("TkAgg"
 
 modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
 
+# use non-interactive plotting backend as default
+# see https://matplotlib.org/tutorials/introductory/usage.html#what-is-a-backend
+modextravars = {'MPLBACKEND': 'Agg'}
+
 moduleclass = 'vis'
--- a/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-foss-2018a-Python-3.6.4.eb
+++ b/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-foss-2018a-Python-3.6.4.eb
@@ -52,4 +52,8 @@ sanity_check_commands = ["""python -c 'import matplotlib; matplotlib.use("TkAgg"
 
 modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
 
+# use non-interactive plotting backend as default
+# see https://matplotlib.org/tutorials/introductory/usage.html#what-is-a-backend
+modextravars = {'MPLBACKEND': 'Agg'}
+
 moduleclass = 'vis'
--- a/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-intel-2018a-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-intel-2018a-Python-2.7.14.eb
@@ -52,4 +52,8 @@ sanity_check_commands = ["""python -c 'import matplotlib; matplotlib.use("TkAgg"
 
 modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
 
+# use non-interactive plotting backend as default
+# see https://matplotlib.org/tutorials/introductory/usage.html#what-is-a-backend
+modextravars = {'MPLBACKEND': 'Agg'}
+
 moduleclass = 'vis'
--- a/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-intel-2018a-Python-3.6.4.eb
+++ b/easybuild/easyconfigs/m/matplotlib/matplotlib-2.1.2-intel-2018a-Python-3.6.4.eb
@@ -52,4 +52,8 @@ sanity_check_commands = ["""python -c 'import matplotlib; matplotlib.use("TkAgg"
 
 modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
 
+# use non-interactive plotting backend as default
+# see https://matplotlib.org/tutorials/introductory/usage.html#what-is-a-backend
+modextravars = {'MPLBACKEND': 'Agg'}
+
 moduleclass = 'vis'


wpoely86:
No problem for me. I assume most users already overwrite this to get it to work.

boegel:
Test report by @boegel
**SUCCESS**
node2706.swalot.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/a1ed67e9052f5d71adf753a27f78d328 for a full test report.

schiotz:
I am pretty sure you are right, this is necessary.  People can then set their own backend as desired.  I will check if ASE needs updates to work interactively after this.

@mikstr Any thoughts?

boegel:
Test report by @boegel
**SUCCESS**
node2036.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/031a2e8a5f632dea87a668fc23bbddd8 for a full test report.

migueldiascosta:
it seems to me that for those doing mostly non-interactive stuff, having made `matplotlib` depend on `Tkinter` defeated the purpose of fleshing it out of `Python` in the first place, but it was't really necessary, was it?

For me, interactive plots can be enabled by additionally loading `Tkinter` even if `matplotlib` was built without it, and the `ASE` gui also works as long as it depends on `Tkinter`...

Admitedly, some users may expect to load `matplotlib` and have interactive plotting enabled without needing to load anything else... but a e.g. `matplotlib-tk` bundle of `matplotlib` and `Tkinter` would solve this, imho...

Am I missing something?

schiotz:
If matplotlib does not depend on Tkinter, the Tkinter backend was not build for me (obviously I cannot 100% exclude that I did something stupid).

As for the idea of a `matplotlib-tk` bundle, that sounds like a good idea!  Ideally, it would depend on matplotlib and Tkinter, and contain the missing Tkinter backend for matplotlib.  Creating this is beyond my EasyBuild abilities, however.

migueldiascosta:
@schiotz strange... this is what I see with `matplotib` built without `Tkinter`:

``` bash
$ module load matplotlib/2.1.2-intel-2018a-Python-2.7.14
$ python -c "import matplotlib; matplotlib.use('TkAgg'); import matplotlib.pyplot" && echo "Works"
...
ImportError: No module named _tkinter
$ module load Tkinter/2.7.14-intel-2018a-Python-2.7.14
$ python -c "import matplotlib; matplotlib.use('TkAgg'); import matplotlib.pyplot" && echo "Works"
Works
```

schiotz:
@migueldiascosta OK, as I said I may have done something stupid.  I will investigate, but perhaps not today :-)

boegel:
I would expect that interactive plotting just works with `matplotlib`, assuming I'm in an GUI-enabled environment (e.g. X11 forwarding). Having to load a separate module for that would cause confusion I think...
I'm mainly talking from the point of view of user support here: how can I avoid getting support questions on `matplotlib`? I think this proposed change is the way forward: have it work by default (plot stuff to file), while leaving the option to enable interactive plotting while changing the backend (which is well documented in the matplotlib documentation).

The main reason for fleshing out `Tkinter` from `Python` was to avoid that `Python` depends on `X11` (indirectly), that was just wrong imho. People that want to install `Python` but have no interest in `matplotlib` (or `ASE`), should not be forced to install `X11` first.

schiotz:
I would tend to agree with @boegel here.  The price of having matplotlib depend on Tkinter is only paid when installing matplotlib (which is quite slow when X11 is pulled in).  Having two different matplotlib packages will in the end cause confusion, as will having a matplotlib that fails with a cryptic error if the user has not realized that he should manually load Tkinter.

By the way, the X11 build time could be reduced significantly.  Around half the build time is because building X11 requires perl, and when building perl it runs an hour-long self-test.  Maybe that is not strictly necessary.

boegel:
I would expect that interactive plotting just works with `matplotlib`, assuming I'm in an GUI-enabled environment (e.g. X11 forwarding). Having to load a separate module for that would cause confusion I think...
I'm mainly talking from the point of view of user support here: how can I avoid getting support questions on `matplotlib`? I think this proposed change is the way forward: have it work by default (plot stuff to file), while leaving the option to enable interactive plotting while changing the backend (which is well documented in the matplotlib documentation).

The main reason for fleshing out `Tkinter` from `Python` was to avoid that `Python` depends on `X11` (indirectly), that was just wrong imho. People that want to install `Python` but have no interest in `matplotlib` (or `ASE`), should not be forced to install `X11` first.

migueldiascosta:
@boegel wouldn't the same argument for `Python` apply to those that are interested in `matplotlib` but not  via a gui (at least, not running out of a remote cluster)?

I do understand the user support point of view, that's why I suggested a "`matplotlib-tk`" bundle

But ok, I suppose commenting `Tkinter` in `matplotlib` and creating a bundle can be left to local configurations :)

migueldiascosta:
more to the point, I also see no problem with this PR, we're already doing that anyway

boegel:
@migueldiascosta In some way, you're right, but for `matplotlib` it makes a lot more sense than `X11` is required (since the default plotting mechanism is to open a window). You can't bring that same argument forward for `Python` imho...

If there are no objections, please merge this PR. :)

migueldiascosta:
Going in, thanks @boegel!

PR #6042, opened 2018-03-19
Hint on file name differences between downloaded and expected by config

jhein32:
Added hint on file name differences


jhein32:
Is there any interest left in this PR?

PR #6048, opened 2018-03-20
update Time::HiRes extension in Perl 5.24.0 easyconfig using foss/2016b to fix BioPerl problems

pescobar:
the bioperl tests were crashing until I upgraded this (very old) extension

Couldn't install bioperl until I applied this upgrade


boegel:
@pescobar I'm not a big fan of applying this now, especially since this easyconfig is quite old.

Maybe it's good enough as a reference in case someone runs into the same issue?

Can you eloborate on the exact problems you were seeing with bioperl? Mention them in the description, so people will hit this when searching for the problem...

pescobar:
@boegel please check the file `changes` inside the tarball. We are providing a library released in 90's ;) 
http://www.cpan.org/authors/id/D/DE/DEWEG/Time-HiRes-01.20.tar.gz

You can also find the dates here
http://search.cpan.org/~deweg/Time-HiRes-01.20/HiRes.pm#REVISION

The new tarball was released 2018. Check `changes` inside
http://search.cpan.org/CPAN/authors/id/J/JH/JHI/Time-HiRes-1.9757.tar.gz

IMHO, we should update this. I am surprised this problem hasn't appeared before. The error message in the logs was not clear. I had to dig for quite some time in the logs and spend some google time and try&error until I figured out the problem.

boegel:
@pescobar Which is exactly why you should mention the errors you were seeing, so people will find this issue and quickly realise what the problem is. ;)

If you feel strongly about it, I'm willing to consider it.

pescobar:
this is the error

```
Exception
------------- EXCEPTION: Bio::Root::Exception -------------
MSG: Failed to load module Bio::ClusterIO::dbsnp. "tv_interval" is not exported by the Time::HiRes module
 "gettimeofday" is not exported by the Time::HiRes module
Can't continue after import errors at Bio/ClusterIO/dbsnp.pm line 72.
BEGIN failed--compilation aborted at Bio/ClusterIO/dbsnp.pm line 72.
Compilation failed in require at Bio/Root/Root.pm line 518.
```

Personally I think it makes sense to upgrade a so old library. Funny that I am trying to convince you to upgrade a library from 90's ;)))  it's usually the other way around when you want to upgrade to latest versions and I am the conservative guy who wants to old versions :))

boegel:
@pescobar Turns out `Time::HiRes` was included by mistake, see #6225...

PR #6055, opened 2018-03-21
{bio} [foss/2016b] Chewbbaca 2.0.8 + deps

pescobar:
I guess this won't be accepted upstream because I am using `use_pip_for_deps = True` but I send the PR in case the main easyconfig or the deps are useful for anyone


akesandgren:
We have BLAST+-2.7.1 for various foss/intel toolchains, why not use one of those as base instead of using a tarball install?
We also already have prodigal-2.6.3-GCCcore-6.4.0.eb perhaps move this one down to GCCcore too?

boegel:
There's little hope for getting this merged now (old toolchain, and needs works as @pescobar already mentioned), so closing as is...

PR #6078, opened 2018-03-27
{bio} [foss/2016b] Racon-1.2.0

pescobar:
I haven't added checksums because the tarball has been generated using `git clone --recursive`

PR #6079, opened 2018-03-27
{bio} [foss/2016b] Unicycler-0.4.4 Python-3.5.2

pescobar:
Racon dependency is in PR https://github.com/easybuilders/easybuild-easyconfigs/pull/6078, and BLAST+ in #6055

### DIFF
--- a/easybuild/easyconfigs/u/Unicycler/Unicycler-0.4.4-foss-2016b-Python-3.5.2.eb
+++ b/easybuild/easyconfigs/u/Unicycler/Unicycler-0.4.4-foss-2016b-Python-3.5.2.eb
@@ -34,6 +34,4 @@ sanity_check_paths = {
     'dirs': []
 }
 
-options = {'modulename': 'os'}
-
 moduleclass = 'bio'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/u/Unicycler/Unicycler-0.4.4-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,37 @@
...
+
+source_urls = ['https://github.com/rrwick/Unicycler/archive/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['e25e7423119aec5e574324c403f8ae36d93c0ce1497853d42c1a49a2a924a1f4']
+
+dependencies = [
+    ('Python', '3.5.2'),
+    ('Bowtie2', '2.3.2'),
+    ('SAMtools', '1.3.1'),
+    ('Pilon', '1.22', '-Java-1.8.0_144', True),
```
akesandgren: Should probably use -Java-1.8 as dependency instead and make use of the Java version wrapper.

easybuild/easyconfigs/u/Unicycler/Unicycler-0.4.4-foss-2016b-Python-3.5.2.eb
```diff
@@ -0,0 +1,37 @@
...
+    ('SPAdes', '3.9.0'),
+    ('Racon', '1.2.0'),
+    ('BLAST+', '2.7.1', '', True),
```
akesandgren: And if the corresponding #6055 changes BLAST+ to use foss toolchain this dep need to be changed too.

akesandgren:
Closing due to old toolchain and the fact that we have Unicycler-0.4.8 available

PR #6111, opened 2018-04-04
Add Lua using standard install instructions, custom easyblock

rubendibattista:
Replace #5893.


rubendibattista:
Needs: https://github.com/easybuilders/easybuild-easyblocks/pull/1396

PR #6106, opened 2018-03-30
Add Clang6 with GCC 7.2.0 + deps

rubendibattista:

PR #6121, opened 2018-04-05
Add GLib 2.56 with GCC-7.2.0 and deps

rubendibattista:
Requires deps from #6106


boegel:
Since this requires #6106 which was just closed, I'll go ahead and close this too, usefulness is limited at this point.

Sorry for not staying on top of this @rubendibattista ...

PR #6122, opened 2018-04-05
include pkg-config as build dep in recent R easyconfigs (required for atleast nloptr)

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #6143, opened 2018-04-11
{toolchain} - gmpich 2017.08

rubendibattista:


easybuild/easyconfigs/g/gmpich/gmpich-20180411.eb
```diff
@@ -0,0 +1,22 @@
+easyblock = "Toolchain"
+
+name = 'gmpich'
+version = '20180411'
```
boegel: @rubendibattista For toolchains we usually use a `<year>.<month>` versioning scheme, based on the release date of the most recent toolchain component (as opposed to the date when the toolchain was defined)

boegel:
Going in, thanks @rubendibattista!

PR #6174, opened 2018-04-18
LAMMPS-20180316-goolf-1.7.20

VilleS1:
(created using `eb --new-pr`)


boegel:
@VilleS1 Thanks for the contribution! `LAMMPS` is a big interest, since it's one of the 'big' applications we don't support yet...

This easyconfig looks quite clean, in contrast to the easyblock we were working on, see https://github.com/easybuilders/easybuild-easyblocks/pull/594 .

Do you know how 'complete' it is, i.e. whether this provides a fully fledged out `LAMMPS` installation?

In addition, it seems like some dependencies are missing, in particular `Python` and `libpng` (and also `tbb`). Any reason you're not including those?

Furthermore, the `version` you're using is currently derived from the source tarball, but it won't sort in the output of `module avail LAMMPS` well when multiple versions of `LAMMPS` are installed...

We were planning to use versions like `20170331` for example, which would map to `31Mar17`.

Finally, any reason you're not using the latest stable version (`20180316` a.k.a. `16Mar2018`) which is available from https://github.com/lammps/lammps? The SourceForge page only includes old releases (as indicates in the README).

VilleS1:
Do I know how complete it is? No.

I tried to fix the depndencies, except tbb, maybe we can live without it at this point?

Made a new pr #6190

boegel:
@VilleS1 Please update this PR rather than opening a new one, you can use `eb --update-pr 6174` and provide it with the changed easyconfig file...

VilleS1:
Ok, done.

VilleS1:
Is the build error a travis crash? "already in digraph"

easybuild/easyconfigs/l/LAMMPS/LAMMPS-20180316-goolf-1.7.20.eb
```diff
@@ -0,0 +1,40 @@
...
+source_urls = ['https://github.com/lammps/lammps/archive/']
+sources = ['stable_16Mar2018.tar.gz']
+
+builddependencies = [
+    ('Python', '2.7.11'),
+    ('libpng', '1.6.21'),
+]
+
+dependencies = [
+    ('Python', '2.7.11'),
```
boegel: @VilleS1 You have `Python` listed both a build dependency and (runtime) dependency, that's why the tests are failing.

easybuild/easyconfigs/l/LAMMPS/LAMMPS-20180316-GCC-6.4.0-2.28.eb
```diff
@@ -0,0 +1,48 @@
+easyblock = "MakeCp"
+
+name = "LAMMPS"
+version = "20180316"
+
+homepage = 'http://lammps.sandia.gov/'
+description = """LAMMPS is a classical molecular dynamics code,
+and an acronym for Large-scale Atomic/Molecular Massively Parallel Simulator."""
+
+toolchain = {'name': 'GCC', 'version': '6.4.0-2.28'}
```
boegel: @VilleS1 Using a compiler-only toolchain like `GCC` for `LAMMPS` doesn't make much sense?

I suggest removing this easyconfig file.

boegel:
@VilleS1 Based on your work here, I've opened a pull request for LAMMPS built with the common `foss/2018` and `intel/2018a` toolchains, see #6233

It's a little bit different/better/cleaner:

* no patch file needed: you can just use arguments to the `make` command instead; that's better because then we don't need to maintain the patch
* uses the generic `$MPI_INC_DIR` and `$MPI_LIB_DIR` which are defined in the build environment by EasyBuild, based on the MPI library used in the toolchain
* also enables support for jpg, png & ffmpeg
* also installs the Python bindings for LAMMPS

VilleS1:
I will close this pr then.

PR #5480, opened 2017-12-06
disable new rfkill feature in util-linux that requires recent kernel

boegel:
Installation of `util-linux` 2.31 fails on systems with older kernel with:

```
sys-utils/rfkill.c:67:12: error: 'RFKILL_TYPE_FM' undeclared here (not in a function)
  { .type = RFKILL_TYPE_FM,        .name = "fm",           .desc = "FM" },
            ^~~~~~~~~~~~~~
  CC       sys-utils/setarch.o
make[2]: *** [sys-utils/rfkill-rfkill.o] Error 1
```

Recent kernel is required for this to work:

```
$ grep RFKILL_TYPE_FM  /usr/include/linux/rfkill.h
 * @RFKILL_TYPE_FM: switch is on a FM radio device.
        RFKILL_TYPE_FM,
$ rpm -q --whatprovides /usr/include/linux/rfkill.h
kernel-headers-3.10.0-693.5.2.el7.ug.x86_64
```

`rfkill` feature (disabling/enabling of wireless devices) is utterly useless in HPC context imho, so OK to disable.

cc @verdurin, @migueldiascosta

PR #6212, opened 2018-04-19
{tools}[GCCcore/6.4.0] util-linux v2.32

JackPerdue:
(created using `eb --new-pr`)


easybuild/easyconfigs/u/util-linux/util-linux-2.32-GCCcore-6.4.0.eb
```diff
@@ -0,0 +1,43 @@
...
+
+source_urls = ['%s/v%%(version_major_minor)s' % homepage]
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['ce43afee3182f1bddb0be83f68bd378770efb9b6fdd8f464333ff8e07903db56']
+
+# disable account related utilities (they need OS dependant pam-devel files)
+# disable wall and friends (requires group changing permissions for install user)
+# install systemd service files in install dir
+# install bash completion files in install dir
+configopts = "--disable-chfn-chsh --disable-login --disable-su "
```
boegel: please also include `--disable-rfkill` here, cfr. #5480

easybuild/easyconfigs/u/util-linux/util-linux-2.32-GCCcore-6.4.0.eb
```diff
@@ -0,0 +1,43 @@
...
+# disable building Python bindings (since we don't include Python as a dep)
+configopts += "--without-python "
+
+builddependencies = [
+    ('binutils', '2.28'),
+]
+
+dependencies = [
+    ('LibUUID', '1.0.3'),
+    ('ncurses', '6.0'),
```
boegel: missing `zlib` dependency?

migueldiascosta:
@JackPerdue, this PR is being closed for the following reason(s): closing and reopening to trigger tests.

boegel:
This same easyconfig is already merged via #6237

Only difference here is that `LibUUID` is a dependency, which doesn't make much sense to me (see #8258)

PR #6221, opened 2018-04-23
 Removing Singularity to not get confused, since this is not a suppo…

titansmc:
Removing Singularity to not get confused, since this is not a suppoted method


boegel:
@titansmc "not a supported method" is not fully accurate...

It's actually OK to install Singularity through a module file, but what's not OK is to leave old Singularity versions around since they may have security issues.

I'm a bit torn about whether or not we should remove the Singularity easyconfig files or not...

@easybuilders/easybuild-easyconfigs-maintainers Thoughts on this?

titansmc:
Hi, sorry for that. I just saw https://github.com/singularityware/singularity/issues/736 and since eb is intended to be used by HPC clusters and most of them would have root_squash enabled .... I just thought it wouldn't make sense to have it here.

verdurin:
@boegel I think the concerns could be met by, for example, adding comments to the older `easyconfigs`. I'm against removing them.
It's not as though it's the only application with security vulnerabilities...

titansmc:
Hi, as I said I am just starting with eb, so I might have a poor view about this, but from an admin that wants to run this software in a cluster, I would find it useful to at least have a note stating that if you have root_squash on the shared filesystem , this is not going to work.

But, do you guys think that are there any people using Singularity from modules environment?

dirkpetersen:
While there are a limited number of Singularity users Singularity has already become an essential part of the HPC stack on many sites and yes we are using it via modules because it is still pretty new and users load different versions of the software. The fact that there are issues should lead to easybuild do-it-the-right-wat approach not to a removal, IMO

titansmc:
so @dirkpetersen would you mind to tell me in which shared storage I can set up Singularity via modules? This is also interesting for me, but I still cannot see how I can implement it, and I have a very standard setup in my cluster, software through NFS.

Cheers.

boegel:
We've archived easyconfigs for Singularity 2.x in #10591 due to security concerns, so closing this one...

PR #6287, opened 2018-05-04
Singularity-2.5.1-GCC-5.4.0-2.26

bilke:


boegel:
Test report by @boegel
**FAILED**
node2071.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/c3ec80eff9f1d244808f0b8b88f9795f for a full test report.

boegel:
@bilke `libarchive` is a required build dependency for Singularity 2.5.x, we'll need to include that as a dependency...

`squasfs` is also required, but we have that installed in our OS; maybe we should include that as a dependency too?

bilke:
@boegel Thanks for testing! I added both as `osdependencies`.

boegel:
@bilke Wouldn't it make sense to add `libarchive` as a proper build dependency, and adding `squashfs` (http://squashfs.sourceforge.net/) as regular (runtime) dependency EasyBuild-wise?

easybuild/easyconfigs/s/Singularity/Singularity-2.5.1-GCC-5.4.0-2.26.eb
```diff
@@ -0,0 +1,36 @@
...
+description = """Singularity is a portable application stack packaging and runtime utility."""
+
+toolchain = {'name': 'GCC', 'version': '5.4.0-2.26'}
+
+source_urls = ['https://github.com/singularityware/singularity/archive/']
+sources = ['%(version)s.tar.gz']
+checksums = ['c1369d1ae0318876cf878eaa249d33f9699c6d77e0cdb2c55522f6714beb306e']
+
+builddependencies = [('Autotools', '20150215')]
+dependencies = [('libarchive', '3.3.1')]
```
boegel: @bilke `libarchive` is only a *build* dependency I think?

easybuild/easyconfigs/s/Singularity/Singularity-2.5.1-GCC-5.4.0-2.26.eb
```diff
@@ -0,0 +1,36 @@
...
+osdependencies = [('squashfs-tools')]
```
boegel: any reason to not include `squashfs` as a proper dependency?

easybuild/easyconfigs/s/Singularity/Singularity-2.5.1-GCC-5.4.0-2.26.eb
bilke: Right!

easybuild/easyconfigs/s/Singularity/Singularity-2.5.1-GCC-5.4.0-2.26.eb
bilke: @boegel I am new to easybuild. Do you mean adding an easyconfig for squashfs too?

boegel: Yes, we should try and avoid to rely on OS dependencies when we can.

I'm not 100% this is an option here though, does `squashfs` (also) need `setuid` permissions or anything like that?

bilke: I have added the squashfs easyconfig. No it does not seem to need `setuid`-stuff but I am not sure either.

boegel:
Test report by @boegel
**SUCCESS**
node2581.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/e04b2b039e6e270c2c691646147a0663 for a full test report.

boegel:
@bilke Any chance of testing this on a system where `squashfs` is not available in the OS (both the build & usage)?

boegel:
Test report by @boegel
**SUCCESS**
node2082.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/a0bf0f3126980e26ae9516308abbbdd8 for a full test report.

boegel:
@bilke Got some feedback from the Singularity Slack channel: including `squashfs` as an EasyBuild dependency doesn't make sense, it has to be available in a system location. Quoting @GodloveD:

```
As long as mksquashfs is found on a _system_ path its fine.
That is as long as it’s in the path after Singularity sanitizes the path for security.
...
mksquashfs runs with elevated privs.  And it runs even if you do a non-privileged build from Docker or something.
So you wouldn’t want that to just be defined by anybody.
```

GodloveD:
BTW.  I misspoke about `squashfs` running with privs.  It does not.  That is to say it DOES run unprivileged.  However, the `$PATH` is sanitized very early in the code so it is still true that Singularity will not find `mksquashfs` if it is installed in an unusual location. 

I did create a [PR](https://github.com/singularityware/singularity/pull/1094) that was merged into Singularity development a while back to allow an admin to configure the location of `mksquashfs`.  But it hasn't made it's way into a release yet, and I'm not sure it would help you anyway.

boegel:
@GodloveD Hmm, not sure that PR makes much of a difference in this situation, indeed, since it only adds support for specifying the location of `mksquashfs` in the Singularity configuration.

How about a compile-time option to specify the location, e.g. a `configure` flag like `--with-squashfs=/path/to/squashfs-install-dir` (indicates that `mksquashfs` is at `/path/to/squashfs-install-dir/bin/mksquashfs`), so that the location of `mksquashfs` is "baked in"?

GodloveD:
@boegel that is probably doable.

https://github.com/singularityware/singularity/issues/1523

boegel:
:+1:

jmstover:
@boegel 
If you want to test it out, here's a PR for specifying the mksquashfs location:

https://github.com/singularityware/singularity/pull/1530

easybuild/easyconfigs/s/squashfs/squashfs-4.3-GCC-5.4.0-2.26.eb
```diff
@@ -0,0 +1,25 @@
...
+description = """Squashfs is a highly compressed read-only filesystem for Linux."""
+toolchain = {'name': 'GCC', 'version': '5.4.0-2.26'}
+
+source_urls = ['http://sourceforge.net/projects/squashfs/files/squashfs/squashfs%(version)s/']
+sources = ['%(name)s%(version)s.tar.gz']
+checksums = ['0d605512437b1eb800b4736791559295ee5f60177e102e4d4ccd0ee241a5f3f6']
+
+start_dir = 'squashfs-tools'
+skipsteps = ['configure']
+installopts = 'INSTALL_DIR=%(installdir)s'
```
boegel: @bilke This should probably be tweaked to `%(installdir)s/bin`, to ensure that `mksquashfs` and `unsquashfs` are in a standard location that gets included into `$PATH`?

Cfr expectations for the `--with-mksquashfs` configure option that's being added to Singularity: https://github.com/singularityware/singularity/pull/1530

boegel:
@bilke How do we proceed with this?

boegel:
@bilke ping?

boegel:
We've archived easyconfigs for Singularity 2.x in #10591 due to security concerns, so closing this one...

PR #6284, opened 2018-05-03
{bio}[intel/2018a] Roary v3.12.0

boegel:
(created using `eb --new-pr`)
requires ~~#6277~~ (BioPerl), ~~#6279~~ (PRANK), ~~#6280~~ (CD-HIT), ~~#6281~~ (parallel), ~~#6282~~ (FastTree), ~~#6283~~ (BLAST+), ~~#6297~~ (extra Perl extensions), ~~#6318~~ (Kraken)

PR #6283, opened 2018-05-03
{bio}[intel/2018a] BLAST+ v2.7.1

boegel:
(created using `eb --new-pr`)

PR #6288, opened 2018-05-04
remove Python dependency from BLAST+ 2.7.1 easyconfig using foss/2018a toolchain

boegel:
After running into a version conflict on `Python` when combining `BLAST+` (which currently includes `Python` `2.7.14` as a dependency) with `R` (which indirectly depends on `Python` `3.6.4`), in the context of `Roary`, cfr. #6284), I looked into why we actually have `Python` listed as a dependency for `BLAST+`...

The only reason I could find is the `windowmasker_2.2.22_adapter.py`, which actually has `#!/usr/bin/python` hardcoded (so the `Python` we provide as a dependency is not used).
Moreover, I doubt the script is very relevant, can't seem to find much about it when searching the BLAST+ documentation... Note that there's also a `windowmasker` binary.

So, I propose we drop `Python` as a dependency to `BLAST+`, and rename the easyconfig accordingly. That way, we can also use this easyconfig in a situation where `Python` `3.6.4` is also pulled in.
The obvious other alternative is adding `BLAST+-2.7.1-*-2018a-Python-3.6.4.eb` easyconfigs, but that's silly imho, since i) the `python` we provide isn't used given the hardcoding, and ii) the `windowmasker_2.2.22_adapter.py` script is not compatible with Python 3.x...

We can easily do this for the easyconfig for `BLAST+` `2.7.1` built with `foss/2018` (which was added very recently in #6015), since that's not used as a dependency anywhere yet and it's not included in an EasyBuild release.
Doing the same for `BLAST+-2.7.1-intel-2017b-Python-2.7.14.eb` is a lot more painful, and since the dependency issue hasn't popped up there, I suggest we leave it as is (not worth the trouble) and only change this going forward.

I'd like to (also) hear @Serpens 's thoughts on this, since he contributed the easyconfig being changed here (cfr. #6015)...

If this gets merged, I'll update #6283 accordingly

### DIFF
--- a/easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1-foss-2018a.eb
+++ b/easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1-foss-2018a.eb
@@ -14,7 +14,6 @@ easyblock = 'ConfigureMake'
 
 name = 'BLAST+'
 version = '2.7.1'
-versionsuffix = '-Python-%(pyver)s'
 
 homepage = 'http://blast.ncbi.nlm.nih.gov/'
 description = """Basic Local Alignment Search Tool, or BLAST, is an algorithm
@@ -36,7 +35,6 @@ dependencies = [
     ('zlib', '1.2.11'),
     ('bzip2', '1.0.6'),
     ('PCRE', '8.41'),
-    ('Python', '2.7.14'),
     ('Boost', '1.66.0'),
     ('GMP', '6.1.2'),
     ('libpng', '1.6.34'),
--- a/easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1-foss-2018a.eb
+++ b/easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1-foss-2018a.eb
@@ -43,7 +43,7 @@ dependencies = [
 ]
 
 configopts = "--with-64 --with-z=$EBROOTZLIB --with-bz2=$EBROOTBZIP2 --with-pcre=$EBROOTPCRE "
-configopts += "--with-python=$EBROOTPYTHON --with-boost=$EBROOTBOOST --with-gmp=$EBROOTGMP --with-png=$EBROOTLIBPNG "
+configopts += "--with-boost=$EBROOTBOOST --with-gmp=$EBROOTGMP --with-png=$EBROOTLIBPNG "
 configopts += "--with-jpeg=$EBROOTLIBJPEGMINTURBO --with-lmdb=$EBROOTLMDB"
 
 sanity_check_paths = {
--- a/easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1-intel-2018a.eb
+++ b/easybuild/easyconfigs/b/BLAST+/BLAST+-2.7.1-intel-2018a.eb
@@ -14,7 +14,6 @@ easyblock = 'ConfigureMake'
 
 name = 'BLAST+'
 version = '2.7.1'
-versionsuffix = '-Python-%(pyver)s'
 
 homepage = 'http://blast.ncbi.nlm.nih.gov/'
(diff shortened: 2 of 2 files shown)


vanzod:
Your argument sounds solid enough to me.
The only thing I am not a huge fan of is not removing it from `BLAST+-2.7.1-intel-2017b-Python-2.7.14.eb` too. Knowing myself I will always end up asking why one has the dependency and the other does not. Does the "pain" come from dependency reasons?

boegel:
@vanzod I see your point, and I'm in doubt for the exact same reasons.
I have two reasons not to change it there:

* i) `BLAST+-2.7.1-intel-2017b-Python-2.7.14.eb` is used as a dependency for `proovread-2.14.1-intel-2017b.eb` (so we need to update that too)
* ii) it's already included in existing EasyBuild releases, so it may surprise people when it's no longer there or even cause problems (when they have own easyconfigs that depend on that `BLAST+` easyconfig)

The latter one is the bigger issue, the former is actually not that big of a deal.

boegel:
Test report by @boegel
**SUCCESS**
node2543.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/4581d8ebe1f1f705ecdbd925eae73607 for a full test report.

Serpens:
I'm for dropping `Python` dependency for `BLAST+`. I kept it because I wanted the new easyconfig to be as close to the old one as possible.

`windowmasker_2.2.22_adapter.py` is only needed because options for `windowmasker` changed with `BLAST+` 2.2.22 release (in 2009). It builds a new command using only built-in Python modules, so it doesn't matter which Python 2 it uses.

boegel:
Test report by @boegel
**SUCCESS**
node2105.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/899a6aab23ec44e0f7537477271b03fa for a full test report.

boegel:
Thanks a lot for the feedback @Serpens!

@easybuilders/easybuild-easyconfigs-maintainers: good to go!

boegel:
#6283 was merged by @verdurin before this one, so I updated this PR to also drop the Python dependency from `BLAST+-2.7.1-intel-2018a.eb` (and I dropped the reference to `$EBROOTPYTHON` from `configopts`)

verdurin:
Test report by @verdurin
**SUCCESS**
easybuild.novalocal - Linux centos linux 7.5.1804, Intel Xeon E312xx (Sandy Bridge), Python 2.7.5
See https://gist.github.com/0193a93fb017758c1c3a5016b7a03c3f for a full test report.

verdurin:
Going in, thanks @boegel!

boegel:
Test report by @boegel
**SUCCESS**
node2501.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/613b51aafb49c0ed28754eca2dc5e426 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2083.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/6275e18b0461112b0cff0b361a9007fe for a full test report.

verdurin:
Test report by @verdurin
**SUCCESS**
rescomp2 - Linux centos 6.8, Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz, Python 2.7.11
See https://gist.github.com/25b155bcde69ca5c8d11f08e8f5088ef for a full test report.

PR #6294, opened 2018-05-05
add README for Java with information on downloading source tarball

cstackpole:
Took me a few minutes to figure out how to install Java with a weird warning about not being able to find the download. I thought I would post my notes for others in the future who might benefit.

### DIFF
--- a/easybuild/easyconfigs/j/Java/README.md
+++ b/easybuild/easyconfigs/j/Java/README.md
@@ -1,4 +1,8 @@
-In order to use these configs you must first download Java from [Oracle](http://www.oracle.com/technetwork/java/javase/downloads/index.html).  
+In order to use these configs you must first download Java from [Oracle](http://www.oracle.com/technetwork/java/javase/downloads/index.html). The file should look like this: jdk-8u162-linux-x64.tar.gz    
 
-To install, you must past eb the sourcepath to find the download. For example, if the file is in /tmp your command might look something like:
-`eb Java-1.8.0_162.eb -r --sourcepath=/tmp/`
+To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
+* Pass it as a option on the command line: `eb Java-1.8.0_162.eb -r --sourcepath=~/EB_Downloads`  
+* Set an environment variable:`export EASYBUILD_SOURCEPATH=~/EB_Downloads ; eb Java-1.8.0_162.eb -r`  
+* Configure a permanent location in ~/.config/easybuild/config.cfg before installing with `eb Java-1.8.0_162.eb -r`
+> [config]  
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/j/Java/README.md
```diff
@@ -0,0 +1,4 @@
+In order to use these configs you must first download Java from [Oracle](http://www.oracle.com/technetwork/java/javase/downloads/index.html).  
+
+To install, you must past eb the sourcepath to find the download. For example, if the file is in /tmp your command might look something like:
```
boegel: You're right that EasyBuild only looks in particular places for 'sources', but the way it's stated now is not how it's typically done.

You usually have configured EasyBuild with a particular `sourcepath` location, either directly via the `--sourcepath` configuration option (via a configuration file or via `$EASYBUILD_SOURCEPATH`), or indirectly via `--prefix` (or `$EASYBUILD_PREFIX` or cfg file).

So typically you'll copy/move the downloaded tarball for Java into `<sourcepath>/j/Java`, since that's also where EasyBuild will download stuff into (when it can).

cstackpole:
Thanks for the feedback. Let me research the other methods you mentioned then I will update the notes.

cstackpole:
I wasn't able to get the method you mentioned about <sourcepath>/j/Java, but I was able to verify three ways that worked and I updated the documentation accordingly. Hopefully this helps.

easybuild/easyconfigs/j/Java/README.md
```diff
@@ -0,0 +1,8 @@
+In order to use these configs you must first download Java from [Oracle](http://www.oracle.com/technetwork/java/javase/downloads/index.html). The file should look like this: jdk-8u162-linux-x64.tar.gz
```
boegel: `file` -> `file name`?

easybuild/easyconfigs/j/Java/README.md
```diff
@@ -0,0 +1,8 @@
+In order to use these configs you must first download Java from [Oracle](http://www.oracle.com/technetwork/java/javase/downloads/index.html). The file should look like this: jdk-8u162-linux-x64.tar.gz    
+
+To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.
```
boegel: `you must past eb the sourcepath to find the download.` ->

```
you must configure EasyBuild to use the right `sourcepath` so it can find the manually downloaded file.
```

cstackpole:
Updated based on the feedback. Thanks!
Again, I think this is best left separate from the Intel simply because the Intel has at least one extra step. I prefer to err on the side of clarity than 'efficient but confuses new comers' when it pertains to documentation. But maybe someone else can come up with a better middle ground. In the mean time, I know that this works. :-)

PR #6308, opened 2018-05-09
{numlib}[foss/2018a] PETSc 3.9.1

RvDijk:
Depends on easybuilders/easybuild-easyblocks/pull/1421

### DIFF
--- a/easybuild/easyconfigs/p/PETSc/PETSc-3.9.1-foss-2018a-downloaded-deps.eb
+++ b/easybuild/easyconfigs/p/PETSc/PETSc-3.9.1-foss-2018a-downloaded-deps.eb
@@ -1,0 +1,31 @@
+name = 'PETSc'
+version = '3.9.1'
+versionsuffix = '-downloaded-deps'
+
+homepage = 'http://www.mcs.anl.gov/petsc'
+description = """PETSc, pronounced PET-see (the S is silent), is a suite of data structures and routines for the
+ scalable (parallel) solution of scientific applications modeled by partial differential equations."""
+
+toolchain = {'name': 'foss', 'version': '2018a'}
+toolchainopts = {'usempi': True, 'pic': True}
+
+source_urls = ['http://ftp.mcs.anl.gov/pub/petsc/release-snapshots']
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['742e838a35d278693e956ed1ca4592c1d663451f6beea0694bf334aeb67681e8']
+
+# (build) dependencies provided through EasyBuild
+builddependencies = [('CMake', '3.11.1')]
+dependencies = [('HDF5', '1.10.1')]
+
+# external packages that are downloaded with --download-<package>=1 and --download-<package>-static=1
+download_deps_static = ['hypre', 'metis', 'ml', 'mumps', 'parmetis', 'prometheus', 'ptscotch', 'suitesparse',
+                        'superlu', 'superlu_dist']
+
+# triangle cannot be downloaded with --download-triangle-static because it fails
+download_deps = ['triangle']
+
+configopts = "--with-hdf5=1 --with-hdf5-dir=$EBROOTHDF5"
+
+shared_libs = True
+
+moduleclass = 'numlib'


easybuild/easyconfigs/p/PETSc/PETSc-3.9.1-foss-2018a-downloaded-deps.eb
```diff
@@ -0,0 +1,35 @@
...
+
+source_urls = ['http://ftp.mcs.anl.gov/pub/petsc/release-snapshots']
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['742e838a35d278693e956ed1ca4592c1d663451f6beea0694bf334aeb67681e8']
+
+# (build) dependencies provided through EasyBuild
+builddependencies = [('CMake', '3.11.1')]
+dependencies = [
+    ('HDF5', '1.10.1'),
+    ('Valgrind', '3.13.0'),
```
boegel: Maybe explicitly add `--with-debugging` as a configure option since we're including `Valgrind` as a dependency (it's enabled by default according to the PETSc documentation, but it doesn't hurt making it explicit, cfr. --with-debugging).

easybuild/easyconfigs/p/PETSc/PETSc-3.9.1-foss-2018a-downloaded-deps.eb
```diff
@@ -0,0 +1,35 @@
...
+    ('Bison', '3.0.4'),
```
boegel: Any reference to using `Bison` as a dependency for `PETSc`?

RvDijk: I was using a singularity container to install PETSc, it complained about bison not being available. This is because bison is not in the OS by default. It resulted in this error:
```
*******************************************************************************
         UNABLE to CONFIGURE with GIVEN OPTIONS (see configure.log for details):
-------------------------------------------------------------------------------
PTScotch needs bison installed
*******************************************************************************
```
Then including Bison solved my problem, so I think it is better to include it...

boegel: OK, thanks for clarifying. But, are you sure that `Bison` should be runtime dependency rather than a build dependency? `Bison` usually is a build-only dependency (but it could be different here, of course).

boegel: Based on what I see in https://github.com/petsc/petsc/blob/master/config/BuildSystem/config/packages/PTScotch.py, it looks like a build-only dependency for `PETSc` as well.

Also, you should include `flex` as well as a build dependency...

RvDijk: I agree, thanks for looking into that!

easybuild/easyconfigs/p/PETSc/PETSc-3.9.1-foss-2018a-downloaded-deps.eb
```diff
@@ -0,0 +1,38 @@
...
+toolchainopts = {'usempi': True, 'pic': True}
+
+source_urls = ['http://ftp.mcs.anl.gov/pub/petsc/release-snapshots']
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['742e838a35d278693e956ed1ca4592c1d663451f6beea0694bf334aeb67681e8']
+
+# (build) dependencies provided through EasyBuild
+builddependencies = [
+    ('CMake', '3.11.1'),
+    ('Flex', '2.6.4'),
```
boegel: @RvDijk This should be `flex` rather than `Flex`...

boegel:
Test report by @boegel
**SUCCESS**
node2556.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/b9a25fcb4e17f5b676c937a5aa0b6426 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2092.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/2aaaaff2efcabfea8269bae8b9c64aa1 for a full test report.

PR #6317, opened 2018-05-11
Create README.md file for installing icc/ifort

cstackpole:
It took me several tries before I got this one right. I will update the other Intel tools when I get a chance to verify them.

Used the same documentation format as I did for Java. Hopefully it is good.

### DIFF
--- a/easybuild/easyconfigs/i/icc/README.md
+++ b/easybuild/easyconfigs/i/icc/README.md
@@ -5,7 +5,7 @@ Next you must specify the license file. There are three options:
 * Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
 * Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
 
-To install, you must configure EasyBuild to use the right `sourcepath` so it can find the manually downloaded file. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
+To install, you must configure EasyBuild to use the right `sourcepath` so it can find the manually downloaded file. There are a number of ways to do this assuming the file is in /home/username/EB_Downloads.  
 * Pass it as a option on the command line: `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r --sourcepath=/home/username/EB_Downloads`  
 * Set an environment variable: `export EASYBUILD_SOURCEPATH=/home/username/EB_Downloads ; eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`  
 * Configure a permanent location in ~/.config/easybuild/config.cfg before installing with `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`
--- a/easybuild/easyconfigs/i/icc/README.md
+++ b/easybuild/easyconfigs/i/icc/README.md
@@ -4,7 +4,7 @@ Next you must specify the license file. There are four options:
 * Place it in the following path: ~/licenses/intel/license.lic
 * Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
 * Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
-* Edit the EB script and modify or add the line `license_file = license_file`
+* Edit the EB script and modify or add the line `license_file = /path/to/file.lic`
 
 To install, you must configure EasyBuild to use the right `sourcepath` so it can find the manually downloaded file. There are a number of ways to do this assuming the file is in /home/username/EB_Downloads.  
 * Pass it as a option on the command line: `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r --sourcepath=/home/username/EB_Downloads`  
--- a/easybuild/easyconfigs/i/icc/README.md
+++ b/easybuild/easyconfigs/i/icc/README.md
@@ -7,7 +7,7 @@ Next you must specify the license file. There are three options:
 
 To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
 * Pass it as a option on the command line: `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r --sourcepath=/home/username/EB_Downloads`  
-* Set an environment variable: `export EASYBUILD_SOURCEPATH=~/EB_Downloads ; eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`  
+* Set an environment variable: `export EASYBUILD_SOURCEPATH=/home/username/EB_Downloads ; eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`  
 * Configure a permanent location in ~/.config/easybuild/config.cfg before installing with `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`
 > [config]  
 > sourcepath=~/EB_Downloads
--- a/easybuild/easyconfigs/i/icc/README.md
+++ b/easybuild/easyconfigs/i/icc/README.md
@@ -5,7 +5,7 @@ Next you must specify the license file. There are three options:
 * Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
 * Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
 
-To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
+To install, you must configure EasyBuild to use the right `sourcepath` so it can find the manually downloaded file. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
 * Pass it as a option on the command line: `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r --sourcepath=/home/username/EB_Downloads`  
 * Set an environment variable: `export EASYBUILD_SOURCEPATH=/home/username/EB_Downloads ; eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`  
 * Configure a permanent location in ~/.config/easybuild/config.cfg before installing with `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`
--- a/easybuild/easyconfigs/i/icc/README.md
+++ b/easybuild/easyconfigs/i/icc/README.md
@@ -1,0 +1,13 @@
+In order to use these configs you must first download the Intel Parallel Studio XE CPP file. NOT the full bundle! The file should look like this: parallel_studio_xe_2018_update2_composer_edition_for_cpp.tgz    
+
+Next you must specify the license file. There are three options:  
+* Place it in the following path: ~/licenses/intel/license.lic
+* Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
+* Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
+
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/i/icc/README.md
```diff
@@ -0,0 +1,13 @@
+In order to use these configs you must first download the Intel Parallel Studio XE CPP file. NOT the full bundle! The file should look like this: parallel_studio_xe_2018_update2_composer_edition_for_cpp.tgz
```
boegel: Actually, you can also use the full bundle as a source file, but then you need to change the `sources` spec in the easyconfig file accordingly...

We list which `components` that should be installed exactly for this reason: to ensure that only `icc` installed even if you use the full bundle to install...

easybuild/easyconfigs/i/icc/README.md
```diff
@@ -0,0 +1,13 @@
+In order to use these configs you must first download the Intel Parallel Studio XE CPP file. NOT the full bundle! The file should look like this: parallel_studio_xe_2018_update2_composer_edition_for_cpp.tgz    
+
+Next you must specify the license file. There are three options:  
+* Place it in the following path: ~/licenses/intel/license.lic
+* Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
+* Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
```
boegel: There's a 4th option: change the `license_file` spec in the easyconfig file you are using

easybuild/easyconfigs/i/icc/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.
```
boegel: `you must past eb the sourcepath to find the download.` ->

```
you must configure EasyBuild to use the right `sourcepath` so it can find the manually downloaded file.
```

easybuild/easyconfigs/i/icc/README.md
```diff
@@ -0,0 +1,13 @@
+In order to use these configs you must first download the Intel Parallel Studio XE CPP file. NOT the full bundle! The file should look like this: parallel_studio_xe_2018_update2_composer_edition_for_cpp.tgz    
+
+Next you must specify the license file. There are three options:  
+* Place it in the following path: ~/licenses/intel/license.lic
+* Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
+* Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
+
+To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
+* Pass it as a option on the command line: `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r --sourcepath=/home/username/EB_Downloads`
```
boegel: you can use `~/EB_Downloads` here?

easybuild/easyconfigs/i/icc/README.md
```diff
@@ -0,0 +1,13 @@
+In order to use these configs you must first download the Intel Parallel Studio XE CPP file. NOT the full bundle! The file should look like this: parallel_studio_xe_2018_update2_composer_edition_for_cpp.tgz    
+
+Next you must specify the license file. There are three options:  
+* Place it in the following path: ~/licenses/intel/license.lic
+* Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
+* Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
+
+To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
+* Pass it as a option on the command line: `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r --sourcepath=/home/username/EB_Downloads`  
+* Set an environment variable: `export EASYBUILD_SOURCEPATH=/home/username/EB_Downloads ; eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`
```
boegel: you can use `~/EB_Downloads` here?

easybuild/easyconfigs/i/icc/README.md
```diff
@@ -0,0 +1,13 @@
...
+* Place it in the following path: ~/licenses/intel/license.lic
+* Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
+* Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
+
+To install, you must past eb the sourcepath to find the download. There are a number of ways to do this assuming the file is in ~/EB_Downloads.  
+* Pass it as a option on the command line: `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r --sourcepath=/home/username/EB_Downloads`  
+* Set an environment variable: `export EASYBUILD_SOURCEPATH=/home/username/EB_Downloads ; eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`  
+* Configure a permanent location in ~/.config/easybuild/config.cfg before installing with `eb icc-2018.2.199-GCC-6.4.0-2.28.eb -r`
+> [config]  
+> sourcepath=~/EB_Downloads
```
boegel: I doubt using `~/EB_Downloads` actually works, you probably need to use the resolved path here?

easybuild/easyconfigs/i/icc/README.md
cstackpole: I didn't know that. But for someone reading this documentation trying to get it working the first time, do we want to complicate things by asking them to modify code first? Or use a simple method to first get it working? Personally, I prefer to just get something that works then explore the other options.

easybuild/easyconfigs/i/icc/README.md
cstackpole: Again, good to know but not sure that is what is wanted in a "heres how you get this to work" document for those getting started.

easybuild/easyconfigs/i/icc/README.md
cstackpole: I agree with the wording change. I will update the request.

easybuild/easyconfigs/i/icc/README.md
cstackpole: No. I thought I could short cut it, but it didn't work when I tested and I updated the other places, but missed this one.

cstackpole:
Thanks for the feedback.
As for a single place for the documentation, that's not a bad idea. I thought about doing that when I realized that the documentation for Java was basically the same for Intel. Then I ran into issues with the license path so I decided to keep them separate.

I was thinking about updating the README files to just reference the Intel folder and putting it all in there since the MKL install was basically the same as the ICC and IFORT. However, I've ran into issues last night and haven't actually gotten the overall Intel bundle to actually work yet. I was actually thinking about opening a ticket on the issues if I don't figure it out today. If I can get that to work reliably then maybe I will update the docs.

However, these instructions I feel do fit the documentation for the time being. I have reproduced them repeatedly on a VM and now I have repeated them on 2 other systems. So I'm pretty confident these are working and can help anyone between now and whenever I figure out the overall Intel install + write up new instructions. :-D

easybuild/easyconfigs/i/icc/README.md
boegel: Well, sure, simple approach is better, but we shouldn't claim it's not possible, since it is (with minimal effort).

easybuild/easyconfigs/i/icc/README.md
boegel: Again, just for completeness sake, it should be mentioned...

easybuild/easyconfigs/i/icc/README.md
boegel: ``--sourcepath ~/EB_Downloads` would work, since then `bash` will expand the `~`, it maybe won't when you're using ``--sourcepath=...`

easybuild/easyconfigs/i/icc/README.md
cstackpole: Good point. I will need some time to prove this out and write up the documentation.

It generates a question though, if a full bundle is supported and if there is a primary [Intel script](https://github.com/easybuilders/easybuild-easyconfigs/tree/master/easybuild/easyconfigs/i/intel) then wouldn't it be better to have the documentation for installing the bundle be in the primary Intel section (that way someone gets ALL of the Intel tools at once) then have the documentation for the individual applications be specific to them?

This question may be a bit idealistic in nature considering that the primary intel scripts don't auto include things like the cluster checker, vtune, ect. But maybe it should?

cstackpole: @boegel I had to do heavy modifications to get the full bundle to install. It didn't like any of the patches, had to change the source, and the sha256sum.

Would it not be better to create a whole new file for installing from the full bundle?

easybuild/easyconfigs/i/icc/README.md
cstackpole: Done. Just added to both.

easybuild/easyconfigs/i/icc/README.md
boegel: I think we currently assume in a couple of places that the `intel` toolchain components like `icc`, `ifort`, `impi` and `imkl` are installed separate (or at least that the corresponding `$EBROOT*` environment variables are set), so it's probably less trivial than it seems...

It's not entirely clear to me what the best approach is here. When do people have access to (only) the full bundle, when do they have access to separate tarballs for each component/tool?

easybuild/easyconfigs/i/icc/README.md
```diff
@@ -0,0 +1,14 @@
+In order to use these configs you must first download the Intel Parallel Studio XE CPP file. NOT the full bundle! The file should look like this: parallel_studio_xe_2018_update2_composer_edition_for_cpp.tgz    
+
+Next you must specify the license file. There are four options:  
+* Place it in the following path: ~/licenses/intel/license.lic
+* Set an environment variable: `export INTEL_LICENSE_FILE=/path/to/file.lic`
+* Set an environment variable: `export LM_LICENSE_FILE=/path/to/file.lic`
+* Edit the EB script and modify or add the line `license_file = license_file`
```
boegel: should probably formatting like (or something like it)?

```
`license_file = "/path/to/licenses/intel.lic"`
```

easybuild/easyconfigs/i/icc/README.md
cstackpole: So it depends on which version of the tools you buy (see #6346 for the work I've done in trying to make sense of it). Each of those is a bundle of tools and I honestly don't know why anyone would want to install just individual components on a system*. I don't think you can buy just an individual tool anymore; so most of the time you have a bundle of at least a compiler, MKL, DAAL, IPP, and ITB. The bigger bundles have more tools and Cluster has them all.

* The only time I've ever just installed a single product on the cluster outside of a bundle is when I got bit bad by a bug and they released a point release which was literally one package update in one tool...so I just updated that one package.

With that said, when you have access to a bundle you have access to the individual products as well. I'm liking the idea of a "master" that installs the bundle by calling the individual projects. I started hacking away on that but life got busy and I've been swamped at work recently. *shrug* it goes that way sometimes. I will get back to hacking on EB soon. :-)

Thanks!

easybuild/easyconfigs/i/icc/README.md
cstackpole: Good catch. Changing now.

PR #6323, opened 2018-05-14
{compiler} PGI v18.4 (REVIEW)

SimonPinches:
Latest PGI compiler, v18.4.

### DIFF
--- a/easybuild/easyconfigs/p/PGI/PGI-18.4-GCC-6.4.0-2.28.eb
+++ b/easybuild/easyconfigs/p/PGI/PGI-18.4-GCC-6.4.0-2.28.eb
@@ -1,0 +1,25 @@
+name = 'PGI'
+version = '18.4'
+
+homepage = 'http://www.pgroup.com/'
+description = "C, C++ and Fortran compilers from The Portland Group - PGI"
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['pgilinux-20%(version_major)s-%(version_major)s%(version_minor)s-x86-64.tar.gz']
+checksums = ['81e0dcf6000b026093ece180d42d77854c23269fb8409cedcf51c674ca580a0f']
+
+gccver = '6.4.0'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/p/PGI/PGI-2018a.eb
```diff
@@ -0,0 +1,16 @@
+easyblock = "Toolchain"
+
+name = 'PGI'
+version = '2018a'
```
boegel: Hmm, what's the point here, why not use `PGI` `18.4` directly as a toolchain?

I don't really see the extra value in this...

Also, we kind of reserve the `<year><month>` versioning scheme for the common toolchains (& derivatives thereof). Maybe `2018.04` is better here?

SimonPinches: Ah, I didn't realise I could use `PGI/18.4` directly as a toolchain.  Do I just put `easyblock = "Toolchain"` at the top of `PGI-18.4-GCC-6.4.0-2.28.eb`?

boegel: You don't need to, you can use the `PGI` module as is as a toolchain in easyconfigs:

```python
toolchain = {'name': 'PGI', 'version': '18.4-GCC-6.4.0-2.28'}
```

The `Toolchain` easyblock is just a fancy `Bundle` easyblock, it does little more than pack multiple modules together under one umbrella, but you don't need that here, since there's only `PGI`.

SimonPinches: Okay, thanks.  I didn't release this (or had forgotten!).  I'll remove `PGI-2018a.eb` file from this PR.

vanzod:
Test report by @vanzod
**SUCCESS**
See https://gist.github.com/3e74699e192f6331442e98c52af87af3 for a full test report.

vanzod:

PR #6178, opened 2018-04-18
{toolchain}[gmpolf-2017.10]

rubendibattista:

PR #6325, opened 2018-05-14
{gmpolf-2017.10} PETSc 3.9.1 + petsc4py + all deps

rubendibattista:
Notes:

- SuiteSparse, Hypre, PETSc are using generic easyblocks instead of specific ones
- Not sure 100% PETSc gets compiled with all the external modules I want, the tests are passing btw
- Boost needs updated easyblock (easybuilders/easybuild-easyblocks#1424)
- Numpy needs updated easyblock (easybuilders/easybuild-easyblocks#1425)


boegel:
@rubendibattista This is a really big PR, which would be a nightmare to get merged, it should be broken up into smaller ones imho...

Also, the use of a bare `Python` as a dependency is an issue, since that makes those modules incompatible with other modules that depend on a non-bare Python...

What's your motivation for doing it like this?

rubendibattista:
Ok, sounds reasonable. 

I suggest a plan: 
1. Approve `gmpolf` toolchain (https://github.com/easybuilders/easybuild-easyconfigs/pull/6178)
2. Let's test Boost (with updated easyblock)
3. Accessory stuff (`bzip2`, `readline`, `CMake`)
4. Linear Algebra (`OpenBLAS`, `SuiteSparse`, `Hypre`, `SuperLU`)
5. Python stuff
6. PETSc

What do you think? 

For the `Python-bare` question, actually I think would be better to provide just the `Python-bare` package + the python modules that need optimizations and rely on users to use `virtualenv` to install al the rest of things that they could need. Actually the fully-fledged distributions are somehow arbitrary in the package they provide, aren't they?

boegel:
@rubendibattista Plan makes sense to me, I'll get #6178 tested & merged. It would be nice to also have a PR for Boost quickly, so we have a test case for https://github.com/easybuilders/easybuild-easyblocks/pull/1424 (so we can include that in the upcoming EasyBuild v3.6.1).

I'm not sure letting people rely on `virtualenv` is a good idea though...

The packages we include in the `Python` installations grew over time, we basically add anything there that makes sense, i.e. that doesn't have any external dependencies. Packages that requires `BLAS` & `LAPACK` (e.g. `numpy`) or packages that require an MPI library (e.g. `mpi4py`) are fine to include since we usually install Python on top of a full toolchain which provides all that...

rubendibattista:
Why do you say `virtualenv` is not a good idea? IMHO yes, everyone can install whatever he wants (provided is not already given by modules) and so the sysadmin can just focus on providing a light Python implementation with only the optimized packages, easier to maintain.

boegel:
@rubendibattista For complex Python packages like `numpy` it's not a good idea to assume that users will install it correctly (i.e. linked to the BLAS/LAPACK libraries properly, etc.).

I strongly double simple `pip install` in a `virtualenv` "jail" will result in a fast `numpy` installation. See the `numpy` easyblock to get a good idea of how complex it can be to install `numpy` properly...

rubendibattista:
@boegelbot Well, I know. I provided a MR for it: https://github.com/easybuilders/easybuild-easyblocks/pull/1425 :)

Ok, I agree with you. Hence why not providing `Python-bare` plus the other modules (e.g. `numpy`, `scipy`, etc...) in separate modules? The user can still run a virtualenv symlinking system modules into the virtualenv (`--system-site-packages` args to `virtualenvwrapper`), so having already in its virtualenv the modules he/her wants but still installing other stuff that does not need optimization by him/her-self. :).

Maybe I'm overseeing important stuff :).

boegel:
@rubendibattista Two potential issues with that approach:

* Users will need to actually load those separate modules, rather than install them using `pip install numpy` for example... How will they know they should load the module instead?
* Installing every Python package as a separate module is going to result in a large amount of extra modules...

rubendibattista:
Hei @boegel, thanks for the answer.

1. It's not what they already need to do, for example, to add Python bindings to Python environment (e.g. for `vtk`)
2. I agree, but it's actually what most package managers do, don't they?

BTW working on the plan we extablished.

boegel:
@rubendibattista Do you plan to get back to this? If not, I suggest we close it...

rubendibattista:
Yep, just close it. If ever, I'll open another one.

PR #6328, opened 2018-05-16
{perf}[gompi/2019a] Score-P 6.0 + Scalasca 2.5 + deps

geimer:
Minimal-toolchain versions (hierarchical MNS) for gompi/2018a.


easybuild/easyconfigs/o/OTF2/OTF2-2.1.1-GCCcore-6.4.0.eb
```diff
@@ -0,0 +1,58 @@
...
+    ('binutils', '2.28'),
+    # SIONlib container support (optional):
+    ('SIONlib', '1.7.2', '-tools'),
+]
+
+dependencies = [
+    # Python bindings (optional)
+    ('Python', '2.7.14', '-bare'),
+    ('future', '0.16.0', '-Python-2.7.14-bare'),
+    ('Six', '1.11.0', '-Python-2.7.14-bare'),
```
boegel: @geimer By using a `-bare` Python here, you're effectively making this incompatible with anything that depends on a full `Python`, which is quite common... Isn't that a concern?

Maybe it makes sense to make the `Python` that gets built with `GCCcore` a trimmed down version of the one we have with a full toolchain, i.e. still include common dependencies like `six`, `future`, `setuptools`, `pip`, etc into it?

easybuild/easyconfigs/s/Six/Six-1.11.0-GCCcore-6.4.0-Python-2.7.14-bare.eb
```diff
@@ -0,0 +1,34 @@
+easyblock = 'PythonPackage'
+
+name = 'Six'
```
boegel: Should be `six`, cfr. http://six.readthedocs.io/: `The name, “six”, comes from the fact that 2*3 equals 6.`

boegel:
@geimer This PR is getting quite big, are all of these inter-dependent? Is it worth fleshing out the `Python` built with `GCCcore` in a separate PR?

easybuild/easyconfigs/o/OTF2/OTF2-2.1.1-GCCcore-6.4.0.eb
geimer: @boegel There doesn't seem to be any consensus how to deal with `Python` on the `GCCcore` level, cfr. the discussion I raised two or three weeks ago on the mailing list. I'm doing *something* here to make progress, but the `Python` stuff needs to be fixed first, and then I can adjust this PR accordingly. But I'm **not** the one who is going to take a lead on this, as there are too many dark corners I don't understand...

boegel: Good point... Maybe you can consider building the Python bindings in a separate module, which can be loaded as needed?

geimer: Not sure our build system supports this right now.  Is there any support in EB for building things twice into the same prefix but with different build dependencies (like with a list of `configopts`)?  This may work too, but I would immediately admit that this is a rather strange feature request ;-)

geimer: Talking to the OTF2 core devs, this currently seems to be the only option: building OTF2 twice into the same prefix with the different Python's as build dependencies, and then providing Python binding modules that simply depend on the common OTF2 installation and only set `PYTHONPATH` accordingly. But I don't think this is currently possible with EB...

boegel: Not in a clean way, no, but it relates closely to the way in which the ComputeCanada folks install Python packages, so we should come up with a clean way of doing that anyway...

@bartoldeman @mboisson Do you have your approach clearly documented somewhere? Maybe it's time to start implementing this...

bartoldeman: What our approach boils down to:
1. Let the Python module set PYTHONPATH to a location with a file `sitecustomize.py`
2. this file has the following contents:
```
import sys
import os
import site

# use prefixes from EBPYTHONPREFIXES, so they have lower priority than
# virtualenv-installed packages

if "EBPYTHONPREFIXES" in os.environ:
    postfix = os.path.join('lib', 'python'+sys.version[:3], 'site-packages')
    for prefix in os.environ["EBPYTHONPREFIXES"].split(os.pathsep):
        sitedir = os.path.join(prefix, postfix)
        if os.path.isdir(sitedir):
            site.addsitedir(sitedir)
```
3. modules that use Python set EBPYTHONPREFIXES. We now do it as follows:
```
modextrapaths = {
    'EBPYTHONPREFIXES': [''],
}
```
4. module have to load different python modules for builds. We now do this as follows:
```
prebuildopts = ['module load python/2.7 && ',
                'module load python/3.5 && ',
                'module load python/3.6 && ']
```
but that's a hack. It would be better to allow builddependencies to be a list of lists.
5. we have this for some modules that absolutely need python (for some it is optional):
modluafooter = """
depends_on("python")
"""
this basically will load the default python if none is loaded and otherwise keep whatever python module is already loaded.

Hope that helps.

boegel:
@geimer How do we proceed with this? Bump to `2019a` where we have `Python` and `Qt5` built with `GCCcore`?

geimer:
@boegel Yes, my idea is to use `2019a` with `Qt5` from #8201 and bump versions to the latest Score-P/Scalasca/Cube releases.  I will start working on this once my `Qt5` build has finished...

geimer:
@boegel As outlined yesterday, I've updated this PR to the latest tool versions for `2019a`. Installations are fairly feature-complete, except for:

 - `OTF2` can be built with Python bindings.  I haven't included `Python` as a dependency yet, as I'm not quite sure how to use the `multi_deps` thing introduced recently.  Any pointer?
 - Score-P's library wrapping feature is not enabled, as it requires `LLVM` + `Clang` libraries (built as one blob). The `LLVM` configs currently available are not sufficient, and `Clang` is of moduleclass `compiler` and thus conflicts with `GCC`.  That is, it would require a redesign of how `LLVM` and `Clang` are provided in EasyBuild.
 - Other optional Score-P features such as OpenACC, OpenCL, and CUDA support have to be enabled on demand.

geimer:
Hmm... Isn't the Travis test too strict here?  IMO a "pure" bundle that only depends on other packages shouldn't be required to specify `sanity_check_paths`.  One way to work around this would be to re- check things that have already been checked in the underlying packages, but this is (a) redundant and (b) error-prone.  Any comments?

boegel:
@geimer You're absolutely right, we should whitelist easyconfigs that use `Bundle` (and don't have extensions listed), much like we whitelisted `Toolchain` in #8216 .

Are you up for looking into that yourself?
This should be helpful: https://github.com/easybuilders/easybuild-easyconfigs/pull/8216/commits/c56010e00532ef110a5b14750a7368463e3988ed

Note that it's a bit more involved than just checking for the use of `Bundle`, since we also have bundles of extensions (where having a custom `sanity_check_paths` is still very much desired).

easybuild/easyconfigs/s/Score-P/Score-P-5.0-gompi-2019a.eb
```diff
@@ -0,0 +1,62 @@
...
+# Copyright:: Copyright 2013-2019 Juelich Supercomputing Centre, Germany
+# Authors::   Bernd Mohr <b.mohr@fz-juelich.de>
+#             Markus Geimer <m.geimer@fz-juelich.de>
+# License::   3-clause BSD
+#
+# This work is based on experiences from the UNITE project
+# http://apps.fz-juelich.de/unite/
+##
+
+easyblock = 'EB_Score_minus_P'
```
boegel: You can actually remove this, since it matches the software name?

easybuild/easyconfigs/s/Score-P/Score-P-5.0-gompi-2019a.eb
```diff
@@ -0,0 +1,62 @@
...
+    # PDT source-to-source instrumentation support (optional):
+    ('PDT', '3.25'),
+]
+
+configopts = '--enable-shared'
+
+sanity_check_paths = {
+    'files': ['bin/scorep', 'include/scorep/SCOREP_User.h',
+              ('lib/libscorep_adapter_mpi_event.a', 'lib64/libscorep_adapter_mpi_event.a'),
+              ('lib/libscorep_adapter_mpi_event.%s' % SHLIB_EXT, 'lib64/libscorep_adapter_mpi_event.%s' % SHLIB_EXT)],
```
boegel: only `lib/` is fine

easybuild/easyconfigs/c/Cube/Cube-20190321-GCCcore-8.2.0.eb
```diff
@@ -0,0 +1,42 @@
...
+# License::   3-clause BSD
+#
+# This work is based on experiences from the UNITE project
+# http://apps.fz-juelich.de/unite/
+##
+
+easyblock = 'Bundle'
+
+name = 'Cube'
+version = '20190321'
```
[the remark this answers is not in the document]
geimer: There is no "main" component.  The former `Cube` package has been split into three disjoint componts: `CubeWriter`, `CubeLib`, and `CubeGUI`.  They now evolve somewhat independently and version numbers may diverge more over time.  In this case, the three components have been released on the same date (chosen as `version`), but this doesn't have to be the case in the future. Dependent packages may require only some of them or in different use cases (`dependency` vs. `builddependency`, see the `Score-P` easyconfig), thus I would really prefer to keep them separate.

In principle, the `Cube` bundle isn't required at all -- I've just added it as a convenience for people who are used to the old `Cube` package.  While we also provide a convenience bundle tarball including all components on our web page, its version is a simple count that has very little to do with the versions included (as the versions are allowed to diverge, it is hard to come up with a common version).  Having said all this, I'm also OK with removing the bundle if you think it causes too much confusion. (This would also magically "fix" the failing Travis check :wink:)

easybuild/easyconfigs/s/Score-P/Score-P-5.0-gompi-2019a.eb
geimer: Wasn't there a policy change at some point in the past to make the easyblock explicit?

geimer:
@boegel W.r.t. the CI check:  I don't have the complete picture yet when the check is desired and when it's not.  If it comes down to whitelisting `Bundle`s with an empty `exts_list` I can probably tackle this.  If it is more complicated than that, I might be out of the game...

geimer:
Test report by @geimer
**SUCCESS**
swat - Linux debian 9.9, Intel(R) Xeon(R) CPU           X5550  @ 2.67GHz, Python 2.7.13
See https://gist.github.com/9915275afb38032ace9b6a4fd3158ae1 for a full test report.

geimer:
This PR has been pending for so long that I could incorporate yet another version bump to the packages released yesterday :smiley: I think it's now (finally) ready for prime time...

easybuild/easyconfigs/s/Score-P/Score-P-5.0-gompi-2019a.eb
boegel: A (long) time, we dropped the automagic fallback to trying with `ConfigureMake` (which was confusing since that always left you wondering which easyblock was actually being used if none was specified).

But as long as the software name matches the easyblock, there's no point in specifying it, since there's no ambiguity.

boegel:
Test report by @boegel
**SUCCESS**
node3103.skitty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/7fe6bb549c118da7cc8013167d5157cc for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2425.golett.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/17d21f2e0b70b84b5514f9495def5cb8 for a full test report.

boegel:
Going in, thanks @geimer!

PR #6330, opened 2018-05-16
{data}[PGI/18.4] HDF5 v1.10.2 (REVIEW)

SimonPinches:
New `ebfiles` for `HDF5` libraries (v1.10.1 and v1.10.2) built with `PGI/18.4`

### DIFF
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
@@ -16,7 +16,8 @@ checksums = [
     'bfec1be8c366965a99812cf02ddc97e4b708c1754fccba5414d4adccdc073866',  # hdf5-1.10.2.tar.gz
 ]
 
-buildopts = 'CXXFLAGS="-noswitcherror"'
+# add -noswitcherror to avoid libtool adding -pthread (which is not known by PGI compiler)
+buildopts = 'CXXFLAGS="$CXXFLAGS -noswitcherror"'
 
 dependencies = [
     ('zlib', '1.2.11'),
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
@@ -16,7 +16,7 @@ checksums = [
     'bfec1be8c366965a99812cf02ddc97e4b708c1754fccba5414d4adccdc073866',  # hdf5-1.10.2.tar.gz
 ]
 
-# add -noswitcherror to avoid libtool adding -pthread (which is not known by PGI compiler)
+# Add -noswitcherror to make PGI compiler ignore the unknown compiler option -pthread
 preconfigopts = 'export CXX="$CXX -noswitcherror" && '
 
 dependencies = [
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
@@ -1,0 +1,26 @@
+name = 'HDF5'
+version = '1.10.2'
+
+homepage = 'https://support.hdfgroup.org/HDF5/'
+description = """HDF5 is a data model, library, and file format for storing and managing data.
+ It supports an unlimited variety of datatypes, and is designed for flexible
+ and efficient I/O and for high volume and complex data."""
+
+toolchain = {'name': 'PGI', 'version': '18.4-GCC-6.4.0-2.28'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-%(version_major_minor)s/hdf5-%(version)s/src']
+sources = [SOURCELOWER_TAR_GZ]
+
+checksums = [
+    'bfec1be8c366965a99812cf02ddc97e4b708c1754fccba5414d4adccdc073866',  # hdf5-1.10.2.tar.gz
+]
+
+buildopts = 'CXXFLAGS="-noswitcherror"'
+
+dependencies = [
+    ('zlib', '1.2.11'),
+    ('Szip', '2.1.1'),
+]
+
+moduleclass = 'data'
--- a/easybuild/easyconfigs/p/PGI/PGI-18.4-GCC-6.4.0-2.28.eb
+++ b/easybuild/easyconfigs/p/PGI/PGI-18.4-GCC-6.4.0-2.28.eb
@@ -1,0 +1,25 @@
+name = 'PGI'
+version = '18.4'
+
+homepage = 'http://www.pgroup.com/'
+description = "C, C++ and Fortran compilers from The Portland Group - PGI"
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['pgilinux-20%(version_major)s-%(version_major)s%(version_minor)s-x86-64.tar.gz']
+checksums = ['81e0dcf6000b026093ece180d42d77854c23269fb8409cedcf51c674ca580a0f']
+
+gccver = '6.4.0'
+binutilsver = '2.28'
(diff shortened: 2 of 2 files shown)


easybuild/easyconfigs/h/HDF5/HDF5-1.10.1-PGI-18.4-GCC-6.4.0-2.28.eb
```diff
@@ -0,0 +1,24 @@
...
+
+toolchain = {'name': 'PGI', 'version': '18.4-GCC-6.4.0-2.28'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-%(version_major_minor)s/hdf5-%(version)s/src']
+sources = [SOURCELOWER_TAR_GZ]
+
+checksums = ['048a9d149fb99aaa1680a712963f5a78e9c43b588d0e79d55e06760ec377c172']
+
+buildopts = "CXXFLAGS='--noswitcherror'"
```
boegel: @SimonPinches Can you clarify why this is required? Maybe we should include a comment in these easyconfig to explain it? Never seen this before (but my experience with PGI is limited).

boegel: Actually, this is even causing problems for me...

```
pgc++-Error-Unknown switch: --noswitcherror
```

boegel: It's definitely a clean way to fix the problem, but I think it's worth clarifying it by including a comnent in the easyconfig itself? You're hard overriding the `$CXXFLAGS` set by EasyBuild now, which is probably not the intention, since you only want to *add* a flag?

```python
# add -noswitcherror to avoid that libtool adds -pthread (which is not known by PGI compiler)
buildopts = 'CXXFLAGS="$CXXFLAGS -noswitcherror"'
```

Note: the switching of quotes is required to ensure that `$CXXFLAGS` gets expanded when the build command is run by EasyBuild...

vanzod:
@SimonPinches What is the rationale of adding two different version of `HDF5` that differ only by one microversion step?

SimonPinches:
I made v1.10.1 and then realised that v1.10.2 had been released so added that.  I'm happy to drop the v1.10.1 version if that's preferred.

boegel:
Yeah, just adding 1.10.2 probably makes sense, nobody is going to use the 1.10.1 one if 1.10.1 is also there. :)

easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
```diff
@@ -0,0 +1,26 @@
...
+toolchainopts = {'pic': True}
+
+source_urls = ['https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-%(version_major_minor)s/hdf5-%(version)s/src']
+sources = [SOURCELOWER_TAR_GZ]
+
+checksums = [
+    'bfec1be8c366965a99812cf02ddc97e4b708c1754fccba5414d4adccdc073866',  # hdf5-1.10.2.tar.gz
+]
+
+buildopts = 'CXXFLAGS="-noswitcherror"'
```
boegel: @SimonPinches As mentioned in the other (now hidden discussion): this needs to be changed to avoid overwriting the `$CXXFLAGS` set by EasyBuild.
Also included a comment to clarify why `-noswitcherror` is needed is probably useful:

```python
# add -noswitcherror to avoid that libtool adds -pthread (which is not known by PGI compiler)
buildopts = 'CXXFLAGS="$CXXFLAGS -noswitcherror"'
```

SimonPinches: This is now done in a new commit.

boegel: Hmm, seems like this isn't working as expected, I'm getting the exact problem you are trying to dance around (see incoming test report).

boegel:
Test report by @boegel
**FAILED**
node2556.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/bf916a9fe8fc4e34fe3a7e07f4addd2d for a full test report.

easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
SimonPinches: Weird!

SimonPinches:
I originally thought this too but the error message is `pgc++-Error-Unknown switch: -pthread` and is not associated with `pgfortran`.  Unless you can see some other reason for passing the `-noswitcherror` to the Fortran compiler as well?

vanzod:
@SimonPinches I was kinda grasping at straws by trying to understand where that `-lpthread` flag comes from. And in the local Makefile it was only in `FCFLAGS`. However, that does not seem to be the issue.
Even removing the `--with-pthread` that is automatically added by the `hdf5.py` easyblock does not solve the problem.

Maybe @akesandgren as some hints for us.

akesandgren:
If this uses openmpi built with slurm then you need to remove -pthread from inherited_linker_flags=' -pthread' in slurms libpmi and libslurm, similar to this

perl -pi -e 's/-pthread//' /lap/slurm/${version}/lib/libpmi.la /lap/slurm/${version}/lib/libslurm.la

SimonPinches:
I don't understand why this now doesn't work again.  Passing the `-noswitcherror` flag to the PGI compiler should get it to ignore any unrecognised flags (like `-pthread` which I think is added by `libtool`).  This works on my system and I think there was already a successful build on Travis with this flag when the `buildopts` were set as `buildopts = 'CXXFLAGS=-noswitcherror'`, i.e. the `eb` set `$CXXFLAGS` were skipped.

boegel:
@SimonPinches Travis doesn't actually test the build, it just runs a couple of simple checks on the easyconfig files (checks for missing dependencies, etc.).

The problem is that the `Unknown switch` occurs when *linking*, not compiling, and `-noswitcherror` isn't getting picked up from `$CXXFLAGS` when linking.

I think this is the correct fix (testing right now to confirm):

```python
buildopts = 'CXXLD="$CXX -noswitcherror"'
```

(the default is basically `CXXLD=$CXX`)

boegel:
And of course using `CXXLD` doesn't do the trick, it seems like that doesn't get picked up, but I'm not sure why.

boegel:
@SimonPinches After a bit of playing around: injecting `-noswitcherror` does work via `$CXX` and `preconfigopts`, see https://github.com/SimonPinches/easybuild-easyconfigs/pull/10

easybuild/easyconfigs/h/HDF5/HDF5-1.10.2-PGI-18.4-GCC-6.4.0-2.28.eb
```diff
@@ -0,0 +1,27 @@
...
+toolchainopts = {'pic': True}
+
+source_urls = ['https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-%(version_major_minor)s/hdf5-%(version)s/src']
+sources = [SOURCELOWER_TAR_GZ]
+
+checksums = [
+    'bfec1be8c366965a99812cf02ddc97e4b708c1754fccba5414d4adccdc073866',  # hdf5-1.10.2.tar.gz
+]
+
+# add -noswitcherror to avoid libtool adding -pthread (which is not known by PGI compiler)
```
boegel: @SimonPinches Strictly speaking the comment is not correct, it should say:

```
# add -noswitcherror to let PGI ignore unknown compiler option -pthread
```

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/d365a19f30e9c41127395e5163532f07 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2400.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/04ac1c89d2f4428cf84601b411764ff4 for a full test report.

boegel:

Issue #6295, opened 2018-05-05
Matlab README needs updating


cstackpole:
Greetings,
I was wanting to install Matlab. I attempted to follow along with the directions. I downloaded the two 2017 iso's, mounted them, rsync'd the contents to a 2017a folder, tar'd it up, and tried doing the install with eb. But it seemed like I was constantly having issues (some of my own doing, some because of Java [shakes fist :-)], some because the README is incomplete). I still don't have it working... I was also really confused about the README saying I needed to update the config file for the license. Huh? There is nothing about the license.

So I went poking around.

Looks like there used to be lines for the license but no longer. Also, it seems that we may not need to jump through unpacking the iso's anymore?? At least that is what I gathered from #408. 

I was so close to pushing up a new version of the README for how I was doing it, but then it failed again...(failing sanity tests and I don't know why). I figured no one wanted updated partial notes that don't work. :-D

Can someone who knows how to install Matlab with EasyBuild update the README for me please? How do I install Matlab with EasyBuild?

Thank you!

guest1604:
Hello @cstackpole,

Here my custom easyconfig file MATLAB-2017b.eb, I hope it can help you. For the source, I have unpacked the ISO files and created a "tar.gz" archive (as stated in README).

`key` parameter is the installation key __required for installing MATLAB__

```
name = 'MATLAB'
version = '2017b'

homepage = 'http://www.mathworks.com/products/matlab'
description = """MATLAB is a high-level language and interactive environment
 that enables you to perform computationally intensive tasks faster than with
 traditional programming languages such as C, C++, and Fortran."""

toolchain = {'name': 'dummy', 'version': ''}

# be sure to copy both DVD content to the SAME directory, 
# including the hidden files, especially .dvd1 and .dvd2 
sources = [SOURCELOWER_TAR_GZ]

dependencies = [('Java', '1.8.0_121')]

key = 'XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX'

modluafooter = """
prepend_path("MLM_LICENSE_FILE", "port@licsever1:port@licserver2:port@licserver3")
"""

moduleclass = 'math'
```

cstackpole:
Thanks! I will try this right now!!

cstackpole:
Thanks very much @guest1604 for that. It helped greatly and I found the issue!

[...]

PR #6341, opened 2018-05-18
update README on constructing source file for MATLAB

cstackpole:
The instructions I've been using that I can reliably reproduce Matlab installs. This is the work from the conversations in #6295 .


verdurin:
@cstackpole - thanks for this. Regarding your questions in #6295 I agree with the idea in the last one of providing a template file.

easybuild/easyconfigs/m/MATLAB/README.md
```diff
@@ -0,0 +1,32 @@
+This example uses Matlab 2018a and MATLAB-2018a.eb  
+Steps to install Matlab:
+1. Install the appropriate Java install which requires manual configuration. See Java documentation for install.
```
crubb: I'd suggest changing "configuration" to "manual download of the source" (I was looking for a configuration option, and didn't find any). What about including a link to the Java documentation?

easybuild/easyconfigs/m/MATLAB/README.md
```diff
@@ -0,0 +1,32 @@
...
+1. Matlab needs a "File Installation Key". There are a few options available:
+   1. Your Matlab license should have a "File Installation Key" in under "Advanced Options" in the "Install and Activate" tab of your "License Center". This will only install the products associated with this license.
+   1. If you need to install all of the toolboxes to support multiple licenses Mathworks can enable a "All Product File Installation Key". This will install all of the client products but will not install server products like MDCS. A second manual installation on top of the first will be required.
+   1. If you have a server product key, like MDCS, the "File Installation Key" may not install client products like Matlab Compilers. A second manual installation on top of the first will be required.
```
verdurin: Could the section about a requirement for a 'second manual installation' be clarified?

easybuild/easyconfigs/m/MATLAB/README.md
```diff
@@ -0,0 +1,32 @@
...
+1. Edit the EasyBuild script.
+   1. Update the version of Java to match the version installed in step 1.
+   1. Add the following three lines
+   ```
+   # This string could be really long
+   key = 'XXXXX-XXXXX-XXXXX-XXXXX-XXXXX-XXXXX'
+   license_server = '<IP or FQDN>'
+   license_server_port = '<port>'
```
verdurin: It would be useful to document how to handle the `Designated Computer` licensing model too.

easybuild/easyconfigs/m/MATLAB/README.md
akesandgren: Yes it is. But you could also ignore it and let MATLAB use its internal Java tree. This is what we are doing now.

easybuild/easyconfigs/m/MATLAB/README.md
akesandgren: More specifically, the latest matlab easyblock can take multiple file installation keys thus allowing you to install the base matlab and MDCS in a single install.

easybuild/easyconfigs/m/MATLAB/README.md
boegel: @verdurin I'll go ahead and merge this PR in it's current state, we can extend this README later...

boegel (approved):

PR #6354, opened 2018-05-23
add patch for Autoconf to enable ppc64le support

omula:
Downloaded from http://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob;f=build-aux/config.guess;hb=2ca0d5755f68ea7a99c39a236ea8d02aa207d594


easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-GCCcore-6.4.0.eb
```diff
@@ -20,6 +20,8 @@ source_urls = [GNU_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
 checksums = ['954bd69b391edc12d6a4a51a2dd1476543da5c6bbf05a95b59dc0dd6fd4c2969']
 
+patches = ['config.guess_ppc64le.patch']
```
boegel: @omula Please rename to something like `Autoconf-2.69_config.guess_ppc64le.patch`, and include a checksum for the patch itself (`eb --inject-checksums` can do this for you).

easybuild/easyconfigs/a/Autoconf/config.guess_ppc64le.patch
```diff
@@ -0,0 +1,645 @@
+--- a/build-aux/config.guess	2012-04-25 04:15:10.000000000 +0200
```
boegel: @omula Please include a short description + URL for upstream source of this patch above this line

easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-GCCcore-6.4.0.eb
```diff
@@ -18,7 +18,12 @@ toolchain = {'name': 'GCCcore', 'version': '6.4.0'}
 
 source_urls = [GNU_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
-checksums = ['954bd69b391edc12d6a4a51a2dd1476543da5c6bbf05a95b59dc0dd6fd4c2969']
+checksums = [
+    '954bd69b391edc12d6a4a51a2dd1476543da5c6bbf05a95b59dc0dd6fd4c2969',  # autoconf-2.69.tar.gz
+    '545b602ccfb3e8f27bc41b5f92a2e6aee672a2eaad73760f3fd31102cf2caf2b',  # Autoconf-2.69_config.guess_ppc64le.patch
+]
+
+patches = ['Autoconf-2.69_config.guess_ppc64le.patch']
```
boegel: @omula Please use `%(version)s` rather than hardcoding `2.69` here, and move this line up above `checksums =`

omula:
@boegel applying it to all installation shouldn't harm. It just helps to detect de system.

Are the changes ok?

easybuild/easyconfigs/a/Autoconf/Autoconf-2.69-GCCcore-6.4.0.eb
```diff
@@ -18,7 +18,13 @@ toolchain = {'name': 'GCCcore', 'version': '6.4.0'}
 
 source_urls = [GNU_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
-checksums = ['954bd69b391edc12d6a4a51a2dd1476543da5c6bbf05a95b59dc0dd6fd4c2969']
+
+patches = ['Autoconf-%(version)s_config.guess_ppc64le.patch']
+
```
boegel: @omula Style nitpicking: please drop the empty line above/below the `patches =` line, we usually keep `source_urls`, `sources`, `patches` and `checksums` "glued" together

boegel:
@omula I agree, always applying the patch should be OK.

It would be nice if we could do that in this PR... Are you up for that? I can take care of re-testing all touched easyconfigs, that's not a problem.

omula:
Done :)

boegel:
Relevant thread from `autoconf` mailing list. One of the follow-up posts seems to suggest there's another way to deal with this?
See https://lists.gnu.org/archive/html/bug-autoconf/2014-01/msg00011.html and https://lists.gnu.org/archive/html/bug-autoconf/2014-01/msg00012.html

omula:
@boegel I close this pr as I think it shouldn't be merged. I found a better solution which I plan to address in an other PR.

PR #6348, opened 2018-05-18
{bio}[intel/2018a] DIAMOND v0.9.22

boegel:
(created using `eb --new-pr`)

PR #6362, opened 2018-05-23
{bio}[GCCcore/6.4.0] DIAMOND 0.9.22 (REVIEW)

hajgato:


boegel:
@hajgato Any specific reason you're using a subtoolchain for this? Cfr. #6348

hajgato:
@boegel Since it is mostly `C++` code, I wanted to use `g++`. It does not use anything else but `g++`, so no `MPI`, no mathlibs, no fftw. Should I use `foss/2018a` instead?

boegel:
@hajgato Well, I'm not sure... We usually only use subtoolchains for stuff that we want to use as dependency for other things, so we can have a single easyconfig that can be used as dependency for something built with `foss` or `intel`, but that doesn't make much sense here?

hajgato:
@boegel ok, then  foss/2018a

pescobar:
Test report by @pescobar
**SUCCESS**
Build succeeded for 10 out of 10 (1 easyconfigs in this PR)
login20.cluster.bc2.ch - Linux centos linux 7.3.1611, Intel(R) Xeon(R) CPU E5-2670 v2 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/aed327f9ed5128954f0fcfdd6149698b for a full test report.

PR #6401, opened 2018-05-31
{data} [intel/2018a] MDSplus v7.46.1 (REVIEW)

SimonPinches:
New eb file for the latest version of MDSplus used widely within ITER.


easybuild/easyconfigs/m/MDSplus/MDSplus-7.7.12-intel-2018a.eb
```diff
@@ -0,0 +1,40 @@
...
+
+name = 'MDSplus'
+version = '7.7.12'
+
+homepage = 'http://mdsplus.org/'
+description = """MDSplus is a set of software tools for data acquisition and storage and a methodology
+ for management of complex scientific data."""
+
+toolchain = {'name': 'intel', 'version': '2018a'}
+toolchainopts = {'pic': True, 'opt': True, 'optarch': True}
```
boegel: @SimonPinches `opt` and `optarch` are enabled by default, so please trim to `toolchainopts = {'pic': True}`

easybuild/easyconfigs/m/MDSplus/MDSplus-7.7.12-intel-2018a.eb
```diff
@@ -0,0 +1,40 @@
...
+source_urls = ['https://github.com/%(name)s/%(namelower)s/archive']
+sources = ['stable_release-%s.zip' % version.replace('.', '-')]
+checksums = ['9b781339b62464419e5c8d8166f262729307d63e6aa18799dd924300a510cdab']
+
+dependencies = [
+    ('HDF5', '1.10.1'),
+    ('libxml2', '2.9.4'),
+    ('zlib', '1.2.11'),
+    ('ncurses', '6.0'),
+    ('libreadline', '7.0')
```
boegel: No more dependency on `Python`?

easybuild/easyconfigs/m/MDSplus/MDSplus-7.7.12-intel-2018a.eb
SimonPinches: Okay, I'll fix and push a new version shortly...

easybuild/easyconfigs/m/MDSplus/MDSplus-7.7.12-intel-2018a.eb
SimonPinches: If a system version is found it will be used but it's not a required dependency and gives more flexibility with what this module can be used with. E.g Python 2/3.

boegel:
@SimonPinches We should indeed rename the Python bindings to something else to avoid a name clash. `MDSplus-Python` works, but is maybe a bit silly since there's also a `-Python-` in the module version... Maybe `pyMDSplus`, for lack of a better option?

SimonPinches:
@boegel My suggestion of `MDSplus-Python` was by analogy with `netCDF` where you have separate modules for all the language interfaces, e.g. `netCDF-Fortran`, etc.  Personally, I would prefer `MDSplus-Python` over `PyMDSplus` since the latter sounds like it should be the name of the Python module or like it's a new interface separate from that shipped with the main distro, neither of which are true.  It may also be possible to separate out the `Java` interface, in which case I'd also create `MDSplus-Java` modules as well which use the libraries provided by the `MDSplus` modules.

boegel:
@SimonPinches Makes total sense to me, so let's go with `MDSplus-Python`

boegel:
@SimonPinches Let me know if you need any help with this

boegel:
More recent versions were merged via #10705, so there's little point in spending more effort on this PR => closing

PR #6411, opened 2018-06-01
{bio}[intel/2018a] RAxML-NG v0.5.1b

ajaust:
Easyconfig-file for `RAxML-NG`


boegel:
@ajaust Please rename the file to what is expect by `eb --robot` (i.e. drop the `v`)

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.5.1b-intel-2018a.eb
```diff
@@ -0,0 +1,35 @@
...
+
+homepage = 'https://github.com/amkozlov/raxml-ng'
+description = """RAxML-NG is a phylogenetic tree inference tool which uses maximum-likelihood (ML) optimality
+    criterion. Its search heuristic is based on iteratively performing a series of Subtree Pruning and
+    Regrafting (SPR) moves, which allows to quickly navigate to the best-known ML tree."""
+
+toolchain = {'name': 'intel', 'version': '2018a'}
+toolchainopts = {'usempi': True}
+
+source_urls = ['https://github.com/amkozlov/raxml-ng/releases/download/0.5.1']
```
boegel: @ajaust We should try to avoid hardcoding the version in the source URL... The `b` bit makes that a bit annoying, but you can use this approach:

```python
source_urls = ['https://github.com/amkozlov/raxml-ng/releases/download/%s' % version[:-1]]
```

ajaust: You are right. I don't know why I had this still in there. 

Is there a preferred way to automatically insert the version number? I have fixed this in the easybuild file by using `v%(version)s`.
```
source_urls = ['https://github.com/amkozlov/raxml-ng/releases/download/v%(version)s']
```

ajaust:
Hi. I just wanted to follow up on this pull request. I have changed the deduction of the version number. Do I have to change anything else?

Thanks in advance!

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.5.1b-intel-2018a.eb
```diff
@@ -0,0 +1,35 @@
...
+
+homepage = 'https://github.com/amkozlov/raxml-ng'
+description = """RAxML-NG is a phylogenetic tree inference tool which uses maximum-likelihood (ML) optimality
+    criterion. Its search heuristic is based on iteratively performing a series of Subtree Pruning and
+    Regrafting (SPR) moves, which allows to quickly navigate to the best-known ML tree."""
+
+toolchain = {'name': 'intel', 'version': '2018a'}
+toolchainopts = {'usempi': True}
+
+source_urls = ['https://github.com/amkozlov/raxml-ng/releases/download/v%(version)s']
```
boegel: @ajaust Download URL isn't correct now, the trailing `b` in the version is not included in the download URL...

Proposal:
```python
source_urls = ['https://github.com/amkozlov/raxml-ng/releases/download/v%s' % version[:-1]]

File: easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.5.1b-intel-2018a.eb [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+sources = ['%(namelower)s_v%(version)s_linux_x86_64_MPI.zip']
```
boegel: This is a tarball with pre-built binaries, shouldn't we download the `%(version)s.tar.gz` tarball instead?

ajaust: You are right. I think I should rename the package at least. I'm not sure if it would make sense to have an `easyconfig` file for the version with prebuilt binaries and one that builds the software completely from source with additional dependencies. For the latter case I have to look into `libpll` and create a easyconfig file for that first.

What would you prefer?

boegel: We generally strongly prefer building from source, since then we can target the underlying architecture, which usually significantly benefits performance.

ajaust: @boegel Hi. Sorry for the late reply. 
The last time I checked it, I had problem fulfilling the dependencies properly. Some library part was rewritten and some required module did not have any release. I will see if this has changed when I find the time.

migueldiascosta:
we now have an easyconfig for RAxML-NG, albeit for a more recent version and toolchain

@ajaust I'm closing this PR, but if this specific version is necessary we can come back to this, possibly adapting the more recent easyconfig to make them more uniform

PR #6420, opened 2018-06-04
{devel}[foss/2018a] Boost v1.63.0

smoors:
(created using `eb --new-pr`)
depends on easyblock PR #1434

PR #6415, opened 2018-06-01
{tools}[foss/2018a] IPython v5.7.0

smoors:
(created using `eb --new-pr`)

PR #6421, opened 2018-06-04
{bio,lang,lib,vis}[foss/2018a] EMAN2 v2.21a, PyQt v4.12.1, SIP v4.19.8, ...

smoors:
(created using `eb --new-pr`)
depends on ~~#6420, #6415~~


verdurin:
Test report by @verdurin
**SUCCESS**
Build succeeded for 5 out of 5 (4 easyconfigs in this PR)
easybuild.novalocal - Linux centos linux 7.5.1804, Intel Xeon E312xx (Sandy Bridge), Python 2.7.5
See https://gist.github.com/147d204ac54dcb92d5753c76da24c0ae for a full test report.

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
```diff
@@ -0,0 +1,56 @@
...
+
+homepage = 'http://blake.bcm.edu/emanwiki/EMAN2'
+description = """EMAN2 is the successor to EMAN1. It is a broadly based greyscale scientific image processing suite
+ with a primary focus on processing data from transmission electron microscopes. """
+
+toolchain = {'name': 'foss', 'version': '2018a'}
+
+source_urls = ['https://github.com/cryoem/eman2/archive']
+sources = ['v%(version)s.tar.gz']
+patches = ['EMAN2-2.21a.patch']
```
boegel: @smoors Please use `%(version)s` rather than hardcoding `2.21a`, and make the patch filename a bit more descriptive, e.g. `EMAN2-%(version)s_fix-custom-install-prefix.patch`

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
```diff
@@ -0,0 +1,56 @@
...
+checksums = [
+    '8a7310b60852220103a6676c7116f4119b290776de572a1510b982783bff46f7',  # v2.21a.tar.gz
+    '7399b4645c7fda87dc102f4ce3f7011764e40833984bb6cac025f840ed177a19',  # EMAN2-2.21a.patch
+]
+
+builddependencies = [('CMake', '3.10.2')]
+
+dependencies = [
+    ('Boost', '1.63.0', versionsuffix),
```
boegel: the test that checks for one single dep variant is going to trip over this, we've moved on to `Boost` `1.66` for the `2018a` toolchains...

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
```diff
@@ -0,0 +1,56 @@
...
+    ('freetype', '2.9'),
+    ('FTGL', '2.1.3-rc5'),
```
boegel: Maybe add a comment to clarify why we're using a release candidate here? Looks a bit fishy at first sight.

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
```diff
@@ -0,0 +1,56 @@
...
+preconfigopts = 'export CONDA_BUILD_STATE=BUILD && '
+preconfigopts += 'export PREFIX=%(installdir)s && '
+preconfigopts += 'export SP_DIR=%(installdir)s/lib/python%(pyshortver)s/site-packages && '
+preconfigopts += 'export FFTW3FDIR="$EBROOTFFTW" && '
+preconfigopts += 'export FFTW3DDIR="$EBROOTFFTW" && '
+preconfigopts += 'export HDF5DIR="$EBROOTHDF5" && '
+preconfigopts += 'export FTGLDIR="$EBROOTFTGL" && '
+
+sanity_check_paths = {
+    'files': [],
```
boegel: @smoors Existing `EMAN2` easyconfig check for specific scripts in `bin`, why stop that?

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
smoors: @boegel it does not work with Boost versions 1.66 and 1.65. would it be possible to get an exception for this?

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
smoors: @boegel it's the latest version since 2008, and is also the version included in ubuntu 16.04 and centos 7.4.

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
smoors: because there are many more files in `bin`, way too many to include, but I can add the existing ones back I you prefer.

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
smoors: @boegel I will make the change, but I'm a little annoyed by it.
What's the benefit of using `%(version)s` if we have to change it back to the hard-coded version in future versions of EMAN2? I suggest we choose one of the following:
1) use `%(version)s` always and make new patches for every new version
2) use the hard-coded version always and try to re-use patches as much as possible

smoors:
@boegel I've seen that you suggest making a new easyconfig with an added extension to solve the single dep issue.
Should I do that here too? If so, which extension to you suggest here? Maybe `-no_glibcxx11_abi` ?

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
boegel: You don't need to check for all of them, but checking for a couple is definitely better than checking for now.

easybuild/easyconfigs/e/EMAN2/EMAN2-2.21a-foss-2018a-Python-2.7.14.eb
boegel: It helps with realizing that the patch may need to be updated too when the version is bumped.

Sometimes you get lucky and you can reuse the patch that was made for a previous version though.

boegel:
@smoors There's no magic solution to the Boost dep issue here... This situation is quite different from the one in #6422 where a custom easyconfig for `ImageMagick` was added, since `ImageMagick` was not used as a dependency of an easyconfig file using a `2018a` toolchain yet.

In this case, `Boost` 1.66.0 is already used as a dependency in easyconfigs using a `2018a` toolchain, which means we're "stuck" to it.

I see two options:

* indicate that `EMAN2` is using an older version of `Boost` via the `versionsuffix` (like you did, but you'll need to fix the filename of the `EMAN2` easyconfigg), combined with adding an exception in the tests. I'm not a big fan of that last part, since it'll lead to more exceptions. Also, this doesn't solve the problem of making the `EMAN2` module compatible with already existing `2018a` easyconfigs though, which is what the failing test is highlighting.

* see if it's possible to patch `EMAN2` to make it compatible with Boost 1.66.0; that's definitely the best way forward, but it may be more work (although it's likely the required effort is minimal). Not only will that avoid us having to make an exception in the tests, it will also effectively make `EMAN2` compatible with other `2018a` easyconfigs that also depend on `Boost`...

smoors:
@boegel: @wpoely86 agrees that adding an exception for Boost is justified. The code is not trivial and they regularly break things with new versions.

boegel:
@smoors See https://github.com/smoors/easybuild-easyconfigs/pull/3

boegel:
Test report by @boegel
**SUCCESS**
node2646.swalot.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/d29b83fea384f38c68381ef255e77f87 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2006.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/3e804b9b5ac36b125d067eb1b7b8a7cf for a full test report.

boegel:
Going in, thanks @smoors!

PR #6424, opened 2018-06-05
{toolchain} foss/2018b

boegel:
(created using `eb --new-pr`)
requires ~~#6519~~ (OpenMPI) + ~~#6518~~ (OpenBLAS)

PR #6409, opened 2018-06-01
{toolchain} intel/2018b

boegel:
(created using `eb --new-pr`)

PR #6285, opened 2018-05-04
add py_expression_eval extension to (recent) Python 2.7.14 & 3.6.4 easyconfigs

smoors:
(created using `eb --new-pr`)

PR #6447, opened 2018-06-12
{lang}[iomkl/2018.02] Python v3.6.4, Tkinter v3.6.4, matplotlib v2.1.2

schiotz:
(created using `eb --new-pr`)

edit: depends on https://github.com/easybuilders/easybuild-easyblocks/pull/1441


schiotz:
I am submitting a version of Python build with the newest iomkl toolchain (2018.02).  I do, however, have a few doubts you may want to address before merging it.

* hashlib in the standard library fails to build out of the box.  I have used the patch provided by Andrew Edmondson in an email on the mailing list February 28, 2018.  He suggests somebody sanity checks his patch, I am not qualified to do that.  But I have tested that now hashlib builds with the sha3 methods, and that sha3_256() gives the same results as when building with the foss toolchain.

* Is iomkl/2018.02 merely a draft for iomkl/2018b, or does it make sense to use it?

easybuild/easyconfigs/p/Python/Python-3.6.3-icc_segfault_workaround.patch
```diff
@@ -0,0 +1,58 @@
+--- ./Modules/_sha3/kcp/KeccakP-1600-opt64.c.orig	2018-02-26 16:29:31.750301802 +0000
```
boegel: @schiotz Please include a pointer to where you got this patch above this line (line above this line are comments)

easybuild/easyconfigs/p/Python/Python-3.6.4-iomkl-2018.02.eb
```diff
@@ -0,0 +1,221 @@
...
+description = """Python is a programming language that lets you work more quickly and integrate your systems
+ more effectively."""
+
+toolchain = {'name': 'iomkl', 'version': '2018.02'}
+toolchainopts = {'pic': True}
+
+source_urls = ['http://www.python.org/ftp/%(namelower)s/%(version)s/']
+sources = [SOURCE_TGZ]
+checksums = ['7dc453e1a93c083388eb1a23a256862407f8234a96dc4fae0fc7682020227486']
+patches = ['Python-3.6.3-icc_segfault_workaround.patch']
```
boegel: @schiotz Please also include a SHA256 checksum for the patch file (`--inject-checksums` will do this for you)

easybuild/easyconfigs/p/Python/Python-3.6.4-iomkl-2018.02.eb
```diff
@@ -0,0 +1,221 @@
...
+    }),
+    ('chardet', '3.0.4', {
+        'source_urls': ['https://pypi.python.org/packages/source/c/chardet/'],
+        'checksums': ['84ab92ed1c4d4f16916e05906b6b75a6c0fb5db821cc65e70cbd64a3e2a5eaae'],
+    }),
+    ('certifi', '2018.1.18', {
+        'source_urls': ['https://pypi.python.org/packages/source/c/certifi/'],
+        'checksums': ['edbc3f203427eef571f79a7692bb160a2b0f7ccaa31953e99bd17e307cf63f7d'],
+    }),
+    ('urllib3', '1.22', {
```
boegel: @schiotz A whole bunch of additional extensions are included here... That's not a big issue, but we'll need to maintain these going forward.

How did you pick these?

We should try and ensure that all `Python` 3.6.4 easyconfigs include the same set of extensions, mainly to avoid confusion

Also, please include `py_expression_eval` as well (cfr. #6285)?

boegel:
@schiotz `iomkl/2018.02` is a toolchain like any other. We indeed also have `iomkl` versions that follow the common toolchain versioning scheme, but officially `iomkl` is not a common toolchain (yet).

For `foss/2018b` and `intel/2018b` there are a couple of significant differences, see #6424 and #6409. Mainly: `GCC` 7.3, `OpenMPI` 3.1.0, Intel tools 2018 update 3, etc.

boegel:
@vanzod The patch included here may be a fix to the problems you've been seeing with `hashlib` & Python 3.x?

easybuild/easyconfigs/p/Python/Python-3.6.4-iomkl-2018.02.eb
schiotz: @boegel I picked the extensions by copying Python-3.6.4-intel-2018a.eb  :-)

Since I did that, the py_expression_eval extension has been added, so I will add that one too, and update.

boegel: Ah, sorry for the confusion, it's the existing `Python-3.6.4-iomkl-2018a.eb` that is missing a bunch of extensions compared to the easyconfigs using `intel-2018a` and `foss-2018a`...

easybuild/easyconfigs/p/Python/Python-3.6.3-icc_segfault_workaround.patch
```diff
@@ -0,0 +1,60 @@
+This patch was provided by Andrew Edmondson in an email on the EasyBuild mailing list February 28, 2018.
```
boegel: @schiotz Please include https://lists.ugent.be/wws/arc/easybuild/2018-02/msg00043.html in the comment to point to the actual message.

What about the comment that this requires `-O0`? Maybe @edmondac can provide some more info on this?

easybuild/easyconfigs/p/Python/Python-3.6.4-iomkl-2018.02.eb
boegel: Inconsistencies in `iomkl/2018a` easyconfig are fixed in #6471

easybuild/easyconfigs/p/Python/Python-3.6.3-icc_segfault_workaround.patch
boegel: I mentioned the patch upstream, see https://bugs.python.org/issue33174, maybe they can provide some feedback on this.

edmondac: My email said: 

> I've just experienced this problem too. It seems to be a compiler bug with recent icc (it's still there in 2018.1.163). Python 3.6.x now includes sha3 which used to be a pypi module - and there's one tiny bit of code in there that creates a huge line in C using a macro. Icc segfaults on that line, presumably because it can't fit it in a buffer somewhere.
> 
> Anyway, I've made a patch (attached) which I'll be contributing back along with a Python-3.6.3-iomkl-2018a.eb in the next couple of weeks. If anyone fancies sanity checking the patch then I'd be very grateful! It does make some assumptions about which version of the macro to use (which it just unpacks explicitly into several lines) so someone might have a better idea. Even then, it still needs '-O0' to build it...

The -O0 was just found experimentally.

boegel: There's some feedback on your patch already, see https://bugs.python.org/issue33174#msg319943 .

In my experience the `-O0` isn't needed at all when the patch is included, at least not with `intel/2018.02` (but then again, I wasn't seeing this problem in the first place, for some reason unclear to me...).

cc @vanzod (who has also run into this)

edmondac: See https://software.intel.com/en-us/forums/intel-c-compiler/topic/780574

schiotz: @boegel The patch itself enables -O0 for the relevant file, so nothing extra needs to be done.

schiotz: I am now testing if -O0 is necessary for me.  Due to a PhD defence, it is possible that I will not get time to reply back until tomorrow.

schiotz: @edmondac Unfortunately, the link you provides is to a closed site.  Even registering does not give access to it :-(   But I assume it means that Intel is working on it.

schiotz: @boegel It *is* necessary to use -O0 to compile the relevant file.  I am uploading the new version with more comments in the patch file.

easybuild/easyconfigs/p/Python/Python-3.6.3-icc_segfault_workaround.patch
```diff
@@ -0,0 +1,65 @@
...
++
+     return originalDataByteLen - dataByteLen;
+ }
+--- setup.py.orig	2018-02-26 22:53:30.915149439 +0000
++++ setup.py	2018-02-26 22:53:01.409070915 +0000
+@@ -916,6 +916,7 @@
+         sha3_deps.append('hashlib.h')
+         exts.append( Extension('_sha3',
+                                ['_sha3/sha3module.c'],
++                               extra_compile_args=['-O0'],
```
boegel: @edmondac Which other options did you try here? `-O1`? `-O2` with specific optimizations disabled to try and dance around the problem?

I agree with the upstream concern that this severely limits the applicability of this patch, it really cripples the Python installation...

On which OSs are you seeing the segmentation fault? I'm wondering why I haven't seen this at all...

Does setting `ulimit -s unlimited` help by any chance to fix the build (without using this patch)?

schiotz: @boegel Why does this cripple the Python installation?  It is only the SHA3 hash module that is compiled with -O0, I guess it is rare that an application relies on performance of sha3 (a part from cracking passwords and perhaps mining cryptocurrencies, neither of which is usually done in Python).

vanzod:
I have encountered the following error with `Python-3.6.4-intel-2018a.eb`:
```
  File "/opt/easybuild/software/MPI/intel/2018.1.163-GCC-6.4.0-2.28/impi/2018.1.163/Python/3.6.4/lib/python3.6/hashlib.py", line 121, in __get_openssl_constructor
    f = getattr(_hashlib, 'openssl_' + name)
AttributeError: module '_hashlib' has no attribute 'openssl_sha3_224'
```

I have tested the patch in this PR and with `-O0` it builds fine. However, if the optimization is raised to `-O1` the error appears again.

easybuild/easyconfigs/p/Python/Python-3.6.3-icc_segfault_workaround.patch
boegel: As mentioned in https://bugs.python.org/issue33174, this will make the `sha3` support very slow.

I agree that this may not be very relevant to scientific software, but I'm not sure that's true or how to check that...

On the other hand, I do not see a good alternative either (although I do keep wondering why we haven't seen this issue at all).

@schiotz @vanzod On which OSs (version) have you run into the compilation problem without this patch?

vanzod:
@boegel Building `Python-3.6.4-intel-2018a.eb` with `ulitmit -s unlimited` makes the `hashlib` error go away without the patch.

easybuild/easyconfigs/p/Python/Python-3.6.3-icc_segfault_workaround.patch
boegel: Turns out `ulimit -s unlimited` is the reason I wasn't seeing this, let's follow up on this in #6484

schiotz:
@boegel have updated the PR, removing the patch.  It now depends on the updated easyblock, of course :)

boegel:
Test report by @boegel
**SUCCESS**
node2476.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/08c612806507d65ff37237b124db582f for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2007.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/ffda5bc4c777bf9f1eb78be589d053a0 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3129.skitty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/ff255d3cd78aef379979dedb8c241624 for a full test report.

boegel:
Going in, thanks @schiotz!

PR #5035, opened 2017-08-18
{lib,vis}[intel/2017a] libepoxy v1.4.3, Mesa v17.1.6

boegel:
(created using `eb --new-pr`)
requires ~~#5034~~

PR #6487, opened 2018-06-21
{lib}[foss/2018a] libepoxy 1.5.2 + enable dri/egl in Mesa 17.3.6

RvDijk:
`--enable-dri --enable-egl ` are required in the Mesa configs in order to succeeed.


easybuild/easyconfigs/l/libepoxy/libepoxy-1.5.2-foss-2018a.eb
```diff
@@ -0,0 +1,33 @@
...
+source_urls = ['https://github.com/anholt/libepoxy/archive/']
+sources = ['%(version)s.tar.gz']
+checksums = ['1ffc3ffe0567f986ff5731d4ef3a3765e0a4f9d642fffdae303362567eecd689']
+
+builddependencies = [
+    ('pkg-config', '0.29.2'),
+]
+
+dependencies = [
+    ('Mesa', '18.1.1'),
```
boegel: @RvDijk We need to stick to `Mesa` 17.3.6 for dependencies of easyconfigs that use a `2018a` toolchain, for compatibility with other easyconfigs using the same toolchain...

RvDijk: I see.. I'll stick to 17.3.6.

RvDijk:
@boegel do you mean that  dri and egl dependencies should also be made available through easybuild?

boegel:
@RvDijk If adding those configure options implies that extra libraries are required by `Mesa`, then yes...

RvDijk:
@boegel enabling these options results in installing additional libraries. Those libraries are provided by Mesa, e.g.:  https://dri.freedesktop.org/wiki/Building/. `libepoxy` does not build without these. It resolves the `No package 'egl' found` error that also shows in #5035.

boegel:
@vanzod Can you test this in your minimal test environment, to check whether switching to `--enable-dri` and `--enable-egl` for Mesa requires additional dependencies?

migueldiascosta:
**FAILED**
Build succeeded for 2 out of 3 (2 easyconfigs in this PR)
grc-cluster1 - Linux centos 6.10, Intel(R) Xeon(R) CPU E5-2640 0 @ 2.50GHz, Python 2.7.14
See https://gist.github.com/7151b4cdabd9ce0621ae9eb7eb34a107 for a full test report.

boegel:
@bartoldeman You mentioned in #7639 that adding an explicit `--enable-dri` is not a good idea?

lexming:
@RvDijk thanks for your PR. Enabling DRI requires a system with drivers for a GPU, which is not always the case in HPC. That's why the installation of Mesa in EB uses a software renderer.

Fortunately `libepoxy` does not require DRI to build. I have updated your PR to install it without the changes to Mesa. Please check https://github.com/easybuilders/easybuild-easyconfigs/pull/11520 to see the updated PR.
I'll close this PR in favour of that one.

RvDijk:
@lexming thanks clarifying and also taking care of this. At the time of writing this EB I wasn't aware of these options for `libepoxy`.

PR #6140, opened 2018-04-10
{chem}[foss/2017b] GPAW v1.3.0

martbhell:
The same as #5613 except we set OMP_NUM_THREADS=1

PR #5613, opened 2018-01-10
GPAW version 1.3.0 with PAW atomic setups

mikstr:
(created using `eb --new-pr`)

PR #5389, opened 2017-11-22
{chem}[intel/2017b] GPAW v1.3.0

Micket:
Patches add MKL support and fix dynamic python linking by removing a redundant -L flag.
libxc required only a toolchain version bump.

PR #5281, opened 2017-10-31
New GPAW 1.3.0 for foss/2017b

akesandgren:
(created using `eb --new-pr`)
GPAW 1.3.0, for foss/2017b. Contains a patch for customized.py that takes into account the earlier PRs for GPAW, PR#4400 and PR#4772

PR #6514, opened 2018-06-29
GPAW 1.4.0 with libvdwxc

Micket:
(created using `eb --new-pr`)


migueldiascosta:
**SUCCESS**
grc-cluster1 - Linux centos 6.9, Intel(R) Xeon(R) CPU E5-2640 0 @ 2.50GHz, Python 2.7.14
See https://gist.github.com/365a311604a5e0753ec40986f2dbaf37 for a full test report.

Micket:
Regarding the libxc-version, I think it will build just fine against 3.0.1, but why are we having such an old version in a 2018a toolchain?

easybuild/easyconfigs/a/ASE/ASE-3.16.0-intel-2018a-Python-3.6.4.eb
```diff
@@ -0,0 +1,30 @@
...
+sources = [SOURCELOWER_TAR_GZ]
+
+dependencies = [
+    ('Python', '3.6.4'),
+    ('Tkinter', '%(pyver)s', '-Python-%(pyver)s'),
+    ('matplotlib', '2.1.2', '-Python-%(pyver)s'),
+]
+
+sanity_check_paths = {
+    'files': ['bin/ase-build', 'bin/ase-db', 'bin/ase-gui', 'bin/ase-info', 'bin/ase-run'],
```
migueldiascosta: and `bin/ase` itself?

easybuild/easyconfigs/a/ASE/ASE-3.16.0-intel-2018a-Python-3.6.4.eb
```diff
@@ -0,0 +1,30 @@
+easyblock = 'PythonPackage'
+
+name = 'ASE'
+version = '3.16.0'
```
migueldiascosta: not `3.16.2`?

easybuild/easyconfigs/g/GPAW/GPAW-1.4.0-intel-2018a-Python-3.6.4.eb
```diff
@@ -0,0 +1,51 @@
...
+    ('libreadline', '7.0'),
+]
+
+prebuildopts = 'echo "libraries.append(\'xc\')" >> customize.py && '
+prebuildopts += 'echo "libvdwxc = True" >> customize.py && '
+prebuildopts += 'echo "libraries.append(\'vdwxc\')" >> customize.py && '
+
+prebuildopts += 'echo "scalapack = True" >> customize.py && '
+prebuildopts += 'echo "libraries.append(\'mkl_scalapack_lp64\')" >> customize.py && '
+prebuildopts += 'echo "libraries.append(\'mkl_blacs_intelmpi_lp64\')" >> customize.py && '
```
migueldiascosta: either get libraries from `$LIBSCALAPACK` or also submit a `foss` easyconfig for completeness?

migueldiascosta:
w.r.t. `libxc`, not sure about these particular versions but some codes lag behind in `libxc` support, so I suppose we should allow multiple versions per common toolchain, @boegel ?

migueldiascosta:
does anyone object to closing #6140, #5613, #5389, #5281, #4586, #4400 once this one is (eventually improved/expanded and) merged? Feedback welcome.

akesandgren:
This differs in method from #5281 which was a merge of several other attempts at GPAW.
Please take a look at that one before merging. I think that patching customize.py is a better way of doing this. And as far as i can see, this one doesn't get it quite right?

easybuild/easyconfigs/a/ASE/ASE-3.16.0-intel-2018a-Python-3.6.4.eb
Micket: Copy paste mistake. I will fix this. I should probably drop at least ase-gui as well, as the newer versions tells you to use "ase gui" instead. Perhaps the same applies to all the other  commands?

easybuild/easyconfigs/a/ASE/ASE-3.16.0-intel-2018a-Python-3.6.4.eb
Micket: GPAW 1.4.0 says that 3.16.0 was the corresponding release:
https://wiki.fysik.dtu.dk/gpaw/releasenotes.html#releasenotes
I don't mind bumping to 3.16.2 if you guys think it would still work.

Micket:
@akesandgren Ugh, I wanted to avoid patches since they are always so fragile. Your patch from 1.3.0 can't be applied to 1.4.0 anymore. If only they had just at least left customize.py empty.
In this example, I just appended the missing libraries, and LIBRARY_PATH, CPATH, MPICC and friends took care of the rest.

I can have a look at updating the patch for 1.4.0 if others prefer this method though. Do we have any method of simply replacing a file (without making an easyblock?). That would seem preferable.

migueldiascosta:
does anyone see a problem with editing the `customize.py` script from the easyconfig instead of with a patch? (I don't)

For an `foss` easyconfig, replacing the `mkl` `scalapack` and `blacs` lines with
```python
prebuildopts += 'echo "libraries.append(\'scalapack\')" >> customize.py && '
prebuildopts += 'echo "libraries.append(\'openblas\')" >> customize.py && '
prebuildopts += 'echo "libraries.append(\'gfortran\')" >> customize.py && '
```
(and adding an `ASE` `foss` easyconfig) seems to work

easybuild/easyconfigs/l/libvdwxc/libvdwxc-0.3.2-foss-2018a.eb
```diff
@@ -0,0 +1,31 @@
...
+    ('FFTW', '3.3.7'),
+]
+
+preconfigopts = 'unset CC && unset FC && '
+
+sanity_check_paths = {
+    'files': ['bin/%s' % x for x in ['libvdwxc_fdtest', 'libvdwxc_maintest',
+                                     'libvdwxc_q0test', 'libvdwxc_q0test2']] +
+             ['lib/lib%s.%s' % (x, y) for x in ['vdwxc', 'vdwxcfort']
+              for y in ['a', SHLIB_EXT]],
```
migueldiascosta: trailing whitespace

easybuild/easyconfigs/l/libvdwxc/libvdwxc-0.3.2-intel-2018a.eb
```diff
@@ -0,0 +1,31 @@
...
+    ('FFTW', '3.3.7'),
+]
+
+preconfigopts = 'unset CC && unset FC && '
+
+sanity_check_paths = {
+    'files': ['bin/%s' % x for x in ['libvdwxc_fdtest', 'libvdwxc_maintest',
+                                     'libvdwxc_q0test', 'libvdwxc_q0test2']] +
+             ['lib/lib%s.%s' % (x, y) for x in ['vdwxc', 'vdwxcfort']
+              for y in ['a', SHLIB_EXT]],
```
migueldiascosta: trailing whitespace

akesandgren:
One "advantage" with a patch is that when it fails it forces you to look at customize.py and figure out what they changed and if any of that has an effect on the build. And creating one isn't that hard.

But the important part is that the resulting customize.py should include all the things it need to include. And as far as I could see when taking a quick look, the currect "prebuildopts" way doesn't contain all that parts it has to in order for the build to pick up all things correctly.

Micket:
@akesandgren I.M.H.O that seems like a rather obscure advantage.

I noticed that there is actually a --customize "filename" flag we could use to point to a separate pre-made "eb_customize.py" instead of keeping it as a patch against the example snippets customize.py. We basically need nothing from the original anyway.

I made another attempt which should hopefully be a bit more robust, relying directly on $LIBSCALAPACK

I couldn't get LIBFFT to compile, I'm getting compilation errors due from numpy-array object. I don't see GPAW_WITH_FFT getting defined anywhere In the GPAW-1.3.0-customize-patch either, did those attempts truly build the FFTW bindings? GPAW specifies NPY_NO_DEPRECATED_API, but `c/fftw.c` uses several outdated APIs.

Right now, the sanity check fails
```
Sanity check failed: command "/apps/Hebbe/software/MPI/intel/2018.1.163-GCC-6.4.0-2.28/impi/2018.1.163/Python/3.6.4/bin/python -c "import gpaw"" failed; output:
Fatal error in MPI_Init: Other MPI error, error stack:
MPI_Init(176): Cannot call MPI_INIT or MPI_INIT_THREAD more than once
```
But I'm not sure it is supposed to be able to import gpaw into the normal python interpreter. After all, GPAW builds it's own "gpaw-python", which worked for a small sample code I had.

Micket:
Alright I was just stupid and misunderstood the GPAW build system, so I missed some aspects of the mpi / non-mpi split in the build. Since the list of prebuildopts would have needed to to grow so inconveniently large, I went ahead with Åkes patch instead, though I created a separate file for it instead so it will hopefully remain compatible

migueldiascosta:
**SUCCESS**
grc-cluster1 - Linux centos 6.9, Intel(R) Xeon(R) CPU E5-2640 0 @ 2.50GHz, Python 2.7.14
See https://gist.github.com/9253a8494061741f85482185f7e3b5b8 for a full test report.

migueldiascosta:
Following https://wiki.fysik.dtu.dk/gpaw/install.html#run-the-tests, I get some errors in the `libvdwxc` tests

- `vdw/libvdwxc_mbeef.py` fails the assertion `err < 1e-6` for both toolchains, maybe simply too stringent, but
- the parallel tests `vdw/libvdwxc_functionals.py` and `vdw/libvdwxc_h2.py` with the `intel` toolchain blow up completely, had to skip them (but with `foss` they run and pass)

Can anyone reproduce this?

migueldiascosta:
w.r.t to the tests, maybe someone at fysik.dtu.dk could have (access to) "inside information"? (@OleHolmNielsen?)

Micket:
@migueldiascosta 
In all 4 cases I get the error > 1e-6 tolerance on `vdw/libvdwxc_mbeef.py`
The libvdwxc_mbeef errors aren't huge, though still ~100 times larger than the permitted precision in the test. Perhaps a bugfix in libvdwxc not yet accounted for in the GPAW-test? Or perhaps we need to scale back the optimizations while building libvdwxc.

In addition, i get:
1. intel, sequential: hangs 100% load on `utilities/eed.py`
2. foss, sequential: nothing else

ran parallel 4 processes (seemed to be what most tests were compatible with):
3. intel, parallel: segfault on `vdw/libvdwxc_functionals.py`, `vdw/libvdwxc_h2.py`
4. foss, parallel: It is very very slow. Hundreds of times slower then the sequential build in many around half the test, while some tests run fine. I see nothing of the soft on the intel build.
While all the tests seem to pass (I'm still waiting for the last bunch, but it'll take another hour to finish at this rate)

migueldiascosta:
weird, @boegel also mentioned slow tests. My serial tests run in ~10 min (with -j 32) for both intel and foss, and the parallel tests (with -P 4) around 80 min both toolchains (but skipping the two failing vdw tests with intel)

With foss I do see a couple of warnings, including from `utilities/eed.py`, about `opal_init:warn-fork` which according to `help-opal-runtime.txt` mean:

> A process has executed an operation involving a call to the "fork()" system call to create a child process.  Open MPI is currently operating in a condition that could result in memory corruption or other system errors; your job may hang, crash, or produce silent data corruption.  The use of fork() (or system() or other calls that create child processes) is strongly discouraged.

Of course, if the problems are with the testing suite and not with our builds, we shouldn't get too hang up on them

schiotz:
Test report by @schiotz
**SUCCESS**
Build succeeded for 0 out of 0 (1 easyconfigs in this PR)
sylg.fysik.dtu.dk - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz, Python 2.7.5
See https://gist.github.com/c2900859f597eb042b08af70832183e4 for a full test report.

schiotz:
Test report by @schiotz
**FAILED**
sylg.fysik.dtu.dk - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz, Python 2.7.5
See https://gist.github.com/90b04feb5fd93610baec5ef2f753c9d0 for a full test report.

schiotz:
@Micket @migueldiascosta @boegel 

As GPAW developer, @mikstr approves of this PR (we are looking at it together now).  

Regarding the failing vdw/libvdwxc_mbeef.py test, we will look into it.  We do not normally use libvdwxc ourselves, so we do not know if the problem is with the test or with this build.  libwdvxc does give slightly different results than the built-in vdW support.

Regarding the crash with the Intel compiler: We do not have the Intel toolchain here, merely the iomkl, but have seen other problems with the 2018a Intel compiler.  It may, however, simply be a problem of using the FFTW multithreaded library together with MKL, that seems to be problematic, too.  

In general, we are a bit sceptical of the use of multi-threaded SCALAPACK, LAPACK and BLAS libraries, since GPAW handles parallelization with MPI and using multithreaded libraries usually oversubscribe the CPU, leading to slightly worse performance.

Regarding the version of libxc:  3.0.1 is pretty old, and there are newer XC functionals not available in it, so we would prefer the newer version in this PR.  However, GPAW will work with both versions.

Once this PR will be merged, I will take responsibility for making easyconfigs for new GPAW / ASE releases and new toolchains together with the main GPAW / ASE developer.

schiotz:
The selftest failure in `vdw/libvdwxc_mbeef.py` turns out to be more serious than it looks.  The problem is not in the mbeef_vdw functional, but in the generation of PAW setups.  This self-test just happens to do that.  There is also an issue with initialization of the wavefunctions, but that is less severe as it goes away during the self-consistency iterations.

The problem is being discussed here: https://gitlab.com/gpaw/gpaw/issues/161

So far, it looks like it is triggered by switching from libxc 3.X to 4.X.  It is unclear if the bug is in libxc-4.X or if it is a GPAW bug being triggered by this switch.

For the time being, I will make a new easyconfig with the new toolchains, but with libxc 3.X.  I will probably mark it WIP until this issue is better understood.

migueldiascosta:
can this be closed in favour of #6984?

PR #6569, opened 2018-07-09
{bio}[foss/2018a] Kraken2 2.0.6-beta (REVIEW)

hajgato:


easybuild/easyconfigs/k/Kraken2/Kraken2-2.0.6-beta_CXXFLAGS.patch
```diff
@@ -0,0 +1,12 @@
+tweak Makefile such that $CXX and $CXXFLAGS are picked up from the environment
+B. Hajgato (Free University Brussels, VUB), Kenneth Hoste (HPC-UGent)
+--- kraken2-2.0.6-beta/src/Makefile.orig	2018-06-26 21:19:26.000000000 +0200
++++ kraken2-2.0.6-beta/src/Makefile	2018-07-09 14:00:54.566891002 +0200
+@@ -1,5 +1,5 @@
+-CXX = g++
+-CXXFLAGS = -fopenmp -Wall -std=c++11 -O3
++CXX ?= g++
++CXXFLAGS ?= -fopenmp -Wall -std=c++11 -O3
```
vanzod: Why not using the following instead?
```
CXX := $(CXX) 
CXXFLAGS := $(CXXFLAGS)
```

hajgato: @vanzod Less typing and the end it is the same result.

hajgato: plus the previous `Kraken` versions were made the same way.

vanzod: My preference would be to use the flags that EB passes, unless there is a specific reason to have them hardcoded like that. Maybe we should summon @boegel here...

hajgato: @vanzod 
`FOO ?= BAR` means in make: if `FOO` is not defined as environmental variable, then `FOO` is `BAR` otherwise it takes the environmental variable (the default). See: https://www.gnu.org/software/make/manual/html_node/Setting.html

vanzod: My bad, I did not see the `?`. Makes sense now

vanzod:
Test report by @vanzod
**SUCCESS**
See https://gist.github.com/726aa7b673c3f082d52c91c637d115f4 for a full test report.

vanzod:
Going in, thanks @hajgato!

PR #6571, opened 2018-07-10
{lang}[foss/2018a] R 3.5.0

RvDijk:


boegel:
@RvDijk We prefer to stick to one `R` version per common toolchain, and we already have R 3.4.4 for `2018a`.

Also, R 3.5.1 is already released.

I have a WIP easyconfig for R 3.5.1 with `intel/2018b` in the pipeline, but that won't be finished before I go on summer leave (so it'll have to wait until early August).

RvDijk:
@boegel right, I'll close this.

PR #6612, opened 2018-07-25
{chem}[foss/2016b] Amber v16

SethosII:
This is my attempt for Amber built with the `foss` toolchain.

It's currenty not working and crashes with `/easybuild/test/software/binutils/2.26-GCCcore-5.4.0/bin/ld.gold: error: cannot find -lmpi` which is rather strange? I can provide the full log if needed. Any tips or hints are highly appreciated.

### DIFF
--- a/easybuild/easyconfigs/a/Amber/Amber-16-foss-2016b-AmberTools-16-patchlevel-24-15-CUDA.eb
+++ b/easybuild/easyconfigs/a/Amber/Amber-16-foss-2016b-AmberTools-16-patchlevel-24-15-CUDA.eb
@@ -1,0 +1,41 @@
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+
+name = 'Amber'
+version = '16'
+ambertools_ver = '16'
+# Patch levels from http://ambermd.org/ATPatches.php and http://ambermd.org/AmberPatches.php
+patchlevels = (24, 15)  # (AmberTools, Amber)
+versionsuffix = '-AmberTools-%s-patchlevel-%s-%s-CUDA' % (ambertools_ver, patchlevels[0], patchlevels[1])
+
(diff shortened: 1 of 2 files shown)


vanzod:
@SethosII I stumbled across the same problem when building it with `fosscuda`. It is related to the fact that the `amber` easyblock injects `-static` in the configure options. Since `OpenMPI` does not have static libraries, it fails. I opened an issue on this to get feedback from users/developers about removing it. See https://github.com/easybuilders/easybuild-easyblocks/issues/1461

SethosII:
@vanzod Thank you for the information. I build it successfully after removing `-static` from the easyblock. I removed `-static-intel` from the patch because this is the [default for the intel compiler](https://software.intel.com/en-us/cpp-compiler-18.0-developer-guide-and-reference-static-intel) and only valid for it. The URLs for the patches description have changed from http://ambermd.org/bugfixes16.html and http://ambermd.org/bugfixesat.html to http://ambermd.org/ATPatches.php and http://ambermd.org/AmberPatches.php. Should I update this in the existing easyconfigs?

migueldiascosta:
@SethosII the Amber easyblock now allows disabling static linking from the easyconfig (see https://github.com/easybuilders/easybuild-easyblocks/issues/1461 and https://github.com/easybuilders/easybuild-easyblocks/issues/1468) 

Also, #6618, for the intel toolchain, includes additional patches that should also be relevant for `foss`. I suppose it would make more sense to work on a `foss` easyconfig on top of it? (adding `cuda` and changing patchlevels seems trivial)

akesandgren:
And I will be doing foss and fosscuda/intelcuda builds once #6618 is merged.

SethosII:
Ok, so I guess we can close this one. I looked into @akesandgren's pull request and it seems much more elaborate than this.

PR #6633, opened 2018-08-02
VCFtools - Added explicit zlib dependency

vanzod:
(created using `eb --new-pr`)


vanzod:
The explicit `zlib` dependency is necessary otherwise `VCFtools` will link against and use the system `zlib` independently if a `zlib` module is loaded. I presume the path gets hardcoded somewhere at build time (not RPATH though).

easybuild/easyconfigs/v/VCFtools/VCFtools-0.1.15-foss-2016b-Perl-5.24.0.eb
```diff
@@ -21,6 +21,7 @@ checksums = ['31e47afd5be679d89ece811a227525925b6907cce4af2c86f10f465e080383e3']
 dependencies = [
     ('Perl', '5.24.0'),
     ('tabix', '0.2.6'),
+    ('zlib', '1.2.8'),
```
boegel: This one results in a conflict on `zlib` resolved by `zlib-1.2.8-foss-2016b.eb` vs one for the dependencies that is resolved via `zlib-1.2.8-GCCcore-5.4.0.eb` (due to the historical mistake of including `zlib-1.2.8-foss-2016b.eb`).

Possible ways out:

* remove `zlib-1.2.8-foss-2016b.eb` (`zlib` dep can still be resolved via `zlib-1.2.8-GCCcore-5.4.0.eb`)
* hardcode `zlib` dep here to use `GCCcore/5.4.0` subtoolchain

vanzod: In the end is mainly an issue of making Travis happy since for the final user it should not matter since the version is `1.2.8` for both toolchains.

boegel: It's broader than just Travis, you'll run into the same problem when actually building this without applying one of the solutions... You can't have `zlib-1.2.8-foss-2016b.eb` and `zlib-1.2.8-GCCcore-5.4.0.eb` loaded at the same time.

vanzod: True. At this point if I have to choose I'd rather remove `zlib-1.2.8-foss-2016b.eb`

boegel: OK, let's go for it, if it does cause problems the tests should tell us.

boegel:
@vanzod Looks like we should trash `zlib-1.2.8-GCC-5.4.0-2.26.eb` too?

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 14 out of 15 (10 easyconfigs in this PR)
node2003.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/d0aa1fe4e414f3bdd299cbb0c13e1cc7 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2110.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/606ee6e537b20cb5b99ef33b3231b61a for a full test report.

boegel:
Going in, thanks @vanzod!

PR #6636, opened 2018-08-06
Update hwloc easyconfigs with libxml2 and libpciaccess dependencies.

akesandgren:
(created using `eb --new-pr`)
Update recent hwloc easyconfigs with libxml2 and libpciaccess dependencies and explicitly disable some configure options.


easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb
```diff
@@ -25,15 +25,20 @@ source_urls = ['https://www.open-mpi.org/software/hwloc/v%(version_major_minor)s
 sources = [SOURCE_TAR_GZ]
 checksums = ['0a2530b739d9ebf60c4c1e86adb5451a20d9e78f7798cf78d0147cc6df328aac']
 
+osdependency = [('libudev-dev', 'libudev-devel')]
```
boegel: @akesandgren As discussed on Slack, I'm not sure this is a good idea...

As I understand, `libudev` is an optional dependency for `hwloc`? Including it as OS dep makes it mandatory (unless `--ignore-osdeps` is used).

Also, it seems like on `CentOS` this should be `libgudev1-devel`?

easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb
```diff
@@ -25,15 +25,20 @@ source_urls = ['https://www.open-mpi.org/software/hwloc/v%(version_major_minor)s
...
     ('binutils', '2.30'),
 ]
 
 dependencies = [
     ('numactl', '2.0.11'),
+    ('libxml2', '2.9.8'),
+    ('libpciaccess', '0.14'),
 ]
 
 configopts = "--enable-libnuma=$EBROOTNUMACTL"
```
boegel: missing space here at the end (same for several other easyconfigs)

easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb
akesandgren: libgudev1-devel is udev for those that use glib, libudev-devel is the normal one.

easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb
akesandgren: missing space??

easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb
akesandgren: And you said: "hmm, if it's an optional dep, we should include it as an OS dep since that'll make it a hard requirement" in the channel... so i included it...

easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb
akesandgren: Ahh for the configopts... fixing...

easybuild/easyconfigs/h/hwloc/hwloc-1.11.10-GCCcore-7.3.0.eb
boegel: We don't have `libudev-devel` installed on our systems, so in our case adding this OS dep would be problematic.

W.r.t. the statement on Slack: I clearly hadn't have enough coffee (or wine...) yet at that point. What I meant to say was that we shouldn't make `libudev-devel` as hard requirement by listed it as an OS dep if it's actually an optional dependency...

boegel:
@akesandgren I know I asked you to try and do this for the most recent `hwloc` easyconfigs, but it turns out that results in a lot of version conflicts for `libxml2`, mainly because we haven't been very thorough in sticking to a single dep version per toolchain generation.

It's really only feasible to make this change now for the most recent `hwloc` easyconfig (`hwloc-1.11.10-GCCcore-7.3.0.eb`), and discard the changes for older `hwloc` easyconfigs (which is OK in this case imho, since this hasn't been a huge issue for anyone)...

akesandgren:
Closing this PR due to too many version conflicts for libxml2 and zlib.
A new slimmer one with only the most recent hwloc versions coming up soon.

boegel:
follow-up in #6639

PR #6652, opened 2018-08-08
{bio}[foss/2018a] SAMtools v1.9

akesandgren:
(created using `eb --new-pr`)
New version of SAMtools for foss/2018a

PR #6664, opened 2018-08-09
{bio}[GCC/6.4.0-2.28] SAMtools v1.9 + HTSlib v1.9

jdemeul:
(created using `eb --new-pr`)


vanzod:
@jdemeul I think both easyconfig files could be lowered to the `GCC-6.4.0-2.28` toolchain

jdemeul:
@vanzod excellent, they can indeed, just updated the PR

jdemeul:
Actually, probably easier for you if I make a separate PR for the easyconfigs with `GCC-6.4.0-2.28`toolchain?

akesandgren:
See also https://github.com/easybuilders/easybuild-easyconfigs/pull/6652

boegel:
@vanzod For end-user stuff that is performance critical we usually go with top-level toolchains, no?

boegel:
@vanzod No, and you're right, as long as we're using `GCC` not `GCCcore` (the latter would imply that the `GCCcore` easyconfig can be used to resolve dependencies when building with an `intel` compiler, which was actually my concern).

@jdemeul To go forward, let's remove the easyconfigs using `foss/2018a`? This should do the trick:

```
$ eb --update-pr 6664 :easybuild/easyconfigs/h/HTSlib/HTSlib-1.9-foss-2018a.eb : easybuild/easyconfigs/s/SAMtools/SAMtools-1.9-foss-2018a.eb
```

jdemeul:
Done, thanks for the input @vanzod @boegel !

vanzod:
Test report by @vanzod
**SUCCESS**
See https://gist.github.com/35edd67ff8b32d2701d5356f39b942b0 for a full test report.

vanzod:
Going in, thanks @jdemeul!

PR #5203, opened 2017-10-10
{lang}[dummy/,dummy/dummy] Java v1.7.0, Java v1.8.0, JDK v1.7.0_80, ...

JackPerdue:
(created using `eb --new-pr`)

PR #6677, opened 2018-08-11
{lib}[fosscuda/2018b] TensorFlow v1.10.0

akesandgren:
(created using `eb --new-pr`)
New version of TensorFlow (1.10.0) for fosscuda 2018b. Depends on ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1453~~

PR #6712, opened 2018-08-16
add Java 1.8 easyconfig that can be updated in-place to more recent JDK 1.8.x

boegel:
cfr. discussion in #5203

I feel we should use this approach going forward (starting with the `*2018b*` toolchain generation), but we shouldn't change older easyconfigs to also adapt to this policy since that will lead to too much confusion.
We can easily switch to this policy for the `2018b` toolchain generation since currently only 1 easyconfig already merged in `develop` is affected (`Bazel`), and it wasn't included in an EasyBuild release yet.

cc @JackPerdue 

@verdurin, @vanzod: Since you were critical about following this approach, I'd like to hear back from you before this gets merged...

edit: now requires ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1503~~


easybuild/easyconfigs/b/Bazel/Bazel-0.16.0-GCCcore-7.3.0.eb
```diff
@@ -12,6 +12,6 @@ sources = ['%(namelower)s-%(version)s-dist.zip']
 checksums = ['c730593916ef0ba62f3d113cc3a268e45f7e8039daf7b767c8641b6999bd49b1']
 
 builddependencies = [('binutils', '2.30')]
-dependencies = [('Java', '1.8.0_162', '', True)]
```
akesandgren: Note that this might not merge if the currently pending TensorFlow 1.10 gets merged first, since it uses _172

boegel: I'm aware of that, I'm hoping #6677 gets merged first, and then I'll resolve the conflict here.

vanzod:
@boegel I have changed my mind about the wrapping. I was young back then...

boegel:
@vanzod So, to be clear, no objections from your side on this?

easybuild/easyconfigs/j/Java/Java-1.8.eb
```diff
@@ -0,0 +1,23 @@
+# Authors:: Jack Perdue <j-perdue@tamu.edu> - TAMU HPRC - http://hprc.tamu.edu
+
+easyblock = 'Bundle'
+
+name = 'Java'
+version = '1.8'
```
boegel: Hmm, would `8` make more sense here? Whatever the source tarball may be named, the actual version is actually `Java` 8, right?

JackPerdue: I don't really care....
[j-perdue@tlogin-0502 (08:50) ~]$ ml Java/1.8.0
[j-perdue@tlogin-0502 (08:50) ~]$ java -version
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)

I just went with the naming standard that EB already used. 1.8 seems to make some sense.

JackPerdue: might be a good idea for 9

[j-perdue@login2 (08:59) ~]$ ml Java/9.0
[j-perdue@login2 (09:00) ~]$ java -version
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)

boegel: This thought came up mostly due to the new Java 10, jumping from 1.8 to 10 would be sort of strange.

Also, I think using `('Java', '8')` would make it more clear that we're using a wrapper, since you'd expect a `major.minor`...

boegel: At www.oracle.com/technetwork/java/javase/downloads they refer to it as `Java SE 8` and `Java 8u181`.

What `java -version` prints is probably mostly a concern with backwards compatibility, which they broke in Java 9.x

JackPerdue: your call... I don't care either way

verdurin:
@boegel I'm a bit confused - why are all these other changes in here now?

boegel:
@verdurin Hmm, not sure what went wrong there... The other commits were all stuff that's already in `develop`. I did a `git rebase develop` + `push --force` for this branch, which seems to have fixed it.

easybuild/easyconfigs/b/Bazel/Bazel-0.16.0-GCCcore-7.3.0.eb
verdurin: @boegel Doesn't look like #6677 will merge before this one.

easybuild/easyconfigs/j/Java/Java-8.eb
```diff
@@ -0,0 +1,23 @@
+# Authors:: Jack Perdue <j-perdue@tamu.edu> - TAMU HPRC - http://hprc.tamu.edu
+
+easyblock = 'Bundle'
+
+name = 'Java'
+version = '8'
```
verdurin: This makes quite a lot of changes all in one go - I feel this should be made more prominent somehow.

boegel: Can you clarify? All we're changing here is using a wrapper for `Java`...

This change in policy will be clearly mentioned in the EasyBuild release notes, and maybe we should also highlight it with a separate post on the mailing list...

verdurin: I don't disagree, just noting that we're making lots of simultaneous changes, so we should communicate this as widely as possible.

boegel:
As highlighted by @vanzod, the current approach using using a `Java/8` (or `Java/1.8`) wrapper around a fully versioned `JDK` module causes problems with Lmod module collections (see https://github.com/easybuilders/easybuild-easyconfigs/pull/5203#issuecomment-414325270 for more details).

So, if we want to go through with this we need to do it without a wrapper: a `Java-8.eb` easyconfig that installs a specific version (e.g. `1.8.0_181`), and which gets updated in place over time.

The downside of this approach is that it will take a little bit longer to do an in-place update using `eb Java-8.eb -df`, since it involves an actual re-installation rather than simply replacing a wrapper module, so changes are a bit higher of causing troubling by briefly removing the `Java/8` module.

I see no other option, so we should discuss whether or not to go through with this.
We still have a window of opportunity here w.r.t. the upcoming EasyBuild release (`3.7.0`) and the `2018b` toolchain generation, so now's a good time to make a decision.

JackPerdue:
I don't like it.  I specifically wrote my version for quick rebuild of the module while in live use.  If "module save" has problems then lets talk to Robert about it.

JackPerdue:
And I can update the damn versions... its only taken 10 months for this issue to come to some (wrong) resolution.

boegel:
@JackPerdue We can't change the policy w.r.t. how we deal with `Java` as a dependency at any random moment, part of the reason why this took so long to be picked up is because we have a window of opportunity now to take the approach you suggested for the `2018b` toolchain generation & beyond.
That, and the fact that we're all struggling to find time to work on things like this unfortunately led to where we are now w.r.t. time...

I consider the problem that @vanzod brought up w.r.t. module collections a blocker for going through with the wrapper approach: this would hit any user relying on module collections as soon as the `Java` wrapper is updated in-place, which is not acceptable imho. Especially because a user could restore a collections weeks or months after the `Java` wrapper was updated in place, so there's no correlation time-wise between the update & the `ml restore` failure...

We can let @rtmclay pitch in on this, but I'm not sure it matters too much: the check that Lmod does makes sense to me, since the user can't restore the module collection anymore exactly as it was saved...

As already mentioned, the other option we have is to simply change the versioning scheme of the `Java` easyconfigs to only use the major version (e.g. `8` or `10` in the near future) in `version`, but use a fully specified version in `sources` and update that over time, and then let sites do a forced reinstall of `Java` whenever they see fit.

Your suggested approach is better since it only involves rewriting a module file for the wrapper, so it's fast, while now updating `Java` in place involves an actual re-installation. But (re-)installing `Java` is fast enough that this is mostly a non-argument imho...

Come to think of it, another downside of going forward without the wrapper is that we'll effectively be *removing* an existing `Java` installation, so the impact is potentially more dramatic compared to just changing a wrapper (where no actual software installations are removed), in the sense that it may affect running jobs that all of a sudden can't find `java` anymore for a couple of seconds.

That may just be an academic concern though, reinstalling `Java` takes about 3 seconds on an (fairly old) Intel Sandy Bridge system, I just timed that...

boegel:
Thanks to a suggestion by @mboisson, it looks like we do have a better option after all: effectively making `Java/1.8` a wrapper for `Java/1.8.0_181` via `module-version` in a `.modulerc` file (see also https://github.com/easybuilders/easybuild-easyblocks/pull/1503)...

boegel:
Test report by @boegel
**SUCCESS**
node3140.skitty.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/0cfcb80fde3a73c66b4e348f621622e9 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2404.golett.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/b030f77ed23d1a9d86043596d70b39a0 for a full test report.

migueldiascosta:
test report coming soon, and also manually tested other easyconfigs (e.g. octave), looks good

@bartoldeman did you mean to review this yourself?

migueldiascosta:
**SUCCESS**
grc-cluster1 - Linux centos 6.9, Intel(R) Xeon(R) CPU E5-2640 0 @ 2.50GHz, Python 2.7.14
See https://gist.github.com/173a52f3c8220195485142a40ff2cb92 for a full test report.

bartoldeman:
I am working on a test report myself, should be there in 2 hours or so

bartoldeman:
Test report by @bartoldeman
**SUCCESS**
aw-4r13-n22 - Linux centos 6.6, Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz, Python 2.6.6
See https://gist.github.com/f761aa5d98339ce5ff90038ee6d2a74c for a full test report.

bartoldeman:
going in, thanks @boegel !

PR #6714, opened 2018-08-16
{tools}[GCCcore/7.3.0] pigz v2.4

jdemeul:
(created using `eb --new-pr`)


easybuild/easyconfigs/p/pigz/pigz-2.4_Makefile.patch
```diff
@@ -0,0 +1,19 @@
+diff -Nru pigz-2.4.orig/Makefile pigz-2.4/Makefile
```
boegel: @jdemeul Please include a short description above this line + mention author (and then fix the checksum in the easyconfig)

easybuild/easyconfigs/p/pigz/pigz-2.4_Makefile.patch
```diff
@@ -0,0 +1,19 @@
...
+--- pigz-2.4.orig/Makefile	2017-12-27 00:35:01.000000000 +0000
++++ pigz-2.4/Makefile	2018-08-13 17:35:26.000000000 +0100
+@@ -1,5 +1,4 @@
+-CC=gcc
+-CFLAGS=-O3 -Wall -Wextra -Wno-unknown-pragmas
++ZLIB=-L$(EBROOTZLIB)/lib
```
boegel: Hmm, I don't see why this is necessary... Shouldn't the right `libz.*` get picked up via `$LIBRARY_PATH`, which specifies locations for the linker to consider?

easybuild/easyconfigs/p/pigz/pigz-2.4_Makefile.patch
```diff
@@ -0,0 +1,19 @@
+diff -Nru pigz-2.4.orig/Makefile pigz-2.4/Makefile
+--- pigz-2.4.orig/Makefile	2017-12-27 00:35:01.000000000 +0000
++++ pigz-2.4/Makefile	2018-08-13 17:35:26.000000000 +0100
+@@ -1,5 +1,4 @@
+-CC=gcc
+-CFLAGS=-O3 -Wall -Wextra -Wno-unknown-pragmas
```
boegel: Rather than wiping this, maybe it's better to use `CC:=` such that the `$CC` set in the environment get precedence?

easybuild/easyconfigs/p/pigz/pigz-2.4-GCCcore-7.3.0.eb
```diff
@@ -0,0 +1,51 @@
...
+
+builddependencies = [
+    ('binutils', '2.30'),
+]
+
+dependencies = [
+    ('zlib', '1.2.11'),
+]
+
+parallel = 1
```
boegel: This is not need (anymore), we don't have it in `pigz-2.4-foss-2018a.eb` either

jdemeul:
@boegel, you're right, works perfectly without the patch, simply passing along $CC and $CFLAGS as build options like in the `foss-2018a` easyconfig does the trick.
Thanks again!

boegel:
Test report by @boegel
**SUCCESS**
node3150.skitty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/893a960034401089ab4f9c00b47ff37f for a full test report.

boegel:
Going in, thanks @jdemeul!

PR #6704, opened 2018-08-15
various tools & libraries built with foss/2018b

fizwit:
adding libraries for foss-2018b

edit: depends on ~~#6656~~

PR #6717, opened 2018-08-16
Libjpeg turbo 2.0

fizwit:
switched to Python 3.6.6 and made all the recommended changes.


boegel:
@fizwit This largely overlaps with #6704, doesn't it?

boegel:
I've confirmed (via `eb --review-pr`) that everything here is already merged into `develop` via #6704, so I'll close this.

PR #6728, opened 2018-08-18
fix source_urls in bzip2 easyconfigs

boegel:
It seems like the http://bzip.org/ domain has expired, and is up for sale...

The only seemingly reliable/appropriate alternative I could find for downloading `bzip2-1.0.6.tar.gz` is `https://fossies.org/linux/misc`...

### DIFF
--- a/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
+++ b/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
@@ -9,7 +9,11 @@ description = """bzip2 is a freely available, patent free, high-quality data com
 toolchain = {'name': 'CrayGNU', 'version': '2015.06'}
 toolchainopts = {'pic': True}
 
+source_urls = [
+    'https://fossies.org/linux/misc/',
+    'http://www.bzip.org/%(version)s/',
+]
 sources = [SOURCE_TAR_GZ]
-source_urls = ['http://www.bzip.org/%(version)s']
+checksums = ['a2848f34fcd5d6cf47def00461fcb528a0484d8edef8208d6d2e2909dc61d9cd']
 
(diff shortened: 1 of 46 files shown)


boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 43 out of 43 (43 easyconfigs in this PR)
node2036.delcatty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/1ae70020a4289b6ac6234ec67e55aee3 for a full test report.

boegel:
see also https://lwn.net/Articles/762264/ (hat tip @chrissamuel)

wpoely86:
https://snapshot.debian.org/package/bzip2/1.0.6-9/ maybe?

I would remove the old url as the domain can be hijacked from now.

boegel:
@wpoely86 What's wrong with the `fossies.org` one?

From https://snapshot.debian.org/package/bzip2/1.0.6-9/, I would need to use `https://snapshot.debian.org/archive/debian/20111204T160411Z/pool/main/b/bzip2/bzip2_1.0.6.orig.tar.bz2` which also involves renaming the tarball...

Removing the old URL makes sense, although the SHA256 checksum should protect against downloading another source tarball...

Proposal for new `homepage` value?

wpoely86:
I've never heard of fossies.org while Debian is better known. The URL is indeed not very nice.

damianam:
+1 for fossies. Alternatively, archive.org....... https://web.archive.org/web/20180624184835/http://www.bzip.org/1.0.6/bzip2-1.0.6.tar.gz but I don't like this option too much.

boegel:
@damianam What about the homepage? I don't have a good alternative there...

vanzod:
Going in, thanks @boegel!

PR #6731, opened 2018-08-19
specify location of dependencies in configure options for libgd (to avoid system libraries being used instead)

boegel:
fix for problem reported by @crubb (via IRC)

PR #6617, opened 2018-07-26
{devel,math,vis}[foss/2018b] OpenFOAM v1806

furstj:
(created using `eb --new-pr`)

PR #6732, opened 2018-08-19
{bio}[foss/2018b] FSL 5.0.11 (upgrade, including some dependencies now with foss/2018b or GCCcore-7.3.0)

crubb:
Build for `FSL-5.0.11` using `foss/2018b`. The patches from FSL <= 5.0.10 were redone to merge without fuzzing (i.e. only line numbers were touches).

The following dependencies now also have `foss/2018b` EBs:

- freeglut-3.0.0
- libgd-2.2.5 (taking into account #6731)
- libpng-1.6.34
- VTK-8.1.1-Python-2.7.15


easybuild/easyconfigs/f/FSL/FSL-5.0.11-foss-2018b.eb
```diff
@@ -0,0 +1,40 @@
...
+]
+checksums = [
+    '8a76d36ff280ab3bfa79c46838a1baf7ea8ffe21d0b84bc7546cb8a1cef46c57',  # fsl-5.0.11-sources.tar.gz
+    'dcf76c2f54eaa729c300a917794a5cd3c37dc46205a7b415eb866e1a871f2dbf',  # FSL-5.0.11_makefile_fixes.patch
+    '10cd2d6c66ee0488244036743e70d211dbbdfadac651e5c95dd1de1f676a4b38',  # FSL-5.0.9_missing_lib.patch
+    '9f29713c4c9610fc9beb20b08f2c3a90499a5d17ec57f6beb6193530fe121b46',  # FSL-5.0.11_build_extras.patch
+]
+
+# Depends on
+# - https://github.com/easybuilders/easybuild-easyconfigs/pull/6700
```
boegel: @crubb No need to include this in the easyconfig file, just mentioning it in the description is fine

easybuild/easyconfigs/l/libgd/libgd-2.2.5-foss-2018b.eb
```diff
@@ -0,0 +1,31 @@
+easyblock = 'ConfigureMake'
+
+name = 'libgd'
+version = '2.2.5'
+
+homepage = 'https://libgd.github.io/'
+description = "GD is an open source code library for the dynamic creation of images by programmers."
+
+toolchain = {'name': 'foss', 'version': '2018b'}
```
boegel: @crubb Using `GCCcore/7.3.0` would make more sense here (then this same easyconfig can be used to resolve dependencies when other toolchains like `intel/2018b` are used too).

If you change this, you'll need to change the easyconfig filename accordingly, and also include `binutils` in `builddependencies`

easybuild/easyconfigs/l/libpng/libpng-1.6.34-foss-2018b.eb
```diff
@@ -0,0 +1,31 @@
+easyblock = 'ConfigureMake'
+
+name = 'libpng'
+version = '1.6.34'
+
+homepage = 'http://www.libpng.org/pub/png/libpng.html'
+
+description = "libpng is the official PNG reference library"
+
+toolchain = {'name': 'foss', 'version': '2018b'}
```
boegel: @crubb You can just drop this one, and use the already existing `libpng-1.6.34-GCCcore-7.3.0.eb`

easybuild/easyconfigs/f/FSL/FSL-5.0.11-foss-2018b.eb
```diff
@@ -0,0 +1,40 @@
...
+
+dependencies = [
+    ('Boost', '1.68.0', '-Python-2.7.15'),
```
boegel: @crubb As indicated by the failing test, please stick to `Boost` 1.67.0 in easyconfigs using a `2018b` toolchain.

We stick to one particular version of a dependency per toolchain generation to try and make sure that easyconfigs (for otherwise unrelated software) using the same toolchain generation are compatible with each other.

Currently, anything that uses `Boost` 1.67.0 would be incompatible with this FSL, if someone would want to use them together for whatever reason...

crubb:
All comments/change requests addressed. Thank you for the review!

easybuild/easyconfigs/f/FSL/FSL-5.0.11-foss-2018b.eb
akesandgren: That's odd, there are more Boost 1.68 easyconfigs for 2018b then there are 1.67 in the develop branch.
And actually only 1.68 builds with Python 2/3

akesandgren: And FSL should probably use a Python enabled Boost.

boegel: @akesandgren The CGAL easyconfig using the `foss/2018b` toolchain and `Boost` `1.67.0` was merged into `develop` a week ago via PR #6617.

Although it's technically still possible to bump the `Boost` dependency to `1.68.0` (since it's still only in `develop`, not in an EasyBuild release yet), I'm not sure it's worth the trouble in this case...
For one, it would impact anyone who has already installed the `OpenFOAM` easyconfig contributed in #6617 (which happens to include HPC-UGent).

Bumping versions in existing easyconfigs should only be done if there's a very good reason for it imho (even if they're only in `develop`), otherwise it's just not worth the pain...
Is there any technical reason to bump to Boost 1.68.0 here, or just a matter of preference?

Also, what if Boost 1.69.0 comes out tomorrow?
We can't keep bumping dependency versions because new stuff comes out, otherwise we wouldn't be doing anything else.

I must admit that the fact that we have (several) easyconfig for `Boost` 1.68.0 with a `2018b` toolchain is confusing, but there was nothing that prevent a PR contributing those easyconfigs from being merged, since they were not used as dependencies for anything...

The dependency variant check was added to make sure that all easyconfigs that use the same toolchain generation (`2018b` in this case) are compatible with each other (w.r.t. their dependencies), not to highlight that we're using slightly outdated software versions for common dependencies, that's very hard to avoid if you want to keep things compatible...

Note that this only applies to runtime dependencies: if `Boost` can be made a build-only dependency it's less of an issue, although you could still run into trouble then when mixing apps that were built with different Boost libraries (it depends on how they integrate with each other, if it all).

akesandgren: There is no reason to use 1.68 for FSL as far as I know, I just saw that there where multiple 1.68's for 2018b toolchains and only one config in develop that used Boost for 2018b toolchain.

But, since previous FSLs was made with a Python enabled Boost, we (not me :-) should check if it is really needed or not before merging this.

So go with 1.67, but verify if FSL needs a Python enabled Boost or not.

crubb: According to the [FSL documentation](https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FslInstallation/SourceCode/FslView#BOOST):

> **BOOST**
>
> We're only using some of the template classes from boost so we don't need anything but header files, therefore, you don't need to compile boost.

boegel: In that case, you should be able to move `Boost` to `builddependencies`, and then there's no problem at all w.r.t. potential conflicts with other easyconfigs.

You could even use `Boost` 1.68.0 without much risk of problems, I think... But I would error on the side of caution and stick with 1.67.0 anyway. ;)

akesandgren: And you do not need a Python enabled Boost if if only needs that little.

crubb: Done.

boegel:
Test report by @boegel
**SUCCESS**
node3113.skitty.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/9220f1d6e00e27cbc13bae8a8328661e for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2412.golett.os - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/e010a69f0cb3d594e495ba477b0a78e1 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2016.delcatty.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/957d7c878b808104c42a55218ce73654 for a full test report.

boegel:
Going in, thanks @crubb!

mstud:
A bit late, I know, but I have an issue with this. If "Tk" is not installed as an OS package, the "wish" executable is not found, because in the EasyBuild Tk module, it is named "wish8.6". FSL, by default, is trying to use "wish" without the version suffix though, as specified in etc/fslconf/fsl.sh:
`FSLWISH=wish`
I propose either to include a patch for the fsl.{csh,sh}, or to add a symlink (wish -> wish8.6) in the Tk package or so.

PR #6767, opened 2018-08-27
Fix FreeFem PETSc install (REVIEW)

hajgato:
Needs https://github.com/easybuilders/easybuild-easyblocks/pull/1488

### DIFF
--- a/easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
+++ b/easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
@@ -1,4 +1,4 @@
-easyblock = 'ConfigureMake'
+easyblock = 'EB_FreeFem'
 
 name = 'FreeFem++'
 version = '3.58'
@@ -13,9 +13,18 @@ toolchainopts = {'usempi': True, 'pic': True}
 
 source_urls = ['http://www.freefem.org/ff++/ftp/']
 sources = [SOURCELOWER_TAR_GZ]
-checksums = ['e205c0a89fe8edc876d97b8a6ba43825b83c7f8dbdd49fdf972fa1f089b62743']
+patches = ['%(name)s-%(version)s_ebvars.patch']
+checksums = [
+    'e205c0a89fe8edc876d97b8a6ba43825b83c7f8dbdd49fdf972fa1f089b62743',  # freefem++-3.58.tar.gz
+    'e1fe8b26ab417e059653538bcd76214ca0ff4fc5486af11014413335de1ba58d',  # FreeFem++-3.58_ebvars.patch
+]
 
-builddependencies = [('Autotools', '20170619')]
+builddependencies = [
+    ('Autotools', '20170619'),
+    ('flex', '2.5.39'),
+    ('Bison', '3.0.4'),
+    ('CMake', '3.10.1'),
+]
 
 dependencies = [
     ('HDF5', '1.10.1'),
@@ -23,15 +32,15 @@ dependencies = [
     ('freeglut', '3.0.0'),
 ]
 
-preconfigopts = "autoreconf -i &&"
-
-configopts = '--with-hdf5="$EBROOTHDF5"/bin '
+configopts = '--with-hdf5="$EBROOTHDF5"/bin/h5pcc '
 configopts += '--with-gsl-include="$EBROOTGSL"/include --with-gsl-ldflags="$EBROOTGSL"/lib '
-configopts += '--with-lapack="$LIBLAPACK" '
+configopts += '--enable-download --disable-pastix '
 
 # dependencies are heavily patched by FreeFem++, we therefore let it download and install them itself
 prebuildopts = 'download/getall -a -o ScaLAPACK,ARPACK,freeYams,Gmm++,Hips,Ipopt,METIS,ParMETIS,MMG3D,mshmet,MUMPS,'
 prebuildopts += 'NLopt,pARMS,PaStiX,Scotch,SuiteSparse,SuperLU_DIST,SuperLU,TetGen,PETSc,SLEPc,hpddm &&'
+
+runtest = "check"
 
 parallel = 1
 
--- a/easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
+++ b/easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
@@ -1,5 +1,3 @@
-easyblock = 'EB_FreeFem_plus__plus_'
-
 name = 'FreeFem++'
 version = '3.58'
 versionsuffix = '-downloaded-deps'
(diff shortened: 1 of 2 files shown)


hajgato:
Test fails on `magnycours`

boegel:
@hajgato It fails without the patch on AMD Magny Cours, or with the patch? ;)

hajgato:
@boegel What do you mean? In the previous version, the test was not performed. And now, the test passed on ivybridge, haswell, broadwell, and skylake (with the patch). I still waiting the new magnycours results, as a system library was missing. (libibverbs-devel). So maybe the test failed because of this.

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
```diff
@@ -1,4 +1,4 @@
-easyblock = 'ConfigureMake'
+easyblock = 'EB_FreeFem'
```
boegel: This is not needed if you fix the class name for the easyblock in https://github.com/easybuilders/easybuild-easyblocks/pull/1488

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
```diff
@@ -13,26 +13,35 @@ toolchainopts = {'usempi': True, 'pic': True}
...
 dependencies = [
     ('HDF5', '1.10.1'),
     ('GSL', '2.4'),
     ('freeglut', '3.0.0'),
 ]
 
-preconfigopts = "autoreconf -i &&"
-
-configopts = '--with-hdf5="$EBROOTHDF5"/bin '
+configopts = '--with-hdf5="$EBROOTHDF5"/bin/h5pcc '
```
boegel: No need for the double qoutes around `$EBROOTHDF5`: `--with-hdf5=$EBROOTHDF5/bin/h5pcc` .

Also, why is this not handled by the `FreeFem++` easyblock in https://github.com/easybuilders/easybuild-easyblocks/pull/1488?

If there's an easyblock, the easyconfig should be as minimal as possible...

Same for rest of `configopts` & also `prebuildopts` below...

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58_ebvars.patch
```diff
@@ -0,0 +1,72 @@
+# Pick up EB provided ScaLAPACK and FFTW in PETSc install 
+# August 23rd 2018 by B. Hajgato (Free University Brussels - VUB)
+--- freefem++-3.58/download/ff-petsc/Makefile.orig	2018-01-03 07:55:10.000000000 +0100
++++ freefem++-3.58/download/ff-petsc/Makefile	2018-08-25 13:21:53.409212134 +0200
+@@ -51,7 +51,7 @@
+ ifeq ($(MPICC),$(W_MPICC))
+   WITH_MPI_DIR=
+ else
+-  WITH_MPI_DIR="--with-mpi-dir='$(MPI_DIR)'"
++  WITH_MPI_DIR=--with-mpi-dir='$(MPI_DIR)'
```
boegel: @hajgato Why this change?

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58_ebvars.patch
```diff
@@ -0,0 +1,72 @@
...
+-	--with-blas-lapack-lib='$(LAPACKLIBS) $(BLASLIBS)' \
+-	--download-scalapack --download-metis --download-ptscotch \
++	--with-blas-lapack-lib="[${BLAS_LAPACK_LIB_DIR}/${BLAS_LAPACK_STATIC_LIBS}]" \
++	--with-scalapack=1 \
++	--with-scalapack-include=${SCALAPACK_INC_DIR} \
++	--with-scalapack-lib="[${SCALAPACK_LIB_DIR}/${SCALAPACK_STATIC_LIBS}]" \
++	--download-metis --download-ptscotch \
+ 	--download-mumps --download-ml --download-hypre --download-parmetis \
+-	--download-superlu --download-pastix --download-suitesparse --download-fftw \
++	--download-superlu --download-suitesparse \
```
boegel: Why drop `--download-pastix`?

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
```diff
@@ -1,4 +1,4 @@
-easyblock = 'ConfigureMake'
+easyblock = 'EB_FreeFem_plus__plus_'
```
boegel: @hajgato This should not be needed, will be auto-derived based on software name...

hajgato: Done

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
hajgato: @boegel Do you want me to put the `runtest`, `sanity_check_path`, and `modextrapaths` as well in the easyblock?

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58_ebvars.patch
hajgato: Beacuse without it the build failed. I did not look into the details. But does it really matters?

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
boegel: Yes, that makes most sense to me...

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58_ebvars.patch
boegel: It's weird that it's needed, I was just wondering why that change was made. I guess it doesn't matter too much as long as it fixes a problem. ;)

hajgato: I think it is the nature of `PETSc` build, it is a `Python` build system. If I run the `PETSc` installation script by hand (outside `EasyBuild`), it works with the double-quotation marks, using `EasyBuild` I had to remove it. (`PETSc` script was complaining something like cannot configure with `--with-mpi-dir` value.) So probably because it is a `Python` in a `Python`

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58_ebvars.patch
hajgato: `PaStiX` does not work parallel, so it is disabled in the configuration (see now the easyblock). Here, it also breaks the `PETSc` parallel build.

easybuild/easyconfigs/f/FreeFem++/FreeFem++-3.58-foss-2017b-downloaded-deps.eb
hajgato: OK.

hajgato:
3 out of 291 tests were failed on `Magny-cours` architecture (AMD Opteron(tm) Processor 6134)). All tests were passed on Intel `Ivybrige (v2)`, `Haswell (v3)`, `Broadwell (v4)` and `Skylake (gold)` architectures. Therefore, I removed the test from the easyblock, so it could be easier put back to the easyconfig. Since `Magny-Cours` is from 2010, I did not spend any time to figure out the reason for the failed tests. All the 3 failed test had the error: `mshptg bug 2` and then points are too close, so I guess it should be some numeric instability.

boegel:
I'm closing this in favor of more recent FreeFEM versions, see #9902 for example.

The `foss/2017b` toolchain is pretty ancient by now, and the easyblock PR that this requires has been superseded by another (see https://github.com/easybuilders/easybuild-easyblocks/pull/1969) which may not work with FreeFEM 3.x.

Not worth the effort to pursue this imho, sorry @hajgato ...

PR #6799, opened 2018-08-31
{devel}[system] Eclipse v4.8 (Photon)

SimonPinches:


easybuild/easyconfigs/e/Eclipse/Eclipse-4.8-parallel.eb
```diff
@@ -0,0 +1,31 @@
...
+versionsuffix = "-parallel-photon-R"  # For scientific computing
+
+homepage = 'http://www.eclipse.org'
+description = """Integrated Development Environment"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+source_urls = [
+    'http://www.eclipse.org/downloads/download.php?'
+    'file=/technology/epp/downloads/release/photon/R/'
```
boegel: @SimonPinches This is short enough to fit in < 120 chars on a single ine?

easybuild/easyconfigs/e/Eclipse/Eclipse-4.8-parallel.eb
```diff
@@ -0,0 +1,31 @@
...
+]
+sources = [{
+    'filename': '%(name)s-%(version)s%(versionsuffix)s.tar.gz',
+    'download_filename': 'eclipse%(versionsuffix)s-linux-gtk-x86_64.tar.gz'
```
boegel: Wait, they don't provide a properly versioned tarball!?

easybuild/easyconfigs/e/Eclipse/Eclipse-4.8-parallel.eb
SimonPinches: Okay - I'll push a fix for this.

easybuild/easyconfigs/e/Eclipse/Eclipse-4.8-parallel.eb
SimonPinches: Doesn't seem so :-(

boegel:
It looks like the "source" file is no longer available for download, so I'll close this, there's little point in merging this...

PR #6807, opened 2018-09-02
Add libuuid (from util-linux 2.32.1) and update fontconfig-2.13.0 to depend on it

crubb:
In ran into some problems on my Ubuntu Bionic release after compiling `LibUUID`. As soon, as that module was loaded, I ran into errors like to following one:

```(...)/LibUUID/1.0.3-GCCcore-7.3.0/lib/libuuid.so.1: version `UUID_1.0' not found (required by /usr/lib/rstudio-server/bin/rserver)```

Other tools, such as `wget`, were affected as well. I mitigated this problem by compiling `LibUUID`  only from `util-linux` 2.32.1 and depending on that. I got the idea from Debian providing a [libuuid1 package](https://packages.debian.org/stretch/libuuid1), but I am not sure this is the way to go.

The original LibUUID package on sf.net was last touched [~4 years ago](https://sourceforge.net/projects/libuuid/files/).

### DIFF
--- a/easybuild/easyconfigs/f/fontconfig/fontconfig-2.13.0-GCCcore-7.3.0.eb
+++ b/easybuild/easyconfigs/f/fontconfig/fontconfig-2.13.0-GCCcore-7.3.0.eb
@@ -26,7 +26,7 @@ builddependencies = [
 dependencies = [
     ('expat', '2.2.5'),
     ('freetype', '2.9.1'),
-    ('LibUUID', '1.0.3'),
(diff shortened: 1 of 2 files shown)


boegel:
@crubb We already have easyconfig files for `util-linux`, why not simply include that as a dependency rather than `LibUUID`?

boegel:
#6963 is a better way to fix this issue, since it leverages the existing `util-linux` easyconfigs, so I'm closing this...

PR #6828, opened 2018-09-10
Support for Inspector-2018_update3 suitable for intel-2018b

mrolli:


vanzod:
Going in, thanks @mrolli!

PR #6841, opened 2018-09-13
work around known Intel license server bug in icc/ifort 2018.3.222

ocaisa:
Requires ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1509~~


boegel:
@ocaisa The bug is actually in the license server software, so it doesn't apply to a particular `icc` or `ifort` version, as far as I understand...

boegel:
This is not the right place to deal with this, so closing...

PR #6734, opened 2018-08-20
{bio}[foss/2018a] BCFtools v1.9

jdemeul:
(created using `eb --new-pr`)

PR #6842, opened 2018-09-13
{bio}[foss/2018a] BCFtools 1.9

bedroge:


bedroge:
I guess this means that I have to change the version of HTSlib to 1.8?

boegel:
Strictly speaking, yes, but does it make sense to combine `HTSlib` 1.8 with `BCFtools` 1.9?

@verdurin Thoughts?

jdemeul:
See also #6734. 

I changed it to 1.8, but also think it makes more sense to keep the versions of BCFtools, SAMtools and HTSlib synced.

bedroge:
I'll close this one, since it's the same as #6734.

boegel:
:+1:, thanks for following up @bedroge!

PR #6868, opened 2018-09-19
patch libxc 3.x and 4.0.x to compile on powerpc

micaeljtoliveira:
Some constants in Libxc are defined with the L suffix to mark them as long doubles, but this causes problems when using those values to initialize other variables.

This has been fixed upstream starting from version 4.0.5.


easybuild/easyconfigs/l/libxc/libxc_3_longdouble.patch
```diff
@@ -0,0 +1,29 @@
+# Patch to remove long double suffix from some constants so the library can compile in POWER architectures.
+# This does not change the results obtained with the library, as long doubles are not used anywhere else.
```
boegel: @micaeljtoliveira Do you have a link to the upstream fix you're backporting here? Same for the other patch (you'll need to fix the checksums for the patches too, `eb --inject-checksums` is your friend here)

easybuild/easyconfigs/l/libxc/libxc-4.0.1-intel-2017b.eb
```diff
@@ -14,7 +14,11 @@ toolchainopts = {'lowopt': True, 'optarch': False}
 
 source_urls = ['http://www.tddft.org/programs/octopus/down.php?file=libxc/%(version)s']
 sources = [SOURCE_TAR_GZ]
-checksums = ['836cb2f529bb9d1979a347d4d5460bf9a18c64f39b8127e88f3004471a72ab30']
+patches = ['libxc_4.0_longdouble.patch']
```
boegel: @micaeljtoliveira Please rename to match with the 3.0 patch

easybuild/easyconfigs/l/libxc/libxc-3.0.0-GCC-5.4.0-2.26.eb
```diff
@@ -13,6 +13,11 @@ toolchainopts = {'lowopt': True}
 
 sources = [SOURCE_TAR_GZ]
 source_urls = ['http://www.tddft.org/programs/octopus/down.php?file=libxc/']
+patches = ['libxc_3_longdouble.patch']
```
boegel: @micaeljtoliveira Maybe rename the patch file to `libxc-3.0_no-longdouble.patch`, and use `%(version_major_minor)s` here to avoid hardcoding the version (you can then use the same construct in the libxc 4.0 easyconfigs):

```python
patches = ['libxc-%(version_major_minor)s_no-longdouble.patch']

easybuild/easyconfigs/l/libxc/libxc_3_longdouble.patch
micaeljtoliveira: The [upstream fix](https://gitlab.com/libxc/libxc/commit/10889c4c32db6c735715782f66895a24b626bfee) to the 3.x.x. and 4.0.x releases is not the same as the one I implemented here, as that was a simple workaround that did not correct the actual problem. This problem with the long doubles reappeared later and was then fixed correctly [here](https://gitlab.com/libxc/libxc/commit/de20fdea12819f3ff90aad71a0e67475c29cce51).

Putting these links in the patch without this longer explanation will probably be more confusing then anything else.

easybuild/easyconfigs/l/libxc/libxc-4.0.1-intel-2017b.eb
micaeljtoliveira: One patch should be applied to all the 3.x.x releases, while the other should be applied only to the 4.0.x releases. What should be the correct naming convention here?

easybuild/easyconfigs/l/libxc/libxc-3.0.0-GCC-5.4.0-2.26.eb
micaeljtoliveira: See my question above.

easybuild/easyconfigs/l/libxc/libxc_3_longdouble.patch
boegel: @micaeljtoliveira I think referring to https://gitlab.com/libxc/libxc/commit/de20fdea12819f3ff90aad71a0e67475c29cce51 is good enough, it confirms that this same change was made upstream at some point, and provides a handle to figure out more...

easybuild/easyconfigs/l/libxc/libxc-4.0.1-intel-2017b.eb
boegel: @micaeljtoliveira We don't have any easyconfigs for `libxc` 3.x where `x > 0`, so using `3.0` and `4.0` in the patch filename makes sense?

micaeljtoliveira: @boegel You're absolutely right. I forgot there is no 3.1.x series (and probably there is never going to be one).

boegel:
Test report by @boegel
**SUCCESS**
node2419.golett.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/dd7e8651eba03d9c6d68e2994a859039 for a full test report.

boegel:
Going in, thanks @micaeljtoliveira!

PR #6917, opened 2018-09-26
Add support for LAMMPS using CMakeMake

ocaisa:
(created using `eb --new-pr`)


ocaisa:
Alternate approach to https://github.com/easybuilders/easybuild-easyblocks/pull/594 using `CMakeMake` since this build system is now available in LAMMPS

ocaisa:
We currently have 2 CUDA versions in `2018b`, the error in the test is not aware of the fact that this is already the case (i.e, I'm not the one causing it)

ocaisa:
There is room for an easyblock here so you can seamlessly handle the currently excluded build options and other specifics but at least this is an initial version.

ocaisa:
I don't have access to the toolchain used here (I successfully built the original easyconfig in the JSC custom env with slightly different deps), if someone else could upload a test report that would be great.

ocaisa:
I've commented out the CUDA support (although it does work with CUDA 10) because you need to provide specifics about the particular hardware you have

boegel:
@ocaisa Test reports coming in (just for proof of concept sake).

This is way too complex for an easyconfig-only approach, but using the CMake build procedure should make a custom easyblock for `LAMMPS` as bit easier indeed...

boegel:
Test report by @boegel
**SUCCESS**
node2404.golett.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/8a56858996bddc14c0fd60be3608c76f for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 13 out of 13 (3 easyconfigs in this PR)
node3124.skitty.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/f3eb97124e6f93495de60ae4076fb0d7 for a full test report.

vanzod:
Test report by @vanzod
**SUCCESS**
Build succeeded for 7 out of 7 (3 easyconfigs in this PR)
See https://gist.github.com/4c02180b8f89fc08687cd1db68b377b9 for a full test report.

ocaisa:
Ok, this has basically everything you need to create the easyblock. The included packages need to cross-referenced with https://lammps.sandia.gov/doc/Build_extras.html#kokkos

easybuild/easyconfigs/l/LAMMPS/LAMMPS-patch_18Sep2018-intel-2018b.eb
```diff
@@ -0,0 +1,114 @@
...
+# CMakeLists.txt is in the cmake subdir
+srcdir = short_sources + '/cmake'
+
+# Make the symlink to solve our long path problem
+preconfigopts = "ln -sf %(builddir)s/%(namelower)s-%(version)s " + "%s &&" % short_sources 
+
+# Initialise configopts so we can comment out at will below
+configopts = ""
+
+parallelisation_packages = ['BUILD_MPI', 'BUILD_OMP', 'PKG_KOKKOS', 'PKG_OPT', 'PKG_USER-OMP', 'PKG_USER-INTEL', 'PKG_GPU']
```
ocaisa: Looks like you can't have PKG_GPU and Kokkos at the same time

easybuild/easyconfigs/l/LAMMPS/LAMMPS-patch_18Sep2018-intel-2018b.eb
```diff
@@ -0,0 +1,114 @@
...
+specific_opts = ' '.join("-D%s" % item for item in kspace_opts + gpu_opts)
+other_opts = ' '.join("-D%s=yes" % item for item in optional_deps)
+
+configopts += packages_opts + ' ' + other_opts + ' ' + specific_opts 
+
+# Kokkos options:
+# Need to use the Kokkos wrapper if we are using CUDA
+configopts += ' -DCMAKE_CXX_COMPILER="%(builddir)s/%(namelower)s-%(version)s/lib/kokkos/bin/nvcc_wrapper" '
+# and tell the wrapper which c++ compiler to use
+configopts += ' -DCMAKE_CXX_FLAGS="-ccbin $CXX $CXXFLAGS" '
```
ocaisa: This seems to cause problems with some packages (which seem to be using the `CXX_FLAGS` with the C compiler)

ocaisa:
@vanzod Can you take this for another spin, I've updated the version and introduced logic. I know @boegel  won't like this but the hardware related stuff will make an easyblock time consuming, it's on my to-do but this at least exposes all the pieces of the puzzle.

easybuild/easyconfigs/l/LAMMPS/LAMMPS-24Oct2018-intel-2018b.eb
```diff
@@ -0,0 +1,143 @@
...
+    ('Python', '2.7.15'),
+    ('libpng', '1.6.34'),
+    ('libjpeg-turbo', '2.0.0'),
+    ('netCDF', '4.6.1'),
+    ('GSL', '2.5'),
+    ('zlib', '1.2.11'),
+    ('gzip', '1.9'),
+    ('HDF5', '1.10.2'),
+    ('tbb', '2018_U5'),
+    #('CUDA', '10.0.130', '', True),
```
ocaisa: It would seem that CUDA+kokkos support only works with GCC, the wrapper causes problems in other modules using (CUDA supported) Intel compilers.

ocaisa:
Personally, I've successfully built this with `cuda = True, kokkos = False`, `cuda = False, kokkos = False` and `cuda = False, kokkos = True`. The only successful build I've had with kokkos+cuda has been by cheating and using the underlying GCCcore as the C++ compiler (by setting a hardcoded `CMAKE_CXX_FLAGS`)

JackPerdue:
"@boegel won't like this but" 

I'm sure he absolutely abhors it but I absolutely love it.   "We don't need no stinkin' easyblocks".   

I never knew that easyconfigs had if: else: clauses.... that's cool.   I'm going to play with this.

JackPerdue:
[LAMMPS-20180316-intel-2017b-Python-3.6.3-CUDA-9.1.85-mpi.eb.txt](https://github.com/easybuilders/easybuild-easyconfigs/files/3004575/LAMMPS-20180316-intel-2017b-Python-3.6.3-CUDA-9.1.85-mpi.eb.txt)

(Our latest build.... local users need VORO++ also)

JackPerdue:
OK... bad example... it uses the lammps.py easyblock PR.  But still, anything that can be pushed into the easyconfig should be.   Don't make packagers have to jump through hoops to update easyblocks just to build the latest version.

fgeorgatos:
@ocaisa , @JackPerdue : I have also found unnatural that you need to *choose* about what goes into an easyconfig and what goes into an easyblock; 

perhaps, I'd rather have a long yaml file configuration format on which python code can be embedded: 
* in any case `easyconfigs` or `easyblocks` are _not_ stand-along code anyhow, so it would be ideal to bundle the logic up, so that git commits preserve related information into one transaction!

Darkless012:
Could this be closed/rejected due to LAMMPS-foss/intell-2019b being merged?
https://github.com/easybuilders/easybuild-easyconfigs/pull/9883
https://github.com/easybuilders/easybuild-easyconfigs/pull/9884

boegel:
closing due to https://github.com/easybuilders/easybuild-easyblocks/pull/1964

PR #6935, opened 2018-10-01
Update the GROMACS-2016.3 nvml lib search patch to be more generic.

akesandgren:
And match the GROMACS 2018 patch.


boegel:
follow-up in #6936

PR #6945, opened 2018-10-02
Add patch to use make instead of gmake in ParMGridGen.

akesandgren:
And update checksums.


boegel:
@akesandgren I don't think this requires a patch at all, just adding `make=make` to `buildopts` should be sufficient?

akesandgren:
Ah, perhaps. Didn't really check when i made the patch and it's been sitting there for some time.

I'll test...

akesandgren:
Yes, much better. Closing this. Better solution in https://github.com/easybuilders/easybuild-easyconfigs/pull/6952

PR #6962, opened 2018-10-05
solve 'version UUID_1.0 not found' problem in LibUUID easyconfigs

ocaisa:
(created using `eb --new-pr`)


boegel:
@ocaisa As discussed on the mailing list, I don't think this is the right way to fix the problem...

We should replace the `LibUUID` dependencies with `util-linux` instead.

I don't mind fixing these `LibUUID` easyconfigs in parallel though...

easybuild/easyconfigs/l/LibUUID/LibUUID-1.0.3-GCCcore-6.4.0.eb
```diff
@@ -11,12 +11,18 @@ toolchain = {'name': 'GCCcore', 'version': '6.4.0'}
 
 source_urls = [SOURCEFORGE_SOURCE]
 sources = [SOURCELOWER_TAR_GZ]
-checksums = ['46af3275291091009ad7f1b899de3d0cea0252737550e7919d17237997db5644']
+patches = [('libuuid.sym', '%(builddir)s')]
```
boegel: @ocaisa You'll need to include this in the PR.

But maybe it's better to download this on the fly, by listing it in `sources` instead?

ocaisa: Can you give me the correct syntax, it's a hunt to figure out how to use `extract_cmd`. Is it
```
source_urls = [
    SOURCEFORGE_SOURCE,
    'https://raw.githubusercontent.com/karelzak/util-linux/master/libuuid/src',
]
sources = [
    SOURCELOWER_TAR_GZ,
    {'filename': 'libuuid.sym', 'extract_cmd': 'mv libuuid.sym %(builddir)s')]
checksums = [
    '46af3275291091009ad7f1b899de3d0cea0252737550e7919d17237997db5644',  # libuuid-1.0.3.tar.gz
    '5932866797560bf5822c3340b64dff514afb6cef23a824f96991f8e9e9d29028',  # libuuid.sym
]
```

boegel: `cp %s %(builddir)s` as extract command should do it

the `%s` gets replaced with the full path download location of `libuuid.sym`

easybuild/easyconfigs/l/LibUUID/LibUUID-1.0.3-GCCcore-6.4.0.eb
```diff
@@ -9,14 +9,25 @@ description = """Portable uuid C library"""
 
 toolchain = {'name': 'GCCcore', 'version': '6.4.0'}
 
-source_urls = [SOURCEFORGE_SOURCE]
-sources = [SOURCELOWER_TAR_GZ]
-checksums = ['46af3275291091009ad7f1b899de3d0cea0252737550e7919d17237997db5644']
+source_urls = [
+    SOURCEFORGE_SOURCE,
+    'https://raw.githubusercontent.com/karelzak/util-linux/master/libuuid/src',
```
boegel: @ocaisa Can we avoid downloading from `master`, and use a particular version/tag instead, since this is effectively a moving target...

boegel:
Test report by @boegel
**SUCCESS**
node2007.delcatty.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/2a1c4f349d73a0019d8578fc206f3974 for a full test report.

boegel:
Going in, thanks @ocaisa!

PR #6983, opened 2018-10-09
removed dead source of bzip2 tarball from bzip.org

jose-d:
bzip.org not providing tarballs as expected so removed from source_urls. (related to #6728 )

### DIFF
--- a/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
+++ b/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
@@ -11,7 +11,6 @@ toolchainopts = {'pic': True}
 
 source_urls = [
     'https://fossies.org/linux/misc/',
-    'http://www.bzip.org/%(version)s/',
 ]
 sources = [SOURCE_TAR_GZ]
 checksums = ['a2848f34fcd5d6cf47def00461fcb528a0484d8edef8208d6d2e2909dc61d9cd']
--- a/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
+++ b/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
@@ -9,9 +9,7 @@ description = """bzip2 is a freely available, patent free, high-quality data com
 toolchain = {'name': 'CrayGNU', 'version': '2015.06'}
 toolchainopts = {'pic': True}
 
-source_urls = [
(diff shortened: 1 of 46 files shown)


boegel:
@jose-d We left the original download URL in place in #6728 since it wasn't clear back then what would happen with `bzip.org` (it seems like it has been revamped now though...).

Is there any reason the `bzip.org` download URL is causing problems for you?

easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
```diff
@@ -11,7 +11,6 @@ toolchainopts = {'pic': True}
 
 source_urls = [
     'https://fossies.org/linux/misc/',
-    'http://www.bzip.org/%(version)s/',
 ]
```
boegel: @jose-d Since this is now a single entry, it makes sense to reformat this to:

```python
source_urls = ['https://fossies.org/linux/misc/']
```

Are you up for taking care of that?

jose-d:
@boegel:
> Is there any reason the `bzip.org` download URL is causing problems for you?

build(s) were failing with:
`== 2018-10-09 13:05:23,195 easyblock.py:2851 WARNING build failed (first 300 chars): Checksum verification for /home/users/jose/.local/easybuild/sources/b/bzip2/bzip2-1.0.6.tar.gz using a2848f34fcd5d6cf47def00461fcb528a0484d8edef8208d6d2e2909dc61d9cd failed.
`
as it was not tar.gz but some html file.

Or is there some consensus to not remove dead http(s) sources and just temporarily black-list them? (how?)

easybuild/easyconfigs/b/bzip2/bzip2-1.0.6-CrayGNU-2015.06.eb
jose-d: sure.

jose-d:
@boegel 
i did the reformat as requested.

> Removing the dead URL isn't going to fix your problem
> you're basically never supposed to hit the old URL 
it fixed actually. I had no clue that the order matter so I didn't any deeper testing. I'll do so.

boegel:
Test report by @boegel
**SUCCESS**
node2602.swalot.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/7d659859b801907e3553e7cca9d25992 for a full test report.

boegel:
Going in, thanks @jose-d!

PR #6989, opened 2018-10-10
wheel 0.31.0 for fosscuda 2018a and Python 2.7.14

FokkeDijkstra:
wheel 0.31.0 for fosscuda 2018a and Python 2.7.14


boegel:
Going in, thanks @FokkeDijkstra!

PR #6940, opened 2018-10-02
add missing extensions for TensorFlow 1.10.1 (REVIEW)

boegel:
required ~~because of the changes to the `TensorFlow` easyblock, see https://github.com/easybuilders/easybuild-easyblocks/pull/1537~~ to fix problems reported in https://github.com/easybuilders/easybuild-easyblocks/issues/1536

cc @akesandgren, @agijsberts

PR #6993, opened 2018-10-10
TensorFlow 1.10.1 for the foss(cuda) 2018a toolchain

FokkeDijkstra:
TensorFlow 1.10.1 for the foss(cuda) 2018a toolchain, including Python 2.7.14 and 3.6.4 versions.

### DIFF
--- a/easybuild/easyconfigs/t/TensorFlow/TensorFlow-1.10.1-foss-2018a-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/t/TensorFlow/TensorFlow-1.10.1-foss-2018a-Python-2.7.14.eb
@@ -1,0 +1,31 @@
+name = 'TensorFlow'
+version = '1.10.1'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://www.tensorflow.org/'
+description = "An open-source software library for Machine Intelligence"
+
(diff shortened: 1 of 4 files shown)


boegel:
@FokkeDijkstra We've made some changes to the existing `TensorFlow` 1.10.1 easyconfigs because of a reported bug, see https://github.com/easybuilders/easybuild-easyblocks/issues/1536, https://github.com/easybuilders/easybuild-easyblocks/pull/1537 and https://github.com/easybuilders/easybuild-easyconfigs/pull/6940 .

If more `TensorFlow` easyconfigs are added for recent versions (1.10.x) or newer, they should follow the same structure...

FokkeDijkstra:
@boegel then I suggest to reject this PR. Once the proposed fixes have stabilized I can create the foss*-2018a versions with these fixes. 
I see that I also need to fix a wheel dependency. I probably should have sticked to wheel 0.30 and add the missing ones. Now there's a mixture of 0.30 and 0.31 for foss*-2018a in the repository.

boegel:
@FokkeDijkstra There may be a mixture of `wheel` versions, but that's not a big issue since those are just build dependencies for `TensorFlow`.

We disallow multiple versions of runtime dependencies (that's checked by the tests), to limit the chances of introducing version conflicts among easyconfigs of a single toolchain generation.

Closing this PR, as suggested...

Issue #7007, opened 2018-10-12
Build of Perl-5.28.0-GCCcore-7.3.0.eb fails in singularityhub/centos:latest


jose-d:
It seems there is some dependency missing in 

[Perl-5.28.0-GCCcore-7.3.0.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/p/Perl/Perl-5.28.0-GCCcore-7.3.0.eb)

Build in container with singularityhub/centos:latest with additional packages 

`yum-utils epel-release ohpc-base-compute lmod-ohpc EasyBuild-ohpc GitPython pysvn graphviz vim zlib-devel libyaml-devel openssl-devel gdbm-devel readline-devel ncurses-devel libffi-devel curl git redis libxml2-devel libxslt-devel libcurl-devel libicu-devel python libuuid-devel
`
and

```
yum -y groupinstall "Development tools"
yum-builddep -y perl
```

is failing in build of **perl-Net-SSLeay** installation step:

```
== 2018-10-12 13:03:28,777 run.py:192 INFO running cmd: tar xzf /sw/local/el7/x86_64/sources/p/Perl/extensions/Net-SSLeay-1.85.tar.gz 
== 2018-10-12 13:03:28,808 run.py:192 INFO running cmd:  perl Makefile.PL PREFIX=/sw/local/el7/x86_64/software/lang/Perl/5.28.0-GCCcore-7.3.0  
== 2018-10-12 13:03:28,999 run.py:192 INFO running cmd:  make -j 64  
== 2018-10-12 13:03:29,048 build_log.py:158 ERROR EasyBuild crashed with an error (at ?:124 in __init__): cmd " make -j 64 " exited with exit code 2 and output:
make: *** No targets specified and no makefile found.  Stop.

```
It is enough to install build-deps of SSLeay to have it working..

`yum-builddep -y perl-Net-SSLeay
`
I work on problem isolation, however I already suspect that we should include [ExtUtils::MakeMaker](https://metacpan.org/pod/ExtUtils::MakeMaker) somewhere into [Perl-5.28.0-GCCcore-7.3.0.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/p/Perl/Perl-5.28.0-GCCcore-7.3.0.eb).

jose-d:
> `yum-builddep -y perl-Net-SSLeay`
[...]

PR #7008, opened 2018-10-12
add openssl OS deps in Perl 5.28.0 easyconfig for Net::ssleay

jose-d:
fix for #7007

### DIFF
--- a/easybuild/easyconfigs/p/Perl/Perl-5.28.0-GCCcore-7.3.0.eb
+++ b/easybuild/easyconfigs/p/Perl/Perl-5.28.0-GCCcore-7.3.0.eb
@@ -18,10 +18,8 @@ dependencies = [
     ('expat', '2.2.5'),  # for XML::Parser
 ]
 
-osdependencies = [
-    ('openssl'),						# for Net::SSLeay
-    ('openssl-devel', 'libssl-dev', 'libopenssl-devel')		# dtto
-]
+# OpenSSL is required for Net::SSLeay
+osdependencies = ['openssl', ('openssl-devel', 'libssl-dev', 'libopenssl-devel')]
 
 # !! order of extensions is important !!
 # extensions updated on June 26th 2018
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/p/Perl/Perl-5.28.0-GCCcore-7.3.0.eb
```diff
@@ -18,6 +18,11 @@ dependencies = [
     ('expat', '2.2.5'),  # for XML::Parser
 ]
 
+osdependencies = [
+    ('openssl'),						# for Net::SSLeay
+    ('openssl-devel', 'libssl-dev', 'libopenssl-devel')		# dtto
+]
```
boegel: @jose-d Since this is really one thing for one particular extension (and you have a typo in `ditto`), why not go with:

```python
# OpenSSL is required for Net::SSLeay
osdependencies = ['openssl', ('openssl-devel', 'libssl-dev', 'libopenssl-devel')]
```

I suspect `'openssl'` is not the right package name across the different Linux distros though... I guess we'll also have to consider `libssl` and `libopenssl`? cc @vanzod @akesandgren

```python
osdependencies = [('openssl', 'libssl', 'libopenssl'), ('openssl-devel', 'libssl-dev', 'libopenssl-devel')]

jose-d: @boegel 

> (typo + code style)

agree.

> I suspect `'openssl'` is not the right package name across the different Linux distros though... I guess we'll also have to consider `libssl` and `libopenssl`? cc @vanzod @akesandgren

`openssl-libs` (`libssl`|`libopenssl`) is something else, and it is actually RPM dependency of already present `openssl-devel` (and I _assume_ that other distros have same concept with their `libssl-dev`|`libopenssl-devel`) - so here we should be OK.

`Net::ssleay` is actually searching real `openssl` - not the libs.

I checked centos7 and recent ubuntu - on both it is package called `openssl`.

boegel: OK, just a reformat should be fine then...

boegel:
Test report by @boegel
**SUCCESS**
gligar02.gligar.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/023e5b5b9e2bd49f7d11c05ba7fabbe2 for a full test report.

boegel:
Going in, thanks @jose-d!

PR #7028, opened 2018-10-15
osdeps for git +last git eb 2.19.1

jose-d:
+missing osdeps for git documentation build as mentioned in https://github.com/jose-d/easybuild-easyconfigs/commit/50ab4fa71d2654abbb54dc382c56b6e9dd1b6dab#commitcomment-30863442.

### DIFF
--- a/easybuild/easyconfigs/g/git/git-2.12.2-foss-2016b.eb
+++ b/easybuild/easyconfigs/g/git/git-2.12.2-foss-2016b.eb
@@ -22,6 +22,9 @@ dependencies = [
     ('Perl', '5.24.0', '-bare'),
 ]
 
+# asciidoc and xmlto are required for git man/doc build
+osdependencies = ['asciidoc', 'xmlto']
+
 preconfigopts = 'make configure && '
 
 # Work around git build system bug.  If LIBS contains -lpthread, then configure
--- a/easybuild/easyconfigs/g/git/git-2.13.1-foss-2016b.eb
+++ b/easybuild/easyconfigs/g/git/git-2.13.1-foss-2016b.eb
@@ -23,6 +23,9 @@ dependencies = [
     ('Perl', '5.24.0', '-bare'),
 ]
 
+# asciidoc and xmlto are required for git man/doc build
+osdependencies = ['asciidoc', 'xmlto']
+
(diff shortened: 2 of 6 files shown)


boegel:
@jose-d Any idea how hard it would be to include these as 'proper' EasyBuild dependencies?

jose-d:
> @jose-d Any idea how hard it would be to include these as 'proper' EasyBuild dependencies?

Mhmm. Of-course, it would be doable. 

Actually the question is, "do we want EB to include all the small toys"?  Or do we expect the distro will give us the basics? I am most likely too new in this project to answer this properly. 

In this case I somehow miss non-existing "_osbuilddependencies_" - we don't need anything like xmlto nor asciidoc for runtime..

jose-d:
> Travis test report: 7/7 runs failed - see https://travis-ci.org/easybuilders/easybuild-
y. there are no shasums in the older ebs.. I'll fix it.

boegel:
@jose-d The policy is to rely on OS dependencies as little as possible. We do it in some cases though, e.g. OpenSSL (because of security updates that are picked up from the OS) and Infiniband libraries like `libibverbs` which are tightly coupled with the drivers. That's also part of the reason why there's no such thing as `osbuilddependencies` yet.

For everything else, ideally we resolve the dependencies through EasyBuild itself, but sometimes that may not be worth the effort. I seem to remember that `asciidoc` is a bit of a mess for example...

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 13 out of 14 (14 easyconfigs in this PR)
gligar02.gligar.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/2ed735963caf938353d7188dc24172a4 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
gligar02.gligar.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/f53fd6461a0cad7d9a14fab0de5b0502 for a full test report.

PR #7033, opened 2018-10-16
add missing checksums for ipaddress Python extension (+ reformatting)

boegel:
Somehow the missing checksum slipped through in #7022, even though the tests are supposed to catch missing checksums in easyconfigs being touched...

Works as expected with `--check-contrib` though, which uses the same code as the tests...

```
$ eb --check-contrib  easybuild/easyconfigs/p/Python/Python-2.7.14-foss-2017b.eb --ignore-osdeps
== temporary log file in case of crash /Users/kehoste/work/TMP/eb-OW6INh/easybuild-5MuZoZ.log

Running style check on 1 easyconfig(s)...

[PASS] /Volumes/work/easybuild-easyconfigs/easybuild/easyconfigs/p/Python/Python-2.7.14-foss-2017b.eb

>> All style checks PASSed!

Checking for SHA256 checksums in 1 easyconfig(s)...

[FAIL] /Volumes/work/easybuild-easyconfigs/easybuild/easyconfigs/p/Python/Python-2.7.14-foss-2017b.eb
Checksums missing for one or more sources/patches of extension ipaddress in Python-2.7.14-foss-2017b.eb: found 1 sources + 0 patches vs 0 checksums

>> One or more SHA256 checksums checks FAILED!

ERROR: One or more contribution checks FAILED!
```

PR #7022, opened 2018-10-15
add missing extensions to Python 2.7.14 easyconfigs using toolchain of 2017b generation

boegel:
This is mainly done because picking up the most recent `PyNaCl` version (1.3.0) is problematic with this generation of Intel compilers (cfr. #6971)

PR #7034, opened 2018-10-16
also run checks on changed files when target branch for PR is something different than 'develop'

boegel:
fix for failure to catch missing checksum issue in #7022 (which is fixed in #7033)


vanzod:
Going in, thanks @boegel!

PR #7052, opened 2018-10-18
+Python3.5.6, recent toolchain and extensions

jose-d:


boegel:
@jose-d We prefer having only a single Python 3.x version per toolchain generation, to avoid confusion. For `2018b` we have both easyconfigs for both 3.6.6 and 3.7.0 because we ran into problems with 3.7.0 and opted for not removing the easyconfig for 3.7.0. All easyconfigs for Python packages use Python 3.6.6 consistently as a dependency though (3.7.0 is not used as a dependency).

Adding another easyconfig for Python 3.5.6 would only add to the confusion...

Can you explain the need for this? Why not use the existing Python 3.6.6 easyconfig file?

jose-d:
> Can you explain the need for this? Why not use the existing Python 3.6.6 easyconfig file?

@boegel python 3.5 was requested by users - they use (sometimes obscure) cosmologic python libs and they want sometimes test them with different python releases (like the 3.5.X). Vendor of the HPC system provided us with manually-builded-manually-installed-python version on the system, but in long term, I'd like to kick all of such things out, and provide all the software with standard (=easybuilded) way.

Users needs to do their science, and if Python 3.5.* will help them to do so, till there will be maintained version, I have no problem with it.

That's why I took recent tool-chain, (I don't want to have ancient tool-chains on my system just because of python 3.5*) wrote some script to check the latest versions of python extensions (as I don't want solved bugs on my system too) and created this EB.

If this eb is (not beneficial for others|confusing the dependencies|*), no need to merge it.

boegel:
@jose-d I think it's better not the include this in the central repository with this toolchain, since it would to further confusion w.r.t. what the recommended Python version is for `foss/2018b`.

If you're fine with that, please close this pull request.

I hope it's clear that this doesn't mean that we don't value your contributions, please keep them coming... ;)

jose-d:
roger that, closing this one.

> please keep them coming

sure, actually this one was the last one missing and crucial one for my system, so i think I'll slowly take look at the topic we discussed in https://github.com/easybuilders/easybuild-easyconfigs/pull/7028#issuecomment-429905302.

PR #7073, opened 2018-10-25
also install shared libraries for LZO

pescobar:
while installing an application that needs the dynamic library `liblzo2.so.2` I realized that the easyconfig for LZO only provides static libraries because that's the default in the build system provided by LZO

```
lzo-2.10]$ ./configure --help |grep -i shared
  --enable-shared[=PKGS]  build shared libraries [default=no]
```

I have enabled the dynamic build for all the easyconfigs and extended the sanity check

@boegel I guess you are the only one who has all these toolchains installed. Can you trigger a test build please?

### DIFF
--- a/easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
+++ b/easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
@@ -25,6 +25,8 @@ sanity_check_paths = {
     'dirs': ['lib', 'include']
 }
 
+configopts = '--enable-shared'
+
 runtest = 'test'
 
 parallel = 1    # this is a very conservative choice
--- a/easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
+++ b/easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
@@ -15,8 +15,9 @@ version = '2.06'
 homepage = 'http://www.oberhumer.com/opensource/lzo/'
 description = "LZO-2.06: Portable lossless data compression library"
 
-sources = [SOURCELOWER_TAR_GZ]
 source_urls = [homepage + 'download/']
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['ff79e6f836d62d3f86ef6ce893ed65d07e638ef4d3cb952963471b4234d43e73']
 
 toolchain = {'name': 'goolf', 'version': '1.4.10'}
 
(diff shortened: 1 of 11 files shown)


easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
```diff
@@ -21,10 +21,12 @@ source_urls = [homepage + 'download/']
 toolchain = {'name': 'goolf', 'version': '1.4.10'}
 
 sanity_check_paths = {
-    'files': [],
+    'files': ['lib/liblzo2.a', 'lib/liblzo2.so'],
     'dirs': ['lib', 'include']
```
akesandgren: Don't really need lib in the dirs check if we explicitly check for files in lib

easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
```diff
@@ -21,10 +21,12 @@ source_urls = [homepage + 'download/']
...
 }
 
+configopts = '--enable-shared'
+
```
akesandgren: At least personally i think configopts should go before sanity_checks

easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
pescobar: I agree. I only added the new check but I didn't remove the old one. I will do it

boegel:
@pescobar One more missing checksum in `LZO-2.10-intel-2017a.eb`... :)

easybuild/easyconfigs/l/LZO/LZO-2.06-goolf-1.4.10.eb
```diff
@@ -15,18 +15,21 @@ version = '2.06'
...
-}
+configopts = '--enable-shared'
+
+parallel = 1    # this is a very conservative choice
 
 runtest = 'test'
 
-parallel = 1    # this is a very conservative choice
+sanity_check_paths = {
+    'files': ['lib/liblzo2.a', 'lib/liblzo2.so'],
```
boegel: @pescobar Overlooked this earlier, we should use `SHLIB_EXT` rather than hardcode `.so`

Fixed in https://github.com/pescobar/easybuild-easyconfigs/pull/68

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 11 out of 11 (11 easyconfigs in this PR)
node2006.delcatty.os - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/50771fc752a4e28409935aa1c9e42710 for a full test report.

boegel:
Going in, thanks @pescobar!

PR #6944, opened 2018-10-02
{bio}[dummy/dummy] novoalign v3.09.00

jdemeul:
(created using `eb --new-pr`)

Requires manual download from http://www.novocraft.com/support/download/

PR #7210, opened 2018-11-28
{bio}[foss/2018b] novoalign 3.09.01 w/ R and Perl (REVIEW)

hajgato:


hajgato:
@jdemeul and @boegel Opinions about this and #6944 ?

vanzod:
Test report by @vanzod
**FAILED**
See https://gist.github.com/f03581e55b2529e3aaf3cecdb9f24888 for a full test report.

easybuild/easyconfigs/n/novoalign/novoalign-3.09.01_fix_hardcoded.patch
```diff
@@ -0,0 +1,42 @@
+# Fix hard-coded R and Perl paths
+# Novermber 29th 2018 by B. Hajgato (Free University Brussels - VUB)
```
akesandgren: novope2bed.pl also need harcoded path fixed

akesandgren:
I think this is better than #6944 since it uses EB's R/Perl

easybuild/easyconfigs/n/novoalign/novoalign-3.09.01_fix_hardcoded.patch
hajgato: @akesandgren Indeed, I looked it over. Fixed.

jdemeul:
@akesandgren @hajgato if I recall correctly, NovoAlign itself doesn't require R/Perl, but other tools in the NovoCraft package do.
Is the aim of this easyconfig to provide the full NovoCraft toolkit, or only NovoAlign?

hajgato:
@jdemeul I strongly suspect that only `novoalign` is needed (but not sure, so in principle `R` and `Perl` are not needed as a dep.). 
At the other hand, the question can be asked from another point of view: If the install provides extra software, should we make those as well operational?  Even if it is not needed right now, we might needed in the future. All in all I am open for different solutions, but since I did not get a quick answer  (and users are always impatient), I went further with the version I found better.

jdemeul:
That's fine, but perhaps the name of the easyconfig should reflect the additional functionality i.e.: NovoCraft for this one instead of simply NovoAlign?

akesandgren:
Precisely what I was about to suggest. The tar file does contain all of it as far as I can tell.

hajgato:
@jdemeul Actually, the existing `novoalign` has the same functionality than this if you have repo installed Perl and R. But even without R and Perl, the existing version has a functional `novoalign`, `novoalignCS`, `novosort`, `novomethyl`. Not to mention, that the existing ebconfig deliberately copies all executables including R and Perl. So in my eyes, both should be called the same.

easybuild/easyconfigs/n/novoalign/novoalign-3.09.01-foss-2018b-R-3.5.1.eb
```diff
@@ -0,0 +1,58 @@
...
+ and Color Spance NGS platforms.
+ NovoAlignCS:
+ Leading aligner packed with features designed to fully support reads from
+ ABI SOLiD Color Space.
+ NovoSort:
+ Custom designed multi-threaded sort/merge tools for BAM files.
+ NovoMethyl:
+ It can analyse a set of alignments to identify methylated cytosine’s.
+ WARNING!
+ You can only ise the sofware without a license (with some features disabled)
```
akesandgren: Small typo: "ise" -> "use"

easybuild/easyconfigs/n/novoalign/novoalign-3.09.01-foss-2018b-R-3.5.1.eb
```diff
@@ -0,0 +1,58 @@
+easyblock = 'PackedBinary'
+
+name = 'novoalign'
+version = '3.09.01'
+versionsuffix = '-R-%(rver)s'
```
akesandgren: Not sure that the -R suffix is useful here.

easybuild/easyconfigs/n/novoalign/novoalign-3.09.01-foss-2018b-R-3.5.1.eb
hajgato: Done

easybuild/easyconfigs/n/novoalign/novoalign-3.09.01-foss-2018b-R-3.5.1.eb
hajgato: I would rather keep this here, as `novoaligin` is a dependency, and I do not want to change them as well. At the end it does not hurt.

akesandgren:
Test report by @akesandgren
**FAILED**
Build succeeded for 1 out of 3 (1 easyconfigs in this PR)
See https://gist.github.com/0a0f1179d83295a5c14e40a949a46d84 for a full test report.

akesandgren:
Test report by @akesandgren
**SUCCESS**
See https://gist.github.com/4cf06fbfcb1b465b32561c45ae17d88e for a full test report.

akesandgren:
Going in, thanks @hajgato!

PR #7220, opened 2018-11-29
[intel/2018a] [Python/3.6.4] OpenCV-Python v3.4.4.19

SimonPinches:


vanzod:
Test report by @vanzod
**SUCCESS**
See https://gist.github.com/bc87dba4bbefebc88e84efe2f0c8d5d2 for a full test report.

vanzod:
@SimonPinches Is there a specific reason you are installing it from a wheel instead of building it from source as we typically do?

SimonPinches:
@vanzod I installed from a wheel through a combination of ignorance and convenience!  If I find time, I can try and build from scratch but when I started out this way I quickly ran into problems and I needed to to make an installation ASAP for our own internal use.

vanzod:
@boegel Thoughts?

SimonPinches:
I think this PR can be declined / closed since I now agree that it adds nothing that isn't already provided by the `OpenCV` modules.

PR #7159, opened 2018-11-14
update SCOTCH 6.0.5 easyconfigs in-place to 6.0.6 to fix wrong download URL

boegel:
fix for problem reported by @smoors in #7154

see also enhanced `SCOTCH` easyblock in ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1580~~

WIP since needs thorough testing (which is under way); test reports welcome!

PR #7245, opened 2018-12-06
Make SCOTCH 6.0.5 install version 6.0.5 (instead 6.0.4)

hajgato:
Just for reference, this is `SCOTCH` `6.0.5a` as `6.0.5` download link is not publicly accessible. Since all previous `6.0.5` versions installed version `6.0.4`, it should not be a problem.

Just for a record, `6.0.5` can be downloaded from https://gforge.inria.fr/frs/download.php/file/37396/scotch_6.0.5.tar.gz


hajgato:
Already submitted #7159

Issue #7425, opened 2019-01-15
sympy test failure II


jose-d:
as found in #4024 and #7405, sympy tests are _sometimes_ failing.
This issue is to keep the track of the investigation.

So far there is observable correlation between length/depth of the buildpath:

_Long_ path, e.g. `--buildpath=/tmp/fakehome/jose/.local/easybuild/build/` will reproduce the error, and _short_ path, e.g. `--buildpath=/dev/shm/jose/ebtest/` seems to be working fine.

I did just few builds, so this can be also just random noise. - to be confirmed.

test failing seems to be line `4231` of `sympy/printing/pretty/tests/test_pretty.py`:

```
4227 def test_pretty_no_wrap_line():
4228     huge_expr = 0
4229     for i in range(20):
4230         huge_expr += i*sin(i + x)
4231     assert xpretty(huge_expr            ).find('\n') != -1
4232     assert xpretty(huge_expr, wrap_line=False).find('\n') == -1
```

```
________________________________________________________________________________
_____ sympy/printing/pretty/tests/test_pretty.py:test_pretty_no_wrap_line ______
Traceback (most recent call last):
  File "/tmp/fakehome/jose/.local/easybuild/build/sympy/1.3/foss-2018b-Python-3.6.6/sympy-1.3/sympy/printing/pretty/tests/test_pretty.py", line 4231, in test_pretty_no_wrap_line
    assert xpretty(huge_expr            ).find('\n') != -1
AssertionError

 tests finished: 7700 passed, 1 failed, 220 skipped, 351 expected to fail, 
19 expected to fail but passed, in 1953.62 seconds 
DO *NOT* COMMIT!
```

I'm currently digging around..

jose-d:
..tests are looking as not stateless, if `eb` build is interrupted (ctrl-C) in test phase, and then the `eb` command started again and EasyBuild attempts to continue, tests will fail a lot. 

second, most likely unrelated problem is appearing as 
```
 sympy/utilities/_compilation/tests/test_compilation.py:test_compile_link_import_strings 
Traceback (most recent call last):
  File "/home/users/jose/.local/easybuild/build/sympy/1.3/foss-2018b-Python-3.6.6/sympy-1.3/sympy/utilities/_compilation/tests/test_compilation.py", line 61
, in test_compile_link_import_strings
    shutil.rmtree(info['build_dir'])
[...]

PR #7436, opened 2019-01-16
+download URL for intel2018a components

jose-d:
`intel2018a` is not so old and as I need it for #7425 anyway, let's do it in _correct_ way..

(created using `eb --new-pr`)


boegel:
Thanks @jose-d!

PR #7438, opened 2019-01-16
{chem}[intel/2018b] CP2K v6.1

boegel:
(created using `eb --new-pr`)
requires #7437 (libxsmm, PLUMED, Libint)

PR #7437, opened 2019-01-16
{chem}[intel/2018b] CP2K v5.1, libxsmm v1.10, PLUMED 2.5.0, Libint v1.1.6

boegel:
(created using `eb --new-pr`)

PR #7447, opened 2019-01-18
retract CP2K easyconfigs using intel/2018b since the resulting installation segfaults on even basic inputs

boegel:
This removes the CP2K easyconfigs that were added in #7437 and #7438.

These installations are pretty much useless, even basic inputs quickly result in segfaults, for example using https://github.com/cp2k/cp2k/blob/master/tests/QS/benchmark/H2O-256.inp

```
$ mpirun -np 2 cp2k.popt -i H2O-256.inp
...
forrtl: severe (174): SIGSEGV, segmentation fault occurred
```

No help from guidelines posted at https://software.intel.com/en-us/articles/determining-root-cause-of-sigsegv-or-sigbus-errors, since:

* `ulimit -s unlimited` already set
* CP2K easyblock already injects `-heap-arrays 64`

Same CP2K versions built with `intel/2018a` do not show this problem.


boegel:
Some useful info at https://xconfigure.readthedocs.io/en/latest/cp2k/README/ (hat tip @klust)

PR #7407, opened 2019-01-10
{cae,data}[foss/2018b] ROOT v6.14.06, GATE v8.1.p01

boegel:
(created using `eb --new-pr`)
requires #7406 (CLHEP, Geant4)

PR #7347, opened 2018-12-24
Solved bug where ROOT did not build with rpath wrappers

casparvl:
(created using `eb --new-pr`)
Solved the wrapper issue described here https://github.com/easybuilders/easybuild-easyconfigs/issues/7339. The patch file contains the same changes that the ROOT devs made in their own repository, but now as an EasyBuild patch file. In addition, I encountered a linking error: our CFITSIO -from local repo- was build with cURL -an optional dependency-, but since ROOT tried to link CFITSIO without linking cURL, this led to undefined reference errors. Solved by adding -DCMAKE_CXX_FLAGS=-Wl,--as-needed -lcurl. Feel free to pick only the patch if you feel so inclined, though the extra linker flags should not hurt anyone due to the --as-needed flag.

PR #7503, opened 2019-01-28
ROOT doesn't build if its CFITSIO dependency is build with cURL support

casparvl:
If dependency CFITSIO is build with optional cURL dependency, and extra configure option -DCMAKE_CXX_FLAGS=-Wl,--as-needed -lcurl is needed in order for ROOT to compile. The --as-needed flag ensures lcurl is only linked to those libraries that need it, so it doesn't hurt anyone - even if people use a CFITSIO that was compiled without curl

### DIFF
--- a/easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
@@ -2,7 +2,7 @@ name = 'ROOT'
 version = '6.10.08'
 versionsuffix = '-Python-%(pyver)s'
 
-homepage = 'http://root.cern.ch/drupal/'
+homepage = 'https://root.cern.ch/drupal/'
 description = """The ROOT system provides a set of OO frameworks with all the functionality
     needed to handle and analyze large amounts of data in a very efficient way."""
 
--- a/easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
@@ -43,6 +43,6 @@ configopts += '-Droofit=ON '
 # Standard ROOT build doesn't account for the fact that CFITSIO could be linked against cURL.
 # Official bug: https://sft.its.cern.ch/jira/browse/ROOT-9755
 # If CFITSIO is linked against cURL, this results in undefined reference errors during builds unless setting:
-configopts += '-DCMAKE_CXX_FLAGS="-Wl,--as-needed -lcurl" '
+configopts += '-DCMAKE_CXX_FLAGS="$CXXFLAGS -Wl,--as-needed -lcurl" '
 
 moduleclass = 'data'
--- a/easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
@@ -11,7 +11,10 @@ toolchainopts = {'pic': True}
 
 source_urls = ['https://root.cern.ch/download/']
 sources = ['%(namelower)s_v%(version)s.source.tar.gz']
-checksums = ['2cd276d2ac365403c66f08edd1be62fe932a0334f76349b24d8c737c0d6dad8a']
+
+checksums = [
+    '2cd276d2ac365403c66f08edd1be62fe932a0334f76349b24d8c737c0d6dad8a', # ROOT source
+]
 
 builddependencies = [
     ('CMake', '3.9.5'),
@@ -38,6 +41,9 @@ dependencies = [
 configopts = '-Dxrootd=OFF -Dmysql=OFF -Dkrb5=OFF -Dodbc=OFF '
 configopts += '-Doracle=OFF -Dpgsql=OFF -Dqt=OFF '
 configopts += '-Dunuran=ON -Dtable=ON -Dexplicitlink=ON -Dminuit2=ON '
-configopts += '-Droofit=ON'
+configopts += '-Droofit=ON '
+
+# Needed if CFITSIO was build with cURL (optional dependency)
+configopts += '-DCMAKE_CXX_FLAGS="-Wl,--as-needed -lcurl" '
 
 moduleclass = 'data'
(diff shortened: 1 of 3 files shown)


casparvl:
Must be Monday morning: mixed up --pr-descr and --pr-commit-msg.

Anyway, this 'fixes' the issue for the latest EasyConfigs, as requested here https://github.com/easybuilders/easybuild-easyconfigs/pull/7347

vanzod:
@casparvl I might be missing something here but neither `CFITSIO-3.42-GCCcore-6.4.0.eb` nor `CFITSIO-3.42-intel-2017b.eb` are built against `cURL`. This is only introduced in more recent `CFITSIO` easyconfig files.

In either case, does adding this flag impact building against `CFITSIO` without `cURL` support?

casparvl:
@vanzod Sorry, you're missing a bit of context now that I resubmitted this (original PR was here https://github.com/easybuilders/easybuild-easyconfigs/pull/7347).

We had a local copy of CFITSIO-3.42-GCCcore-6.4.0.eb that was build with cURL support. Hence, I ran into this issue.

The solution should not negatively affect people that build against a CFITSIO without cURL support: because of the '--as-needed' flag, the linker doesn't link cURL if no symbols are being used (the only penalty is a slightly longer link time because of the extra logic - not a big issue in my opinion).

I actually wasn't aware that newer CFITISOs in the official repo DO have cURL support, in which case this change to the ROOT easyconfig in fact becomes more important: people tend to base their new ROOT easyconfigs on the existing ones, so it would be nice if the existing ones already contains this flag a.s.a.p. :)

easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
```diff
@@ -38,6 +41,9 @@ dependencies = [
 configopts = '-Dxrootd=OFF -Dmysql=OFF -Dkrb5=OFF -Dodbc=OFF '
 configopts += '-Doracle=OFF -Dpgsql=OFF -Dqt=OFF '
 configopts += '-Dunuran=ON -Dtable=ON -Dexplicitlink=ON -Dminuit2=ON '
-configopts += '-Droofit=ON'
+configopts += '-Droofit=ON '
+
+# Needed if CFITSIO was build with cURL (optional dependency)
```
boegel: @casparvl The comment in https://github.com/easybuilders/easybuild-easyconfigs/pull/7347/files#diff-3f9708d03f22d0e104c6c7dc0017d7e2R51 was actually more descriptive, please use it heretoo?

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 0 out of 3 (3 easyconfigs in this PR)
node3105.skitty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/54259f51f41688ad376d8c7ec8b4dc83 for a full test report.

boegel:
@casparvl Looks like this isn't so harmless as you think it is, I see some of these builds now failing with:

```
ld: cannot find -lcurl
```

easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
```diff
@@ -38,6 +38,11 @@ dependencies = [
 configopts = '-Dxrootd=OFF -Dmysql=OFF -Dkrb5=OFF -Dodbc=OFF '
 configopts += '-Doracle=OFF -Dpgsql=OFF -Dqt=OFF '
 configopts += '-Dunuran=ON -Dtable=ON -Dexplicitlink=ON -Dminuit2=ON '
-configopts += '-Droofit=ON'
+configopts += '-Droofit=ON '
+
+# Standard ROOT build doesn't account for the fact that CFITSIO could be linked against cURL.
+# Official bug: https://sft.its.cern.ch/jira/browse/ROOT-9755
+# If CFITSIO is linked against cURL, this results in undefined reference errors during builds unless setting:
+configopts += '-DCMAKE_CXX_FLAGS="-Wl,--as-needed -lcurl" '
```
boegel: @casparvl Please change this to the following, to avoid that `$CXXFLAGS` gets ignored completely...

```python
configopts += '-DCMAKE_CXX_FLAGS="$CXXFLAGS -Wl,--as-needed -lcurl" '
```

boegel:
@casparvl What's also puzzling me a bit is why this change is needed while we were able to build `ROOT` build on top of a `CFITSIO` built on top of `cURL` in #7407...

easybuild/easyconfigs/r/ROOT/ROOT-6.10.08-foss-2017b-Python-2.7.14.eb
casparvl: Finally done... =)

casparvl:

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 5 (3 easyconfigs in total)
node2602.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/6be5163c3d96755b1fc977c905bf7178 for a full test report.

casparvl:
It looks like setting the `-Wl,--as-needed -lcurl` can actually hurt if the library cannot be found:

`ld: cannot find -lcurl`

Thus, we will not merge this PR.

This issue should be resolved from ROOT 6.20 (see https://sft.its.cern.ch/jira/browse/ROOT-9755). For anyone who hits this error when trying to install any of the older versions

```
Undefined symbols for architecture x86_64:
"_curl_easy_cleanup", referenced from:
_https_open_network in libcfitsio.a(drvrnet.o)
"_curl_easy_init", referenced from:
_https_open_network in libcfitsio.a(drvrnet.o)
"_curl_easy_perform", referenced from:
_https_open_network in libcfitsio.a(drvrnet.o)
"_curl_easy_setopt", referenced from:
_https_open_network in libcfitsio.a(drvrnet.o)
"_curl_global_cleanup", referenced from:
_ffchtps in libcfitsio.a(cfileio.o)
"_curl_global_init", referenced from:
_ffihtps in libcfitsio.a(cfileio.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [lib/libFITSIO.so] Error 1
make[1]: *** [graf2d/fitsio/CMakeFiles/FITSIO.dir/all] Error 2
```

Then you can try to run the install using the EasyConfig from this PR.

PR #7529, opened 2019-02-01
{math} [GCC/6.4.0-2.28, iccifort/2018.1.163-GCC-6.4.0-2.28] SLATEC v4.1

SimonPinches:


vanzod:
@SimonPinches The same software cannot be on both `GCCcore/6.4.0` and `intel/2018a` since the former is a sub-toolchain of the latter and this would create some obvious conflicts.
Since this library contains math routines, where performance and results might change depending on the compiler, and given that it does not seem to need anything at the `foss`/`intel` toolchain level, how about adding two easyconfig files for `GCC` and for `iccifort` instead?

SimonPinches:
@vanzod I've made the changes you suggested to provide easyconfig files for `GCC` and for `iccifort` instead.  Please let me know if there are any other changes you'd like to see.

boegel (approved):
lgtm

boegel:

PR #4247, opened 2017-03-03
{chem}[dummy/dummy] Gaussian v09.e.01-AVX, Gaussian v16.A.03-AVX2

akesandgren:
(created using `eb --new-pr`)
Gaussian 09 and 16 binary installations.

PR #7553, opened 2019-02-06
{chem}[dummy/dummy] Gaussian g16.a03-AVX2

HPC-UniOldenburg:
(created using `eb --new-pr`)


vanzod:
@HPC-UniOldenburg In the official software website it is capitalized as `Gaussian`. It may then be worth keeping the same here and in the file name.

I was aware that Gaussian came in different flavors depending on the vectorization instructions used to build it (SSE4.2, AVX, AVX2). Can you please check on this since you are using a single archive here?

HPC-UniOldenburg:
@vanzod You're absolutely right, I just noticed that this is specifically the AVX2 version. I've also noticed that my colleague who built this originally, recompressed the archive because the files name on the CD was 'E6B-113X.tbz' and did not include the manual file.

I built and pushed a new eb-file, which refers to the original .tbz-archive and which has AVX2 included in the versionsuffix.
If you approve, I would delete the old gaussian-g16.a03.eb and rename this PR to gaussian-g16.a03-AVX2.eb.

boegel:
If we want to get this merged at all, we need a test report at the very least...

boegel:
(close/re-open to get updated test suite run)

boegel:
@akesandgren Can you take a look at this one during the merge sprint?

easybuild/easyconfigs/g/Gaussian/Gaussian-g16.a03-AVX2.eb
```diff
@@ -0,0 +1,46 @@
...
+that includes various methods for electronic structure calculations."""
+homepage = "http://www.gaussian.com/"
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+# license needed, download manually
+sources = ['E6B-113X.tbz']
+checksums = ['50aa2c7fd7912da7f77745407b5fdadc7aa7d6b78bb14733d67a246d0a7cbd0f']
+
+postinstallcmds = ['chgrp -R %(namelower)s %(installdir)s/g16']
```
akesandgren, line 18: This is prone to fail and not really necessary unless you don't have a site license, in which case it's probably better to do that step outside of the easyconfig.

See #4247 for an alternative.

boegel: In addition the name of the group is very much site-specific anyway...

boegel:
@akesandgren Shall we close this in favor of #4247?

akesandgren:
I think so yes.
Closing in favor of #4247

PR #7558, opened 2019-02-06
{bio}[foss/2017b] GffCompare v0.10.6

HPC-UniOldenburg:
(created using `eb --new-pr`)


easybuild/easyconfigs/g/GffCompare/GffCompare-0.10.6-foss-2017b.eb
```diff
@@ -0,0 +1,41 @@
...
+toolchainopts = {'pic': True}
+
+source_urls = [
+    'https://github.com/gpertea/%(namelower)s/archive/',
+    'https://github.com/gpertea/gclib/archive/',
+]
+sources = [
+    'https://github.com/gpertea/%(namelower)s/archive/v%(version)s.tar.gz',
+    'https://github.com/gpertea/gclib/archive/v0.10.3.tar.gz',
+]
```
vanzod: @HPC-UniOldenburg Please remove the full URL and leave only the file name in the `sources` section. Also prefer `%(version)s` to the hardcoded version in the names.

easybuild/easyconfigs/g/GffCompare/GffCompare-0.10.6-foss-2017b.eb
```diff
@@ -0,0 +1,44 @@
...
+
+toolchain = {'name': 'foss', 'version': '2017b'}
+toolchainopts = {'pic': True}
+
+source_urls = [
+    'https://github.com/gpertea/%(namelower)s/archive/',
+    'https://github.com/gpertea/gclib/archive/',
+]
+
+gclib_version = '0.10.3'
```
boegel: @HPC-UniOldenburg Any reason you're downloading an old version of `gclib` here? There is a `0.10.6` available too (now)...

We also need to be careful w.r.t. downloading from these two repositories, maybe we better put the source URL in `sources` in this particular case?

See also https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#common-easyconfig-param-sources-alt

HPC-UniOldenburg: I stuck with the old version because otherwise both would be version 0.10.6 which caused huge problems during the fetching process, which I could not resolve with 'filename' and/or 'download-filename'. I Found another source file for gffcompare which contains the programs' name in the download file but I got problems during the building process when using this version combined with gclib 0.10.6, which I could not solve.
On the upside, the mentioned source file contains a folder with gclib, which makes the installation very easy but it is still version 0.10.3.
I hope this solution is okay to you.

zao:
Test report by @zao
**FAILED**
eb-u1604 - Linux ubuntu 16.04, x86_64, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz, Python 2.7.12
See https://gist.github.com/500385f69a24a4f8425ea66564c879d3 for a full test report.

boegel:
I don't think this is still relevant: old `GffCompare` version, ancient (now deprecated) toolchain...

My apologies for never getting back to this @HPC-UniOldenburg...

PR #7460, opened 2019-01-21
{chem,phys}[foss/2018b,intel/2018b] ASAP3 v3.11.3, OpenKIM-API v2.0.0-beta.3

schiotz:
(created using `eb --new-pr`)

PR #7569, opened 2019-02-08
{chem,phys}[foss/2018b,intel/2018b] ASAP3 v3.11.6, OpenKIM-API v2.0.0

schiotz:
(created using `eb --new-pr`)


schiotz:
ASAP is just a version bump.
OpenKIM-API is a rewritten easyconfig, as they have switched to using CMake.

This obsoletes PR #7460

schiotz:
The ASE version conflict is not really relevant, since it is only ASAP3 and GPAW using the old version, and I have uploaded new versions of both using the new ASE. See PR #7570.

The OpenKIM conflict is even less relevant, as it is with an older version of ASAP3.

schiotz:
Hi @boegel 

Is there any hope that this PR and #7570 can make it into the next EasyBuild release?  

The reason the self-test fails is that the new versions of GPAW and ASAP3 use a new version of ASE.  But if the user uses the newest versions of both, then everything is consistent.  Nothing else seems to be using ASE.

Best regards
Jakob

easybuild/easyconfigs/o/OpenKIM-Models/OpenKIM-Models-20190221-foss-2018b.eb
```diff
@@ -0,0 +1,49 @@
...
+    ('OpenKIM-API', '2.0.1'),
+]
+
+source_urls = ['https://s3.openkim.org/archives/collection/']
+sources = ['OpenKIM-Models-v2-2019-02-21.txz']
+checksums = ['3bd30b0cf2bab314755a66eed621a77c72d3f990818d08366874149be39f208e']
+
+separate_build_dir = True
+configopts = '-DKIM_API_INSTALL_COLLECTION=SYSTEM '
+configopts += '-DCMAKE_C_COMPILER=`which gcc` '
```
ocaisa: Not sure how to escape this properly but you should use the environment variables `$CC`, `$CXX`, `$FC` rather than specific compilers, that way the configopts are identical in each toolchain (and less room for copy\paste mistakes)

easybuild/easyconfigs/o/OpenKIM-API/OpenKIM-API-2.0.1-intel-2018b.eb
```diff
@@ -0,0 +1,40 @@
...
+version = '2.0.1'
+
+homepage = 'https://openkim.org/'
+description = """Open Knowledgebase of Interatomic Models.
+
+OpenKIM is an API and a collection of interatomic models (potentials) for
+atomistic simulations.  It is a library that can be used by simulation programs
+to get access to the models in the OpenKIM database.
+
+This EasyBuild only installs the API, the models have to be installed
```
ellio167: You may want to update this text block to point to the openkim models easybuild

easybuild/easyconfigs/o/OpenKIM-Models/OpenKIM-Models-20190221-foss-2018b.eb
schiotz: Good point.  I will do that and update.

easybuild/easyconfigs/o/OpenKIM-API/OpenKIM-API-2.0.1-intel-2018b.eb
schiotz: I'll update the comment block.

As for the names, I think it is better to stay consistent with the naming conventions of EasyBuild, just as it is usually better for an API to be consistent with the conventions of the programming language rather than being consistent across languages.

@boegel Any opinion on this?

akesandgren:
@schiotz The easyblock PR has been merged

easybuild/easyconfigs/o/OpenKIM-API/OpenKIM-API-2.0.1-intel-2018b.eb
schiotz: @ellio167 @boegel 
I think I should reconsider - after all, there are lots of all-lowecase names in EasyBuild, and I seem to remember some semantic difference between OpenKIM and KIM (standard and implementation, perhaps).

I still really dislike the -v2 part of the name.  It is redundant, given the version number, and there is no need to keep v1 and v2 side-by-side as v1 is already on its way out.  In the same way, I regret not dumping the 3 in the ASAP3 name at the latest when I made it pip-installable (once upon a time, ASE and ASAP version 2 could coexist with versions 3, but the twos are long gone).  Now it is too late, another package has claimed the asap name on PyPI.

Python 2 and 3 have a need to be able to coexist, but even here EasyBuild has departed from the tradition to name the package Python3!

I suggest ``kim-api``  and ``kim-models`` (or is it ``openkim-models``?).

ellio167: It would be `kim-api` and `openkim-models`.

I totally understand your aversion to the "-v2".  I think my arguments for its necessity for side-by-side usage still stand.  And model comparison (even between various versions of the same model) is one of the goals of the KIM project.  However, your comments and my observations of how the various Simulator codes are adopting the transition (basically dropping v1 support and having only v2 support) are making me rethink the importance of having seamless side-by-side support.  

If we were to officially drop the "-v2" we would need to do it quickly.  I've initiated an internal discussion on this.  @schiotz, I'll keep you posted either here or via other means.

schiotz: @ellio167 OK, I'll defer to the collective wisdom of your group :-)

schiotz: @ellio167 Did you reach a conclusion on the name?

ellio167: @schiotz Thanks for the ping.  Yes, indeed.  We just finished our discussion and decided to make tje switch to a more conventional “kim-api” package name.  We’ll be implementing and releasing this quite soon.

ellio167: @schiotz OK, kim-api-2.0.2 is released and there is a new version of the models

https://s3.openkim.org/kim-api/kim-api-2.0.2.txz
https://s3.openkim.org/archives/collection/openkim-models-2019-03-29.txz

schiotz: @ellio167 @boegel 
I am making a new version of the openkim modules with the suggested module names (version 2.0.2, including a patch for the "refex" bug).  This made a new release of Asap necessary, too.

I am aware that it may never be merged due to the conflicting ASE versions.  Next week, I will make similar builds for the 2019a toolchain, which should be unproblematic.

boegel:
@schiotz Sorry for not following up on this...

Is it still worth pursuing this, or should we just close this PR?

schiotz:
We have moved on from the 2018b toolchains to the 2019b and soon 2020b.  I do not think there is a strong need to keep this open, but I am willing to go back and look at it if you think it makes sense.  I doubt it is much work.  But just closing it is of course the easiest option for all of us :-)

boegel:
@schiotz We already have easyconfigs for newer versions of ASAP3, but not for OpenKIM-API v2.0.0 yet, so it could be worth the effort to try and get this merged...

ellio167:
It appears that there is now a `kim-api` easyconfig
https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/k/kim-api
to replace the `OpenKIM-API` one.  I would suggest that the `OpenKIM-API` be removed/retired.

Separately, I would suggest that an `openkim-models` easyconfig be created to provide the OpenKIM library of interatomic potentials to be used with the `kim-api`.  It seems to me this could/should be done in a separate PR.

schiotz:
I have versions of kim-api and openkim-models almost ready for the 2020b toolchains, will submit them just after New Year.

schiotz:
I suggest closing this PR, it is obsolete.  kim-api and openkim-models were both merged in PR #12086

PR #6985, opened 2018-10-09
{devel}[foss/2017b,intel/2017b] GConf v3.2.6

vanzod:
(created using `eb --new-pr`)

PR #7570, opened 2019-02-08
{chem}[foss/2018b,intel/2018b] GPAW v1.5.2

schiotz:
(created using `eb --new-pr`)


schiotz:
A new version of GPAW has been released.

schiotz:
The ASE version conflict is not really relevant, since it is only ASAP3 and GPAW using the old version, and I have uploaded new versions of both using the new ASE.  See PR #7569

boegel:
@schiotz We can consider adding an exception in the tests for `ASE` and `OpenKIM-API`, but this goes against our policy a bit of keeping a single dependency version per toolchain generation.

The idea is that we try to keep easyconfigs that use the same toolchain compatible with each other. In some cases that doesn't make much sense of course, like there's no point in keeping `GPAW` 1.4.0 and `GPAW` 1.5.1 compatible, since they can't be used together anyway.

It's more complex than this though, since `ASE` is not only a dependency for `GPAW`, but also for `ASAP3`, as shown by the tests...

We're aware this is quite limiting, but we've learned the hard way that not using this policy results in a lot of pain later (see for example the problems with `expat` in #6985).

Of course, `ASE` is by no means as common as `expat` is, so maybe it is acceptable to add an exception, it's a bit of a judgement call...

schiotz:
@boegel 
GPAW and ASAP are rather tightly coupled with ASE, so a new version of GPAW usually comes with a new version of ASE (less is happening for ASAP, it just follows along).  And since GPAW evolves significantly, getting new releases out is relatively important.

We could of course just make the new version with a new toolchain, but that requires installing new versions of *everything*.  Locally we have informally decided to skip every second toolchain to keep the number of modules down.  

On the other hand, I see your points about version consistency.  And I fully understand if you are reluctant to modify the test suite to special-case ASE, since that kind of stuff tends to become unmaintainable.  Of course, if it is just adding a line to a config file...

If need be, we can locally install from this PR (already done, actually), and I can make a new PR for the 2019a toolchains.

schiotz:
Test report by @schiotz
**SUCCESS**
svol.fysik.dtu.dk - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/e0e647ef7a69cb28164b87b75f90b9f0 for a full test report.

boegel:
@schiotz If `ASE` and `GPAW` are that tightly coupled, we can make and exception in the tests to allow for multiple versions of `ASE` being used as a dependency.

Or maybe we should mention the `ASE` version in the `versionsuffix` of `GPAW`, to avoid confusion/causing surprising conflicts?

schiotz:
@boegel I think the risk of ASE conflicts is relatively small so far, so I would prefer to not add an ASE version in the versionsuffix.  But it is not a strong preference, your call.

But in any case, @akesandgren is right: this version should be replaces with 1.5.2 due to a serious bug in 1.5.1.  I am currently making and testing new easyblocks, and can either update this PR or make a new one.

schiotz:
@boegel @akesandgren 
I have updated the PR to GPAW 1.5.2.  This is a minimal bug-fix release fixing a critical issue with 1.5.1.

**WARNING:**
The infamous OpenBLAS Skylake bug is also affecting GPAW (all versions).  The symptom is mainly random crashes, I have not seen wrong results.  Please use the Intel version on Skylake CPUs, that also gives a major performance boost.

akesandgren:
Test report by @akesandgren
**SUCCESS**
Build succeeded for 10 out of 10 (2 easyconfigs in this PR)
See https://gist.github.com/edd67b76036460ae12104059c25a373f for a full test report.

easybuild/easyconfigs/g/GPAW/GPAW-1.5.2-foss-2018b-Python-3.6.6.eb
```diff
@@ -0,0 +1,45 @@
...
+    ('ASE', '3.17.0', versionsuffix),
+    ('libxc', '3.0.1'),
+    ('libvdwxc', '0.3.2'),
+    ('GPAW-setups', '0.9.20000', '', True),
+]
+
+buildcmd = 'build --customize=eb_customize.py'
+install_target = 'install --customize=eb_customize.py'
+
+modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
```
akesandgren: this PYTHONPATH is not needed since PythonPackage adds it by default.
It also needs
download_dep_fail = True
use_pip = False

The same for the other one.

easybuild/easyconfigs/g/GPAW/GPAW-1.5.2-slurm-submit-fix.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/gpaw/cli/sbatch.py b/gpaw/cli/sbatch.py
```
akesandgren: Please add description and author info to the patch, and don't forget to update the checksum for it in the easyconfigs.

easybuild/easyconfigs/g/GPAW/GPAW-1.5.2-foss-2018b-Python-3.6.6.eb
schiotz: Done!  Thanks for the fixes.

easybuild/easyconfigs/g/GPAW/GPAW-1.5.2-slurm-submit-fix.patch
schiotz: Done

schiotz:
Of course, my bad.  Thanks, @akesandgren

migueldiascosta:
we now have easyconfigs for the same toolchain but versions 1.4.0 and 19.8.1

@schiotz I'm closing this PR, but if this specific version is relevant, do feel free to reopen it

PR #7619, opened 2019-02-14
{bio}[iccifort/2019.1.144] G-PhoCS v1.2.3

Micket:
(created using `eb --new-pr`)


vanzod:
@Micket From the description I feel that performance of this software can be affected by the compiler used to build it. What do you think about building two separate versions, one with `GCC` and one with `iccifort` instead?

Micket:
@vanzod I have requested some files to test from the user who wanted the software. I'll update when I have checked.

Micket:
@vanzod I thought it was going to be I/O bound, but, turns out not (at least for the small test case I got). The intel version was significantly better, so I don't see any need to supply a GCC version.

easybuild/easyconfigs/g/G-PhoCS/G-PhoCS-1.2.3-iccifort-2019.1.144-GCC-8.2.0-2.31.1.eb
```diff
@@ -0,0 +1,31 @@
...
+
+toolchain = {'name': 'iccifort', 'version': '2019.1.144-GCC-8.2.0-2.31.1'}
+
+source_urls = ['http://compgen.cshl.edu/GPhoCS/downloads/']
+sources = ['GPhoCS-%(version)s.tar.gz']
+checksums = ['e3854591f68b2111d5095f64221007456ae7a601bf760a21bd00ca5c21273776']
+
+parallel = 1
+
+prebuildopts = 'sed -i "/^CC=.*/d" Makefile && '
```
boegel: @Micket What's this for exactly? To make sure the right compiler command is picked up?

That's usually deal with 'clesnly' using:

```python
buildopts = 'CC="$CC"'

File: easybuild/easyconfigs/g/G-PhoCS/G-PhoCS-1.2.3-iccifort-2019.1.144-GCC-8.2.0-2.31.1.eb [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+
+files_to_copy = [(['bin/readTrace', 'bin/G-PhoCS-1-2-3'], 'bin')]
```
boegel: The hardcoding of the version here is... annoying. Can we avoid that? Should we rename the binary?

boegel:
Although this is quite old (and using a deprecated toolchain), it was relatively low effort to get this ready to merge, so I did...

boegel (approved):

boegel:
Going in, thanks @Micket!

PR #7626, opened 2019-02-15
{bio}[dummy/] MALT v0.4.1

sassy-crick:
(created using `eb --new-pr`)


easybuild/easyconfigs/m/malt/malt-0.4.1.eb
```diff
@@ -0,0 +1,40 @@
...
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['malt-%(version)s.tar.gz']
+checksums = ['68ac3c1b0a862a595eaa13ae72a49974292bcea486cf5fff531113993afd6b61']
+
+# specify dependency on Java/1.8 "wrapper", rather than a specific Java version
+dependencies = [('Java', '1.8', '', True)]
+
+# modloadmsg = "To execute %(name)s run: ImageJ-linux64"
```
vanzod: @sassy-crick Can you please clarify on this line?

sassy-crick: @vanzod Do you mean the line with the dependencies? If that is the line you are after, please see here:
https://easybuild.readthedocs.io/en/latest/Wrapping_dependencies.html
If you mean the # modloadmsg line, I really should have removed that. Sorry. 
If you have a better idea how to donwload the installer and get that working I am open for that. As I said, that is a bit of a hack and I am not too happy with that.

boegel: The comment was made on the `modloadmsg` line, so please drop that. :)

easybuild/easyconfigs/m/malt/malt-0.4.1.eb
```diff
@@ -0,0 +1,40 @@
...
+homepage = 'http://ab.inf.uni-tuebingen.de/software/malt/'
+description = """MALT (MEGAN alignment tool) is an extension of MEGAN (metagenome analyzer).
+
+MALT performs alignment of metagenomic reads against a database of reference sequences 
+(such as NR, GenBank or Silva) and produces a MEGAN RMA file as output. The software is 
+currently under development. """
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+sources = ['malt-%(version)s.tar.gz']
```
boegel: @sassy-crick Where did you download this from? I don't see it at http://ab.inf.uni-tuebingen.de/data/software/malt/download/welcome.html ?

easybuild/easyconfigs/m/malt/malt-0.4.1.eb
sassy-crick: @boegel Thanks. It was not quite clear to me which line was meant here. I will upload the new version in a moment. :-)

easybuild/easyconfigs/m/malt/malt-0.4.1.eb
sassy-crick: @boegel I was downloading the installation script from that URL, installed it in my space and tarballed the folder. As I said in the comment, that is a hack. Let me know how to do that better and I am happy to do this. As I am still having problems with the latest GAMESS to get that configured I did not want to create another issue for me, if you see what I mean. That is the reason for my comment on top of the file. :-)

boegel: Sounds like a case for the `CmdCp` easyblock to me...

sassy-crick: There is actually more to it. The author does not do a proper version, so everything is 'master' which makes it impossible to rebuild anything. Unless somebody has a good idea here, I would suggest to close that pull request.

sassy-crick:
I had another look into that: the downloaded MALT_unix_0_4_1.sh is firing up an installer GUI. Any idea how to handle that?

sassy-crick:
I tried it with the CmdCp eaysblock but that does not work. Originally the installer file (MALT_unix_0_4_1.sh) is launching a GUI and that does then the installation. So the MALT_unix_0_4_1.sh file contains the instructions and the tarball. I am not sure how to handle that in EasyBuild. Any suggestions?

boegel:
@sassy-crick There's no way to obtain the source tarball without starting a GUI? Where's the documentation on the installation procedure?

sassy-crick:
@boegel Not that I am aware of. This is the page to download the software:
http://ab.inf.uni-tuebingen.de/data/software/malt/download/welcome.html
You can get the bits from GitHub:
https://github.com/danielhuson/malt
You would need to build the java-files yourself. I just have tried that yesterday and apart from you will need to download more programs (jloda, megan-ce) I also got quite a lot of error messages during the compilation and it eventually failed. 
I am not sure if the software is suitable for EB to be honest but maybe you know more about this (building Java applications with EB) than I do. I only know how to run Java (and brew it).

boegel:
@sassy-crick From the manual at https://ab.inf.uni-tuebingen.de/data/software/malt/download/manual.pdf (section 3):

```
To install the program in non-gui console mode,type./MALTunix0.4.1.sh -c
```

Have you tried that?

easybuild/easyconfigs/m/malt/malt-0.4.1.eb
```diff
@@ -0,0 +1,40 @@
...
+# The malt directory is bundled into a tarball which will be installed
+# This is a bit of a hack!
+# The homepage is here:
+# http://ab.inf.uni-tuebingen.de/software/malt/
+# Contribution from the Crick HPC team
+# uploaded by J. Sassmannshausen
+
+easyblock = 'PackedBinary'
+
+name = 'malt'
```
boegel: should be `MALT`?

easybuild/easyconfigs/m/malt/malt-0.4.1.eb
```diff
@@ -0,0 +1,40 @@
...
+version = '0.4.1'
+
+homepage = 'http://ab.inf.uni-tuebingen.de/software/malt/'
+description = """MALT (MEGAN alignment tool) is an extension of MEGAN (metagenome analyzer).
+
+MALT performs alignment of metagenomic reads against a database of reference sequences 
+(such as NR, GenBank or Silva) and produces a MEGAN RMA file as output. The software is 
+currently under development. """
+
+toolchain = {'name': 'dummy', 'version': ''}
```
migueldiascosta: Use of 'dummy' toolchain is deprecated, use 'system' toolchain instead; see https://github.com/easybuilders/easybuild-framework/pull/2877 for more information

boegel: @sassy-crick Please remove this easyconfig (since you added another using `MALT` as name):

```
eb --update-pr 7626 :malt-0.4.1.eb  --pr-commit-msg "remove malt easyconfig (since name should be uppercase)"

File: easybuild/easyconfigs/m/MALT/MALT-0_4_1.eb [hunk 5]
```diff
@@ -0,0 +1,43 @@
...
+# This is a bit of a hack!
+# The homepage is here:
+# http://ab.inf.uni-tuebingen.de/software/malt/
+# Contribution from the Crick HPC team
+# uploaded by J. Sassmannshausen
+
+easyblock = 'CmdCp'
+
+name = 'MALT'
+version = '0_4_1'
```
boegel: Please use `0.4.1` as version

easybuild/easyconfigs/m/MALT/MALT-0_4_1.eb
```diff
@@ -0,0 +1,43 @@
...
+
+source_urls = ['https://ab.inf.uni-tuebingen.de/data/software/malt/download']
+sources = ['MALT_unix_%(version)s.sh']
+checksums = ['8ccf5f073be6d0960c0886b1fdfc6237515dfaf7d4b71a470db805d72bfc17f3']
+
+# specify dependency on Java/1.8 "wrapper", rather than a specific Java version
+dependencies = [('Java', '1.8', '', True)]
+
+extract_sources = False
+install_cmd = ['./MALT_unix_0_4_1.sh -c']
```
boegel: You should avoid hardcoding the version here... This should work I think:

```python
install_cmd = ['./MALT_unix_*.sh -c']

File: easybuild/easyconfigs/m/MALT/MALT-0_4_1.eb [hunk 7]
```diff
@@ -0,0 +1,43 @@
...
+homepage = 'http://ab.inf.uni-tuebingen.de/software/malt'
+description = """MALT (MEGAN alignment tool) is an extension of MEGAN (metagenome analyzer).
+MALT performs alignment of metagenomic reads against a database of reference sequences 
+(such as NR, GenBank or Silva) and produces a MEGAN RMA file as output. The software is 
+currently under development. """
+
+toolchain = SYSTEM
+
+source_urls = ['https://ab.inf.uni-tuebingen.de/data/software/malt/download']
+sources = ['MALT_unix_%(version)s.sh']
```
boegel: If you use `0.4.1` as version, you'll need to change this to:

```python
sources = ['MALT_unix_%s.sh' % version.replace('.', '_')]

sassy-crick:
I close this PR as I will no longer be able to work on it at all. Sorry for this!

PR #7762, opened 2019-02-28
{chem,tools}[foss-2018b,intel-2018b] CheMPS2-1.8.9, deepdiff-3.3.0, pytest-4.3.0

tovrstra:
This is the low-hanging fruit. The new PSI4 build itself is a piece of work and will be done in a separate PR.


tovrstra:
@boegel The Error message was not very clear but I think I get it. Does it mean the following?

'PythonBundle' easyblock **must be** used for bundle of Python packages in pytest-4.3.0-intel-2018b-Python-3.6.6.eb

akesandgren:
There is already libxc-4.2.3 for foss/intel 2018b
And there is no dependency on libxc in these easyconfigs so it doesn't really belong in here.

There is also the added problem of only allowing one version of dependencies like libxc per version of toolchain (unless the package is at GCCcore level)

Should libxc be pulled from this PR? @boegel thoughts on that?

boegel:
Yeah, the `libxc` should be fleshed out, ideally...

There's no big issue with having the easyconfig itself, but using it as a dep may introduce conflicts when combining multple tools.

boegel:
@tovrstra We assert that `PythonBundle` is used, and that is failing. So if you see `AssertionError: PythonBundle is used ...` that makes sense imho (I can see why it's confusing, but we do this consistently like this in all tests)

boegel:
Since this review is pretty clean, I think we can merge this as is if the incoming test reports pass.

`libxc` isn't needed as a dep here, but it does fit the scope of tools and it could be useful...

tovrstra:
ok. I was trying to build PSI4 1.3 against an external libxc, but that turned out to be impossible because of: https://github.com/psi4/psi4/issues/1302, so I'll take them back out.

akesandgren:
The CheMPS2 tests are sensitive to OMP_NUM_THREADS and sometimes (often/always) hangs if it isn't set to 1.
Currently this means that one need to do
```
env OMP_NUM_THREADS=1 eb ...
```
to make it pass.

boegel:
@akesandgren Indeed, all the tests reports I had running timed out after many hours...

akesandgren:
Test report by @akesandgren
**SUCCESS**
Build succeeded for 6 out of 6 (6 easyconfigs in this PR)
See https://gist.github.com/56a343d982cf52d713e7330407458b60 for a full test report.

tovrstra:
@akesandgren Thanks for pointing this out. Would the following line in the config fix this?

```python
runtest = 'OMP_NUM_THREADS=1 test'
```

akesandgren:
No it won't because the arg to runtest is fed "make args", we have a solution in the works though. I'll let you know when it's merged and what to change.

tovrstra:
Thanks. Good to know. Is there a way (for other configs, e.g. PSI4) to specify the full test command, e.g. to call pytest? Alternatively, could we also just add a patch to change the test target in the CMake files?

akesandgren:
Please add
```
pretestopts = 'export OMP_NUM_THREADS=1 && '
```
to the CheMPS2 easyconfigs.
It doesn't help right now but with some newly added changes to framework and easyblocks it will take effect with the next release of EasyBuild.

akesandgren:
@tovrstra Can you add the pretestopts line please?

tovrstra:
@akesandgren I'll do so next week. I'm on vacation at the moment.

tovrstra:
@akesandgren Done. Sorry for the delay.

akesandgren:
Test report by @akesandgren
**SUCCESS**
Build succeeded for 6 out of 6 (6 easyconfigs in this PR)
See https://gist.github.com/9a6b574ad6d213828db499c6a2e3de91 for a full test report.

akesandgren:
Going in, thanks @tovrstra!

tovrstra:
you're welcome. :)

PR #7773, opened 2019-03-01
{tools}[intel/2018b] NCO v4.7.9 (+ deps)

boegel:
(created using `eb --new-pr`)


easybuild/easyconfigs/l/libdap/libdap-3.20.3-GCCcore-7.3.0.eb
```diff
@@ -0,0 +1,34 @@
...
+
+name = 'libdap'
+version = '3.20.3'
+
+homepage = 'http://opendap.org/download/libdap'
+description = """A C++ SDK which contains an implementation of DAP 2.0
+ and the development versions of DAP3, up to 3.4.
+ This includes both Client- and Server-side support classes."""
+
+toolchain = {'name': 'GCCcore', 'version': '7.3.0'}
```
akesandgren: If this is a C++ SDK, then GCCcore is probably not a good fit. It probably need to be at GCC/iccifort level.

boegel:
superseded by #8108 + #7818

boegel:
Oh, hold on, different toolchain...

easybuild/easyconfigs/l/libdap/libdap-3.20.3-GCCcore-7.3.0.eb
boegel: Same as in https://github.com/easybuilders/easybuild-easyconfigs/pull/7818/files#r263772876, we haven't seen any issues with `libdap` built with `GCCcore` until now, so I think this can stay as-is.

boegel:
Managed to mess this up real good, so replaced with #8316

PR #7111, opened 2018-11-02
{vis}[foss/2017b,intel/2017b] Graphviz v2.40.1

vanzod:
(created using `eb --new-pr`)

PR #7784, opened 2019-03-01
{vis}[GCC/6.4.0-2.28,iccifort/2017.4.196-GCC-6.4.0-2.28] GraphicsMagick v1.3.28

murfreesboro:
(created using `eb --new-pr`)


murfreesboro:
Test report by @murfreesboro
**SUCCESS**
dev806.vampire - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU           E5620  @ 2.40GHz, Python 2.7.5
See https://gist.github.com/4f000f43953c05d6a9b9c76759eb34c9 for a full test report.

easybuild/easyconfigs/g/GraphicsMagick/GraphicsMagick-1.3.28-GCC-6.4.0-2.28.eb
```diff
@@ -0,0 +1,41 @@
...
+    ('libpng', '1.6.32'),
+    ('libjpeg-turbo', '1.5.2'),
+    ('LibTIFF', '4.0.9'),
+    ('libxml2', '2.9.4'),
+    ('XZ', '5.2.3'),
+    ('zlib', '1.2.11'),
+    ('Ghostscript', '9.22'),
+]
+
+modextrapaths = {'CPATH': ['include/GraphicsMagick']}
```
akesandgren: Is this really needed? Code that uses this is really supposed to do
```
#include <GraphicsMagick/somefile.h>
```

murfreesboro: I checked the GraphicsMagick codes. It seems at least freetype and X11 are explicitly needed for fundamental pieces in this module. Additionally, this package also has the function to convert/output image format like TIFF, png etc. so that in general they also need these running time library to support these function. Therefore I think it's reasonable to have such a list as dependency.

akesandgren: I wasn't talking about the dependencies but about the
```
modextrapaths = {'CPATH': ['include/GraphicsMagick']}
```
line.
By default we always add %(installdir)s/include to CPATH.

murfreesboro: I see your point. That's reasonable way to define the include path. I changed it into %(installdir)s/include.

murfreesboro: finally I changed it into modextrapaths = {'CPATH': 'include'} to drop the GraphicsMagick.

murfreesboro:
Test report by @murfreesboro
**SUCCESS**
dev806.vampire - Linux centos linux 7.4.1708, Intel(R) Xeon(R) CPU           E5620  @ 2.40GHz, Python 2.7.5
See https://gist.github.com/3b1eeae40bc998fe37e22de4a7c1e63c for a full test report.

easybuild/easyconfigs/g/GraphicsMagick/GraphicsMagick-1.3.28-GCC-6.4.0-2.28.eb
akesandgren: Still not what i meant.
```
modextrapaths = {'CPATH': 'include'}
```
is what is done by default, which means you don't need it at all.

If you check the resulting module file you should, with the above still in place, see CPATH getting set twice with the same path.

murfreesboro: OK. Then what do you want me to do? Do you want me to delete this line? 
`modextrapaths = {'CPATH': ['include/GraphicsMagick']}`

akesandgren: As I started with, if code that includes files from graphicsmagick does the right thing and uses
```
#include <GraphicsMagick/filename.h>
```
then you do not need the modextrapaths at all. However, if the common way to include files from graphicsmagick is to use
```
#include <filename.h>
```
then you do need
```
modextrapaths = {'CPATH': ['include/GraphicsMagick']}
```
I.e., if we need the modextrapaths or not, depends on how other codes, that use graphicsmagick, does their includes.

Does that make it clear?

murfreesboro: I understand how the codes look like for the include files, I just do not know what you wants me to do. Next time you can just tell me "delete this line" then I know what you want.

murfreesboro: delete the include dir, use the default one now.

zao: Given the file `include/GraphicsMagick/magick/api.h` - samples on their website [1] uses `#include <magick/api.h>` and the tool `GraphicsMagick-config --cppflags` says `-I/foo/GraphicsMagick`.

As such, the original formulation of `modextrapaths = {'CPATH': ['include/GraphicsMagick']}` would probably be appropriate to mimic the behaviour of existing usage.

[1] http://www.graphicsmagick.org/api/api.html

akesandgren: According to the documentation and looking at Octave we need to keep the
```
modextrapaths = {'CPATH': ['include/GraphicsMagick']}
```
line.

akesandgren:
Due to libxml2 and Ghostscript version conflicts we might have to lift GraphicsMagick up to foss/intel level.
@boegel, any thoughts here?

easybuild/easyconfigs/g/GraphicsMagick/GraphicsMagick-1.3.28-GCC-6.4.0-2.28.eb
murfreesboro: OK. Add the line back.

murfreesboro:
I tried to upgrade the toolchain but it does not work.

akesandgren:
For that to have a chance to work you would also need to lower libxml2 to 2.9.4, which is what 2017b toolchains use.

vanzod:
@akesandgren I am afraid that the problem is not the toolchein level but the fact that we have conflict at the `GCCcore` level for `libxml2` since it is shared between `2017b` and `2018a`. The situation seems the same as in #6985 and #7111.

akesandgren:
I thought that we  don't check for conflicts at GCCcore level.

boegel:
@akesandgren There's a difference between checking for introduced conflicts, where you require two versions in a particular dependency tree, and the dependency variants check we do to try and ensure that *any* two easyconfigs in the same generation (e.g. `2018b`) are compatible with each other (without depending on each other).

PR #7791, opened 2019-03-02
avoid use of .items() in R (bundle) easyconfigs

boegel:
Note: adding checksums for these old `R-*` easyconfigs that involve a lot of extensions (many of which the sources have disappeared into thin air already, I'm looking at you Bioconductor) is close to impossible, so I've temporarily disabled the checksum requirement for `R-*` easyconfigs in 
bc11d6a (I'll roll that back as soon as this PR is merged).

PR #7792, opened 2019-03-02
restore checksums check for R-* easyconfigs 

boegel:
follow-up of checksums check for `R-*` that was temporarily disabled in #7791


akesandgren:
Going in, thanks @boegel!

PR #7796, opened 2019-03-02
archive ancient CUDA 5.0.35 easyconfigs with creative way of determining sources

boegel:
Easyconfig tests on top of Python 3.x fail on these easyconfigs, trivial change...


akesandgren:
Going in, thanks @boegel!

PR #7798, opened 2019-03-03
fix Travis config to test with both Python 2.6 & 2.7

boegel:
Turns out only *one* configuration is being tested after #7795 got merged, which was certainly not the intention...


akesandgren:
Going in, thanks @boegel!

PR #7815, opened 2019-03-06
{bio}[foss/2018b] Trim_Galore v0.6.0

sassy-crick:
(created using `eb --new-pr`)


easybuild/easyconfigs/t/Trim_Galore/Trim_Galore-0.6.0.eb
```diff
@@ -0,0 +1,36 @@
...
+source_urls = ['https://github.com/FelixKrueger/TrimGalore/archive/']
+sources = [{
+    'filename': '39e1ad7.tar.gz',
+}]
+
+checksums = ['aad008ff2e9605079655013bd91615ca2272048b151b2fe7ee197d18b1b87908']
+
+dependencies = [
+    ('FastQC', '0.11.8-Java-1.8'),
+    ('cutadapt', '1.18-foss-2018b-Python-2.7.15')
```
akesandgren: Trim_Galore actually depends on Perl and needs a patch to not use perl from /usr/bin/perl but instead
#!/usr/bin/env perl
And thus don't really belong in dummy, unless an explicit osdependency for the respecitve perl packages for debian/RH is added.

akesandgren: And, since I was slightly blind last time i looked at this...
that cutadap dependency is wrong.
Trim_Galore must use the foss/2018b toolchain so it can depend on cutadapt correctly.
And further more, w already have Trim_Galore-0.6.0-foss-2018b-Python-2.7.15.eb merged so this PR is likely not needed any longer.
@sassy-crick Close it?

sassy-crick: Hi akesandgren,
thanks for the feedback and sorry for my slow response. I have done the requested changes and created a patch-file for the Perl issue. I have noticed there is a newer version out by now. If you are happy with this EB build file, I would suggest we can close this one and I upload the EB build file for the new version. Are you happy with this?

sassy-crick: @akesandgren 
The same issue regarding Perl was also mentioned here:
https://github.com/easybuilders/easybuild-easyconfigs/pull/8019
How shall we proceed regarding Perl in Trim_Galore?
As I said, I have no opinion here at all.

easybuild/easyconfigs/t/Trim_Galore/Trim_Galore-0.6.0-foss-2018b.eb
```diff
@@ -0,0 +1,42 @@
...
+}]
+
+checksums = [
+    'aad008ff2e9605079655013bd91615ca2272048b151b2fe7ee197d18b1b87908',  # 39e1ad7.tar.gz
+    'd4d812854755daf91c2443d268afabdacbd1f0820891aaeafb6ddd056571b3a3',  # perl.patch
+]
+
+dependencies = [
+    ('FastQC', '0.11.8', '-Java-1.8', True),
+    ('cutadapt', '1.18', '-Python-2.7.15')
```
boegel: @sassy-crick Since `Python` is a runtime dependency, this should trickle up via `versionsuffix`, at which point we basically get the easyconfig already merged via #7893, so closing this.

In addition, the `sources` spec shouldn't be using a commit ID...

easybuild/easyconfigs/t/Trim_Galore/Trim_Galore-0.6.0.eb
boegel: @sassy-crick In #7893, it's dealt with via a `sed` command, which is more portable w.r.t. maintaining easyconfigs for future `Trim_Galore` versions...

PR #7754, opened 2019-02-27
run various checks on easyconfigs that are touched in pull requests and involve Python packages

boegel:
tested via #7753, works as intended...

PR #7603, opened 2019-02-13
{devel}[foss/2018b] PyTorch v1.0.1 w/ Python 3.6.6

branfosj:
(created using `eb --new-pr`)

PR #7717, opened 2019-02-22
{devel}[fosscuda/2017b] PyTorch v0.3.1

vanzod:
(created using `eb --new-pr`)

PR #7844, opened 2019-03-09
add PyTorch to whitelist for not having 'use_pip' enabled

boegel:
relevant for #7603 and various other current (& future) `PyTorch` PRs since we've put #7754 in place...


boegel:
@vanzod did an effort to try and get it working with `pip`, see https://github.com/easybuilders/easybuild-easyconfigs/pull/7717#issuecomment-467477325

akesandgren:
Going in, thanks @boegel!

Issue #7951, opened 2019-03-25
New version of BEDTools includes htslib


verdurin:
Note that the new version of `BEDTools` includes `htslib`. My first quick test build failed, so some changes will be required:

https://github.com/arq5x/bedtools2/releases/tag/v2.28.0

ghuls:
@verdurin This pull request https://github.com/easybuilders/easybuild-easyconfigs/pull/7962 works for me.

One thing I noticed is that I think BEDtools is never compiled with `-march=native`

boegel:
@verdurin Can you elaborate on the problems you ran into?

Other than the changes in `sanity_check_paths`, the easyconfig contributed by @ghuls in #7962 looks like a pretty straightforward update (unless I'm missing something)...

verdurin:
I was trying with `foss-2018b` and first quick attempt saw lots of errors like this:
```
  * compiling chromsweep.cpp
In file included from bedFile.cpp:12:0:
bedFile.h: In member function ‘void BedFile::reportBedTab(const T&)’:
bedFile.h:47:21: error: expected ‘)’ before ‘PRId64’
 #define PRId_CHRPOS PRId64
                     ^
bedFile.h:942:33: note: in expansion of macro ‘PRId_CHRPOS’
                 printf ("%s\t%" PRId_CHRPOS "\t%" PRId_CHRPOS "\t", bed.chrom.c_str(), start, end);
                                 ^~~~~~~~~~~
bedFile.h:47:21: error: expected ‘)’ before ‘PRId64’
 #define PRId_CHRPOS PRId64
                     ^
bedFile.h:945:33: note: in expansion of macro ‘PRId_CHRPOS’
                 printf ("%s\t%" PRId_CHRPOS "\t%" PRId_CHRPOS "\t%s\t",
                                 ^~~~~~~~~~~
bedFile.h:47:21: error: expected ‘)’ before ‘PRId64’
 #define PRId_CHRPOS PRId64
                     ^
bedFile.h:949:33: note: in expansion of macro ‘PRId_CHRPOS’
                 printf ("%s\t%" PRId_CHRPOS "\t%" PRId_CHRPOS "\t%s\t%s\t",
                                 ^~~~~~~~~~~
```

It may be that there's no problem with `htslib`, it's just that the announcement made it sound as though there was more bundling.

ghuls:
[...]

PR #7962, opened 2019-03-26
{bio}[foss/2018a] BEDTools v2.28.0

ghuls:
This should fix issue #7951

### DIFF
--- a/easybuild/easyconfigs/b/BEDTools/BEDTools-2.28.0-foss-2018a.eb
+++ b/easybuild/easyconfigs/b/BEDTools/BEDTools-2.28.0-foss-2018a.eb
@@ -1,0 +1,44 @@
+# Author: Maxime Schmitt, University of Luxembourg
+# Author: Adam Huffman, The Francis Crick Institute
+#
+# Based on the work of: Pablo Escobar Lopez
+# Swiss Institute of Bioinformatics (SIB)
+# Biozentrum - University of Basel
+
+easyblock = 'MakeCp'
+
+name = 'BEDTools'
+version = '2.28.0'
+
+homepage = "https://github.com/arq5x/bedtools2"
+description = """The BEDTools utilities allow one to address common genomics tasks such as finding feature overlaps
+ and computing coverage. The utilities are largely based on four widely-used file formats: BED, GFF/GTF, VCF,
+ and SAM/BAM."""
+
+toolchain = {'name': 'foss', 'version': '2018a'}
+
+source_urls = ['https://github.com/arq5x/bedtools2/releases/download/v%(version)s/']
+sources = ['bedtools-%(version)s.tar.gz']
+checksums = ['15af6d10ed28fb3113cd3edce742fd4275f224bc06ecb98d70d869940220bc32']
+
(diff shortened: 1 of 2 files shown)


boegel:
(close/re-open to re-trigger Travis after target branch was changed to `develop`)

easybuild/easyconfigs/b/BEDTools/BEDTools-2.28.0-foss-2018a.eb
```diff
@@ -0,0 +1,38 @@
...
+buildopts = 'CXX="$CXX"'
+
+dependencies = [
+    ('zlib', '1.2.11'),
+]
+
+files_to_copy = ["bin", "docs", "data", "genomes", "scripts", "test"]
+
+sanity_check_paths = {
+    'files': ['bin/%s' % x for x in ['annotateBed', 'bamToBed', 'bamToFastq', 'bed12ToBed6', 'bedpeToBam', 'bedToBam', 'bedToIgv', 'bedtools', 'closestBed', 'clusterBed', 'complementBed', 'coverageBed', 'expandCols', 'fastaFromBed', 'flankBed', 'genomeCoverageBed', 'getOverlap', 'groupBy', 'intersectBed', 'linksBed', 'mapBed', 'maskFastaFromBed', 'mergeBed', 'multiBamCov', 'multiIntersectBed', 'nucBed', 'pairToBed', 'pairToPair', 'randomBed', 'shiftBed', 'shuffleBed', 'slopBed', 'sortBed', 'subtractBed', 'tagBam', 'unionBedGraphs', 'windowBed', 'windowMaker']],
```
boegel: @ghuls Do you mind breaking wrapping this line to have lines that are maximally 120 characters wide?

Something like:

```python
sanity_check_paths = {
    'files': ['bin/%s' % x for x in ['annotateBed', 'bamToBed', 'bamToFastq', 'bed12ToBed6',
                                     'bedpeToBam',, 'bedToBam', 'bedToIgv', 'bedtools', 'closestBed', 
                                     'clusterBed', 'complementBed', , 'coverageBed', ...]],
    'dirs': files_to_copy,
}

vanzod:
@ghuls Previous versions of `BEDTools` required `BamTools` as a dependency, which is now missing here. Is this dependency not required anymore?

boegel:
@vanzod Does #7951 answer your question?

ghuls:
@vanzod I am not sure how much it still uses from Bamtools now that they use htslib. They include their own copy anyway: https://github.com/arq5x/bedtools2/tree/master/src/utils/BamTools

vanzod:
@boegel It was not 100% clear to me from that issue. Plus it is useful to have an explicit reference for a dependency change directly into the PR for the love of our future selves.

@ghuls Ok, so the fact that they still include it does it mean that it is still needed? We always try to avoid internal builds and prefer to use components built by EasyBuild directly instead.

boegel:
The toolchain being used here is now deprecated, and we have more recent versions of `BEDTools` already merged, so I'll close this PR; sorry for not actively following up on this @ghuls!

PR #7994, opened 2019-04-02
Fix LWM2, OTF2, OPARI2, and Score-P download URLs

geimer:
The URLs have changed due to an CMS upgrade :unamused:

### DIFF
--- a/easybuild/easyconfigs/__archive__/l/LWM2/LWM2-1.1-gompi-1.4.12-no-OFED.eb
+++ b/easybuild/easyconfigs/__archive__/l/LWM2/LWM2-1.1-gompi-1.4.12-no-OFED.eb
@@ -29,9 +29,9 @@ description = """The lightweight measurement module (LWM2) is a low overhead pro
 toolchain = {'name': 'gompi', 'version': '1.4.12-no-OFED'}
 toolchainopts = {"usempi": True}
 
-# http://www.vi-hps.org/upload/projects/hopsa/lwm2-1.1.tar.gz
+# http://www.vi-hps.org/cms/upload/projects/hopsa/lwm2-1.1.tar.gz
 sources = [SOURCELOWER_TAR_GZ]
-source_urls = ['http://www.vi-hps.org/upload/projects/hopsa/']
+source_urls = ['http://www.vi-hps.org/cms/upload/projects/hopsa/']
 
 # compiler toolchain depencies
 dependencies = [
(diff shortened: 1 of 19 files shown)


geimer:
Test report by @geimer
**SUCCESS**
Build succeeded for 19 out of 19 (19 easyconfigs in this PR)
swat - Linux debian 9.7, Intel(R) Xeon(R) CPU           X5550  @ 2.67GHz, Python 2.7.13
See https://gist.github.com/cca57b9346dda842bb11833626864dc8 for a full test report.

geimer:
Hmm... Since I only changed the download URLs in existing easyconfigs, I have no clue what Travis is complaining about.  Any hint would be appreciated.

boegel:
@geimer Tests fail because of a bug in framework, fix is being added via https://github.com/easybuilders/easybuild-framework/pull/2825, we'll re-trigger the tests here as soon as the fix is merged...

Two things though:

* there's little point in fixing the archived easyconfigs

* for all easyconfigs touched in PRs, we now require that SHA256 checksums are in place for all sources/patches... (and that's actually very relevant here, we want to make sure the downloaded tarballs remain the same as before)

vanzod:
Re-trigger travis after fix

geimer:
@boegel: SHA256 checksums are now in place.  Regarding archived easyconfigs: I think there is little point in keeping them if they are known to be broken.  So either we fix them (if it's about checksums and URLs) or delete them entirely.

boegel:
@geimer The archived easyconfigs will be trashed in the upcoming EasyBuild v4.0, but there other EasyBuild releases where they are still included.

I don't strongly object fixing them as well, but it may be more hassle than it's worth (they're only picked up if people configure EasyBuild with `--consider-archived-easyconfigs`).

I've re-triggered the tests now the problems in framework have been resolved, but it seems like there's one more issue you should take care of:

```
AssertionError: Failed to find parsed easyconfig for LWM2-1.1-gompi-1.4.12-no-OFED.eb
```

(let's see if that's still there when the tests complete again though)

geimer:
@boegel The archived easyconfigs are now fixed as well, so no additional work to do in this regard.  The easyconfig mentioned in the assertion is actually one of them -- and it is there:
```
$ find . -name LWM2-1.1-gompi-1.4.12-no-OFED.eb
./easybuild/easyconfigs/__archive__/l/LWM2/LWM2-1.1-gompi-1.4.12-no-OFED.eb
```
Not sure what is going one here...

boegel:
@geimer It may be a side effect of touching archived easyconfigs, some tests may not be aware of that.

Let me take a look at it when it fails again, could be an easy fix.

boegel:
@geimer Please see if https://github.com/geimer/easybuild-easyconfigs/pull/5 fixes the tests tripping over the fact that you're changing archived easyconfigs...

geimer:
@boegel This seems to do the trick.  Thanks!

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 209 out of 209 (19 easyconfigs in this PR)
node2055.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/a5174bfdda2ad45b31533c42af4255d4 for a full test report.

boegel:
Going in, thanks @geimer!

PR #8005, opened 2019-04-04
{numlib} [iimpi iompi] Set up PKG_CONFIG_PATH for MKL library

SimonPinches:
Seems like a useful feature to me which can be ignored by anyone not interested in the Intel shipped `pkg-config` files.

### DIFF
--- a/easybuild/easyconfigs/i/imkl/imkl-2018.1.163-iimpi-2018a.eb
+++ b/easybuild/easyconfigs/i/imkl/imkl-2018.1.163-iimpi-2018a.eb
@@ -36,4 +36,8 @@ modextravars = {
     'MKL_EXAMPLES': '%(installdir)s/mkl/examples/',
 }
 
+modextrapaths = {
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/i/imkl/imkl-2018.1.163-iimpi-2018a.eb
```diff
@@ -36,4 +36,8 @@ modextravars = {
     'MKL_EXAMPLES': '%(installdir)s/mkl/examples/',
 }
 
+modextrapaths = {
+    'PKG_CONFIG_PATH': '%(installdir)s/compilers_and_libraries_2018.1.163/linux/mkl/bin/pkgconfig',
```
akesandgren: %(installdir)s/mkl/bin/pkgconfig
is shorter and version agnostic.

SimonPinches: @akesandgren, I fully agree.  I've pushed an update to make this change to both `eb` files.

PR #8018, opened 2019-04-05
{math}[foss/2018b] Dakota v6.9

JackPerdue:
(created using `eb --new-pr`)


easybuild/easyconfigs/g/GSL/GSL-2.5-GCCcore-7.3.0.eb
```diff
@@ -0,0 +1,25 @@
...
+
+homepage = 'http://www.gnu.org/software/gsl/'
+
+description = """
+ The GNU Scientific Library (GSL) is a numerical library for C and C++
+ programmers.  The library provides a wide range of mathematical routines
+ such as random number generators, special functions and least-squares fitting.
+"""
+
+toolchain = {'name': 'GCCcore', 'version': '7.3.0'}
```
akesandgren: We already have GSL-2.5 aat GCC level (GSL-2.5-GCC-7.3.0-2.30.eb) so this one can be dropped.

JackPerdue: Why wasn't it reduced  to the minimal toolchain  so it could be used with Intel toolchains too???  
I deleted it here but can't delete it on our clusters since we've used it for multiple toolchains (the icc toolchain reduces to GCCcore, not GCC)

JackPerdue:
try again

boegel:
superseded by #11072 and #12275

PR #7929, opened 2019-03-19
{devel}[GCCcore/8.2.0] Mako v1.0.7

Micket:
(created using `eb --new-pr`)

PR #8038, opened 2019-04-09
fix checking of binutils build dep in easyconfig tests

boegel:
When `multi_deps` is involved, we must use the `.builddependencies()` method rather than accessing the `builddependencies` easyconfig parameter directly, to avoid hitting a list of lists of build dependencies, cfr. https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/framework/easyconfig/easyconfig.py#L888

This will fix the failing test in #7929 (cc @Micket)


migueldiascosta:
Going in, thanks @boegel!

PR #8123, opened 2019-04-24
PythonBundle already has a decent customised sanity_check_paths for bundles of Python packages

boegel:
follow-up for #8101

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -428,8 +428,8 @@ class EasyConfigTest(TestCase):
     def check_sanity_check_paths(self, changed_ecs):
         """Make sure a custom sanity_check_paths value is specified for easyconfigs that use a generic easyblock."""
 
-        # PythonBundle already has a decent customised sanity_check_paths for bundles of Python packages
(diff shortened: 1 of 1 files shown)


test/easyconfigs/easyconfigs.py
```diff
@@ -428,10 +428,16 @@ def check_python_packages(self, changed_ecs):
     def check_sanity_check_paths(self, changed_ecs):
         """Make sure a custom sanity_check_paths value is specified for easyconfigs that use a generic easyblock."""
 
+        # PythonBundle already has a decent customised sanity_check_paths for bundles of Python packages
+        whitelist = ['PythonBundle']
```
smoors: `PythonPackage` also has a customized `sanity_check_paths`
isn't it better to check if the `sanity_check_paths` is equal to the default instead of having to maintain a whitelist?

boegel: Good point on `PythonPackage`, but checking against the default value is tricky because of how `PythonBundle` & `PythonPackage` add the default, see https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/generic/pythonbundle.py#L110

PR #8251, opened 2019-05-06
temporarily only test with Python 2.6 + Lmod

boegel:
I would like to temporarily run the easyconfigs test suite with only Python 2.6 + Lmod, to avoid that Travis becomes a bottleneck during the merge sprint we have planned Wed May 8th 2019.

@migueldiascosta Please merge this when you start your day, since you'll be the first... ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ matrix:
   fast_finish: true
   include:
     - python: 2.6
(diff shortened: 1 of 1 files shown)


migueldiascosta:
Going in, thanks @boegel!

Issue #3697, opened 2016-10-20
DBus-1.10.12-intel-2016b.eb needs expat/2.2.0 as a dependency


JackPerdue:
I had to add it to get a successful build.

geimer:
@JackPerdue This issue should be fixed with #8283, that is, it can be closed?

boegel:
Indeed fixed, so closing...

PR #8283, opened 2019-05-09
DBus: Add `pkg-config` and `expat` as (build) dependency

geimer:
On a test system without `pkg-config` installed on the system, building these `DBus` configs fails as `configure` is unable to find `expat`.  Adding `pkg-config` as a build dependency fixes this.

`DBus-1.13.0-intel-2017b.eb` is likely to also suffer from this issue, however, since I don't have this version of the Intel toolchain on my test system, I'm unable to verify this.  Maybe someone else can check and let me know?

Older versions (e.g., `DBus-1.10.20-GCCcore-6.4.0.eb`) seem to build OK.


geimer:
Test report by @geimer
**SUCCESS**
swat - Linux debian 9.7, Intel(R) Xeon(R) CPU           X5550  @ 2.67GHz, Python 2.7.13
See https://gist.github.com/1032bfe7d20d732d09ba806a7b2774a4 for a full test report.

akesandgren:
Please add the fix to all versions of DBus 1.13.6 and newer that exist in develop branch and we'll run the tests.

geimer:
@akesandgren Ehm... That's what I've done already. Or am I missing something?

akesandgren:
oops, sorry. I though I saw more of them in the ls output :-)

akesandgren:
So please add it to anything with a toolchain 2016a or newer. That's what I should have said...

geimer:
The question is rather whether it is needed or not...  The situation is as follows.  Currently, there are 9 `DBus` easyconfigs in `develop`.  I have tested and with this PR fixed the following:

 - `DBus-1.13.6-GCCcore-6.4.0.eb`
 - `DBus-1.13.6-GCCcore-7.3.0.eb`
 - `DBus-1.13.8-GCCcore-8.2.0.eb`

I have also tested

 - `DBus-1.10.20-GCCcore-6.4.0.eb`

but it didn't seem to require the fix; it found the `expat` library even without `pkg-config` for whatever reason.  Of the remaining configs, only

 - `DBus-1.13.0-intel-2017b.eb`

specifies `expat` as a dependency, and since it is of the 1.13 series, likely also requires the fix.  The others, i.e.,

 - `DBus-1.10.8-foss-2016a.eb`
 - `DBus-1.10.8-intel-2016a.eb`
 - `DBus-1.10.12-intel-2016b.eb`
 - `DBus-1.11.20-intel-2017a.eb`

don't specify `expat` as a dependency, thus `pkg-config` is very likely not required if this is really correct (#3697 might be an indication that it is not, but this should be verified).

Bottom line: I'm happy to include fixes for the easyconfigs I'm unable to test, but I'd like to get some indication what is actually needed to avoid adding unnecessary dependencies.

akesandgren:
At least dbus 1.10.8 uses pkg-config during configure, so anything from that onwards needs it.

geimer:
Added. What about the `expat` dependency?  Is it required in all configs or not?

akesandgren:
From 1.10.8 yes.

geimer:
Hmm... Just noticed that the dependency conflict check is failing with a "No conflicts detected" assertion (see https://travis-ci.org/easybuilders/easybuild-easyconfigs/builds/530758391). This seems to be a logic error...

boegel:
@geimer No, the assertion is that there are no conflicts, and that is failing (because there are conflicts).

It's a bit confusing, but the other way around could be confusing too (we use this same pattern consistently in the test suite)

boegel:
Test report by @boegel
**SUCCESS**
node2079.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/c8c87457c0b208ddd29f667a481fc367 for a full test report.

boegel:
Thanks a lot for your persistence on this one @geimer!

boegel:
Going in, thanks @geimer!

PR #8299, opened 2019-05-12
fix Python classifiers in setup.py, should be (only) Python 2.6 & 2.7

boegel:


setup.py
```diff
@@ -113,14 +113,15 @@ def get_data_files():
...
-                   "Environment :: Console",
-                   "Intended Audience :: System Administrators",
-                   "License :: OSI Approved :: GNU General Public License v2 (GPLv2)",
-                   "Operating System :: POSIX :: Linux",
-                   "Programming Language :: Python :: 2.4",
-                   "Topic :: Software Development :: Build Tools",
-                  ],
+        "Development Status :: 5 - Production/Stable",
+        "Environment :: Console",
+        "Intended Audience :: System Administrators",
```
akesandgren: Doesn't match the change you did in https://github.com/easybuilders/easybuild/pull/507

Intentional or?

(Same goes for the one in framework which I failed to notice)

PR #8403, opened 2019-05-29
{devel,lang}{dummy/}ReFrame-HPC v2.18, Python-bare v3.6.8

rsarm:


boegel:
@rsarm Making `ReFrame` depend on a `Python-bare` is going to limit its usefulness quite a bit I think.

Wouldn't it make more sense to specify an OS dependency on `python3` instead, much like we do for EasyBuild itself (cfr. https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/e/EasyBuild/EasyBuild-3.9.0.eb#L43)?

vkarak:
@boegel Good point, thank you!

vkarak:
@boegel Is there a way to check in the EB recipe if a specific OS version of python is available?

vkarak:
@boegel You may close this PR cos it's replaced by #8481.

PR #8458, opened 2019-06-07
Enabling shared toolchainopt solves numpy test errors

edmondac:
Perhaps a better solution to fixing the numpy tests than https://github.com/easybuilders/easybuild-easyblocks/pull/1743


edmondac:
See also https://github.com/easybuilders/easybuild-easyblocks/pull/1744

edmondac:
It seems that this *doesn't work* for a PythonBundle (like SciPy-bundle) although it would work for a straight numpy easyconfig.

It looks like https://github.com/easybuilders/easybuild-easyblocks/pull/1743 might be better after all!

edmondac:
@ocaisa it looks like the 'shared' isn't passed through from toolchainopts for SciPy-bundle...

edmondac:
Closing in favour of easybuilders/easybuild-easyblocks#1744

PR #8462, opened 2019-06-07
Boost v1.66.0 for GCCcore/6.4.0 {REVIEW}

SimonPinches:
By building without Python libraries and `boost_mpi`, `Boost` can be compiled against `GCCcore/6.4.0` which is useful for building other software that works with both `intel/2018a` and `foss/2018a` toolchains.  This limits the number of modules that have to be installed and increases flexibility.

### DIFF
--- a/easybuild/easyconfigs/b/Boost/Boost-1.66.0-GCCcore-6.4.0-serial.eb
+++ b/easybuild/easyconfigs/b/Boost/Boost-1.66.0-GCCcore-6.4.0-serial.eb
@@ -1,5 +1,6 @@
 name = 'Boost'
 version = '1.66.0'
+versionsuffix = '-serial'
 
 homepage = 'http://www.boost.org/'
 description = """Boost provides free peer-reviewed portable C++ source libraries."""
--- a/easybuild/easyconfigs/b/Boost/Boost-1.66.0-GCCcore-6.4.0-no_mpi.eb
+++ b/easybuild/easyconfigs/b/Boost/Boost-1.66.0-GCCcore-6.4.0-no_mpi.eb
@@ -1,6 +1,6 @@
 name = 'Boost'
 version = '1.66.0'
-versionsuffix = '-serial'
+versionsuffix = '-no_mpi'
 
 homepage = 'http://www.boost.org/'
 description = """Boost provides free peer-reviewed portable C++ source libraries."""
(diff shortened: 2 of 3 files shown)


boegel:
@SimonPinches Your reasoning makes sense, but merging this will cause problems for people that have `--minimal-toolchains` enabled, see https://easybuild.readthedocs.io/en/latest/Manipulating_dependencies.html#using-minimal-toolchains-for-dependencies .

If both `Boost-1.66.0-foss-2018a.eb` and `Boost-1.66.0-GCCcore-6.4.0.eb` around, then the latter will be used when `--minimal-toolchains` is enabled whenever `Boost` 1.66.0 needs to be resolved as dependency.

And since that installation is "crippled" (no Python bindings, but mainly no Boost MPI), that is likely to lead to problems...

One way around that could be to use `versionsuffix = '-serial'` (or something like that) here, but that may limit the usefulness of this with your use case in mind?

SimonPinches:
This all sounds reasonable to me so I'll adopt your suggested suffix and push an update.

easybuild/easyconfigs/b/Boost/Boost-1.66.0-GCCcore-6.4.0-serial.eb
```diff
@@ -0,0 +1,29 @@
+name = 'Boost'
+version = '1.66.0'
+versionsuffix = '-serial'
```
smoors: this Boost can still work in parallel with threading, right?
in that case the `-serial` suffix is misleading. maybe use `-no_mpi` instead?

smoors (changes requested):

easybuild/easyconfigs/b/Boost/Boost-1.66.0-GCCcore-6.4.0-serial.eb
SimonPinches: @smoors, another good suggestion so I'll push an update ASAP.

smoors:
Test report by @smoors
**SUCCESS**
nic426 - Linux centos linux 7.5.1804, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/00ed6e59c215282229008f845e854b34 for a full test report.

smoors (approved):
lgtm

smoors:

PR #8475, opened 2019-06-11
{chem}[intel/2018b] ABINIT v8.10.2

jmbeuken:
use libXC 3.0.1

### DIFF
--- a/easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
+++ b/easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
@@ -18,8 +18,9 @@ checksums = ['4ee2e0329497bf16a9b2719fe0536cc50c5d5a07c65e18edaf15ba02251cbb73']
 
 dependencies = [
     ('libxc', '3.0.1'),
-    ('netCDF', '4.6.1','HDF5-1.10.2'),
-    ('netCDF-Fortran', '4.4.4','HDF5-1.10.2'),
+    ('netCDF', '4.6.1'),
+    ('netCDF-Fortran', '4.4.4'),
+    ('HDF5','1.10.2'),
 ]
 
 # ensure mpi and intel toolchain
--- a/easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
+++ b/easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
@@ -1,0 +1,45 @@
+easyblock = 'ConfigureMake'
+
+name = 'ABINIT'
+version = '8.10.2'
+
+homepage = 'https://www.abinit.org/'
+description = """ABINIT is a package whose main program allows one to find the total energy,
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
```diff
@@ -0,0 +1,47 @@
...
+dependencies = [
+    ('libxc', '3.0.1'),
+    ('netCDF', '4.6.1'),
+    ('netCDF-Fortran', '4.4.4'),
+    ('HDF5', '1.10.2'),
+]
+
+# ensure mpi and intel toolchain
+configopts = '--enable-mpi '
+configopts += '--enable-gw-dpc '
```
smoors: @jmbeuken is double precision really needed for GW calculations? note that it is marked 'EXPERIMENTAL'.

easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
```diff
@@ -0,0 +1,47 @@
...
+configopts += '--with-linalg-flavor=mkl --with-fft-flavor=dfti --with-fft-libs="$LIBFFT" '
```
smoors: @jmbeuken is there any performance benefit with `dfti` vs. `fftw3-mkl`?

easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
gmatteo: The documentation is outdated.  Double precision is ok in GW
Actually, this is the option we use in our testfarm for testing. 
The only drawback is that GW in double precision is slower and more memory demanding...

easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
gmatteo: These two options are equivalent when the code is linked with MKL as  fftw3-mkl is just wrapping  the native MKL FFT library (DFTI) with an API that is compatible with FFTW3.
fft-flavor DFTI calls the MKL library directly so it's supposed to be slightly faster.

Note, however, that fftw3-mkl should not be used if the threaded FFT algorithm is wanted as  
the wrappers are not thread-safe.

smoors:
Test report by @smoors
**SUCCESS**
nic426 - Linux centos linux 7.5.1804, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/2b9ecda7d1e1387016729c4b0df2eb11 for a full test report.

easybuild/easyconfigs/a/ABINIT/ABINIT-8.10.2-intel-2018b.eb
smoors: @jmbeuken can you add a comment that this option is for enabling double precision?

smoors:
Going in, thanks @jmbeuken!

PR #8481, opened 2019-06-12
{devel}{dummy/} ReFrame-HPC v2.18

vkarak:
Easyconfig file for ReFrame.

Replaces #8403.


easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
```diff
@@ -0,0 +1,29 @@
+easyblock = 'PythonPackage'
+
+req_py_majver = 3
+req_py_minver = 5
+
+name = 'ReFrame-HPC'
```
boegel: @vkarak Why not just `ReFrame`, which is the actual project name?

easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
```diff
@@ -0,0 +1,29 @@
+easyblock = 'PythonPackage'
+
+req_py_majver = 3
+req_py_minver = 5
```
boegel: Please move this down below the `checksums =` line, above the `options =` line (that's just a matter of style, order doesn't matter w.r.t. functionality), and add a comment to clarify:

```python
# ReFrame requires Python 3.5, may only be available via 'python3'

File: easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb [hunk 3]
```diff
@@ -0,0 +1,29 @@
...
+
+options = {'modulename': 'reframe'}
+
+sanity_check_paths = {
+    'files': ['bin/reframe'],
+    'dirs': []
+}
+sanity_check_commands = [
+    ('reframe', '-V'),
+    ('python3', "-c 'from reframe import RegressionTest'")
```
boegel: You can simply these by just making them regular strings (the tuple format is old and may get deprecated at some point):

```python
sanity_check_commands = [
    "reframe -V",
    "python3 -c 'from reframe import RegressionTest'",
]
```

Rather than having the second one, why not use `options = {'modulename': 'reframe.RegressionTest'}` above?

easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
```diff
@@ -0,0 +1,29 @@
...
+]
```
boegel: missing empty line below (again just a minor style issue)

easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
```diff
@@ -0,0 +1,29 @@
...
+homepage = 'https://github.com/eth-cscs/reframe'
+description = """ReFrame is a framework for writing regression tests for HPC systems."""
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+source_urls = [PYPI_SOURCE]
+sources = [SOURCE_TAR_GZ]
+checksums = ['ba6527f9036d7028d6af8d69e8e74603da239abd056e3b0cebac7ce8290a6e31']
+
+
```
boegel: minor style issue, avoid duplicate empty lines

boegel:
@vkarak Please also keep in mind to target the `develop` branch (I've changed this for you just now for this PR).

The issues with Travis & Python 2.6 are fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/8483

easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
vkarak: What does the `options` variable do actually? I was wondering.

easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
vkarak: The problem is that we were forced to use `reframe-hpc` as a name for the PyPI, cos there is another `reframe` project, completely unrelated... If there is a way to name to the easyconfig file and the resulting module `reframe`, it would be great for me, but I couldn't find a way to do this.

boegel: Well, you can, but you just can't use the `PYPI_SOURCE` template.

Here's what I would do:

```python
name = 'ReFrame'
...
source_urls = ['https://pypi.python.org/packages/source/R/ReFrame-HPC']
```

Shouldn't need to change anything else.

(see also output of `eb -a --avail-easyconfig-templates | grep PYPI_SOURCE`)

easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
boegel: It's a way to specify values for parameters for "extensions" (like Python packages are).

We should have a more direct way, since many "extensions" are now installed stand-alone as well (like here).

easybuild/easyconfigs/r/ReFrame-HPC/ReFrame-HPC-2.18.eb
vkarak: @boegel Coo, this is what I needed!

boegel:
(close/re-open to make Travis CI happy w.r.t. Python 2.6)

easybuild/easyconfigs/r/ReFrame/ReFrame-2.18.eb
```diff
@@ -0,0 +1,26 @@
...
+name = 'ReFrame'
+version = '2.18'
+
+homepage = 'https://github.com/eth-cscs/reframe'
+description = '''ReFrame is a framework for writing regression tests for HPC systems.'''
+
+toolchain = {'name': 'dummy', 'version': ''}
+
+source_urls = ['https://pypi.python.org/packages/source/R/ReFrame-HPC']
+sources = ['ReFrame-HPC-%s.tar.gz' % version]
```
vkarak: @boegel I tried to use `%(version)s` here but didn't work, so I reverted to the standard way.

boegel: There's absolutely no reason why it shouldn't work... How did it fail exactly?

```python
sources = ['ReFrame-HPC-%(version)s.tar.gz']

**Comment by boegel** [comment]
@vkarak You should add the following too in the easyconfig:

```python
download_dep_fail = True
use_pip = False
```

Enabling `download_dep_fail` will make `eb` check whether any dependencies were being auto-downloaded from PyPI (which is bad w.r.t. reproducibility).

I wouldn't use `pip` to install `ReFrame` since that'll only introduce another OS dependency (and you may run into problems because of the clash between `pip` and `python3`), but you must specify something to make the contribution tests happy.

easybuild/easyconfigs/r/ReFrame/ReFrame-2.18.eb
vkarak: It works. I don't know what I've done wrong before. I fixed that.

vkarak:
> I wouldn't use pip to install ReFrame since that'll only introduce another OS dependency (and you may run into problems because of the clash between pip and python3), but you must specify something to make the contribution tests happy.

@boegel As far as I can understand EB doesn't use `pip` to install this kind of `PythonPackage`, right?

boegel:
Not currently by default, but it will if you set “use_pip True”

vkarak:
@boegel Can you re-review this one? CI now passes.

boegel:
Test report by @boegel
**SUCCESS**
boegels-MacBook-Pro.local - Darwin UNKNOWN UNKNOWN, Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz, Python 2.7.10
See https://gist.github.com/94ff5a4e3720c8eacebda1b2ec31c7fd for a full test report.

boegel:
Test report by @boegel
**FAILED**
node3115.skitty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/5c8898df95871f196bd6ec30d56355c6 for a full test report.

boegel:
Test report by @boegel
**FAILED**
node2013.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/95a1a8c432fc6b93adc9963a2c6c10ad for a full test report.

boegel:
Test report by @boegel
**FAILED**
node2403.golett.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/c956e8ca79042683070aa1c88fd6cfe0 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3139.skitty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/1989dff95bba2fa3e82680153f6a9321 for a full test report.

boegel:
@vanzod Can you submit a test report from a Debian system that has `python3`?

I don't expect any problems, but doesn't hurt.

vanzod:
Test report by @vanzod
**FAILED**
cermis - Linux debian 9.9, Intel(R) Xeon(R) CPU           X5670  @ 2.93GHz, Python 2.7.13
See https://gist.github.com/d12089fc21682ffa54413ed0eaa9ebbf for a full test report.

vanzod:
@vkarak @boegel `setuptools` is required to install `ReFrame`. We could put is as an OS dependency but it would be of little help aside from probably terminating the build a little more gracefully.

boegel:
Or we can consider adding `setuptools` as a build dependency here...

Thoughts @vkarak, @victorusu?

vkarak:
@boegel I think it's easier to go with an OS dependency to `python-setuptools`. We will update the recipe. If `setuptools` was already in the `requirements.txt` of ReFrame, I guess this wouldn't be needed, right? We will update this in the next release of ReFrame.

boegel:
@vkarak Then you'll also need to provide `setuptools` via EasyBuild as a build dep.

That's doable, but it's messy when relying on the system Python...

Adding `setuptools` as an OS dep is a bit messy too though, since it's the `setuptools` on top of Python 3 you're after, and the package name may be different depending on the OS (CentOS vs Debian vs ...)

vkarak:
@boegel What if I use this as build dep? https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/s/setuptools/setuptools-1.4.2.eb

boegel:
@vkarak Using `setuptools` as a build dep could work, but you should definitely use a more recent version of `setuptools`, since 1.4.2 is pretty ancient...

vkarak:
@boegel Yeah, sure. I'll do that.

vkarak:
@boegel @vanzod I've just pushed an updated recipe for ReFrame that adds `setuptools` as a build dependency. Since I'm requiring Python 3 for ReFrame, I'm doing the same for `setuptools` as well. Can you have a look?

easybuild/easyconfigs/s/setuptools/setuptools-41.0.1.eb
```diff
@@ -0,0 +1,25 @@
...
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+source_urls = [PYPI_SOURCE]
+sources = [SOURCE_ZIP]
+checksums = ['a222d126f5471598053c9a77f4b5d4f26eaa1f150ad6e01dcf1a42e185d05613']
+
+download_dep_fail = True
+use_pip = False
+
+req_py_majver = 3
```
boegel: Please add a comment to clarify why you do this (it makes sense, but it'll look weird when looking at this outside of the ReFrame context).

It also makes sense to use a versionsuffix like `-py3` here imho...

easybuild/easyconfigs/s/setuptools/setuptools-41.0.1.eb
```diff
@@ -0,0 +1,25 @@
...
+description = """Easily download, build, install, upgrade, and uninstall Python packages"""
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+source_urls = [PYPI_SOURCE]
+sources = [SOURCE_ZIP]
+checksums = ['a222d126f5471598053c9a77f4b5d4f26eaa1f150ad6e01dcf1a42e185d05613']
+
+download_dep_fail = True
+use_pip = False
```
boegel: I'm not sure installing `setuptools` without `pip` is actually still supported (but setting it to `True` may cause problems too, since we don't have any logic in place to favor `pip3` over `pip`)...

easybuild/easyconfigs/r/ReFrame/ReFrame-2.18.eb
```diff
@@ -0,0 +1,31 @@
...
+
+name = 'ReFrame'
+version = '2.18'
+
+homepage = 'https://github.com/eth-cscs/reframe'
+description = '''ReFrame is a framework for writing regression tests for HPC systems.'''
+
+toolchain = {'name': 'dummy', 'version': 'dummy'}
+
+builddependencies = [('setuptools', '41.0.1')]
```
boegel: @vkarak Please move this down, below `checksums` (style nitpicking)

easybuild/easyconfigs/s/setuptools/setuptools-41.0.1.eb
vkarak: > It also makes sense to use a versionsuffix like -py3 here imho...

I had the same feeling, but I was expecting your feedback. :-)

easybuild/easyconfigs/s/setuptools/setuptools-41.0.1.eb
vkarak: Actually, I changed also the EasyBlock used to `PythonPackage` and set those after your recommendation for the ReFrame easyconfig, cos otherwise the CI was failing.

vkarak:
@boegel Could you have another look? @vanzod Standard CI passes now, could you check if the recipe works in your case as well?

boegel:
Test report by @boegel
**SUCCESS**
boegel.ugent.be - Darwin UNKNOWN UNKNOWN, Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz, Python 2.7.10
See https://gist.github.com/646a74e3146cfc4fda82f8ea338cc78d for a full test report.

vanzod:
Test report by @vanzod
**FAILED**
Build succeeded for 0 out of 2 (2 easyconfigs in this PR)
cermis - Linux debian 9.9, Intel(R) Xeon(R) CPU           X5670  @ 2.93GHz, Python 2.7.13
See https://gist.github.com/39d1dfd24a474b3965b84b763b17fe4e for a full test report.

boegel:
@vanzod Remember, you need `python3` (and accompanying `setuptools`) installed in the OS for this one...

boegel:
Test report by @boegel
**SUCCESS**
node2081.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/5ec8c74bda169ad9cefdeab30c631e9d for a full test report.

vanzod:
Test report by @vanzod
**FAILED**
cermis - Linux debian 9.9, Intel(R) Xeon(R) CPU           X5670  @ 2.93GHz, Python 2.7.13
See https://gist.github.com/60cfe865975c973a9eb5aa8c6b7ffd71 for a full test report.

vanzod:
Test report by @vanzod
**FAILED**
cermis - Linux debian 9.9, Intel(R) Xeon(R) CPU           X5670  @ 2.93GHz, Python 2.7.13
See https://gist.github.com/63d9728f76c9ffe215e9ceedc8d5f57e for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
boegel.ugent.be - Darwin UNKNOWN UNKNOWN, Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz, Python 2.7.10
See https://gist.github.com/95e67ed09e9575e32b82548eace6adba for a full test report.

boegel:
@vanzod's last test report fails with `ImportError: No module named 'pkg_resources'` because `setuptools` for Python 3.x is not installed on his system.

This shows that `setuptools` (which provides `pkg_resources`) is actually a *runtime* dependency of ReFrame, which complicates things a bit...

We could just promote `setuptools` to a regular dep in the ReFrame easyconfig, but that is not ideal (since you'll always have a `setuptools` module loaded).

@vkarak @victorusu The best road forward is to make the `ReFrame` easyconfig a `PythonBundle` with both `setuptools` and `ReFrame` listed in `exts_list`, imho...

boegel:
When trying to use `PythonBundle` myself, I first ran into the requirement that `PythonBundle` has currently where it expects `Python` to be listed as a (proper) dependency.

That can be fixed quite easily, but then I run into this:

```
pkg_resources.DistributionNotFound: The 'ReFrame-HPC==2.18' distribution was not found and is required by the application
```

That problem does not pop up when simply promoting `setuptools` to a runtime dependency...

vanzod:
@boegel I have tested the installation by adding `setuptools` as build dependency and it completes correctly.

boegel:
Test report by @boegel
**SUCCESS**
boegel.ugent.be - Darwin UNKNOWN UNKNOWN, Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz, Python 2.7.10
See https://gist.github.com/e373328f0e8cbbfe67dc3257413097d8 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2045.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/b3986f9ac967588720926b545c8e0318 for a full test report.

boegel:
Promoted `setuptools` to a runtime dep, after discussing this with @vkarak, so we can get this merged... Hopefully we can change this to use `PythonBundle` later and avoid the `setuptools` dep.

boegel:
Going in, thanks @vkarak!

vkarak:
Thanks @boegel for merging!

PR #8546, opened 2019-07-07
{lang}[foss/2019a] Update nodejs to version 10.15.3 and build libnode and libv8 shared libs

lexming:
The easyconfigs of `nodejs-10.15.1-foss-2018b` and `nodejs-10.15.3-foss-2019a` are modified to also build the `libnode` shared libs and create fake `libv8.so`, `libv8_libbase.so` and `libv8_libplatform.so` linked to `libnode`. This is necessary to be able to use nodejs for packages requiring the V8 engine.


migueldiascosta:
**SUCCESS**
sms - Linux centos linux 7.6.1810, AMD EPYC 7601 32-Core Processor, Python 2.7.5
See https://gist.github.com/c7d04204e9174565862d663a87837e89 for a full test report.

lexming:
Had en error with another package not yet in develop. Node.js has a small partial implementation of ICU that can raise errors on anything relying on a full ICU implementation. This update fully disables ICU in Node.js. If any package needs ICU it should be added as a dependency of that package.

migueldiascosta:
**SUCCESS**
sms - Linux centos linux 7.6.1810, AMD EPYC 7601 32-Core Processor, Python 2.7.5
See https://gist.github.com/ae8643e419718e1a18329ba42f423590 for a full test report.

migueldiascosta:
@lexming maybe we should take the opportunity to move this (well, the new easyconfig) down to the `GCCcore` level, since it doesn't depend on `MPI` or linear algebra?

lexming:
Good idea, I'll do that

lexming:
Done :+1: 
I also removed `toolchainopts = {'lowopt': True}` as it seems to be inherited from very old versions of Node.js. It builds and runs fine on my side with regular optimizations.

migueldiascosta:
@lexming I meant moving only the new easyconfig to `GCCcore`, as renaming the existing one is a regression

in fact, unless you really need these enhancements in the existing easyconfig, it would be safer to only add the new one and leave the existing one unchanged (?)

lexming:
I checked and there is nothing in the tree that depends on nodejs-10.15.1, so I think that it is very safe to apply this changes to it as well. I'm also currently working on a new package that depends on nodejs-10.15.1 and it needs these changes.

migueldiascosta:
@lexming regarding the enhancements on 10.15.1, that's ok, just checking if they were really necessary

regarding moving an existing config to `GCCcore`, I'm not sure... although nothing depends on it in the easybuild tree, in principle sites could depend on it (but then again, it's `nodejs`, not exactly a cornerstone of an HPC stack :p )

Can I have a second opinion from this week's MOTWs, @smoors and @ocaisa?

lexming:
Ok :+1: I added again `nodejs-10.15.1-foss-2018b.eb` including the same build changes introduced by its `GCCcore` counterpart.

migueldiascosta:
Going in, thanks @lexming!

PR #8560, opened 2019-07-09
Remove ciftify for 2017b toolchains

vanzod:
The current `ciftify` easyconfigs in the repository are lacking some required dependencies, including `FSL`. However `FSL` brings in software built against `Python/2.7.14`, while `ciftify` only supports `Python/3.x`.
It is then easier to remove `ciftify` from the 2017b toolchains and add them in the 2019a toolchains where `Python` dependencies are correctly managed through `multi_deps`.


boegel:
This is essentially reverting.

@vanzod Is this beyond fixing? I strongly prefer not removing easyconfigs, if we can avoid it...

vanzod:
@boegel Yes, but as is the `ciftify` easyconfig is useless since it lacks `FSL` as dependency and it cannot be added for the reason I explained before.

PR #8562, opened 2019-07-10
Update to Salmon 0.14.1

robertamezquita:


migueldiascosta:
@robertamezquita in order not to break dependencies and for reproducibility, we don't update/rename existing easyconfig files, we create new easyconfig files for the new versions

so in this case, please re-add the original `Salmon-0.12.0-foss-2018b.eb` file.

for future contributions, you may also want to have a look at https://easybuild.readthedocs.io/en/latest/Integration_with_GitHub.html, namely `--new-pr` (and `--update-pr` and `--upload-test-report`

Issue #5183, opened 2017-09-28
OpenFOAM-4.x-cleanup.patch doesn't set WM_CC (etc)


JackPerdue:
Changes like this break OpenFOAM 4 for users since $CC isn't set by any compiler module:

-            export WM_CC='gcc'
+            export WM_CC=$CC

e.g.

```
$ ml purge
$ ml OpenFOAM/4.1-intel-2017A-Python-2.7.12
$ source $FOAM_BASH
$ echo $WM_CC
```

nothing.... WM_CC isn't set because CC wasn't set when the user sourced $FOAM_BASH.

If of course works during building becase EB sets CC.   But EB does not put CC in any compiler modules.  This basically means users can't use wmake.

The proper solution (besides adding CC/CXX/FC to the compiler modules) it to replace 

export WM_CC=$(CC)

with:

export WM_CC=gcc 
or
export WM_CC=icc

Same for other compiler vars.

boegel:
Hmm, this is supposed to be fixed with https://github.com/easybuilders/easybuild-easyblocks/pull/692... (see also https://github.com/easybuilders/easybuild-easyblocks/blob/master/easybuild/easyblocks/o/openfoam.py#L168)

Which EasyBuild version are you using here @JackPerdue? Are you using the stock OpenFOAM easyblock, or a customised one?

JackPerdue:
Currently 3.4.0  stock block, though I see my last my last build of FOAM was on 3.3.0 in late June.  I'll try again with 3.4.0.

boegel:
I don't think that will make a difference, the PR that is supposed to fix this has been included ever since EasyBuild v2.4.0.

I just took a quick look, it seems like the compiler commands are still hardcoded:

```
$ diff -u OpenFOAM/4.1-intel-2017a/OpenFOAM-4.1/wmake/rules/linuxIcc/c++.orig.eb OpenFOAM/4.1-intel-2017a/OpenFOAM-4.1/wmake/rules/linuxIcc/c++
--- OpenFOAM/4.1-intel-2017a/OpenFOAM-4.1/wmake/rules/linuxIcc/c++.orig.eb      2016-10-16 16:11:45.000000000 +0200
+++ OpenFOAM/4.1-intel-2017a/OpenFOAM-4.1/wmake/rules/linuxIcc/c++      2017-09-26 02:02:43.826343000 +0200
@@ -6,7 +6,7 @@
[...]

PR #8584, opened 2019-07-16
{cae}[foss/2018b] OpenFOAM v7

mboisson:


mboisson:
This build requires https://github.com/easybuilders/easybuild-easyblocks/pull/1773

JackPerdue:
First off, thanks for the easyconfig and easyblock mods.  We have a customer here wanting 7 and this was a good start.  

Alas, I found some problems.

1) The cleanup patch still suffers from the problem discussed in https://github.com/easybuilders/easybuild-easyconfigs/issues/5183    Some users want to load OF, run 'source $FOAM_BASH', and then be able to use $WM_CC to compile their OF module.  This fails because $CC is not set when OF is loaded.  It's an old problem, but really needs to be dealt with.   Currently testing a build without that part of the patch for etc/config.sh/settings.  My workaround in the past has been: 
prebuildopts = """
    sed -e "s/WM_CC=\$CC/WM_CC='$CC'/" -i.eb.cc etc/config.sh/settings
    sed -e "s/WM_CXX=\$CXX/WM_CXX='$CXX'/" -i.eb.cxx etc/config.sh/settings
    sed -e "s/WM_CFLAGS=\$CFLAGS/WM_CFLAGS='$CFLAGS'/" -i.eb.cflags etc/config.sh/settings
    sed -e "s/WM_CXXFLAGS=\$CXXFLAGS/WM_CXXFLAGS='$CXXFLAGS'/" -i.eb.cxxflags etc/config.sh/settings
    sed -e "s:WM_LDFLAGS=\$LDFLAGS:WM_LDFLAGS='$LDFLAGS':" -i.eb.ldflags etc/config.sh/settings
"""
2) for some reason on RHEL7, I need to add:
prebuildopts += """
    sed -e 's,src/Pstream,general/src/Pstream,' -i.eb src/Pstream/Allwmake  # not sure why needed
    sed -e 's,src/parallel,general/src/parallel,' -i.eb src/parallel/decompose/Allwmake # ditto
"""
3) Not having any luck on CentOS 6 yet.  Unlike CentOS7, it doesn't seem to have a platforms/linux64GccDPInt32OptEASYBUILDMPI directory, still investigating that

In any case, all I really care about is 1) at this point.  It's a known problem to EB and should be resolved by now, one way or another (personally, I'd set CC the C compiler module is loaded but that's just me).

mboisson:
@JackPerdue, @boegel, for OpenFOAM, at Compute Canada, we include the action of OpenFOAM's bashrc into the module, as is listed here : 
https://github.com/ComputeCanada/easybuild-easyconfigs/blob/5fa14e050690ab4aafc6526861f596c30997aa16/easybuild/easyconfigs/o/OpenFOAM/OpenFOAM-7-gompi-2018.3.312.eb

Should EB upstream start doing that ? This avoids requiring users to source a shell script when needing OpenFOAM.

JackPerdue:
setenv("WM_CC", "")

How does WM_CC get set???

mboisson:
Mmmm, you may have found a bug, but a quick look reveales that OpenFOAM's bashrc does not set `WM_CC` either. I should probably review these environment variables.

mboisson:
Yep, you found a bug in our recipes. I wonder where it got lost... I had it in my old installations, and somehow, it got lost when I transitioned my old installations to EB recipes.

JackPerdue:
[OpenFOAM-7-intel-2018b.eb.txt](https://github.com/easybuilders/easybuild-easyconfigs/files/3513400/OpenFOAM-7-intel-2018b.eb.txt)

That works here on RHEL7.  6 is giving me problems.

akesandgren:
I'm all for adding the stuff from CC into the easyconfig file for OF. And I'd really like to get this one finished asap...

akesandgren:
And if the function of FOAM_BASH is incorporated into the module, FOAM_BASH should be replaced by an empty file, so people that use it won't destroy their environment. Getting people to change their behavior is difficult.

mboisson:
I am thinking that if we go this route, it might be best to set these in the easyblock rather than the easyconfig. We did it in the easyconfig because we could have our own customized one, but there is no reason why this information should need to be copied in every easyconfig file.

boegel:
One reason to *not* replace `$FOAM_BASH` with static stuff in the module file is that it's... static.

Doesn't `$FOAM_BASH` pick up on environment variables that may be defined by the user before `$FOAM_BASH` is sourced?
In theory it's possible to mimic that in a module file, but in practice it's a maintenance nightmare...

mboisson:
Mmm, like what ?   
  

  
  

  
  
>

jhein32:
HI, 

I have two users who need to use OpenFOAM 7.  Can we try to get this finalised?  Anything I could help with?

Joachim

mboisson:
@jhein32 
Did you try it as is ? I unfortunately don't have the specificed toolchain installed, so I cannot test it. I did install OpenFOAM 7 on our system with a different toolchain though.

jhein32:
@mboisson Hi, thanks for asking back.  I have build it on our system and am about to engage with the users, for them to test whether or not it works. 

Which toolchain are using for building this?

mboisson:
@jhein32 Compute Canada does not use foss or intel. We always use OpenMPI (even with Intel), and always use MKL (even with GCC). We also heavily push things toward minimal toolchains, so we don't use full-featured Compiler+MPI+MKL toolchains for many software, we use subtoolchains. Finally, we have many dependencies that get ignored because they are not install as modules, but rather in the base compatibility layer that we provide to all clusters.

jhein32:
Hi,
I heard back from my user.  They ran a few tutorials and the installation appears to be working.

Best wishes
   Joachim

boegel:
closing this, since foss/2018b is now a deprecated toolchain (see https://docs.easybuild.io/en/latest/Deprecated-easyconfigs.html#deprecated-toolchains)

Sorry for never getting back to this @mboisson :(

PR #8600, opened 2019-07-22
Updated version of Mesa3D

furstj:
(created using `eb --new-pr`)


vanzod:
@furstj Can you please justify this change? We tend not change the version of dependencies of easyconfig already released since this could break the dependencies tree of people that have already used such easyconfig files. It would need to be an extremely important bug fix for us to accept this PR.

furstj:
Mesa 19.0.1 does not export some OpenGL symbols and it breaks e.g. the compilation of ParaView (I'm currently working on easyconfig for ParaView 5.6.1). The problem has been probably resolved in Mesa 19.0.5 (see https://www.mesa3d.org/relnotes/19.0.5.html). 

I'm aware of big changes due to upgrade of Mesa and I will be happy if we can find a smooth solution (e.g. postpone the update to foss-2019b or to next release of EB).

Micket:
I hate to be a downer, but changing versions of existing easyconfigs that's been part of a EB release is not a good idea, it would wreak havoc for most existing module trees. We avoid multiple versions of software within the same toolchain, and that is extra important for core deps like Mesa.

So, I can't really suggest any changes here, other than; lets not do this.
The 2019b toolchains are coming though, and it seems they will be based on GCCcore 8.3.0, so, PR's for populating this tree with all up-to-date deps would be highly appreciated though.

furstj:
@Micket , I agree that we should rather wait for 2019b toolchains with new mesa. Should I close this PR? How can I create new PR which depends on foss-2019b? Should I wait until foss-2019b will be merged?

Micket:
> How can I create new PR which depends on foss-2019b? Should I wait until foss-2019b will be merged?

A whole lot of configs will need to be created for the next GCCcore level before we get to Mesa.
Python, X11, etc.

We should only wait for 2019b, because we need to know what version of GCCcore to target. @vanzod mentioned it was going to be discussed on tomorrows conference call.

furstj:
I will wait then ...

PR #8651, opened 2019-07-31
{lib}[dummy/dummy] SeqAn3 v3.0.0

cmeesters:
(created using `eb --new-pr`)


cmeesters:
I erased the gcc dependency, 'cause it is a quasi dependency, but not a strict one.

easybuild/easyconfigs/s/seqan/seqan-3.0.0.eb
```diff
@@ -0,0 +1,41 @@
+easyblock = "Tarball"
+
+name = 'seqan'
```
Micket: I think it should be called `SeqAn3`.
Using the lower case version would cause conflicts with the existing `SeqAn` configs.

easybuild/easyconfigs/s/seqan/seqan-3.0.0.eb
```diff
@@ -0,0 +1,41 @@
...
+
+files_to_copy = ['include',
+                 'submodules',
+                 'build_system']
+
+sanity_check_paths = {
+    'files': [],
+    'dirs':  ['include', 'submodules']}
+
+dependencies = [('CMake', '3.9.6')]
```
Micket: Should be a build dep

cmeesters:
@Micket 

Thank you. You are right.

As for the naming: I did my own distinction between apps (CamelCase) and the lib (lower case) but that does not apply to eb.

As for the dependency: The SeqAn developers state gcc >7 and cmake as dependencies, yet these are dependencies to work *with* the (header only) library and not to install it. Please do not confuse the library with the SeqAn-2.4.0-xxx-eb configs in the 'bio' moduleclass - these install applications.

It is my mistake having included dependencies where non are.

Micket:
@cmeesters You forgot to remove the old eb config from the PR.

Sometimes header-only libraries still actually use CMake just to install itself. That's what I thought was going on here (and, in which case, you could use a "build" dep for this). Is it not?

cmeesters:
@Micket 

No, it's just a copy-step.

How can one remove an old file within this thread?

boegel:
@cmeesters You can prefix the filename to remove with a colon `:` when using `eb --update-pr`, see also https://easybuild.readthedocs.io/en/latest/Integration_with_GitHub.html#deleting-easyconfig-files-or-patches

easybuild/easyconfigs/s/SeqAn3/SeqAn3-3.0.0.eb
```diff
@@ -0,0 +1,39 @@
...
+
+sanity_check_paths = {
+    'files': [],
+    'dirs':  ['include', 'submodules']}
+
+modextrapaths = {
+    'INCLUDEPATH': 'include/',
+    'INCLUDEPATH': 'submodules/range-v3',
+    'INCLUDEPATH': 'submodules/sdsl-lite',
+    'INCLUDEPATH': 'submodules/cereal',
```
boegel: @cmeesters This won't work, you'll only have one of these actually used (and to make it even more fun, it'll be a random one, since keys in Python dictionaries don't have any order...).

You can fix this as follows:

```python
modextrapaths = {
    'INCLUDEPATH': ['include', 'submodules/range-v3', 'submodules/sdsl-lite', 'submodules/cereal'],
}

cmeesters: thanks and outch, outch, outch ...

cmeesters: done

boegel:
Test report by @boegel
**SUCCESS**
node3150.skitty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/c00a38705a1dd77a8f1eae50a95f8614 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2412.golett.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/ddd36f245388c24b57937fba0b099149 for a full test report.

boegel:
Going in, thanks @cmeesters!

Issue #4171, opened 2017-02-21
CMake: ncurses dependency not found during build


geimer:
Building `CMake` on a Debian 8.7 (jessie) box without the `libncurses5-dev` OS package installed resulted in an installation lacking the `ccmake` utility, even though `ncurses/6.0` is listed as a dependency in the easyconfig file.  The only diagnostic output of the `configure_step` regarding this is
```
Curses libraries were not found. Curses GUI for CMake will not be built.
```

I was able to convince `CMake` to find the `ncurses` stuff and to build `ccmake` by adding
```
configopts += '-DCURSES_INCLUDE_PATH=$EBROOTNCURSES/include '
configopts += '-DCURSES_LIBRARY=$EBROOTNCURSES/lib/libcurses.so '
configopts += '-DCURSES_NCURSES_LIBRARY=/lib/libncurses.so'
```
to the `CMake` easyconfig file, though I'm not yet convinced that this is the way to go. (`CMake` should just do its job properly and find the `ncurses` installation!)  Also, even with this "fix", the log file shows some weirdness:
```
-- Looking for cbreak in /lib/libncurses.so
-- Looking for cbreak in /lib/libncurses.so - not found
```

boegel:
@geimer Which CMake version was being used here? Did you try using a different CMake version? Cfr. #4170

geimer:
@boegel I was installing CMake 3.7.2, but 3.7.1 shows the same behavior.  I did not test older versions yet.

geimer:
@boegel It's the same for CMake 3.6.2 and trying with ncurses 5.9 didn't help either...

geimer:
@boegel Maybe this issue is also an incentive to add `ccmake` to the sanity check (once it is resolved)? :wink:

geimer:
Hmm... Alright, it looks like the right way to do it is to set `CMAKE_PREFIX_PATH=$EBROOTNCURSES` (see https://cmake.org/cmake/help/v3.7/variable/CMAKE_PREFIX_PATH.html#variable:CMAKE_PREFIX_PATH).  Then the `FindCurses` command will correctly pick up `ncurses` from the EasyBuild install.  Now I'm wondering whether EB-generated modulefiles for libraries should prepend to this environment variable automatically?  Though it looks to me like CMake is pretty much borked...

boegel:
[...]

PR #8654, opened 2019-08-01
{devel}[ssytem] CMake v3.15.1

cmeesters:
(created using `eb --new-pr`)


verdurin:
Test report by @verdurin
**FAILED**
Build succeeded for 1 out of 2 (1 easyconfigs in this PR)
dell.lan - Linux fedora 30, Intel(R) Core(TM) i7-4600U CPU @ 2.10GHz, Python 2.7.16
See https://gist.github.com/c70d753f1e38ddd83976a0587c6735bf for a full test report.

cmeesters:
@verdurin
The partial log does not contain any build information (the 'scanning' and 'building' parts are not contained. It is therefore not hard to understand *why* ccmake is not present in this particular build and what to do about it.

verdurin:
Test report by @verdurin
**FAILED**
Build succeeded for 1 out of 2 (1 easyconfigs in this PR)
easybuild.novalocal - Linux centos linux 7.6.1810, Intel Xeon Processor (Skylake, IBRS), Python 2.7.5
See https://gist.github.com/93c05e1c34bd811fcbe21ee9ea7afdb9 for a full test report.

verdurin:
@cmeesters full build log at https://gist.github.com/verdurin/50ad75e6f76c4328af194670010a639c

cmeesters:
@verdurin 

Thank you!

I am not sure what to make of it: Your build log does not state "Scanning dependencies of target ccmake" (not valgrind and a few others).

But am I not sure of the cause. The configure differences state that (note: the same bare system compiler gcc 4.8.5!):

Checking for curses support - Failed 

This is only in your log - yet I do not understand the implication, nor do I see curses anywhere stated as a build dependency for cmake. So I did not make any progress. Ideas?

cmeesters:
@verdurin 

Hm, http://cmake.3232098.n2.nabble.com/ccmake-is-not-getting-installed-td7583001.html and https://askubuntu.com/questions/121797/how-do-i-install-ccmake

and particularly

indicate, that indeed curses support is needed. In that case curses should be a build-dependency of the easyconfig. It was not in my template, but ncurses is in our filter-deps and installed on the system. This, apparently is a problem.

I will test this tomorrow and report back, potentially with an updated pr.

Thanks again for pointing this out!

https://github.com/easybuilders/easybuild-easyconfigs/issues/4171

pescobar:
Test report by @pescobar
**FAILED**
Build succeeded for 1 out of 2 (1 easyconfigs in this PR)
worker01.cluster.bc2.ch - Linux centos linux 7.5.1804, Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz, Python 2.7.5
See https://gist.github.com/6ed802441e7804de4b11360645e10be2 for a full test report.

pescobar:
my build failed in the sanity_check step

```
: build failed (first 300 chars): Sanity check failed: no file found at 'bin/ccmake'
```

cmeesters:
@pescobar 

Yes, we are at that point - I still did not get a conclusive answer and probably only will find the time to dive into this tomorrow. Any ideas for the underlying reason are highly appreciated. (the curses-idea seems a probable cause).

boegel:
@cmeesters `ncurses` should indeed be included as a (build) dependency in order for `ccmake` to get built, see also other `CMake` easyconfigs that use the `dummy` toolchain, for example https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/c/CMake/CMake-3.12.1.eb

cmeesters:
As you can see: the 'ncurses' dependency was already present in the original pr (contrary to my own believe).

The really embarrassing observation for me is: I tested with and without ncurses on a node it build regardless (as 'ncurses' is in our filter-deps we do not provide a module, so eb has no chance to pick up the lib from a module). 

If the missing 'ncurses' leads to the build fails, why then do these build attempts not pick up the requirement and consider it? Anyone?

boegel:
@cmeesters You're being bit by using `dummy` as a toolchain version, which implies *not* loading the dependencies in the build environment; see also https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#loading-of-modules-for-dependencies-with-a-dummy-toolchain .

We're fully aware that this is weird behavior, and the reason why it was implemented like this originally is lost to the sands of time (the most likely motivation was to discourage the use of the `dummy` toolchain, but that's a weak argument at best).
Hence, we're getting rid of this in the upcoming EasyBuild 4.0, by replacing the `dummy` toolchain with the (better named) `system` toolchain, see https://github.com/easybuilders/easybuild-framework/pull/2877 .

For this PR: just use an empty toolchain version, which will result in loading the `ncurses` dependency in the build environment; that should fix the problem:

```python
toolchain = {'name': 'dummy', 'version': ''}
```

Also note that `ncurses` should only be a *build* dependency and make sure you link statically (using a runtime dependency may cause problems later when this `CMake` is used to build something that depends on `ncurses` with a non-dummy toolchain); see the existing easyconfig https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/c/CMake/CMake-3.12.1.eb

cmeesters:
@boegel 
Yes, you mentioned that - and I've forgotten. I'll have time today to give a try and report back. Thanks!

cmeesters:
I updated the pr according to Kenneth's remark - now the behavior is as expected. Please go ahead and test on your systems to be sure.

boegel:
Test report by @boegel
**SUCCESS**
node3103.skitty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/423cd7f57cde4a6126d1b43dec7515be for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2031.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/c72b3c6e13311b85519f236eb6ab7803 for a full test report.

verdurin:
Test report by @verdurin
**SUCCESS**
rescomp2 - Linux centos linux 7.3.1611, Intel(R) Xeon(R) Gold 6126 CPU @ 2.60GHz, Python 2.7.5
See https://gist.github.com/a2fc8d042b9d6cd19880ca0f5575949a for a full test report.

boegel:
@cmeesters Please make sure `ncurses` is statically linked, by specifying additional configure option, like we do in https://github.com/easybuilders/easybuild-easyconfigs/blob/master/easybuild/easyconfigs/c/CMake/CMake-3.12.1.eb:

```
-DCURSES_FORM_LIBRARY=$EBROOTNCURSES/lib/libform.a -DCURSES_NCURSES_LIBRARY=$EBROOTNCURSES/lib/libncurses.a
```

boegel:
@cmeesters Without static linking, you'll just end up with a broken `ccmake`:

```
$ module load CMake/3.15.1
$ ccmake --version
ccmake: error while loading shared libraries: libncurses.so.6: cannot open shared object file: No such file or directory
$ echo $?
127
$ ldd $(which ccmake) | grep ncurses
        libncurses.so.6 => not found
```

Maybe we should include `ccmake --version` as a sanity check command as well?

boegel:

boegel:
Test report by @boegel
**SUCCESS**
node3501.doduo.os - Linux RHEL 8.2, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/9ab83b697db1952ccd538414c4b848d3 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2631.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/b1f86725877322de7ee2113ae596ced5 for a full test report.

boegel (approved):

boegel:
Going in, thanks @cmeesters!

PR #8710, opened 2019-08-09
fix names of local variables in c/*/*.eb easyconfigs

boegel:
required to avoid warnings when these easyconfigs are parsed because of changed made in https://github.com/easybuilders/easybuild-framework/pull/2938

PR #8709, opened 2019-08-09
fix names of local variables in b/*/*.eb easyconfigs

boegel:
required to avoid warnings when these easyconfigs are parsed because of changed made in https://github.com/easybuilders/easybuild-framework/pull/2938

PR #8702, opened 2019-08-09
fix names of local variables in a/*/*.eb easyconfigs

boegel:
required to avoid warnings when these easyconfigs are parsed because of changed made in https://github.com/easybuilders/easybuild-framework/pull/2938

PR #8369, opened 2019-05-22
replace use of deprecated 'dummy' toolchain with 'toolchain = SYSTEM'

boegel:
+ check for use of 'dummy' toolchain in tests

These changes were made automatically using  `eb --fix-depr easybuild/easyconfigs`.
Afterwards the changes to the easyconfigs in `easybuild/easyconfigs/__archive__` were rolled back before committing the changes, since there's little point in tweaking the archived easyconfigs.

The tests were updated to catch future use of the `dummy` toolchain (which will happen when we sync `develop` into the `4.x` branch regularly until we get rid of `4.x` and only continue with `develop` in preparation of the EasyBuild v4.0 release), see changes in `test/easyconfigs/easyconfigs.py`.

PR #8711, opened 2019-08-09
re-enable checksum requirement for easyconfig files touched in PRs

boegel:
This reverts commit acd4f38a4c3f406a724d23e5388db7507f5d46fe (was added in #8369).

We'll leave the checksum requirement off for now, while we fix the local variable naming issues (see #8702, #8709, #8710, etc.)

This should be merged when that effort has been completed, certainly before `4.x` is collapsed into `develop`...


boegel:
re-triggering Travis checks...

PR #8764, opened 2019-08-14
Add missing builddep for pkg-config to Gdk-Pixbuf.

akesandgren:
Also move GObject-Introspection to builddeps where appropriate.

### DIFF
--- a/easybuild/easyconfigs/g/Gdk-Pixbuf/Gdk-Pixbuf-2.36.0-foss-2016b.eb
+++ b/easybuild/easyconfigs/g/Gdk-Pixbuf/Gdk-Pixbuf-2.36.0-foss-2016b.eb
@@ -16,12 +16,16 @@ toolchain = {'name': 'foss', 'version': '2016b'}
 source_urls = [FTPGNOME_SOURCE]
 sources = [SOURCELOWER_TAR_XZ]
 
+builddependencies = [
+    ('pkg-config', '0.29.1'),
+    ('GObject-Introspection', '1.49.1')
+]
+
 dependencies = [
     ('GLib', '2.49.5'),
     ('libjpeg-turbo', '1.5.0'),
     ('libpng', '1.6.24'),
     ('LibTIFF', '4.0.6'),
-    ('GObject-Introspection', '1.49.1')
 ]
 
 configopts = "--disable-maintainer-mode --enable-debug=no --enable-introspection=yes "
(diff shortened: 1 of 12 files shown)


boegel:
Test report by @boegel
**FAILED**
Build succeeded for 11 out of 12 (12 easyconfigs in this PR)
node2040.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/a0b765efdd9463e5e6f221f17deea176 for a full test report.

boegel:
@akesandgren `Gdk-Pixbuf-2.36.8-intel-2017a.eb` fails now because the `shared-mime-info` is required now (`pkg-config` can't find it).

Can we avoid that requirement?

akesandgren:
Don't know. I'll try to get to test building of it soon.

akesandgren:
Doesn't look like it. But shared-mime-info is something we should probably have as an osdep. There's just so much on a system that depends on it...

boegel:
Hmm, we actually *have* `shared-mime-info-1.8-4.el7.x86_64` installed.

I'll check on it again, it shouldn't have failed (and yeah, I guess we should include `shared-mime-info` as an OS dep).

boegel:
Or we build our own, it's a simple `configure-make-make install`, see https://freedesktop.org/wiki/Software/shared-mime-info/

akesandgren:
The problem with that is that it might make things point for mime to the EB tree only, and stuff won't see OS installed stuff. There's no clear and obvious path forward here that I can see.

boegel:
@akesandgren I just looked into the issue with `shared-mime-info`...

`pkg-config --exists shared-mime-info` fails because the system locations are not considered by the `pkg-config` we include as a build dep. We can fix that by adding the system locations via `$PKG_CONFIG_PATH`.

But in this particular case, the `configure` output gives the suggestion when the check for `shared-mime-info` failed to set `$SHARED_MIME_INFO_LIBS` and `$SHARED_MIME_INFO_CFLAGS` to avoid the need for using `pkg-config` to check for `shared-mime-info`.

The funny part is that both `pkg-config --cflags shared-mime-info` and `pkg-config --libs shared-mime-info` both yield nothing, so this works (not that those environment variables *must* be non-empty for this to work, so a single space is required):

```python
preconfigopts = "export SHARED_MIME_INFO_LIBS=' ' SHARED_MIME_INFO_CFLAGS=' ' && "
```

Thoughts on including this too, perhaps even consistently across all these easyconfigs (even though only `Gdk-Pixbuf-2.36.8-intel-2017a.eb` actually shows this issue, for some reason unclear to me...

akesandgren:
All versions of 2.36 from .8, at least, and forward that i've looked at tries to find shared-mime-info. SO for those it does make sense.

I can take a look at that in a while.

boegel:
@akesandgren fixed with https://github.com/akesandgren/easybuild-easyconfigs/pull/11

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 12 out of 12 (12 easyconfigs in this PR)
node2071.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/99eac6d40331fdac39e652323a753e76 for a full test report.

boegel:

PR #8768, opened 2019-08-14
{chem}[foss/2019a,intel/2019a] ASE v3.18.0, GPAW v19.8.1 w/ Python 3.7.2

schiotz:
(created using `eb --new-pr`)

PR #8786, opened 2019-08-16
{chem}[foss/2019a,intel/2019a] ASAP3 v3.11.10, kim-api v2.1.2, openkim-models v20190725, ... w/ Python 3.7.2

schiotz:
(created using `eb --new-pr`)


schiotz:
This makes ASAP3 available for the foss and intel 2019a toolchains.  It also includes the newest KIM API and the OpenKIM models (ping @ellio167).

I have included the same version of ASE as in #8768 as otherwise the self-test would fail.  Git should handle this duplication gracefully in the PR merged last.

smoors:
@schiotz I'm not a fan of changing software names. in any case it should be consistent.
I propose one of the following:
`kim-api` and `kim-models`,
`KIM-API` and `KIM-MODELS`, or
`OpenKIM-API` and `OpenKIM-MODELS`

schiotz:
Hi @smoors 

I certainly see your point, but I think changing the names to be consistent with all other package managers should be done now, while there is not a bunch of software depending on it.  So we avoid situations like the packages for Jupyter being called IPython because once upon a time that was the name.

The names `kim-api` and `openkim-models` are the names chosen by the developers, and used in other package managers.  They distinguish between the KIM, which is an API specification (and an implementation of it) and OpenKIM which is one particular collection of KIM models, with many others being possible.

@ellio167 could you comment on this, please?

ellio167:
Yes, what @schiotz says is correct.  

We have been able to largely standardize the package names across system:
kim-api:
[Packaging status](https://repology.org/project/kim-api/versions)

openkim-models:
[Packaging status](https://repology.org/project/openkim-models/versions)

As well as within https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages/kim-api and https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages/openkim-models

It seems desirable to have easybuild use the same names...

smoors:
@schiotz ok, fair enough.

schiotz:
Closing and reopening to trigger new self-test (for EB 4.0).

schiotz:
@boegel Can I do anything to get this one merged?

schiotz:
@smoors Thank you for looking into it.  I have removed the ASE configs (I though git handled that gracefully, but perhaps it would have introduced a change in the files).  Next time I will avoid such duplication.

smoors:
Test report by @smoors
**SUCCESS**
Build succeeded for 9 out of 9 (6 easyconfigs in this PR)
See https://gist.github.com/0ea9a9a8661e7f72fb7c36eacf1d42f0 for a full test report.

smoors:
Going in, thanks @schiotz!

easybuild/easyconfigs/a/ASAP3/ASAP3-3.11.10-foss-2019a-Python-3.7.2.eb
```diff
@@ -0,0 +1,41 @@
...
+    ('pkgconfig', '1.5.1', '-python'),
+]
+
+dependencies = [
+    ('Python', '3.7.2'),
+    ('ASE', '3.18.0', versionsuffix),
+    ('kim-api', '2.1.2'),
+]
+
+use_pip = False
```
boegel: @schiotz Just wondering, which issues did you run into with `pip install` that made you use `use_pip = False` here?

schiotz: Asap installs a version of the python executable linked with MPI (called asap-python, needed for parallel simulations only), this does not get installed properly when using pip, since pip assumes it is a script, and tries to read the file and replace the hash-bang in the first line with the right path to the python executable.  But since it is not a script but an executable, it goes wrong.

The main GPAW developer is working on fixing this in gpaw so building a special executable is not necessary, but there still seems to be problems on some combinations of architectures and MPI versions.  Eventually, I am going to borrow his solution for asap, but it does require some work.

boegel: Thanks a lot for the clarification!

PR #8865, opened 2019-08-30
{tools} [GCCcore/6.4.0] libssh v0.9.0 {REVIEW}

SimonPinches:


easybuild/easyconfigs/l/libssh/libssh-0.9.0-GCCcore-6.4.0.eb
```diff
@@ -0,0 +1,39 @@
...
+
+source_urls = ['https://www.libssh.org/files/0.9/']
+sources = ['%(name)s-%(version)s.tar.xz']
+checksums = ['25303c2995e663cd169fdd902bae88106f48242d7e96311d74f812023482c7a5']
+
+builddependencies = [
+    ('CMake', '3.12.1'),
+    ('binutils', '2.28'),
+]
+
```
Micket: At the very least, this probably has a OS dependency on OpenSSL?

Micket (changes requested):
I'd like a second opinion on this; Do we even want libssh as a module, or do we prefer relying on OS to keep it in regards to security updates? Perhaps it's enough to keep openssl as OS level.

akesandgren:
I'm for keeping things like this out of EB and relying on OS builds due to the frequent security updates.

boegel:
We have easyconfig files for `OpenSSL` too, but we actively discourage using them to resolve dependencies.

I think it's OK to add an easyconfig for this, it could help people that don't have admin access to install a missing dep?

easybuild/easyconfigs/l/libssh/libssh-0.9.0-GCCcore-6.4.0.eb
Micket: So, this should either add a dep. on OpenSSL or add a os dependency on it.

smoors: @SimonPinches please add an os dependency for SSL, as @Micket requested?
```
osdependencies = [('openssl-devel', 'libssl-dev', 'libopenssl-devel')]
```

SimonPinches: Done - apologies for the delay!

smoors:
Test report by @smoors
**SUCCESS**
See https://gist.github.com/59156ab53d1f20c1219edfdc463c5cf5 for a full test report.

smoors (approved):
lgtm

Micket (approved):

smoors:

PR #8870, opened 2019-09-03
fix incorrect checksums for Amber patches

akesandgren:
And add some missing deps.


boegel:
@akesandgren What will we do with the missing `versionsuffix` that is now enforced by the tests?

Also, please submit a test report, the incorrect patch checksum should never have happened in the first place... I don't have access to the sources for Amber 16 myself.

akesandgren:
test report is on it's way... it just takes a long time to build 4 versions of Amber :-)

As for the Python suffix requirement I think we should override it in this case. It's an update to fairly old easyconfigs.

akesandgren:
Test report by @akesandgren
**SUCCESS**
b-cn0823.hpc2n.umu.se - Linux ubuntu 16.04, Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz, Python 2.7.12
See https://gist.github.com/b544bdb3b11b97f486810e7edaa2dad5 for a full test report.

boegel:
@akesandgren The same issue applies to recent `Amber` easyconfigs like `Amber-18-foss-2018b-AmberTools-18-patchlevel-10-8.eb`.

Should we whitelist `Amber` as not needing a `-Python-*` `versionsuffix` in the tests?

akesandgren:
Might be a good idea at the moment. My plan for the next version is to both change how the version looks (more like CC does it if i remember correctly) and to add -Python

dianaiusan:
There are several *eb files that have a wrong checksum for Amber-16-AT-17_fix_fixed_size_cmd_in_nab.patch:

Amber-16-foss-2018b-AmberTools-17-patchlevel-10-15.eb
Amber-16-fosscuda-2018b-AmberTools-17-patchlevel-10-15.eb
Amber-16-intel-2017b-AmberTools-17-patchlevel-8-12.eb
Amber-16-intel-2018b-AmberTools-17-patchlevel-10-15.eb

boegel:

Issue #8082, opened 2019-04-16
Lets merge icc and ifort and just have iccifort


Micket:
I didn't spot any other issues for this, other then easybuilders/easybuild-framework#1076
Rather then trying to fix that, can't we just use iccifort instead?
It doesn't look like intel will be splitting the release schedule for them anymore.
Starting with 2019b maybe?

With HMNS, it's particularly bad to have them split. It lets users load just the icc to get access to modules. But this leads to broken software, as anything built with ifort requires the user to load ifort as well.
Worse yet, `module spider` actually gives these this as a suggestion (as it just tells you what is needed to load the module).

I have been manually modifying my old modules, moving the modulepath into the iccifort module instead, but that's tedious, and ugly, and a bit fragile (it i ever had to rebuilt modules etc.)

bartoldeman:
I would be happy with that, but in a conversation with @ocaisa he mentioned that Intel can sell icc and ifort separately. You can actually get "Intel System Studio" for free (with some strings attached) but that only includes C and C++, not Fortran.

Both FZ Julich and us at Compute Canada use a custom HMNS where icc and ifort modules are hidden dependencies of iccifort and only the iccifort module (renamed to intel, but that is an implementation detail) extends `$MODULEPATH`

akesandgren:
Could you share that HMNS def?

Micket:
@bartoldeman I suppose we don't need to remove the individual `icc`, `ifort` (if someone else really wants it)?
Just make `iccifort` standalone. Because, anyone who gets just `icc` or `ifort` standalone, would have to construct their own toolchains. It wouldn't be of `intel`-ish toolchains anymore anyway.

My manual fiddling with configs, got me into trouble when building intelcuda/2019a (which prompted me to write this issue).

bartoldeman:
@akesandgren 
The Julich MNS is here:
https://github.com/easybuilders/JSC/tree/master/Custom_MNS
Compute Canada's is here:
https://github.com/ComputeCanada/easybuild-computecanada-config

bartoldeman:
@Micket I agree. Something similar was done in the past to with Clang: there are ClangGCC toolchains where the Fortran is provided by gfortran; and this was before GCCcore was introduced. With GCCcore you could in principle promote `icc` to a compiler-only toolchain where GCCcore provides `gfortran`, and keep `iccifort` independent of `icc`.

boegel:
[...]

PR #8681, opened 2019-08-07
{toolchain} intel/2019b

boegel:
(created using `eb --new-pr`)

PR #8567, opened 2019-07-10
{toolchain} foss/2019b

boegel:
(created using `eb --new-pr`)

PR #8879, opened 2019-09-05
Install iccifort directly using the "composer" tarball.

bartoldeman:
The advantage over a bundle of icc and ifort is that this avoids
duplication of files between icc and ifort.

I included easyconfigs with and without the -2.32 suffix for
easy testing against develop.

Note that the use of the EB_ifort easyblock is temporary.


bartoldeman:
This is my first attempt to fix #8082, although it needs some support in the framework and an iccifort easyblock.

easybuild/easyconfigs/i/iccifort/iccifort-2019.4.227-GCC-8.3.0-2.32.eb
```diff
@@ -0,0 +1,42 @@
+# This is an easyconfig file for EasyBuild, see http://easybuilders.github.io/easybuild
+
+name = 'iccifort'
+version = '2019.4.227'
```
boegel: @bartoldeman Correct subversion is `243`, not `227` (otherwise the sanity check fails).

How did this work for you?

easybuild/easyconfigs/i/iccifort/iccifort-2019.4.227-GCC-8.3.0.eb
```diff
@@ -0,0 +1,42 @@
+# This is an easyconfig file for EasyBuild, see http://easybuilders.github.io/easybuild
+
+name = 'iccifort'
+version = '2019.4.227'
+versionsuffix = '-GCC-8.3.0'
```
Micket: If we are dropping the binutils suffix because it's always just one version anyway; can't we go all the way here and drop the entire suffix as the same applies for iccifort -> GCC.

boegel: Makes sense, see also discussion in https://github.com/easybuilders/easybuild-easyconfigs/pull/8681/files#r311570349

easybuild/easyconfigs/i/iccifort/iccifort-2019.4.227-GCC-8.3.0-2.32.eb
bartoldeman: Actually I took the 227 from the intel-2019b proposal @8681.
This exact easyconfig did not work for me because our license server is not up to date... I tested with 2019.3.

bartoldeman: argh, Intel has changed the tarball in the mean time. I'll adjust the 227 to be 243.

easybuild/easyconfigs/i/iccifort/iccifort-2019.4.243.eb
```diff
@@ -0,0 +1,41 @@
...
+patches = ['iccifort-%(version)s_no_mpi_rt_dependency.patch']
+checksums = [
+    # parallel_studio_xe_2019_update4_composer_edition.tgz
+    '1915993445323e1e78d6de73702a88fa3df2036109cde03d74ee38fef9f1abf2',
+    # iccifort-2019.4.243_no_mpi_rt_dependency.patch
+    '929ffa72acd4b3e107791a7cf74324aaefd88f24923c0da862b1e43f6d52a731',
+]
+
+local_gccver = '8.3.0'
+versionsuffix = '-GCC-%s' % local_gccver
```
Micket: I found another `versionsuffix` :)

easybuild/easyconfigs/i/iccifort/iccifort-2019.4.243-GCC-8.3.0-2.32.eb
```diff
@@ -0,0 +1,42 @@
+# This is an easyconfig file for EasyBuild, see http://easybuilders.github.io/easybuild
+
+name = 'iccifort'
+version = '2019.4.243'
+versionsuffix = '-GCC-8.3.0-2.32'
```
Micket: Is there a reason why this file is included? It seems we are all in favor of dropping the versionsuffix, so this file isn't very relevant

bartoldeman: It's just there so people can test with current develop until foss-2019b is merged.

boegel: @bartoldeman It's a bit confusing to have both in here. Please remove this one, and clearly mention in the PR description that this depends on `GCC-8.3.0.eb` from #8567?

boegel: Hold on, the other easyconfig doesn't require `GCC-8.3.0.eb` at all? How exactly does this depend on #8567? It doesn't I think?

bartoldeman: You're right, the other easyconfig depends on `GCCcore, 8.3.0` in `dependencies`, not on `GCC`.

boegel:
Test report by @boegel
**SUCCESS**
node3100.skitty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/2a73a8ce2c05d8e549b6108b1064fecc for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2628.swalot.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/2eb075a8c0669b7e109c314104b3317d for a full test report.

boegel:
Going in, thanks @bartoldeman!

PR #8892, opened 2019-09-10
avoid that easy_install installs easyconfigs package as a zipped egg

boegel:
Even though we're not using `setuptools` anymore in `setup.py`, people could still use `easy_install` (provided by `setuptools`).
We need to make sure that the easyconfigs is not installed as a zipped egg, since we need the easyconfig files + patches to be accessible separately...


boegel:
This didn't work as expected, but #8894 does...

PR #9008, opened 2019-09-26
Create gnucobol-3.0-rc1-GCC-4.9.2.eb

levlafayette:
It would be good to have cobol.

### DIFF
--- a/easybuild/easyconfigs/g/gnucobol-3.0-rc1-GCC-4.9.2.eb
+++ b/easybuild/easyconfigs/g/gnucobol-3.0-rc1-GCC-4.9.2.eb
@@ -1,0 +1,28 @@
+# Lev Lafayette, University of Melbourne, 20190808
+easyblock = 'ConfigureMake'
+
+name = 'gnucobol'
+version = '3.0-rc1'
+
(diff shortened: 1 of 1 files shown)


boegel:
@levlafayette Are there other pull requests open for the missing dependencies, or did you forget to include the dependencies?

Also, I'm baffled as to why nobody has added this yet, how have people been doing science without COBOL?! ;)

levlafayette:
Yes, still to add the other pull requests. Need to add DB/5.0.26-GCC-4.9.2, ncurses/6.0-GCC-4.9.2

Whilst it is true that COBOL doesn't generate much love in the scientific software world (for good reason), GnuCOBOL is a transpiler that converts COBOL code into C. 

Which means multithreaded COBOL! Even MPI COBOL!

boegel:
@levlafayette You can add the additional easyconfigs in this PR, no need to make separate PRs.

Micket:
But.. why such a ancient version of GCC?

boegel:
No updates since Sept 2019, and required dependency PRs #9039 and #9040 never got dealt with, so I'll assume this is no longer relevant and close this PR

PR #8999, opened 2019-09-25
{cae}[intel/2019a] OpenFOAM v2.2.x

boegel:
(created using `eb --new-pr`)

PR #9006, opened 2019-09-25
{data,math}[intel/2019a] GEOS v3.7.2, GDAL v3.0.0 w/ Python 3.7.2

boegel:
(created using `eb --new-pr`)

PR #8997, opened 2019-09-25
{bio}[gompi/2019b] HMMER v3.2.1

smoors:
(created using `eb --new-pr`)

PR #9012, opened 2019-09-26
add check for http:// URLs in easyconfig files added/changed in PRs

boegel:
works as expected, see #8999 where I squeezed in this change too by matter of proof of concept...

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -454,11 +454,11 @@ class EasyConfigTest(TestCase):
         self.assertFalse(failing_checks, '\n'.join(failing_checks))
 
     def check_https(self, changed_ecs):
-        """Make sure https:// is used for homepage/source_urls (rather than http://)."""
+        """Make sure https:// URL is used (if it exists) for homepage/source_urls (rather than http://)."""
 
         whitelist = []
 
-        http_regex = re.compile('http://[^"\'\n]*', re.M)
+        http_regex = re.compile('http://[^"\'\n]+', re.M)
 
         failing_checks = []
         for ec in changed_ecs:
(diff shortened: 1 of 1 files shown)


boegel:
@Micket Once this is merged, tests in Travis will fail for (new/updated) PRs if easyconfigs still have `http://` in them.

We may need to whitelist for some stuff where `https://` doesn't work (should we auto-check if the specified URLs exist?)

Micket:
@boegel Having looked through a bunch of them the past week, I've vaguely recall seeing at least a couple of exceptions already. I think autochecking would be a good idea before requiring it
If they aren't using https, and there is a positive response from `requests.get(url, verify=True)`, then they should prefer it.

boegel:
@Micket Do you happen to recall an example of an exception, to test with?

Micket:
https://github.com/easybuilders/easybuild-easyconfigs/pull/8997/files

boegel:
enhanced test works as expected, see #8999 and #9006

test/easyconfigs/easyconfigs.py
```diff
@@ -453,6 +453,28 @@ def check_sanity_check_paths(self, changed_ecs):
 
         self.assertFalse(failing_checks, '\n'.join(failing_checks))
 
+    def check_https(self, changed_ecs):
+        """Make sure https:// is used for homepage/source_urls (rather than http://)."""
+
+        whitelist = []
+
+        http_regex = re.compile('http://[^"\'\n]*', re.M)
```
wpoely86: + instead of * ?

test/easyconfigs/easyconfigs.py
```diff
@@ -453,6 +453,28 @@ def check_sanity_check_paths(self, changed_ecs):
 
         self.assertFalse(failing_checks, '\n'.join(failing_checks))
 
+    def check_https(self, changed_ecs):
+        """Make sure https:// is used for homepage/source_urls (rather than http://)."""
```
wpoely86: add: if it exists?

test/easyconfigs/easyconfigs.py
boegel: You mean `+`?

PR #8404, opened 2019-05-29
Amber: update urls, add checksums and use variables

SethosII:
I don't have access to `Amber14.tar.bz2` and `AmberTools15.tar.bz2` so I can't add checksums for them.

PR #9023, opened 2019-09-27
Amber: update urls, use variables and add checksums

SethosII:
Replacement for https://github.com/easybuilders/easybuild-easyconfigs/pull/8404.

### DIFF
--- a/easybuild/easyconfigs/a/Amber/Amber-16-foss-2017b-AmberTools-17-patchlevel-10-15-Python-2.7.14.eb
+++ b/easybuild/easyconfigs/a/Amber/Amber-16-foss-2017b-AmberTools-17-patchlevel-10-15-Python-2.7.14.eb
@@ -1,11 +1,11 @@
 name = 'Amber'
 version = '16'
 local_at_ver = '17'
-# Patch levels from http://ambermd.org/bugfixes16.html and http://ambermd.org/bugfixesat.html
+# Patch levels from http://ambermd.org/AmberPatches.php and http://ambermd.org/ATPatches.php
 patchlevels = (10, 15)  # (AmberTools, Amber)
 versionsuffix = '-AmberTools-%s-patchlevel-%s-%s-Python-%%(pyver)s' % (local_at_ver, patchlevels[0], patchlevels[1])
 
-homepage = 'http://ambermd.org/amber.html'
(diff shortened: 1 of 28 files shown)


SethosII:
Should I include Python in the `versionsuffix` and easyconfig name?

SethosII:
I added Python to the versionsuffix where it was missing.

I don't know where the certificate errors are coming from.

zao:
@SethosII Browsers also complain about their certificate and if you force it through, the website you end up on is for Institute for Quantitative Biomedicine (IQB), not Amber.

Upstream has broken itself :tada:

akesandgren:
The "home" url is wrong, it should only be http://ambermd.org/ (and please don't use %(name)s in the home url)

The problem here is that easybuild's tests check for the existence of a working https://... url and ambermd.org's certificate lacks the required SAN for ambermd.org

I'll poke them a bit to see if they can fix that.

akesandgren:
Amber devs poked.

zao:
An EB workaround would be to add the software to the whitelist in https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/test/easyconfigs/easyconfigs.py#L463

SethosII:
@akesandgren What is problematic about using variables in the `homepage`? I think that's what they are there for? And why not use a more specific URL? In case of Amber and AmberTools they would have the same `homepage`, but there are specific ones available.

akesandgren:
the "problem" is that it is rarely usable and when trying to check things before making a new easyconfig one would like to just copy&paste the home page url from an earlier easyconfig without having to manually convert %(name)s.

As for Amber, the ambermd.org IS the home page for amber and ambertools, it clearly says so on the page. The others are for the current version of the packages. And http://ambermd.org/AmberMD.php clearly says that "Amber Home" is http://ambermd.org

SethosII:
@akesandgren Ok, I will change this. Should we wait for the amber devs to decide whether to use https or not?

akesandgren:
Amber now has a functioning cert for their https://

boegel:
This PR only touches easyconfigs with a now [deprecated toolchain](https://docs.easybuild.io/en/latest/Deprecated-easyconfigs.html#deprecated-toolchains), so closing this...

PR #9065, opened 2019-10-03
{lib}[foss/2019a] TensorFlow v2.0.0, w/ Python 3.7.2 (+ Bazel v0.26.1)

boegel:
(created using `eb --new-pr`)
requires https://github.com/easybuilders/easybuild-easyblocks/pull/1824 + ~~#9070~~ (`git`)

PR #6751, opened 2018-08-22
also install docs/man pages in recent git easyconfigs (versions 2.1x.y)

boegel:
fix for #6746 reported by @chrissamuel

Issue #6746, opened 2018-08-22
Recipe to install git doesn't install manual pages & help


chrissamuel:
Hi there,

After a query from a user we've noticed that the EB recipe for git doesn't seem to do the extra targets needed to install the manual pages.  We've found that this change to our local recipe fixes that issue.

```
+# Necessary to install documentation
+buildopts = "doc"
+installopts = "install-doc"
```
Hope this helps!

All the best,
Chris

boegel:
fixed in #6751

PR #9067, opened 2019-10-03
comment out installation of git docs, since it requires additional OS deps

boegel:
We've started also installing the docs with `git` in #6751 as a response to the issue reported by @chrissamuel in #6746, but I think we should reconsider that because of the required OS dependencies it implies (see #7028), which I was not aware of initially...

The reason I bring this up now is because TensorFlow 2.0 requires a sufficiently recent `git` to be installed, which we take of by including `git` as a build dependency (see #9065).

If `asciidoc` and `xmlto` (which are not easy to provide in EasyBuild, see https://github.com/easybuilders/easybuild-easyconfigs/pull/7028#issuecomment-430185379) are required for installing `git`, then anyone who wants to build TensorFlow 2.x from source will need to install those packages in the OS first, which is silly...

For sites who want to have `git` installed with documentation included, they can opt in by putting back the lines in the `git` easyconfig I'm commenting out here...

### DIFF
--- a/easybuild/easyconfigs/g/git/git-2.12.2-foss-2016b.eb
+++ b/easybuild/easyconfigs/g/git/git-2.12.2-foss-2016b.eb
@@ -24,7 +24,7 @@ dependencies = [
 ]
 
 # asciidoc and xmlto are required for git man/doc build
-osdependencies = ['asciidoc', 'xmlto']
+# osdependencies = ['asciidoc', 'xmlto']
 
 preconfigopts = 'make configure && '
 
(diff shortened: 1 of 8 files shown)


boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 8 out of 8 (8 easyconfigs in this PR)
node2092.delcatty.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz, Python 2.7.5
See https://gist.github.com/f308d7e0cdf75a4805c0a838fd261235 for a full test report.

migueldiascosta:
alternatively, if users do expect the docs when loading the `git` module as mentioned in #6751, maybe we could create a separate `-nodocs` easyconfig for use as a `builddependency`?

boegel:
@migueldiascosta Agreed, that's a better approach. `-nodocs` easyconfig for `git` that can be used as a build dep for TensorFlow 2.0.0 added in #9070, so closing this...

PR #9156, opened 2019-10-17
{bio}[GCCcore/8.2.0] FragGeneScan v1.31

Darkless012:
(created using `eb --new-pr`)


Darkless012:
Test report by @Darkless012
**SUCCESS**
node3241.victini.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 2.7.5
See https://gist.github.com/cdaef11bbc5e98d49ef54b56a8c8f56e for a full test report.

migueldiascosta:
**SUCCESS**
See https://gist.github.com/6464c31947cb646108918bd2b62f813f for a full test report.

migueldiascosta:
@Darkless012 this can be moved down to either `GCC` or `GCCcore`, no?

easybuild/easyconfigs/f/FragGeneScan/FragGeneScan-1.31-GCCcore-8.2.0.eb
```diff
@@ -0,0 +1,32 @@
...
+
+easyblock = 'MakeCp'
+
+name = 'FragGeneScan'
+version = '1.31'
+
+homepage = 'https://omics.informatics.indiana.edu/FragGeneScan/'
+description = "FragGeneScan is an application for finding (fragmented) genes in short reads."
+
+toolchain = {'name': 'GCCcore', 'version': '8.2.0'}
```
Micket: In case software is (might) be affected by the compiler performance, it's nicer to keep them at the GCC and iccifort level so that users can pick. So, even if we can move it to GCCcore, doesn't always mean we should.

If this is just a small utility, or when we know that performance isn't significantly better in iccifort, then GCCcore is the best choice. Is this the case here?

easybuild/easyconfigs/f/FragGeneScan/FragGeneScan-1.31-GCCcore-8.2.0.eb
```diff
@@ -0,0 +1,32 @@
...
+
+source_urls = [SOURCEFORGE_SOURCE]
+sources = ['%(name)s%(version)s.tar.gz']
+checksums = ['cd3212d0f148218eb3b17d24fcd1fc897fb9fee9b2c902682edde29f895f426c']
+
+builddependencies = [('binutils', '2.31.1')]
+
+buildopts = 'CC="$CC" CFLAG="$CFLAGS" fgs && chmod -R go+rx *.pl train example'
+
+files_to_copy = ['FragGeneScan', 'run_FragGeneScan.pl', 'example', 'train']
```
Micket: Doesn't `run_FragGeneScan.pl` indicate that this ought to have a `Perl` dependency?

easybuild/easyconfigs/f/FragGeneScan/FragGeneScan-1.31-GCCcore-8.2.0.eb
Darkless012: The only tests I could do were example runs with included data. Measuring the execution time while having this compiled under GCC vs GCCcore didn't make any measurable difference. Thus leaving to GCCcore.

easybuild/easyconfigs/f/FragGeneScan/FragGeneScan-1.31-GCCcore-8.2.0.eb
Darkless012: Good catch. It had hard-coded shebang and was using system Perl. I've added shebang fix and Perl deps.
I didn't pay attention to that, as I was updating from older configfile.

easybuild/easyconfigs/f/FragGeneScan/FragGeneScan-1.31-GCCcore-8.2.0.eb
Micket: There would never be any difference between GCC and GCCcore, the question is whether or not gcc and icc shows any performance differences. If icc is better, we don't want a version at GCCcore. 

I had a look at compiling both, and, either way, it seems like a small utility, so it might not be worth the hassle, and icc didn't perform any better anyway.

Micket:
Test report by @Micket
**SUCCESS**
vera-c1 - Linux centos linux 7.7.1908, Intel Xeon Processor (Skylake), Python 2.7.5
See https://gist.github.com/64b5aed02d50b08b2f5d63595abf8f22 for a full test report.

Micket:
Going in, thanks @Darkless012!

PR #9208, opened 2019-10-24
added recipes to move cuda at the core level

mboisson:
These two recipes move CUDA at the core level through a new easyconfig, called CUDAcore, which is hidden. The new CUDA recipe use the ModuleOnly EasyBlock which is from https://github.com/easybuilders/easybuild-easyblocks/pull/1839

### DIFF
--- a/easybuild/easyconfigs/c/CUDA/CUDA-10.1.eb
+++ b/easybuild/easyconfigs/c/CUDA/CUDA-10.1.eb
@@ -1,0 +1,15 @@
+easyblock = "ModuleOnly"
+name = 'CUDA'
+version = '10.1'
+
+homepage = 'https://developer.nvidia.com/cuda-toolkit'
+description = """CUDA (formerly Compute Unified Device Architecture) is a parallel
+ computing platform and programming model created by NVIDIA and implemented by the
+ graphics processing units (GPUs) that they produce. CUDA gives developers access
+ to the virtual instruction set and memory of the parallel computational elements in CUDA GPUs."""
+
+toolchain = {'name': 'GCC', 'version': '7.3.0'}
+
+dependencies = [ ('CUDAcore','10.1.243') ]
+
+moduleclass = 'system'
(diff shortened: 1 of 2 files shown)


mboisson:
Actually, this is Compute Canada-specific because of our setrpaths.sh script.

mboisson:
@akesandgren, while this PR was too ComputeCanada-specific, I think it may be of interest to you.

easybuild/easyconfigs/c/CUDA/CUDA-10.1.eb
```diff
@@ -0,0 +1,15 @@
...
+
+homepage = 'https://developer.nvidia.com/cuda-toolkit'
+description = """CUDA (formerly Compute Unified Device Architecture) is a parallel
+ computing platform and programming model created by NVIDIA and implemented by the
+ graphics processing units (GPUs) that they produce. CUDA gives developers access
+ to the virtual instruction set and memory of the parallel computational elements in CUDA GPUs."""
+
+toolchain = {'name': 'GCC', 'version': '7.3.0'}
+
+dependencies = [ ('CUDAcore','10.1.243') ]
```
akesandgren: Doesn't that need hiddendependencies?

akesandgren:
Yeah, I had ideas around something similar...

easybuild/easyconfigs/c/CUDA/CUDA-10.1.eb
mboisson: Mmm, well, it installed and found the dependency just fine on our system, and I don't see anything in our config.cfg related to hidden dependencies ?

akesandgren: Ok, haven't used hidden_deps so don't really know hot it works.

bartoldeman: hiddendependencies isn't really necessary, if you use it it's site-specific, it basically enforces that the module name should have a dot, but normal dependencies will also look for hidden modules (whether dotted or configured so via Lmod config files, doesn't matter)

PR #9217, opened 2019-10-31
{bio}[iccifort/2019.1.144-GCC-8.2.0-2.31.1] bwa-meth v0.2.2, BWA v0.7.17 w/ Python 2.7.15 + 3.7.2

Darkless012:
(created using `eb --new-pr`)

PR #9224, opened 2019-11-04
ignore commented out lines in easyconfig files when checking for http:// URLs

### LINKED ISSUE #9180: HTTPS URL check should ignore commented lines
The current URL http -> https checks are not very nice, they should ignore commented lines, but don't. See, for example, https://github.com/easybuilders/easybuild-easyconfigs/pull/9146#issuecomment-543125211:

```
> Travis test report: 5/5 runs failed - see https://travis-ci.org/easybuilders/easybuild-easyconfigs/builds/599019907
> 
> Only showing partial log for 1st failed test suite run 17707.1;
> full log at https://travis-ci.org/easybuilders/easybuild-easyconfigs/jobs/599019908
> 
> ```
> ...
> FAIL: Specific checks only done for the (easyconfig) files that were changed in a pull request.
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/travis/build/easybuilders/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 541, in test_changed_files_pull_request
>     self.check_https(changed_ecs)
>   File "/home/travis/build/easybuilders/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 485, in check_https
>     self.assertFalse(failing_checks, '\n'.join(failing_checks))
> AssertionError: Found http:// URL in LibTIFF-4.0.9-GCCcore-7.3.0.eb, should be https:// : http://hpcbios.readthedocs.org/en/latest/
```

That URL that is tripped over is part of a commented line:
https://github.com/easybuilders/easybuild-easyconfigs/blob/b031c6a57a1417676d0109be37dd2d2d427f22b7/easybuild/easyconfigs/l/LibTIFF/LibTIFF-4.0.9-GCCcore-7.3.0.eb#L10

So even though the description of `check_https()` says it checks URLs used for "homepage/source_urls" it does something different by simply matches *any part of the easyconfig text* without looking at lines and comments:

https://github.com/easybuilders/easybuild-easyconfigs/blob/5b85d4d009ca93d58cdf80dd390a6b91de0860b0/test/easyconfigs/easyconfigs.py#L459-L485

Should be fairly easy to fix, by splitting the raw contents into lines and matching one those, except on commented lines.

boegel:
fixes #9180 reported by @paulmelis

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -472,7 +472,10 @@ class EasyConfigTest(TestCase):
             if any(ec_fn.startswith(x) for x in whitelist):
                 continue
 
-            for http_url in http_regex.findall(ec.rawtxt):
+            # ignore commented out lines in easyconfig files when checking for http:// URLs
+            ec_txt = '\n'.join(l for l in ec.rawtxt.split('\n') if not l.startswith('#'))
+
+            for http_url in http_regex.findall(ec_txt):
                 https_url = http_url.replace('http://', 'https://')
                 try:
                     https_url_works = bool(urlopen(https_url))


verdurin:
@boegel Haven't tested it though?

test/easyconfigs/easyconfigs.py
```diff
@@ -472,7 +472,10 @@ def check_https(self, changed_ecs):
             if any(ec_fn.startswith(x) for x in whitelist):
                 continue
 
-            for http_url in http_regex.findall(ec.rawtxt):
+            # ignore commented out lines in easyconfig files when checking for http:// URLs
+            ec_txt = '\n'.join(l for l in ec.rawtxt.split('\n') if not l.startswith('#'))
```
akesandgren: The correct thing is to ignore comments altogether, not just lines that start with #.

boegel: I think this is good enough, I doubt we'll hit `http://` URLs in inline comments?

Also, stripping off all comments isn't that trivial, since there may appear `#` characters in other places (like in string values)

boegel:
@verdurin This is hard to test without it being merged. I plan to re-trigger the tests in #9217 once this is merged to confirm the fix.

test/easyconfigs/easyconfigs.py
akesandgren: Agree that it is non-trivial and should be left for someday when we have nothing else to do...

akesandgren:
Going in, thanks @boegel!

PR #9236, opened 2019-11-07
speed up easyconfigs test suite by avoiding re-parsing and re-ordering of easyconfigs

boegel:
It turns out that the guards that are supposed to avoid that the easyconfig files are parsed & ordered (via `resolve_dependencies`) multiple times don't work, because `self.parsed_easyconfigs` and `self.ordered_specs` refer to *instance* variables.

It seems that `unittest` creates a new instance of `EasyConfigTest` for every test being run, so checking `self.parsed_easyconfigs` and `self.ordered_specs` to cache things between tests is totally useless.

By switching to using the `EasyConfigTest.parsed_easyconfigs` and `EasyConfigTest.ordered_specs` *class* variables, which survive across instances, we're effectively caching parsed easyconfigs as intended...

This significantly speeds up the test suite (in GitHub CI):

* with Python 2.7: ~25% speedup (~9m to ~7m)
* with Python 3.5: ~60% (~17m to ~10.5m)
* with Python 3.7: ~65% (~57m to ~34m)

Still a bit puzzled by the large differences between Python versions (will do some more profiling to try and figure that out).

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -106,17 +106,17 @@ class EasyConfigTest(TestCase):
         specs = glob.glob('%s/*/*/*.eb' % easyconfigs_path)
 
         # parse all easyconfigs if they haven't been already
-        if not self.parsed_easyconfigs:
+        if not EasyConfigTest.parsed_easyconfigs:
             for spec in specs:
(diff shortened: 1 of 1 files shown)


migueldiascosta:
Going in, thanks @boegel!

PR #9248, opened 2019-11-11
{data}[fosscuda/2019b] h5py v2.9.0 w/ Python 3.7.4

Flamefire:
(created using `eb --new-pr`)

Requires:
- ~~#9251~~ (pkgconfig)
- ~~#9253~~ (SciPy-bundle) 
- ~~#9249~~ (HDF5)

PR #8133, opened 2019-04-25
{devel}[GCCcore/8.2.0] pkgconfig v1.5.1

akesandgren:
(created using `eb --new-pr`)

PR #9251, opened 2019-11-11
{devel}[GCCcore/8.3.0] pkgconfig v1.5.1 w/ Python 3.7.4

Flamefire:
(created using `eb --new-pr`)


Flamefire:
Is a suffix of `Python-x.y.z` requirement new? This is based of the EC in https://github.com/easybuilders/easybuild-easyconfigs/pull/8133 where travis did not complain. Shall I change this?

boegel:
@Flamefire If you depend on a single Python, then a `-Python-<pyver>` versionsuffix is indeed required.
We've added a check for that a while ago, but note that the easyconfig in #8133 uses `multi_deps` to make it compatible with both Python and 3 (and hence it doesn't require a `versionsuffix`).

Is there a specific reason why you didn't do that here?

Flamefire:
I'm unsure how multi-dep works. I'd expect it to create 2 modules. What will they be called? Wouldn't each of those module require the same suffix as if it was 2 ECs?

> Is there a specific reason why you didn't do that here?

As Python 2 will drop out of support by the end of this year I'd rather not create new ECs/modules which install it in a 2019b based toolchain.

boegel:
@Flamefire When `multi_deps` is used like in #8133, it will install a single module that is compatible with both Python 2 & 3. We added support for this due to the large number of easyconfig files that were identical other than the Python version they were using.

I understand your point w.r.t. Python 2, we shouldn't encourage researchers to still use it.

In this particular case, since `pkgconfig` is almost exclusively a dependency for `h5py` (which depends on a specific Python version anyway in the 2019b generation of easyconfig because it depends on `SciPy-bundle`, see #9248), we can go forward by simply adding the `versionsuffix` to make the tests happy.
That'll indicate that the module depends on a specific Python version, and allows to add another easyconfig for `pkgconfig` module with Python 2 if the need arises...

boegel:
Hmm, I should check why the tests on fail in Travis, and not in GitHub Actions...

boegel:
@Flamefire Please remove the `pkgconfig-1.5.1-GCCcore-8.3.0-python.eb ` too.

```
eb --update-pr 9251 :pkgconfig-1.5.1-GCCcore-8.3.0-python.eb --pr-commit-msg "remove pkgconfig with wrong versionsuffix"

Flamefire:
Just noticed that `--update-pr` does not replace the file. Done better in #9248. Please squash on merge

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/df3aef4da3a41c676ef8ae591a367e2a for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3300.joltik.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6242 CPU @ 2.80GHz, Python 2.7.5
See https://gist.github.com/c9dd7892de15b9b42039a22944dbbf9b for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3151.skitty.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/afa223f2a68eca93e44600ed0e8be415 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2437.golett.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/81c2eefb748a24a5bb8c72982865e71a for a full test report.

boegel:
Going in, thanks @Flamefire!

PR #9256, opened 2019-11-12
archive old patches for Xmipp 3.1

boegel:
These patches are only used by archived `Xmipp` easyconfigs, so they should be moved too...


akesandgren:
Going in, thanks @boegel!

PR #9102, opened 2019-10-09
{lang}[foss/2018b] Mono v6.4.0.198

sciroc:
(created using `eb --new-pr`)

PR #9281, opened 2019-11-15
{bio}[foss/2018b] MaxQuant v1.6.10.43

SethosII:
This PR uses https://github.com/easybuilders/easybuild-easyconfigs/pull/9102.

I could need some guidance on this one. Is `modaliases` the way to go here? ~Is a `sanity_check_commands` with `maxquantcmd --help` helpful?~ (I tried this but the alias doesn't seem to be known during the sanity check) Should `source_urls`  be removed because you need to register and get a link with some hash and an expiration date so a normal download will fail due to error 403?

### DIFF
--- a/easybuild/easyconfigs/m/MaxQuant/MaxQuant-1.6.10.43-foss-2018b.eb
+++ b/easybuild/easyconfigs/m/MaxQuant/MaxQuant-1.6.10.43-foss-2018b.eb
@@ -11,7 +11,6 @@ description = """MaxQuant is a quantitative proteomics software package designed
 toolchain = {'name': 'foss', 'version': '2018b'}
 
 # requires registration at https://www.maxquant.org/download_asset/maxquant/latest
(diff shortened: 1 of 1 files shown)


smoors:
@SethosII `modaliases` is indeed fine here.
modaliases are only added to the module file, so if you want `sanity_check_commands` you need to specify the full command (you could define a local_* variable to avoid having to type it twice).
`source_urls` should indeed be removed.

SethosII:
@smoors I removed `source_urls`. `--help` and `--version` don't return 0 but 1 so there is no easy `sanity_check_commands`.

smoors:
@SethosII as a workaround, you could grep the output of the --help output, like is done in `ACT-18.0.3-Java-11.eb` (not saying that you should).

smoors:
Test report by @smoors
**SUCCESS**
See https://gist.github.com/d29da23d61aef2902eee4a71fcc6ff0a for a full test report.

smoors:
Going in, thanks @SethosII!

Issue #9182, opened 2019-10-20
Linker problems with Bazel-0.25.2-GCCcore-8.2.0 on RHEL8


terjekv:
Building Bazel causes linker problems, similar to what has been seen when using ld.gold in other situations. Building with the SYSTEM toolchain works. The relevant part of the logs are:

```
ERROR: /tmp/eb-v7mb3x3f/bazel_bmkm1c5Q/out/external/bazel_tools/third_party/ijar/BUILD:74:1: Linking of rule '@bazel_tools//third_party/ijar:zipper' failed (Exit 1)
: gcc failed: error executing command
  (cd /tmp/eb-v7mb3x3f/bazel_bmkm1c5Q/out/execroot/io_bazel && \
  exec env - \
    LD_LIBRARY_PATH=/opt/uio/modules/rhel8/easybuild/software/Java/1.8.0_221/lib:/opt/uio/modules/rhel8/easybuild/software/binutils/2.31.1-GCCcore-8.2.0/lib:/opt/uio/modules/rhel
8/easybuild/software/zlib/1.2.11-GCCcore-8.2.0/lib:/opt/uio/modules/rhel8/easybuild/software/GCCcore/8.2.0/lib/gcc/x86_64-pc-linux-gnu/8.2.0:/opt/uio/modules/rhel8/easybuild/soft
ware/GCCcore/8.2.0/lib64:/opt/uio/modules/rhel8/easybuild/software/GCCcore/8.2.0/lib \
    PATH=/opt/uio/modules/rhel8/easybuild/software/Java/1.8.0_221:/opt/uio/modules/rhel8/easybuild/software/Java/1.8.0_221/bin:/opt/uio/modules/rhel8/easybuild/software/binutils/
2.31.1-GCCcore-8.2.0/bin:/opt/uio/modules/rhel8/easybuild/software/GCCcore/8.2.0/bin:/opt/uio/modules/rhel8/easybuild/packages/EasyBuild/4.0.1/bin:/site/bin:/site/common/hosts:/u
sr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin \
    PWD=/proc/self/cwd \
  /opt/uio/modules/rhel8/easybuild/software/GCCcore/8.2.0/bin/gcc -o bazel-out/host/bin/external/bazel_tools/third_party/ijar/zipper -Wl,-S '-fuse-ld=gold' -Wl,-no-as-needed -Wl,
[...]

PR #9312, opened 2019-11-16
Fix #9274, Qt5/5.13.1-GCCcore-8.3.0 depends on JasPer.

### LINKED ISSUE #9274: Qt5/5.13.1-GCCcore-8.3.0 depends on JasPer.
I'm trying to build Qt5/5.13.1-GCCcore-8.3.0 on RHEL8 and it fails on:

```
qjp2handler.cpp:47:10: fatal error: jasper/jasper.h: No such file or directory
 #include <jasper/jasper.h>
          ^~~~~~~~~~~~~~~~~
```

JasPer is not set as a dependency for Qt5/5.13.1-GCCcore-8.3.0 and there is no JasPer/2.0.14-GCCcore-8.3.0 available. I'm mostly wondering if this is a RHEL8 issue since this hasn't been seen before? There is a jasper-devel package available in the OS, but there are no osdeps either in the package.

For what its worth, JasPer/2.0.14 builds fine with GCCcore 8.3.0 and Qt5/5.13.1-GCCcore-8.3.0 builds fine once JasPer/2.0.14-GCCcore-8.3.0 is installed and added as a dependency.

terjekv:
- Qt5 depends on JasPer, which is missing from the spec.
  - Add JasPer for GCCcore-8.3.0.
  - Resolves #9274
  - No longer relies on locally built binutils that were installed due to testing of the issues found in #9182

### DIFF
--- a/easybuild/easyconfigs/h/h5py/h5py-2.9.0-foss-2019b-Python-3.7.4.eb
+++ b/easybuild/easyconfigs/h/h5py/h5py-2.9.0-foss-2019b-Python-3.7.4.eb
@@ -1,0 +1,33 @@
+easyblock = 'PythonPackage'
+
+name = 'h5py'
+version = '2.9.0'
+versionsuffix = '-Python-%(pyver)s'
+
+homepage = 'https://www.h5py.org/'
+description = """HDF5 for Python (h5py) is a general-purpose Python interface to the Hierarchical Data Format library,
+ version 5. HDF5 is a versatile, mature scientific software library designed for the fast, flexible storage of enormous
+ amounts of data."""
+
+toolchain = {'name': 'foss', 'version': '2019b'}
+toolchainopts = {'usempi': True}
+
+source_urls = [PYPI_SOURCE]
+sources = [SOURCE_TAR_GZ]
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/q/Qt5/Qt5-5.13.1-GCCcore-8.3.0.eb
```diff
@@ -34,7 +34,8 @@ builddependencies = [
     ('gperf', '3.1'),
     ('Ninja', '1.9.0'),
     ('Python', '3.7.4'),
-    ('re2c', '1.2.1')
+    ('re2c', '1.2.1'),
+    ('JasPer', '2.0.14'),
```
boegel: @terjekv You shouldn't add `JasPer` both as build dep and runtime dep.

If a dependency is listed in `dependencies`, it will already be loaded at build time too.

This is also why the test fails, you're adding `JasPer` twice in the same dependency graph.

terjekv: Aha, thanks! I'll fix it later today!

PR #9378, opened 2019-11-26
{bio}[foss/2018b] FSL 6.0.2

casparvl:
(created using `eb --new-pr`)

Add new FSL EasyConfig and patches for FSL-6.0.2. It only runs in conjunction with an updated EasyBlock from this PR ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1860~~


casparvl:
Darn... internally, we don't really use CUDA as part of a toolchain, but as prefix, so we can update it more often than the rest of the toolchain...

I guess there shouldn't be a problem in simply removing the CUDA dep and the toolchain with fosscuda.

Regarding the missing libs: I'll update the PR to include those EasyConfigs as well.

casparvl:
Ok... just changing to foss cuda is not THAT easy, because some of my deps were for foss. Now, I need to create the equivalent EasyConfigs for the fosscuda toolchain (or move them to GCCcore-based).

No more time to do that today :(

casparvl:
Ok, finally passed the checks. I've made this PR an FSL build for CPU.

I will do a separate PR for a GPU build because I can either do 
- an FSL-6.0.2-foss-2018b-CUDA-10.0.130.eb based PR (this is what we use internally, we don't use the full ```fosscuda``` toolchain since it often saves us building a lot of deps with ```fosscuda```)
- an FSL-6.0.2-fosscuda-2018b.eb, but that requires many other deps to be built with either ```fosscuda``` or the underlying ```GCCcore```, and I didn't create EasyConfigs for those (yet).

verdurin:
Test report by @verdurin
**FAILED**
Build succeeded for 3 out of 4 (4 easyconfigs in this PR)
See https://gist.github.com/d6382649922c056398ba8560b38cbfc7 for a full test report.

verdurin:
Test report by @verdurin
**SUCCESS**
Build succeeded for 1 out of 1 (4 easyconfigs in this PR)
See https://gist.github.com/f8aee9b8cbfda54a3a7f7ec1edfc7286 for a full test report.

easybuild/easyconfigs/f/FSL/FSL-6.0.2-foss-2018b-Python-2.7.15.eb
```diff
@@ -0,0 +1,78 @@
...
+# Some tools like icmp are missing since they are built with the fslpython_install.sh script
+# which locally installs Miniconda and creates a virtual environment. This is too messy and
+# an alternative solution is necessary to install them.
+# https://github.com/easybuilders/easybuild-easyconfigs/issues/7899
+# https://fsl.fmrib.ox.ac.uk/fsl/fslwiki/FslInstallation#imcp.2Fimglob.2Fimmv_errors_after_installation
+
+name = 'FSL'
+version = '6.0.2'
+local_pyver = '2.7.15'
+versionsuffix = '-Python-%s' % local_pyver
```
verdurin: @casparvl why this unusual way of specifying the Python version?

casparvl: @verdurin You mean, why not use the template pyver?

Good question, I wasn't sure how to solve this properly, but the situation is this:

Python is not a direct dependency of ```FSL```, but of ```VTK```. ```VTK``` is then a dep for ```FSL```. Because Python is not a direct dependency the template (I think it's ```pyver```, right?) is not defined.

Nevertheless, it seemed to make sense to me to propagate the Python suffix of VTK to FSL, since loading FSL would load a python 2, and thus make it incompatible with any python 3.

It's a bit messy, but I don't know a better way to resolve it. Do you?

casparvl: Oh, and to make things even messier: FSL pulls in a virtualenv based on python3, which also get's added to the path (otherwise, the software doesn't work...). I'm not sure... I could also leave the suffix out, but neither solution feels 'clean'.

casparvl: @verdurin Could you let me know which way you prefer, including the python suffix to let ppl now it's pulling in a VTK with python suffix? Or leave it out?

boegel: @casparvl Just this is fine imho:

```python
versionsuffix = '-Python-2.7.15'
...
dependencies = [
    ...
    ('VTK', '8.1.1', versionsuffix),
```

There's no point in introducing a separate local variable for the Python version, it's mostly about only having to change it in a single place.

easybuild/easyconfigs/f/FSL/FSL-6.0.1_Fix_fsl_exec_script.patch
```diff
@@ -0,0 +1,11 @@
+diff -Nru fsl.orig/src/misc_tcl/fsl_exec.tcl fsl/src/misc_tcl/fsl_exec.tcl
```
boegel: @casparvl Filename should be `FSL-6.0.2_*.patch`?

Also, please include a short description of this patch on top (+ mention patch author or upstream reference).

Make sure you update the SHA256 checksums for the patches accordingly.

easybuild/easyconfigs/f/FSL/FSL-6.0.2_Build_extras.patch
```diff
@@ -0,0 +1,11 @@
+--- fsl.orig/extras/build	2019-11-15 10:48:52.249162071 +0100
```
boegel: @casparvl Please add missing patch description + author or upstream ref

easybuild/easyconfigs/f/FSL/FSL-6.0.2_Makefile_fixes.patch
```diff
@@ -0,0 +1,245 @@
+diff -Nru fsl.orig/config/buildSettings.mk fsl/config/buildSettings.mk
```
boegel: @casparvl Please add missing patch description + author or upstream ref

easybuild/easyconfigs/f/FSL/FSL-6.0.2-foss-2018b-Python-2.7.15.eb
```diff
@@ -0,0 +1,78 @@
...
+    ('GSL', '2.5'),
+    ('Qwt', '6.1.4'),
+]
+
+# FSLDIR needs to be defined when running postinstall to get the correct shebang
+# https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=ind1910&L=FSL&P=R86209
+postinstallcmds = [(
+    'FSLDIR=%(installdir)s PATH=%(installdir)s/fsl/fslpython/bin:$PATH '
+    '%(installdir)s/fsl/etc/fslconf/post_install.sh -f %(installdir)s/fsl; '
+)]
```
boegel: @casparvl Doesn't this belong in the `FSL` easyblock?

easybuild/easyconfigs/f/FSL/FSL-6.0.2-foss-2018b-Python-2.7.15.eb
```diff
@@ -0,0 +1,78 @@
...
+
+modextravars = {
+    'FSLOUTPUTTYPE': 'NIFTI_GZ',
+    'FSLMULTIFILEQUIT': 'TRUE',
+    'FSLTCLSH': 'tclsh',
+    'FSLWISH': 'wish8.6'
```
boegel: This can be tackled by the easyblock too?

easybuild/easyconfigs/f/FSL/FSL-6.0.2-foss-2018b-Python-2.7.15.eb
```diff
@@ -0,0 +1,78 @@
...
+}
+
+# Adding the bin from the virtualenv was the only way to get things working...
+modextrapaths = {
+    'PATH': 'fsl/fslpython/bin',
+    'PYTHONPATH': 'fsl/fslpython/envs/fslpython/lib/python3.7/site-packages/',
```
boegel: More stuff for the easyblock?

This `virtualenv` stuff is making me eyes bleed though (but I think @vanzod also broke his teeth on it recently)

casparvl: I know, it is absolutely horrid. Despite assurances from the devels that 'this should not be needed', it was the only way to make things work. I think for the devs it simply works on their system, because they use a hard-coded she-bang to the python from the virtualenv. That didn't work for me, possibly because my prefix is quite a bit longer (and has more special characters) than their 'default' prefix.

Anyway, I decided I'd rather have a horrible, working solution, than no solution at all ;-)

I believe that @verdurin was planning to get in touch with the devs at some point to improve their installation procedures...

Regarding all requests for moving things to the EasyBlock: I agree that the EasyConfig is getting a lot of stuff in there that could be in an EasyBlock. Especially since FSL already has it's own, customized EasyBlock. However, right now, I really don't have time to move that, test it, etc (I have another project that needs to be finished end of this year...).

Considering that:
- FSL seems to change it's installation repeatedly anyway (configuring the build changed quite a bit between FSL-6.0.1 and FSL-6.0.2... see the changes in the Makefile_fixes), so I'm not sure if what we would put in today would still be relevant tomorrow...
- I understood some people have been asking for an FSL-6.0.2 PR
would it be an option to simply accept this somewhat cluttered EasyConfig? We can always 'beautify' it later... Or maybe one of the core EasyBuild maintainers could pick it up and do it - though I assume you're also all super busy =)

easybuild/easyconfigs/f/FSL/FSL-6.0.2-foss-2018b-Python-2.7.15.eb
casparvl: You're right. The introduction of an extra local variable is a relic from my GPU-build, where the versionsuffix for FSL also includs some ```CUDA-...``` (which should NOT be passed on to VTK of course). I'll do it your way for this PR.

easybuild/easyconfigs/f/FSL/FSL-6.0.1_Fix_fsl_exec_script.patch
casparvl: I developed the patch originally for FSL-6.0.1, where it also applies. It fixes an error when running one of the FSL binaries, quite possibly, people simply hadn't encountered it before. I'd propose to check it in like this. If desired, we can include the patch in any FSL-6.0.1 EasyConfig later on...

boegel: :+1:

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 3 out of 4 (4 easyconfigs in this PR)
generoso-2 - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2660 v4 @ 2.00GHz, Python 3.6.8
See https://gist.github.com/d2a103fc99ae0a3e7f6e155cd86f7cfe for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3124.skitty.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/59e238b33702405cc84b2e2e7284fbe3 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 20 out of 20 (4 easyconfigs in this PR)
node2454.golett.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/b68183c6e06dd4a39ea427f91aeecf12 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
generoso-2 - Linux centos linux 7.6.1810, Intel(R) Xeon(R) CPU E5-2660 v4 @ 2.00GHz, Python 3.6.8
See https://gist.github.com/4ba4a4dd4ec10a47b877df98144103ff for a full test report.

boegel:
Going in, thanks @casparvl!

PR #9491, opened 2019-12-13
{bio}[GCCcore/8.3.0] Varlociraptor v1.5.0

cmeesters:
(created using `eb --new-pr`)


cmeesters:
What is going on, here? Why should a contributor consider something like https://packages.microsoft.com/repos/microsoft-ubuntu-bionic-prod and the respective setups? And if so, how?

boegel:
@cmeesters That was an issue in GitHub itself, fixed with https://github.com/easybuilders/easybuild-easyconfigs/pull/9492 (will be merged soon).

Once it is merged, `eb --sync-pr-with-develop 9491` is an easy way to fix the issue in this PR (or just fix the style issues once #9492 is merged).

cmeesters:
not sure how to satisfy dependencies with futures and current easyconfigs overlapping, see PR 9523

boegel:
I'm closing this PR, since this easyconfig should be redone using the new `Cargo` easyblock, so we have full control over the installation.
If not, the installation will break over time, as was the case with `bamtofastq` (cfr. #15636) before it changed to use the `Cargo` easyblock (see #17595 + https://github.com/easybuilders/easybuild-easyblocks/pull/2902)

PR #9503, opened 2019-12-13
add PETSc foss-2019b easyconfig

jijn:
Add easyconfig file for PETSc with foss 2019b toolchain, together with related easyconfig files: ParMETIS-4.0.3, Hypre-2.18.2.


boegel:
@jijn Thank you very much for your contribution.
GitHub tells me this is your first pull request ever, we're honored!

Why did you close the pull request again? It seems like a couple of minor fixes would be enough to make the easyconfigs test suite pass...

Do let us know if you need any help.

Issue #9306, opened 2019-11-15
Missing/Wrong requirements for TF2


Flamefire:
I ran into crashes and issues like https://github.com/tensorflow/tensorflow/issues/32319 using the current TF 2 EC and traced it back to wrong dependencies. Most notably it requires gast==0.2.2

I'm currently in the process of fixing them. For reference the dependencies required for TF2 are:

```
REQUIRED_PACKAGES = [
    'absl-py >= 0.7.0',
    'astor >= 0.6.0',
    'gast == 0.2.2',
    'google_pasta >= 0.1.6',
    'keras_applications >= 1.0.8',
    'keras_preprocessing >= 1.0.5',
    'numpy >= 1.16.0, < 2.0',
    'opt_einsum >= 2.3.2',
    'six >= 1.10.0',
    'protobuf >= 3.6.1',
    'tensorboard >= 2.0.0, < 2.1.0',
    'tensorflow_estimator >= 2.0.0, < 2.1.0',
    'termcolor >= 1.1.0',
    'wrapt >= 1.11.1',
    'grpcio >= 1.8.6',
    'wheel >= 0.26'
]
```

And current master (future):

```
REQUIRED_PACKAGES = [
    'absl-py >= 0.7.0',
    'astor >= 0.6.0',
    'gast == 0.2.2',
    'google_pasta >= 0.1.8',
    'keras_applications >= 1.0.8',
    'keras_preprocessing >= 1.1.0',
    'numpy >= 1.16.0, < 2.0',
    'opt_einsum >= 2.3.2',
    'protobuf >= 3.8.0',
    'tensorboard >= 2.0.0, < 2.1.0',
    'tensorflow_estimator >= 2.0.0, < 2.1.0',
    'termcolor >= 1.1.0',
    'wrapt >= 1.11.1',
    # python3 requires wheel 0.26
    'wheel >= 0.26;python_version>="3"',
    'six >= 1.12.0',
    'grpcio >= 1.8.6'
]
```

I'll create a PR on Monday with a updated and verified version.

Ccing @boegel 

Edit:
Using the current EC and issuing `pip install gast==0.2.2` yields:
```
Collecting gast==0.2.2
ERROR: tensorflow 2.0.0 requires google-pasta>=0.1.6, which is not installed.
ERROR: tensorflow 2.0.0 requires opt-einsum>=2.3.2, which is not installed.
```

[...]

PR #9516, opened 2019-12-15
require that sanity_pip_check is enabled in new/changed easyconfigs

boegel:
Support for running `pip check` during the sanity check to ensure that all required Python packages are installed was added in EasyBuild v4.1.0 (see #1853).

We should start requiring it in contributed easyconfigs, since it can prevent shipping broken easyconfigs (see https://github.com/easybuilders/easybuild-easyconfigs/issues/9306 for example).

You can enable the running of `pip check` either via:

```python
sanity_pip_check = True
```

or

```python
exts_default_options = {'sanity_pip_check': True}
```

(it can also be enabled for individual extensions, but there's little point in doing that imho, so no need to cover that in the check here)

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -542,6 +542,7 @@ class EasyConfigTest(TestCase):
             ec_fn = os.path.basename(ec.path)
             easyblock = ec.get('easyblock')
             exts_defaultclass = ec.get('exts_defaultclass')
+            exts_default_options = ec.get('exts_default_options', {})
 
             download_dep_fail = ec.get('download_dep_fail')
             exts_download_dep_fail = ec.get('exts_download_dep_fail')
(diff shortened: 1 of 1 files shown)


Flamefire:
Great! But we should also fix https://github.com/easybuilders/easybuild-easyblocks/issues/1877 to avoid failures due to python using the user home.

boegel:
https://github.com/easybuilders/easybuild-easyblocks/issues/1877 is being fixed with https://github.com/easybuilders/easybuild-easyblocks/pull/1891

@smoors This is good to go imho...

Flamefire:
Thanks, LGTM. However I suggest to rework the assert messages. The current one here is e.g. "sanity_pip_check is enabled in X". So if you see the test failing on travis this will be what is printed. The problem: The opposite is the case: The check failed because it is not enabled. This made me loose some time chasing for the failure because I couldn't understand the message. Better: "sanity_pip_check should be enabled in X" or "sanity_pip_check is not enabled in X"

smoors:
@Flamefire I understand your confusion, but all other error messages are constructed in the same way.
So maybe file an issue on that?

Flamefire:
Ok: https://github.com/easybuilders/easybuild-easyconfigs/issues/9542

smoors:
Test report by @smoors
**SUCCESS**
login2.cerberus.os - Linux centos linux 7.6.1810, Intel(R) Xeon(R) Gold 6126 CPU @ 2.60GHz, Python 2.7.5
See https://gist.github.com/c8ff78e72b7d37aa98ea8d20075fff7f for a full test report.

smoors:
Going in, thanks @boegel!

PR #9504, opened 2019-12-14
{numlib}[GCC/8.3.0,iccifort/2019.5.281] GSL v2.6

boegel:
(created using `eb --new-pr`)

PR #9523, opened 2019-12-17
{numlib}[GCCcore/8.3.0] GSL v2.5

cmeesters:
(created using `eb --new-pr`)


cmeesters:
Hi,

This is strange: binutils 2.32 ist NOT a build-dep of this easyconfig. It is an implicit dep in GCCcore 8.3.0, yet this part of eb v 4.0.1 (GCCcore-8.3.0.eb) and in turn satisfied by binutils-2.32-GCCcore-8.3.0.eb

So, why is the test complaining about a missing dependency, one, which is not at all specified by the respective easyconfig, yet should be satisfied by the toolchain?

Best,
Christian

boegel:
@cmeesters This PR is superseded by the GSL 2.6 easyconfigs in #9504

W.r.t. the `binutils` in the failing tests: `binutils` is only a *build* dep for `GCCcore` (we use a `binutils` built with the system compiler to build `GCCcore`).

When using `GCCcore` as a toolchain, you should include the `binutils` build with that `GCCcore` as a build dependency, that's what the tests are asserting.

In this case however, the `GCC` bundle (which is just `GCCcore`+`binutils`) is a better fit for `GSL`, which is performance sensitive (so we also install it with `iccifort` as toolchain).

Using `GCCcore` as toolchain should only be done for tools that are not performance-sensitive (since then we won't rebuild with Intel compilers if the tool is required as a dependency, but use the build done with `GCCcore` both for applications built with `foss` or `intel` toolchain (or any other that is based on `GCCcore`).

cmeesters:
As the pull request is superseeded by the PR for v2.6 (release in August release, which I overlooked), I am closing this PR. Sorry for the effort and thanks for the explanation.

PR #9547, opened 2019-12-19
WIP: Fixes for Bazel and TensorFlow for glibc 2.30 gettid function

edmondac:
(created using `eb --new-pr`)


edmondac:
That patch for TensorFlow doesn't seem to do the business :-( I'll have to leave this for a bit now.

PR #9556, opened 2019-12-19
{lang}[GCCcore-8.2.0] Perl-5.30.0

robqiao:
Perl-5.30.0


boegel:
@robqiao Do you have a particular use case in mind for this, since we already have `Perl-5.30.0-GCCcore-8.3.0.eb`?

We prefer sticking to a single `Perl` easyconfig per "generation" of easyconfigs in the central repository...

robqiao:
> @robqiao Do you have a particular use case in mind for this, since we already have `Perl-5.30.0-GCCcore-8.3.0.eb`?
> 
> We prefer sticking to a single `Perl` easyconfig per "generation" of easyconfigs in the central repository...

We had some issues with the newly released bioperl lately, GCC-8.2.0 seems to be more stable in this case. Again GCC-8.3.0 seems a bug-fix release, so I suspect this may be a local config issue on GCC-8.3.0. I am happy to forgo this PR. @boegel feel free to close this PR as you see fit.

boegel:
@robqiao Which `BioPerl` easyconfig did you have trouble with, and what kind of problems did you run into, do you have more details?

PR #9583, opened 2019-12-27
{bio}[foss/2019a] SAMtools 1.9

jakevc:
Add samtools-foss-2019a, and update cURL dependency.

### DIFF
--- a/easybuild/easyconfigs/c/cURL/cURL-7.66.0-foss-2019a.eb
+++ b/easybuild/easyconfigs/c/cURL/cURL-7.66.0-foss-2019a.eb
@@ -1,0 +1,31 @@
+easyblock = 'ConfigureMake'
+
+name = 'cURL'
+version = '7.66.0'
+
+homepage = 'http://curl.haxx.se'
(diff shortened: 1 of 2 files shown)


boegel:
@jakevc Thank you very much for your contribution!

However, this `SAMtools` easyconfig is basically identical to the already existing `SAMtools-1.9-GCC-8.2.0-2.31.1.eb` easyconfig. Since `SAMtools` doesn't require MPI or BLAS/LAPACK, there's no reason to use a full toolchain like `foss` rather than a compiler-only toolchain like `GCC`.
Did you have a particular reason for using `foss/2019a` as a toolchain rather than using the existing `SAMtools-1.9-GCC-8.2.0-2.31.1.eb`?

W.r.t. the failing tests: we try to stick to a single version for dependencies per "generation" of easyconfigs (`2019a` in this case), since that significantly reduces the change of running into version conflicts when modules installed with an easyconfig file of the same "generation" are loaded together.

Do let us know if something isn't clear...

jakevc:
Okay Excellent. Thanks for your explanation that clears things up a bit. There is not need for this then.

PR #9628, opened 2020-01-08
Make postinstallcmds independent of PWD in OpenCV 3.1.0 easyconfigs

Flamefire:
(created using `eb --new-pr`)

- [x] Requires but contains ~~#9634~~

PR #9634, opened 2020-01-09
Allow missing python versionsuffix for existing easyconfig files changed in PRs

Flamefire:
Otherwise this would break a lot of ECs

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -530,7 +530,7 @@ class EasyConfigTest(TestCase):
         checksum_issues = check_sha256_checksums(retained_changed_ecs, whitelist=whitelist)
         self.assertTrue(len(checksum_issues) == 0, "No checksum issues:\n%s" % '\n'.join(checksum_issues))
 
-    def check_python_packages(self, changed_ecs):
+    def check_python_packages(self, changed_ecs, added_ecs_filenames):
         """Several checks for easyconfigs that install (bundles of) Python packages."""
 
         # These packages do not support installation with 'pip'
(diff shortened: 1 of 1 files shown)


Flamefire:
This is a solution for the failure in https://github.com/easybuilders/easybuild-easyconfigs/pull/9628 and being tested there.

test/easyconfigs/easyconfigs.py
```diff
@@ -653,6 +659,11 @@ def check_https(self, changed_ecs):
 
     def test_changed_files_pull_request(self):
         """Specific checks only done for the (easyconfig) files that were changed in a pull request."""
+        def get_eb_files_from_diff(filter):
```
boegel: @Flamefire `filter` is a Python keyword (see syntax highlighting in the diff), so let's use `filter_char` instead?

test/easyconfigs/easyconfigs.py
```diff
@@ -682,16 +693,18 @@ def test_changed_files_pull_request(self):
...
-                cmd = "git diff --name-only --diff-filter=AM %s...HEAD" % target_branch
-                out, ec = run_cmd(cmd, simple=False)
-                changed_ecs_filenames = [os.path.basename(f) for f in out.strip().split('\n') if f.endswith('.eb')]
-                print("\nList of changed easyconfig files in this PR: %s" % '\n'.join(changed_ecs_filenames))
+                changed_ecs_filenames = get_eb_files_from_diff(filter='M')
+                added_ecs_filenames = get_eb_files_from_diff(filter='A')
+                if changed_ecs_filenames:
+                    print("\nList of changed easyconfig files in this PR: %s" % '\n'.join(changed_ecs_filenames))
+                if added_ecs_filenames:
+                    print("\nList of adeed easyconfig files in this PR: %s" % '\n'.join(added_ecs_filenames))
```
boegel: `adeed` -> `added`

test/easyconfigs/easyconfigs.py
Flamefire: I used diff_filter as per the git command

test/easyconfigs/easyconfigs.py
Flamefire: fixed

PR #9643, opened 2020-01-09
{lib}[GCCcore/8.2.0] ICU 64.2 - Depend on Python3 instead of 2

crubb:
`ICU-64.2` has a build dependency on `Python-2.7.15`, but prefers Python 3 in the build process. This is a problem, when there is a bare Python 3 installed somewhere in the System, and will break the build process:

```
Not rebuilding data/rules.mk, assuming prebuilt data in data/in
Spawning Python to generate test/testdata/rules.mk...
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/software/easybuild4/build/ICU/64.2/GCCcore-8.2.0/icu/source/data/buildtool/__main__.py", line 19, in <module>
    import BUILDRULES
  File "/software/easybuild4/build/ICU/64.2/GCCcore-8.2.0/icu/source/test/testdata/BUILDRULES.py", line 4, in <module>
    from distutils.sysconfig import parse_makefile
ModuleNotFoundError: No module named 'distutils.sysconfig'
configure: error: Python failed to run; see above error.
```

Therefore, I propose to update the `builddependencies` to `Python-3.7.2`.

### DIFF
--- a/easybuild/easyconfigs/g/Ghostscript/Ghostscript-9.50-GCCcore-8.3.0.eb
+++ b/easybuild/easyconfigs/g/Ghostscript/Ghostscript-9.50-GCCcore-8.3.0.eb
@@ -1,0 +1,46 @@
+easyblock = 'ConfigureMake'
+
+name = 'Ghostscript'
+version = '9.50'
+
+homepage = 'https://ghostscript.com'
+description = """Ghostscript is a versatile processor for PostScript data with the ability to render PostScript to
+ different targets. It used to be part of the cups printing stack, but is no longer used for that."""
+
+toolchain = {'name': 'GCCcore', 'version': '8.3.0'}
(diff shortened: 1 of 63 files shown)


smoors:
@crubb if I understand correctly, this will still use the system Python 3.6 (if available), right?
we have to somehow make sure the EB version is used.

crubb:
@smoors ICU's `./configure` just checks for a python3 binary somewhere in the `$PATH`. As soon as an EB Python 3 module was loaded, that one was used.

crubb:
Botched, cleanup in new PR.

PR #9648, opened 2020-01-10
{bio}[foss/2019a] gradunwarp v1.1.0 w/ Python 3.7.2

crubb:
(created using `eb --new-pr`)

PR #9653, opened 2020-01-11
{bio}[foss/2019a] gradunwarp 1.1.0 w/ Python 3.7.2 - Add Python 3 compatibility patches

crubb:
Add missing Python 3 compatibility patches for #9648.

### DIFF
--- a/easybuild/easyconfigs/g/gradunwarp/gradunwarp-1.1.0-foss-2019a-HCP-Python-3.7.2.eb
+++ b/easybuild/easyconfigs/g/gradunwarp/gradunwarp-1.1.0-foss-2019a-HCP-Python-3.7.2.eb
@@ -11,7 +11,15 @@ toolchain = {'name': 'foss', 'version': '2019a'}
 
 source_urls = ['https://github.com/Washington-University/gradunwarp/archive']
 sources = ['v%(version)s.tar.gz']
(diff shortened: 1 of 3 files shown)


boegel:
@crubb Hmm, wouldn't it be better to stick to Python 2 for now, and wait for a new release with official Python 3 support (once https://github.com/Washington-University/gradunwarp/pull/4 has been merged)?

crubb:
The PR https://github.com/Washington-University/gradunwarp/pull/4 is from 2015. I pinged them, let's see.

boegel:
@crubb Since we're very close to releasing EasyBuild v4.1.1, we have two options:

i) merge this PR and ship/maintain a fairly big patch that is not approved yet by upstream (I'm not in favor of doing so)
ii) removing the `gradunwrap` easyconfig that was added in #9648 from `develop` (so it's not included in the EasyBuild v4.1.1 release), since it's broken without the patch

Thoughts?

crubb:
@boegel I'd go for ii) and keep bugging them (even though it has been a few years). Should I submit a PR to remove the broken `gradunwarp`?

boegel:
> Should I submit a PR to remove the broken `gradunwarp`?

@crubb Please do!

boegel:
`gradunwarp` easyconfig using Python 3 was removed in #9683, moving this PR to the next release (after v4.1.1)...

crubb:
A bug in the Python 3 port was found and fixed. Will not be released as 1.1.0 anymore.

PR #9659, opened 2020-01-13
[double-conversion] Use CMake for building

Flamefire:
Reasoning:   
Using Scons (like currently done) will be using the system compiler instead of the toolchain set in the EasyConfig. This will completely fail during compilation when there is no system compiler (namely g++) installed. It may also fail, when this is linked into another application using a different compiler or compiler version due to different flags, ABI etc.

Furthermore Scons doesn't install the headers so that was done manually but to a wrong subdirectory. The correct path is `include/double-conversion/double-conversion.h` not  `include/double-conversion.h`. As the result dependents likely won't find it resulting in either build errors (e.g. when using plain `#include <double-conversion/double-conversion.h>`) or silently using an alternative version. E.g. Qt has a version bundled with its sources which is used when no "system installed" (in this case: by EasyBuild) version is found.

I verified that prior to this change Qt uses the bundled version and after this change the EB version.

### DIFF
--- a/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
+++ b/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
@@ -1,3 +1,7 @@
+# NOTE:
+#    This does use the system compiler if any and WILL fail if no system compiler exists.
+#    Dependents may or may not be able to find this, e.g. Qt will silently fallback to a version included in Qt
+#    For future ECs use the CMake based build
 easyblock = 'SCons'
 
 name = 'double-conversion'
--- a/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
+++ b/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
@@ -1,4 +1,4 @@
-easyblock = 'CMakeMake'
+easyblock = 'SCons'
 
 name = 'double-conversion'
 version = '3.0.3'
@@ -13,24 +13,15 @@ sources = ['v%(version)s.tar.gz']
 checksums = ['ac84e00c19fdb5ee20c01feb18b0fa2e938bf4b8cf71d43bab62502c8b65420a']
 
 builddependencies = [
-    ('CMake', '3.12.1', '', ('GCCcore', '6.4.0')),
+    ('SCons', '3.0.1', '-Python-3.6.4'),
 ]
 
-separate_build_dir = True
-
-build_type = 'Release'
-
-# Build static lib, static lib with -fPIC and shared lib
-configopts = [
-    '',
-    '-DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_STATIC_LIBRARY_SUFFIX_CXX=_pic.a',
-    '-DBUILD_SHARED_LIBS=ON'
-]
+installopts = "DESTDIR=%(installdir)s prefix='' && "
+installopts += "mkdir %(installdir)s/include && cp double-conversion/*.h %(installdir)s/include"
 
 sanity_check_paths = {
-    'files': ['include/double-conversion/double-conversion.h', 'include/double-conversion/utils.h',
-              'lib/libdouble-conversion.a', 'lib/libdouble-conversion_pic.a',
-              'lib/libdouble-conversion.%s' % SHLIB_EXT],
+    'files': ['include/double-conversion.h', 'include/utils.h', 'lib/libdouble-conversion.a',
+              'lib/libdouble-conversion.%s' % SHLIB_EXT, 'lib/libdouble-conversion_pic.a'],
     'dirs': [],
 }
 
--- a/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
+++ b/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
@@ -1,4 +1,4 @@
-easyblock = 'SCons'
+easyblock = 'CMakeMake'
 
 name = 'double-conversion'
 version = '3.0.3'
@@ -13,15 +13,24 @@ sources = ['v%(version)s.tar.gz']
 checksums = ['ac84e00c19fdb5ee20c01feb18b0fa2e938bf4b8cf71d43bab62502c8b65420a']
 
 builddependencies = [
-    ('SCons', '3.0.1', '-Python-3.6.4'),
+    ('CMake', '3.12.1', '', ('GCCcore', '6.4.0')),
 ]
 
-installopts = "DESTDIR=%(installdir)s prefix='' && "
-installopts += "mkdir %(installdir)s/include && cp double-conversion/*.h %(installdir)s/include"
+separate_build_dir = True
+
+local_buildtype = '-DCMAKE_BUILD_TYPE=Release '
+configopts = [
+    local_buildtype + c for c in (
+        '',
+        '-DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_STATIC_LIBRARY_SUFFIX_CXX=_pic.a',
+        '-DBUILD_SHARED_LIBS=ON'
+    )
+]
 
 sanity_check_paths = {
-    'files': ['include/double-conversion.h', 'include/utils.h', 'lib/libdouble-conversion.a',
-              'lib/libdouble-conversion.%s' % SHLIB_EXT, 'lib/libdouble-conversion_pic.a'],
+    'files': ['include/double-conversion/double-conversion.h', 'include/double-conversion/utils.h',
+              'lib/libdouble-conversion.a', 'lib/libdouble-conversion_pic.a',
+              'lib/libdouble-conversion.%s' % SHLIB_EXT],
     'dirs': [],
 }
 
--- a/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
+++ b/easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
@@ -20,6 +20,7 @@ separate_build_dir = True
 
 build_type = 'Release'
 
+# Build static lib, static lib with -fPIC and shared lib
 configopts = [
     '',
     '-DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_STATIC_LIBRARY_SUFFIX_CXX=_pic.a',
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
```diff
@@ -13,15 +13,24 @@ sources = ['v%(version)s.tar.gz']
...
+        '',
+        '-DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_STATIC_LIBRARY_SUFFIX_CXX=_pic.a',
+        '-DBUILD_SHARED_LIBS=ON'
+    )
+]
 
 sanity_check_paths = {
-    'files': ['include/double-conversion.h', 'include/utils.h', 'lib/libdouble-conversion.a',
-              'lib/libdouble-conversion.%s' % SHLIB_EXT, 'lib/libdouble-conversion_pic.a'],
+    'files': ['include/double-conversion/double-conversion.h', 'include/double-conversion/utils.h',
```
boegel: @Flamefire How come the header files are now in a `double-conversion` subdirectory?

Flamefire:
Sure:
1. It is using `scons` and nobody should be using this ;)
2. It doesn't do what the EC makes you believe it does: It doesn't build with the selected toolchain due to its use of `scons`. See discussions on Slack: Scons resets the complete environment
3. Instead of painfully beating scons to do what you want, use CMake 

(Meant to be a humorous attempt in relieving my headache I got from smashing my head against the desk after running into this)

easybuild/easyconfigs/d/double-conversion/double-conversion-3.0.3-foss-2018a.eb
Flamefire: Because this is the correct directory. Previously scons didn't even install the headers and they were copied manually to a directory which I see no reasoning of, see https://github.com/easybuilders/easybuild-easyconfigs/pull/9659/files#diff-113da115196d50f313926fb7149ba44dL20

This is the directory CMake installs the headers to and this library itself used to include the headers by `double-conversion/double-conversion.h` format, although this has been changed, see https://github.com/google/double-conversion/commit/04c6f0a70336b4c4a71e25a48f9ee58ec0a9321d. Hence I'm sure that this is now the correct directory layout.

boegel: Hmm, if this was wrong, I wonder why this didn't break the `Qt5` easyconfigs where `double-conversion` is included as a dependency?

Flamefire: Let me double-check (no pun intended)

Flamefire: Examples from grepping the source for "double-conversion":

```
qtbase/src/3rdparty/double-conversion/bignum-dtoa.cc:#include <double-conversion/ieee.h>
qtbase/src/corelib/configure.json:            "headers": "double-conversion/double-conversion.h",

```
But
`qtwebengine/src/3rdparty/chromium/third_party/icu/BUILD.gn:    "source/i18n/double-conversion.h",`

From the lines in configure.json I deduce that this here is the correct layout and previously it used the included version of double-conversion

Flamefire: Hm, has the Qt5 EC/EB been recently changed? I can't see the configure summary for the Qt5 I installed recently but rebuilding shows it and correctly mentions "Using system DoubleConversion ........ yes" But I'm comparing the log created during build in/tmp and the one installed in the easybuild folder. Maybe the latter doesn't contain the config summary for some reason? Currently rebuilding both QTs using double-conversion and will check what they install and use

Flamefire: Ok seems whether Qt configure outputs a summary depends on the moon phase. Run it twice, got one with and one without the summary :/

Flamefire: Anyway: I rebuild Qt on top of this and ldd shows that the correct libdouble-conversion is used for both Qt5 ECs that have this dependency

Flamefire: And finally: TensorFlow can use this and it does use `#include "double-conversion/double-conversion.h"`

Flamefire:
@boegel Any update on this? This is more correct and works as (initially) intended but has been sitting for a month without update now.

Flamefire:
Test report by @Flamefire
**SUCCESS**
taurusi6258.taurus.hrsk.tu-dresden.de - Linux RHEL 7.7, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 2.7.5
See https://gist.github.com/352f985641f8caea5137ffb59143273f for a full test report.

easybuild/easyconfigs/d/double-conversion/double-conversion-3.1.4-GCCcore-8.3.0.eb
```diff
@@ -14,15 +14,24 @@ checksums = ['95004b65e43fefc6100f337a25da27bb99b9ef8d4071a36a33b5e83eb1f82021']
...
+separate_build_dir = True
+
+local_buildtype = '-DCMAKE_BUILD_TYPE=Release '
+configopts = [
+    local_buildtype + c for c in (
+        '',
+        '-DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_STATIC_LIBRARY_SUFFIX_CXX=_pic.a',
+        '-DBUILD_SHARED_LIBS=ON'
+    )
+]
```
lexming: I'm all in for avoiding repetition, but in this case I would just use a pure list, it is shorter an easier to read. Additionally, could you add a comment explaining what is each compilation doing? I see that last one adds shared libs, but what is the different between first and second one?
```suggestion
configopts = [
    '-DCMAKE_BUILD_TYPE=Release',
    '-DCMAKE_BUILD_TYPE=Release -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_STATIC_LIBRARY_SUFFIX_CXX=_pic.a',
    '-DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON'
]
```

easybuild/easyconfigs/d/double-conversion/double-conversion-3.1.4-GCCcore-8.3.0.eb
```diff
@@ -14,15 +14,24 @@ checksums = ['95004b65e43fefc6100f337a25da27bb99b9ef8d4071a36a33b5e83eb1f82021']
...
 
 sanity_check_paths = {
-    'files': ['include/double-conversion.h', 'include/utils.h', 'lib/libdouble-conversion.a',
-              'lib/libdouble-conversion.%s' % SHLIB_EXT, 'lib/libdouble-conversion_pic.a'],
+    'files': ['include/double-conversion/double-conversion.h', 'include/double-conversion/utils.h',
+              'lib/libdouble-conversion.a', 'lib/libdouble-conversion_pic.a',
+              'lib/libdouble-conversion.%s' % SHLIB_EXT],
```
lexming: I would expand the sanity checks
```suggestion
    'files': ['include/double-conversion/%s.h' % h for h in ['bignum', 'cached-powers', 'diy-fp', 'double-conversion',
                                                             'fast-dtoa', 'fixed-dtoa', 'ieee', 'strtod', 'utils']] +
             ['lib/libdouble-conversion.%s' % e for e in ['a', SHLIB_EXT]] + ['lib/libdouble-conversion_pic.a'],
```

Flamefire: I actually disagree here. It should be a sanity check that stuff is installed to correct location. Checking "include/double-conversion/double-conversion.h" is enough for the headers (one can assume the build systems installs the rest of the headers to the same location)

As for the list comprehension in lib/*: With the same argument as you have in https://github.com/easybuilders/easybuild-easyconfigs/pull/9659#discussion_r393257522 I'd keep it a simple list for clarity

easybuild/easyconfigs/d/double-conversion/double-conversion-3.1.4-GCCcore-8.3.0.eb
Flamefire: I agree. Actually there is a better way now using `build_type` which wasn't available when I made this. Changed here: https://github.com/easybuilders/easybuild-easyconfigs/pull/9659/commits/5b23e0b6578cf140422f45f29206eba4ee0f3ab9

> I see that last one adds shared libs, but what is the different between first and second one?

I added a comment. The difference is that the first builds a shared library without -fPIC. There seems to be use for a `libdouble-conversion.a` and `libdouble-conversion_pic.a` and I don't want to break this (done by the original Scons too)

easybuild/easyconfigs/d/double-conversion/double-conversion-3.1.4-GCCcore-8.3.0.eb
lexming: Ok, I will not insist as it is a minor issue. But just to be clear, the point is to detect a partial installation of the files that may occur due to network or i/o issues. It might be very unlikely, but it does not cost much to add those checks.

lexming:
Test report by @lexming
**SUCCESS**
login2.cerberus.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6126 CPU @ 2.60GHz, Python 2.7.5
See https://gist.github.com/72943bc16e39195a4778de6d4ef4a839 for a full test report.

lexming:
Build of `Qt5-5.12.3-GCCcore-8.2.0.eb` properly binds to `double-conversion` of this PR.

```
[...]
Checking for statx() in libc... no
Checking for 64 bit atomics... yes
Checking for DoubleConversion... yes
Checking for PCRE2... yes
Checking for O_CLOEXEC... yes
[...]
```

And `ldd libQt5Core.so` shows the link with `double-conversion` from EB.

lexming:
Same results for `Qt5-5.13.1-GCCcore-8.3.0.eb`

```
Checking for statx() in libc... no
Checking for 64 bit atomics... yes
Checking for DoubleConversion... yes
Checking for PCRE2... yes
Checking for O_CLOEXEC... yes
[...]
Qt Core:
  DoubleConversion ....................... yes
    Using system DoubleConversion ........ yes
```

lexming:
Going in, thanks @Flamefire !

PR #9683, opened 2020-01-15
remove gradunwarp easyconfig with Python 3 (since it needs a patch)

crubb:
Remove the `*-Python-3.7.2.eb` until the Python 3 compatibility patches for gradunwarp 1.1.0 are accepted upstream (#9648 and #9653).


boegel:
merging into `4.1.x` branch which is used for the upcoming EasyBuild v4.1.1 release, `develop` will be updated accordingly after the release...

PR #9774, opened 2020-01-23
Adding CalculiX easyconfig file

rtripath89:


migueldiascosta:
thanks @rtripath89 

the source code seems to be available, it would be better to build from it?

Even if there is a reason not to build from source, this binary requires shared library libgfortran.so.4, so using the `SYSTEM` toolchain only works when the system compiler is GCC 7

jfgrimm:
@rtripath89 Apologies for this PR going stale. I'm closing this since it's been superceded by a from-source build in #16805

PR #9781, opened 2020-01-24
Fix CPPFLAGS includes for build script of libsndfile

Simontuk:
libsndfile was not building for me with the two following errors occuring:
Making all in M4
make[1]: Entering directory `~/.local/easybuild/build/libsndfile/1.0.28/GCCcore-8.3.0/libsndfile-1.0.28/M4'
make[1]: Nothing to be done for `all'.
make[1]: Leaving directory `~/.local/easybuild/bu (took 28 sec)

sndfile.cc:21:10: fatal error: sndfile.h: No such file or directory
 #include "sndfile.h"

The include path for the sndfile.h needed for the make command in the build step.
Appending the Include for the build directory to the build command solved the issue for me, but please check if this is viable as I'm a beginner with easybuild.

### DIFF
--- a/easybuild/easyconfigs/l/libsndfile/libsndfile-1.0.28-GCCcore-8.3.0.eb
+++ b/easybuild/easyconfigs/l/libsndfile/libsndfile-1.0.28-GCCcore-8.3.0.eb
@@ -13,7 +13,7 @@ source_urls = ['http://www.mega-nerd.com/libsndfile/files/']
 sources = [SOURCE_TAR_GZ]
 checksums = ['1ff33929f042fa333aed1e8923aa628c3ee9e1eb85512686c55092d1e5a9dfa9']
 
-buildopts = ' CPPFLAGS="-I%(installdir)s/include/ $CPPFLAGS" '
+preconfigopts = 'CPPFLAGS="-I%(builddir)s/%(name)s-%(version)s/src/ $CPPFLAGS"'
 
 builddependencies = [('binutils', '2.32')]
 
--- a/easybuild/easyconfigs/l/libsndfile/libsndfile-1.0.28-GCCcore-8.3.0.eb
+++ b/easybuild/easyconfigs/l/libsndfile/libsndfile-1.0.28-GCCcore-8.3.0.eb
@@ -13,6 +13,8 @@ source_urls = ['http://www.mega-nerd.com/libsndfile/files/']
 sources = [SOURCE_TAR_GZ]
 checksums = ['1ff33929f042fa333aed1e8923aa628c3ee9e1eb85512686c55092d1e5a9dfa9']
 
+buildopts = ' CPPFLAGS="-I%(installdir)s/include/ $CPPFLAGS" '
+
 builddependencies = [('binutils', '2.32')]
 
 sanity_check_paths = {


migueldiascosta:
@Simontuk thanks for reporting this

on one hand, I think the reason you are hitting this problem and others haven't is because `sndfile.cc` relates to Octave support, you probably have it installed in your system (?) 

In the generic case where this is not intended, we should probably add `--enable-octave=no` to the `configopts`

If you are (or someone else is) interested in the Octave support, then we should create a new easyconfig with Octave as a dependency

Simontuk:
Hi @migueldiascosta 
thanks for your reply, after checking my commit I figured it actually did not fix my issue.

I checked the relation to octave, and as far as I see my issue is not directly caused by octave. 
Octave is installed in my system, but I do not require it. 
I checked the configure and build process again, and as far as I can see, the sndfile.h, that I found was missing was put into the builddir at following location:
src/sndfile.h

The default configure and make process without providing the prefix or other flags were were able to locate the file.
But when configuring with easybuild this file could not be located. I was not sure if this is an issue with my setup, but I found the following fix to work with a completely new easybuild module folder:

`preconfigopts = 'CPPFLAGS="-I%(builddir)s/%(name)s-%(version)s/src/ $CPPFLAGS"'
`

I think this reliably fixes my issue, regardless of octave support. 
But explicitly creating an easyconfig of this with octave support could also be useful for some.
Maybe this can help others as well.

migueldiascosta:
@Simontuk sorry to insist, but I can reproduce this problem only when octave installed, `libsndfile`'s configure script detects and enables support for it automatically.

In that case, adding `configopts = '--enable-octave=no'` solves the problem, and since octave support is not intended, this seems to be the best solution here. Can you try it?

It's true that in order to enable octave support one would have to add the path to the header file as you propose, but that should go in a separate easyconfig, as discussed.

joopjaeng-outlook:
> libsndfile was not building for me with the two following errors occuring:
> Making all in M4
> make[1]: Entering directory all'.
> make[1]: Leaving directory `~/.local/easybuild/bu (took 28 sec)`~/.local/easybuild/build/libsndfile/1.0.28/GCCcore-8.3.0/libsndfile-1.0.28/M4' make[1]: Nothing to be done for `
> 
> sndfile.cc:21:10: fatal error: sndfile.h: No such file or directory
> #include "sndfile.h"
> 
> The include path for the sndfile.h needed for the make command in the build step.
> Appending the Include for the build directory to the build command solved the issue for me, but please check if this is viable as I'm a beginner with easybuild.

boegel:
@Simontuk I agree with @migueldiascosta here, if Octave support is not intended we should instruct `libsndfile` to not pick up on it if it happens to be available through an OS package.

If Octave support is desired, when Octave should be added as an EasyBuild dependency (so it's a recent version, and it's properly optimized for the processors it will be used on, as opposed to the OS package which is 99% a generic build, and hence potentially much slower).
In that case your fix would still be relevant of course...

Please let us know how you'd like to proceed. If you won't pursue this, we can take over by adding `configopts = '--enable-octave=no'` in all (recent) `libsndfile` easyconfigs to prevent that others hit the issue you were seeing...

Simontuk:
Dear @migueldiascosta and @boegel,
thank you for checking in again, and thanks for making this clear. 
I think you're both right, I finally had time to track the issue down and understood the Octave point. Wasn't aware it was automatically detected.
In my opinion a default `configopts = '--enable-octave=no'` should be clear to keep the build process straightforward.  

I'm honestly not sure how to proceed to only include my changes for when Octave is desired.
Also I don't know about the Octave building process, so I wouldn't feel comfortable adding it.
Would appreciate your input again, but would be totally happy with the changed default to simplify the process.

lexming:
@Simontuk Thanks for reporting this issue. I'm closing this PR as it has been fixed in #10027 by disabling support for Octave in our `libsndfile` easyconfigs.

PR #6963, opened 2018-10-05
Eliminate dependency on problematic LibUUID library.

schiotz:
This removes LibUUID which causes several system programs to fail, see discussion on the mailing list:
https://lists.ugent.be/wws/arc/easybuild/2018-10/msg00011.html

The dependency of LibUUID is replaced by util-linux.  The change only affects fontconfig-2.13.0 as older versions do not depend on LibUUID.  Rebuilding this module is enough to solve the issues, no need to rebuild the entire X11 stack.

CC: @OleHolmNielsen

PR #9792, opened 2020-01-28
{lib}[system] XALT v2.8.4

samcmill:
Adds EasyConfigs for [XALT](https://github.com/xalt/xalt).

A few items of note:
- XALT requires a site specific configuration file at build time (`--with-config`).  In the EasyConfig I have documented that the user should use `--try-amend=config_py=<file>` to specify this file.  If this option is not used, then an error message with this information is printed.
- XALT has a slightly different interpretation of the install prefix than most packages.  It expects the prefix to the base install directory (e.g., `/opt/easybuild/software`) and it appends the package name and version itself (e.g., `xalt/2.7.27`).  The corresponding EasyBlock handles this.
- The GPU enabled variant installs in the same directory as the non-GPU variant.  This is because `versionsuffix` does not work due to the second item.  

Depends on https://github.com/easybuilders/easybuild-easyblocks/pull/1942.  

cc @rtmclay


easybuild/easyconfigs/x/xalt/xalt-2.7.27-gpu.eb
```diff
@@ -0,0 +1,61 @@
...
+
+osdependencies = [
+    ('bc'),
+    ('coreutils'),
+    ('libelf-dev', 'elfutils-libelf-devel'),
+    ('libssl-dev', 'openssl-devel'),
+    ('libxml2'), # required for CUDA-10.1.243
+]
+
+sanity_check_paths = {
```
boegel: @samcmill This can/should be moved to the `XALT` easyblock too, by customizing `sanity_check_step`

easybuild/easyconfigs/x/xalt/xalt-2.7.27-gpu.eb
```diff
@@ -0,0 +1,61 @@
...
+
+source_urls = ['https://github.com/xalt/xalt/archive/']
+sources = [SOURCE_TAR_GZ]
+checksums = ['d0dcefe5d8b342ca232bd1ce8ae675fed714591e49fa96eef9a13b392786b5d8']
+
+builddependencies = [
+    ('CUDA', '10.1.243'),
+    ('cURL', '7.60.0-GCCcore-7.3.0'),
+    ('flex', '2.6.4-GCCcore-7.3.0'),
+    ('LibUUID', '1.0.3-GCCcore-7.3.0'),
```
boegel: `LibUUID` should be avoided, please use `util-linux` instead (see #6963 for example)

easybuild/easyconfigs/x/xalt/xalt-2.7.27-gpu.eb
```diff
@@ -0,0 +1,61 @@
...
+# The XALT site filter config file is site specific and must be specified
+# to successfully build XALT.  Use "--try-amend=config_py=<file>" when
+# building this EasyConfig to specify the site filter config file.
+config_py = None
+
+# Extra options
+#syshost = 'hardcode:cluster'
+#transmission = 'syslog'
+
+configopts = '--with-MySQL=no --with-trackGPU=yes'
```
boegel: Should we have another easyconfig parameter for `--with-MySQL` in the `XALT` easyblock?

easybuild/easyconfigs/x/xalt/xalt-2.7.29.eb
```diff
@@ -0,0 +1,61 @@
+easyblock = 'EB_XALT'
+
+name = 'xalt'
```
boegel, line 3: @samcmill We should use `XALT`, which is the official software name?

When doing so, the `easyblock = ` line can be removed, since EasyBuild will correctly derive the easyblock to use from the software name.

easybuild/easyconfigs/x/xalt/xalt-2.7.29.eb
```diff
@@ -0,0 +1,61 @@
...
+version = '2.7.29'
```
boegel, line 4: worth bumping to `2.8.1`?

easybuild/easyconfigs/x/xalt/xalt-2.7.29.eb
```diff
@@ -0,0 +1,61 @@
...
+    ('cURL', '7.60.0-GCCcore-7.3.0'),
+    ('flex', '2.6.4-GCCcore-7.3.0'),
+    ('util-linux', '2.32.1-GCCcore-7.3.0'),
+]
+
+osdependencies = [
+    ('bc'),
+    ('coreutils'),
+    ('libelf-dev', 'elfutils-libelf-devel'),
+    ('libssl-dev', 'openssl-devel'),
```
boegel, line 58: For this, we have a dedicated constant now: `OS_PKG_OPENSSL_DEV` (see https://github.com/easybuilders/easybuild-framework/pull/3309).

That avoids introducing many variants of this across easyconfigs. Here, `libopenssl-devel` is missing as an alternative for example, so using `OS_PKG_OPENSSL_DEV` instead fixes that.

You'll need EasyBuild v4.2.1 (or current `develop` version) for this to work.

boegel:
@samcmill Installation with the latest update of the `XALT` easyblock in https://github.com/easybuilders/easybuild-easyblocks/pull/1942 works, but it doesn't produce a working XALT installation...

I was running some commands after loading the XALT module, and noticed that nothing was being logged by XALT. When trying to figure it out, I ran into this:

```
$ xalt_configuration_report
/software/XALT/2.8.1/bin/../libexec/xalt_configuration_report.x: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /software/XALT/2.8.1/bin/../libexec/xalt_configuration_report.x)
/software/XALT/2.8.1/bin/../libexec/xalt_configuration_report.x: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /software/XALT/2.8.1/bin/../libexec/xalt_configuration_report.x)
```

The issue here is that `GCCcore/7.3.0`, which provides a newer `libstdc++.so.6` than the OS does, was loaded at the time that XALT was being built, because it got loaded via the build dependencies (`cURL`, etc.).
Since `GCCcore/7.3.0` isn't loaded by the XALT module, we end up with a broken installation.

We really don't want the `XALT` module to load a specific `GCCcore` though, since that would severely limit the compatibility of XALT with modules built with an older/newer toolchain.

So, I think we should:

* run `xalt_configuration_report` as a sanity check command in the XALT easyblock, to catch broken installations like this;
* either:
  * use dependencies built with the `system` toolchain (but that leaves us at the mercy of the system compiler, which may be problematic on some system);
  * if possible, make sure that `libstdc++` is statically linked in XALT (maybe this is something for @rtmclay to help out with?)

rtmclay:
Installing XALT with anything but the system gcc is a mistake.  What happens when you build things with a new gcc than what is on the system and then run things with just xalt?

samcmill:
Since `toolchain = SYSTEM` does not behave as expected, I think there are 3 options.

1. Modify XALT to statically link libstdc++, etc.  The counter-argument is that XALT should just be built with the system gcc, period. Static linking is just masking the larger problem.

2. Replace the build dependencies with OS dependencies.  The downside is that this would break for regular users if those dependencies are not installed.  Some of the dependencies are not commonly installed.

3. Replace the `GCC-7.3.0` build dependencies with system compiler versions.  This would mean creating new easyconfigs in some cases.

4. Replace the `GCC-7.3.0` build dependencies with older versioned toolchain versions.  This would mean creating new easyconfigs in some cases.  The other downside is that it would still mean XALT is built with something other than the system compiler.

If 1 is not viable, then 3 seems like the best choice.  If there is agreement, then I will add those additional easyconfigs to this PR.

boegel:
> Installing XALT with anything but the system gcc is a mistake. What happens when you build things with a new gcc than what is on the system and then run things with just xalt?

@rtmclay That likely works as expected, but let me flip it around: what if XALT is built with a system compiler that is newer than what software was built with?

Fedora 32 ships with GCC 10.1 as system compiler. If XALT is built with the system compiler, and then software is built with GCC 8.3 (foss/2019b) or GCC 9.3 (foss/2020a), then XALT may still be broken due to incompatibilities between libstdc++.so?

libc.so is a different story, EasyBuild doesn't build it's own libc, it always uses the one from the system.

So, option 1 still seems the best way forward to me. Option 3 would work, but only if the system compiler is older than what software is being built with.

samcmill:
The XALT EasyBlock has been updated to add the `static_cxx` option.  If set to True, then XALT will be statically linked with libstdc++ and libgcc_s.

easybuild/easyconfigs/x/XALT/XALT-2.8.1.eb
```diff
@@ -0,0 +1,60 @@
+name = 'XALT'
+version = '2.8.1'
```
boegel, line 2: @samcmill Can we update this to `2.8.4` (and fix the SHA256 checksum), which includes important fixes (in particular, it should resolve the fork bomb issue I ran into...)

samcmill: Done

boegel:
Test report by @boegel
**SUCCESS**
node2413.golett.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz (haswell), Python 2.7.5
See https://gist.github.com/8c608f6ac2d3f0e7da991450dd324fee for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3402.kirlia.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), Python 2.7.5
See https://gist.github.com/c0e168a38e34a4354529fc0e06cad508 for a full test report.

boegel:
Going in, thanks @samcmill!

PR #9859, opened 2020-02-12
{tools}[system] pip v20.0.2

boegel:
(created using `eb --new-pr`)

requires ~~https://github.com/easybuilders/easybuild-easyblocks/pull/1961~~ in case `eb` is run with `python3` but `python` is also available in the OS...


boegel:
Test report by @boegel
**SUCCESS**
node3116.skitty.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/5163026d15fb24d5f537b29fc7f9a162 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
boegel.ugent.be - Darwin UNKNOWN UNKNOWN, Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz, Python 2.7.16
See https://gist.github.com/8f8031a83dbe0d67ce7c964dec2cf1d7 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
See https://gist.github.com/6e0ba303be7b4ebd5d7158390285bb18 for a full test report.

easybuild/easyconfigs/p/pip/pip-20.0.2.eb
```diff
@@ -0,0 +1,39 @@
...
+
+exts_default_options = {'source_urls': [PYPI_SOURCE]}
+
+exts_list = [
+    ('setuptools', '44.0.0', {
+        'source_tmpl': 'setuptools-%(version)s.zip',
+        'checksums': ['e5baf7723e5bb8382fc146e33032b241efc63314211a3a120aaa55d62d2bb008'],
+    }),
+    # stick to latest 0.33.x, 0.34.x does not work (UnicodeEncodeError, see https://github.com/pypa/wheel/issues/331)
+    ('wheel', '0.33.6', {
```
Flamefire: This is supposedly fixed in 0.34.2, why not use that?

boegel: I tried using `0.34.2`, didn't work for me...

Flamefire: To not regret this later can we get a couple test reports with 34.2 first? Multiple people reported success with that so it might also be helpful for upstream and we have an easy way to get reproducible results from a couple environments.

Flamefire: I investigated a bit and found this to be solely a Python2 issue (I guess we'll see more of them), see https://github.com/pypa/wheel/issues/331. A possible solution is to patch setuptools: https://github.com/pypa/setuptools/issues/2003

boegel: I don't think it's worth the effort to patch `setuptools` for this.

Also, I don't expect `setuptools` to care much about Python 2 these days...

Flamefire: Well they introduced a deprecation notice for 45.x with a promise to not drop it before April(?), so I'd wait a couple days for a response on the issue(s)

Flamefire: I discussed this a bit with the guys from setuptools and they were pretty clear that `setup.py install` is no longer supported/deprecated and `pip install` shall be used. Also this method of installing pip is not recommended and `get-pip.py` shall be used instead. And indeed I was able to install all this with:

```
python get-pip.py --no-setuptools --no-wheel --ignore-installed --prefix /tmp/pipeb
PATH=/tmp/pipeb/bin PYTHONPATH=/tmp/pipeb/lib/python2.7/site-packages:$PYTHONPATH pip install --prefix=/tmp/pipeb setuptools==44.0.0
PATH=/tmp/pipeb/bin PYTHONPATH=/tmp/pipeb/lib/python2.7/site-packages:$PYTHONPATH pip install --prefix=/tmp/pipeb .
```

Hence my suggestion:
- use pip to install these packages
- when no pip is found (or always?), install the one from get-pip.py in the preinstallopts (e.g.), which will then be replaced by the one installed in the last step (pip installing pip)

verdurin: @boegel @Flamefire would be good to find a resolution for this one.

boegel:
Not planning to pursue this, installing on top of system Python is bloody painful to get right in every possible context, so closing....

PR #9986, opened 2020-03-05
{tools}[GCC/8.3.0] Singularity v3.5.3

robqiao:
(created using `eb --new-pr`)


easybuild/easyconfigs/s/Singularity/Singularity-3.5.3-GCC-8.3.0.eb
```diff
@@ -0,0 +1,63 @@
...
+
+configure_cmd = "cd '%(builddir)s/%(namelower)s-%(version)s' && "
+configure_cmd += "echo '%(version)s' > VERSION && "
+configure_cmd += "./mconfig -p '%(installdir)s' -b ./buildtree"
+
+build_cmd = "cd ./buildtree && "
+build_cmd += "make"
+
+install_cmd = "cd '%(builddir)s/%(namelower)s-%(version)s/buildtree' && "
+install_cmd += "make install"
```
smoors: @robqiao it's not a good idea to overwrite the _cmd variables unless really necessary. I suggest to change this as follows:
```
start_dir = '%(namelower)s-%(version)s'

configure_cmd = "echo '%(version)s' > VERSION && "
configure_cmd += "./mconfig -p '%(installdir)s' -b ./buildtree"

prebuildopts = "cd buildtree && "

preinstallopts = "cd buildtree && "
```

easybuild/easyconfigs/s/Singularity/Singularity-3.5.3-GCC-8.3.0.eb
```diff
@@ -0,0 +1,56 @@
...
+]
+
+start_dir = '%(namelower)s-%(version)s'
+
+configure_cmd = "echo '%(version)s' > VERSION && "
+configure_cmd += "./mconfig -p '%(installdir)s' -b ./buildtree"
+
+prebuildopts = "cd ./buildtree && "
+
+preinstallopts = "cd '%(start_dir)s/buildtree' && "
```
smoors:
```suggestion
preinstallopts = "cd buildtree && "
```

smoors:
Test report by @smoors
**FAILED**
Build succeeded for 2 out of 3 (2 easyconfigs in this PR)
node368.hydra.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/20667facdee77c870c2868cc75df242e for a full test report.

smoors:
@robqiao I get errors like this one:
```
/tmp/vsc10009/ebinstall/9986/software/Go/1.14/src/encoding/hex/hex.go:9:2: can't open import: "errors": open /user/brussel/100/vsc10009/.cache/go-build/29/292748909c450b971c9d9111821599fa81c44d24ca918a7e1504834d88e948d2-d: no such file or directory
```
this can probably be fixed by adding `parallel = 1` to the easyconfig.

robqiao:
> 
> 
> @robqiao I get errors like this one:
> 
> ```
> /tmp/vsc10009/ebinstall/9986/software/Go/1.14/src/encoding/hex/hex.go:9:2: can't open import: "errors": open /user/brussel/100/vsc10009/.cache/go-build/29/292748909c450b971c9d9111821599fa81c44d24ca918a7e1504834d88e948d2-d: no such file or directory
> ```
> 
> this can probably be fixed by adding `parallel = 1` to the easyconfig.

Thx @smoors, that's right. Single thread `Golang` package installation should be proceded. Fixed 👍

smoors:
Test report by @smoors
**FAILED**
Build succeeded for 2 out of 3 (2 easyconfigs in this PR)
node362.hydra.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/4b1765973cd10d21c9f90c2cf9bea33a for a full test report.

easybuild/easyconfigs/s/Singularity/Singularity-3.5.3-GCC-8.3.0.eb
```diff
@@ -0,0 +1,59 @@
...
```
smoors: @robqiao now the test fails because `start_dir` is not a template. you probably need to change it to this:
```suggestion
preinstallopts = "cd %(namelower)s-%(version)s/buildtree && "
```

easybuild/easyconfigs/s/Singularity/Singularity-3.5.3-GCC-8.3.0.eb
```diff
@@ -33,14 +33,12 @@ builddependencies = [
...
 
-start_dir = '%(namelower)s-%(version)s'
-
 configure_cmd = "echo '%(version)s' > VERSION && "
 configure_cmd += "./mconfig -p '%(installdir)s' -b ./buildtree"
 
 prebuildopts = "cd ./buildtree && "
 
-preinstallopts = "cd '%(start_dir)s/buildtree' && "
+preinstallopts = "cd '%(namelower)s-%(version)s/buildtree' && "
```
smoors: @robqiao my bad, it should be:
```suggestion
preinstallopts = prebuildopts
```

robqiao: Thx @smoors. It seems that `prebuildopts = "cd ./buildtree && "` is relative to  `%(builddir)%(name)s-%(version)s ` but `preinstallopts` is relative to `%(builddir)s`.

```
            if self.dry_run and os.path.samefile(topdir, self.builddir):
                topdir = os.path.join(self.builddir, '%s-%s' % (self.name, self.version))
                self.log.info("Modified parent directory of start dir in dry run mode to likely path %s", topdir)
                # make sure start_dir subdir exists (cfr. check below)
                mkdir(os.path.join(topdir, start_dir), parents=True)            abs_start_dir = os.path.join(topdir, start_dir)
            if topdir.endswith(start_dir) and not os.path.exists(abs_start_dir):
                self.cfg['start_dir'] = topdir
            else:
                if os.path.exists(abs_start_dir):
                    self.cfg['start_dir'] = abs_start_dir
                else:
                    raise EasyBuildError("Specified start dir %s does not exist", abs_start_dir)
```
I suspect that `preinstallopts = "cd '%(namelower)s-%(version)s/buildtree' && "` would be approperate in this case. Any thoughts?

Micket:
So, new Singularity configs have come up multiple times now, and, since they contain setuid binaries, it's not really recommended to install into module systems at all, as you really want to git rid of old installations when security fixes are released.

But, is this even a functioning build, since it's not build as root?
I made a PR to finally archive those old singularity configs, as those should definitely really not be used, https://github.com/easybuilders/easybuild-easyconfigs/pull/10591, which might help discourage people from trying to update them in EB.

robqiao:
> 
> 
> So, new Singularity configs have come up multiple times now, and, since they contain setuid binaries, it's not really recommended to install into module systems at all, as you really want to git rid of old installations when security fixes are released.
> 
> But, is this even a functioning build, since it's not build as root?
> I made a PR to finally archive those old singularity configs, as those should definitely really not be used, #10591, which might help discourage people from trying to update them in EB.

I think you got a point. There is a manual activation step to set permissions to `root` at the end. Also at the same time, I think it can be beneficial when we have a dozen machines to go through at once. See what @Micket thinks.

Micket:
> See what @Micket thinks.

Tempting or not, I still think it's a bad idea. Primarily for the security; you don't want outdated builds of this to stick around on your systems.
There is also the numerous technical issues, as I can only imagine the sudo steps totally break `--job` and `--package` and possible root-squashed filesystems and whatnot.

Installing this software via the OS package manager just seems like the best idea i.m.o., and this is what I do for all the clusters i manage, and I'm pretty sure most if not all the other EB maintainers do the same.

robqiao:
> 
> 
> > See what @Micket thinks.
> 
> Tempting or not, I still think it's a bad idea. Primarily for the security; you don't want outdated builds of this to stick around on your systems.
> There is also the numerous technical issues, as I can only imagine the sudo steps totally break `--job` and `--package` and possible root-squashed filesystems and whatnot.
> 
> Installing this software via the OS package manager just seems like the best idea i.m.o., and this is what I do for all the clusters i manage, and I'm pretty sure most if not all the other EB maintainers do the same.

Agreed on the security vulnerability, Thx @Micket. Ansible script with OS package manager should provide scalability. Now PR closed.

PR #9990, opened 2020-03-05
{bio}[foss/2019b] RAxML-NG v0.9.0

albertogfernan:
(created using `eb --new-pr`)


migueldiascosta:
**FAILED**
See https://gist.github.com/8666d0a980581f5cb413003b164c42d5 for a full test report.

migueldiascosta:
thanks @albertogfernan!

The test failure is due a failing checksum, which is likey due to the use of `git_config`, see note in https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html#common-easyconfig-param-sources-git-config. I suppose using the release packages is not an option?

Related to https://github.com/easybuilders/easybuild-framework/pull/2727 and https://github.com/easybuilders/easybuild-framework/pull/2727

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-foss-2019b.eb
```diff
@@ -0,0 +1,56 @@
...
+    ('binutils', '2.32'),
+    ('CMake', '3.15.3'),
+    ('gtest', '1.8.1'),
+]
+
+dependencies = [
+    ('flex', '2.6.4'),
+    ('Bison', '3.3.2'),
+    ('GMP', '6.1.2'),
+    ('OpenMPI', '3.1.4'),
```
migueldiascosta: this is not necessary, it's already included in the toolchain

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-foss-2019b.eb
```diff
@@ -0,0 +1,56 @@
...
+name = 'RAxML-NG'
+version = '0.9.0'
+
+homepage = "https://github.com/amkozlov/raxml-ng"
+description = """RAxML-NG is a phylogenetic tree inference tool which
+uses maximum-likelihood (ML) optimality criterion. Its search heuristic
+ is based on iteratively performing a series of Subtree Pruning and Regrafting
+ (SPR) moves, which allows to quickly navigate to the best-known ML tree."""
+
+toolchain = {'name': 'foss', 'version': '2019b'}
```
migueldiascosta: the toolchain can probably be lowered to `gompi`, since this doesn't seem to depend on the linear algebra libraries

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-foss-2019b.eb
```diff
@@ -0,0 +1,56 @@
...
+
+sources = [{
+    'filename': 'raxml-ng-%(version)s.tar.gz',
+    'git_config': {
+        'url': 'https://github.com/amkozlov',
```
boegel: When downloading from GitHub, you don't need to use `git_config` at all (which is troublesome w.r.t. checksums, as @migueldiascosta pointed out).

I suggest to use this instead:

```python
source_urls = ['https://github.com/amkozlov/raxml-ng/archive/']
sources = ['%(version)s.tar.gz']
```

(and fix the checksum)

This corresponds to the exact same sources as when using tag `0.9.0` tag, it's just a static tarball provided by GitHub (static in the sense that it'll be the exact same download for everyone, which is not the case when using `git_config` where the tarball is created with the local `tar` and `gzip` commands).

If you care strongly about the source filename, you can make EasyBuild rename it after download:

```python
sources = ['download_filename': '%(version)s.tar.gz', 'filename': 'raxml-ng-%(version)s.tar.gz']
```

or

```python
sources = ['download_filename': '%(version)s.tar.gz', 'filename': SOURCELOWER_TAR_GZ]
```

albertogfernan: Hello @migueldiascosta , hello @boegel ,

Thanks both for your feedback. Concerning the tool chain and the list of dependencies, i will try to do that and update the pull request.

Concerning the checksum, we knew that could cause some problems, but we don't know how to proceed otherwise. The archive doesn't contain the full source code of the solution. Some libraries are also necessary. That's why we added the recursive True to the git_config.

We tried to specify all the sources in the list of sources and source_urls, but that doesn't work out. I don't know if there's a means of specifying manually where to decompress a source within the directory tree of another source.

Another alternative could be to create easy config files for all the additional libraries and then set them up as dependencies, but i don't think that makes sense, since they don't have enough specific weight to be compiled or deployed independently, and the final result of RAxML-NG compilation is a single executable statically linked.

Would you have any suggestions about this?

Best Regards

boegel: You can list everything in `sources`, and the move things into the correct place before you start the build.

Something like this (untested and clearly needs work, but I hope you get the idea):

```python
sources = [
    {
        'source_urls': ['https://github.com/amkozlov/raxml-ng/archive/'],
        'filename': '%(version)s.tar.gz',
    },
    {
        'source_urls': ['https://github.com/amkozlov/terraphast-one/archive',
        'download_filename': '6dcb061.tar.gz',
        'filename': 'terraphast-one-20180618.tar.gz',
    },
]

# copy dependencies into place
preconfigopts = "cp -a %(builddirs)/terraphast-one %(builddir)s/raxml-ng/libs/terraphast && "
```

The directory names `terraphast-one` and `raxml-ng` are probably not fully correct, they may have a version/commit ID attached, but you may be able to work around that using `terraphast-one*` and `raxml-ng*`.

I hope this helps, but do let us know if you're stuck or have questions.

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+    {
+        'source_urls': ['https://github.com/xflouris/libpll-2/archive/'],
+        'download_filename': local_libpll_tree + '.zip',
+        'filename': 'libpll-2.zip',
+    },
+]
+
+# copy dependencies into place
+preconfigopts = ("cp -a %(builddir)s/terraphast-one-" + local_terrap_tree +
+                 "/* %(builddir)s/raxml-ng-%(version)s/libs/terraphast/ && ")
```
migueldiascosta: please move all `preconfigopts` to before `configopts`, for readability, and consider using formatted strings, e.g. `"cp -a %%(builddir)s/terraphast-one-%s/*" % local_terrap_commit`

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+
+preconfigopts += ("cp -a %(builddir)s/libpll-2-" + local_libpll_tree +
+                  "/* %(builddir)s/raxml-ng-%(version)s/libs/pll-modules/" +
+                  "libs/libpll/ && ")
+
+checksums = [
+    '43b95ee1bddae0daee84644e9ee760a77f28bb16e2071e95cca79a30f39373e5',
+    'eef5e2f365619da3bb1679f07920b29dfcd5202191fdf37b9f8fec4a52c1286b',
+    '486f55cc2b4a63ddba8d00db79afcbf03f2499e35687eef56e4730f23d96b90a',
+    '6c90af957d4b06b88ef6568f6ba53ac3507285d042dc9ebc95f9837a7077d044',
```
migueldiascosta: fyi, you can use `eb --inject-checksums -f`, it will place them after `sources` and add comments with each filename

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+files_to_copy = [
+    (["bin/raxml-ng-mpi"], "bin"),
+]
+
+sanity_check_paths = {
+    'files': ['bin/raxml-ng-mpi'],
+    'dirs':  [],
+}
+
+separate_build_dir = True
```
migueldiascosta: please move this close to `build_type`, for readability

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+    '486f55cc2b4a63ddba8d00db79afcbf03f2499e35687eef56e4730f23d96b90a',
+    '6c90af957d4b06b88ef6568f6ba53ac3507285d042dc9ebc95f9837a7077d044',
+]
+
+build_type = 'Release'
+
+builddependencies = [
+    ('binutils', '2.32'),
+    ('CMake', '3.15.3'),
+    ('gtest', '1.8.1', '', ('GCCcore', '8.3.0')),
```
migueldiascosta: why not simply `('gtest', '1.8.1')`?

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
+easyblock = 'CMakeMakeCp'
+
+name = 'RAxML-NG'
+version = '0.9.0'
+
+local_terrap_tree = '6dcb061ceb6ed41a7ebe1a8af1f3354e0126125c'
+local_pllmod_tree = '840a19525ed133ad243afb23124911b849d4685d'
+local_libpll_tree = '6874327c9ca9544e27b426c0e7205c8b6f4ed297'
```
migueldiascosta: `_commit` should be more widely understood than `_tree`

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
albertogfernan: With ('gtest', '1.8.1') and with no extra parameter it tries to find _gtest-1.8.1-gompi_  which doesn't exist and fails.

That's why I force the use of _gtest-1.8.1-Gcccore-8.3.0_, which is part of this pr and replicated from the _gtest-1.8.1-Gcccore-8.2.0_.

migueldiascosta: that shouldn't be necessary, `GCCcore` is a subtoolchain of `gompi`... what version of EB are you using?

Also, I just noticed that https://github.com/easybuilders/easybuild-easyconfigs/pull/10216 adds gtest 1.10.0 for the same toolchain, please use the same as we enforce a single dependency variant per toolchain generation 

(I'm sorry I missed that before)

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+checksums = [
+    '43b95ee1bddae0daee84644e9ee760a77f28bb16e2071e95cca79a30f39373e5',  # 0.9.0.tar.gz
+    'eef5e2f365619da3bb1679f07920b29dfcd5202191fdf37b9f8fec4a52c1286b',  # terraphist-one.zip
+    '486f55cc2b4a63ddba8d00db79afcbf03f2499e35687eef56e4730f23d96b90a',  # pll-modules.zip
+    '6c90af957d4b06b88ef6568f6ba53ac3507285d042dc9ebc95f9837a7077d044',  # libpll-2.zip
+]
+
+# copy dependencies into place
+preconfigopts = "cp -a %%(builddir)s/terraphast-one-%s" % local_terrap_commit
+preconfigopts += "/* %(builddir)s/raxml-ng-%(version)s/libs/terraphast/ && "
```
migueldiascosta:
```suggestion
preconfigopts = "cp -a %%(builddir)s/terraphast-one-%s/* " % local_terrap_commit
preconfigopts += "%(builddir)s/raxml-ng-%(version)s/libs/terraphast/ && "
```

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+
+preconfigopts += "cp -a %%(builddir)s/pll-modules-%s" % local_pllmod_commit
+preconfigopts += "/* %(builddir)s/raxml-ng-%(version)s/libs/pll-modules/"
+preconfigopts += " && "
```
migueldiascosta:
```suggestion
preconfigopts += "cp -a %%(builddir)s/pll-modules-%s/* " % local_pllmod_commit
preconfigopts += "%(builddir)s/raxml-ng-%(version)s/libs/pll-modules/ && "
```

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+
+preconfigopts += "cp -a %%(builddir)s/libpll-2-%s" % local_libpll_commit
+preconfigopts += "/* %(builddir)s/raxml-ng-%(version)s/libs/pll-modules/"
+preconfigopts += "libs/libpll/ && "
```
migueldiascosta:
```suggestion
preconfigopts += "cp -a %%(builddir)s/libpll-2-%s/* " % local_libpll_commit
preconfigopts += "%(builddir)s/raxml-ng-%(version)s/libs/pll-modules/libs/libpll/ && "
```

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
albertogfernan: Hello Miguel,

Just adding more information about the version and the error I get without specifying the toolchain for gtest.

This is EasyBuild 4.1.0 (framework: 4.1.0, easyblocks: 4.1.0) on host b10d98675de4.

eb -r RAxML-NG-0.9.0-gompi-2019b.eb gtest-1.8.1-GCCcore-8.3.0.eb
== temporary log file in case of crash /tmp/eb-C8Kb_o/easybuild-nIQeai.log
== resolving dependencies ...
ERROR: Missing dependencies: gtest/1.8.1-gompi-2019b (no easyconfig file or existing module found)

migueldiascosta: what about if you do, without specifying the toolchain for `gtest`, `eb RAxML-NG-0.9.0-gompi-2019b.eb -r .`? (so that the `gtest` easyconfig is in the robot path)

migueldiascosta: also, now both versions of `gtest` are in the PR, you can remove the old one with something like

 `eb --update-pr :gtest-1.8.1-GCCcore-8.3.0.eb --pr-commit-msg="remove old version of gtest"`

(note the colon before the filename, which means delete that file)

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
migueldiascosta: hm, you reverted these changes when you updated the PR (I guess that's the downside of using github code suggestions...), you'll need to incorporate them in your local file

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
```diff
@@ -0,0 +1,87 @@
...
+toolchain = {'name': 'gompi', 'version': '2019b'}
+
+sources = [
+    {
+        'source_urls': ['https://github.com/amkozlov/raxml-ng/archive/'],
+        'filename': '%(version)s.tar.gz',
+    },
+    {
+        'source_urls': ['https://github.com/amkozlov/terraphast-one/archive/'],
+        'download_filename': local_terrap_commit + '.zip',
```
migueldiascosta: style nitpicking, not a blocker imo, but we typically use formatted strings, e.g. `'%s.zip' % local_terrap_commit`, same below

albertogfernan: Implemented!

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
albertogfernan: That worked! So I removed the toolchain from the dependency and I removed the gtest-1.8.1 easyconfig from the pr too.

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
migueldiascosta: almost done :) please note

> hm, you reverted these changes when you updated the PR (I guess that's the downside of using github code suggestions...), you'll need to incorporate them in your local file

almost done! :) please note the above though (the changes we made to preconfigopts via github suggestions got reverted because they are not in your local eb file)

albertogfernan: F@#k. Sorry about that. My mistake.

I will update the pr again with the changes commited. But before doing that I have a question. I don't see any problem with the preconfig opts of terraphist-one and pll-modules. 

But the preconfig opts of libpll that you proposed will make the line wider than 80 characters. I think that doesn't comply with the pep8 style and the automatic tests fail because of that. Should I keep it in 3 lines of less than 80 or should I apply your proposal?

migueldiascosta: in EB the limit is 120 lines, so it should be ok (you can check yourself with `eb --check-contrib` before updating the PR)

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
migueldiascosta: hm, the result is not quite what I meant, I meant move to `preconfigopts` down to before `configopts`, not `configopts` up to after `preconfigopts` (the reason is that we try to order the easyconfig in roughly the same order of the steps that EB takes)

so now I would move both `preconfigopts` and `configopts` to just after `dependencies`, before `files_to_copy` - but I understand you've already put a lot of work into this, and it looks good enough to me, so up to you

migueldiascosta:
**SUCCESS**
sms - Linux centos linux 7.6.1810, AMD EPYC 7601 32-Core Processor, Python 2.7.5
See https://gist.github.com/1fa38d85e0ff03fce4cfb975d33e4e04 for a full test report.

easybuild/easyconfigs/r/RAxML-NG/RAxML-NG-0.9.0-gompi-2019b.eb
albertogfernan: No worries, I don't mind to update the pr, so we get it perfectly done.

But correct me if I'm wrong, the order would be starting from sources: checksums, dependencies, preconfigopts, configopts, build_type, separate_build, builddependencies, files_to_copy, sanity_check, and moduleclass.

So actually the clause out of order is dependencies, which should be just after checksums. 

Like this you have the phases: sources, preconfig, config, build, install, check.

Would it be like this?

migueldiascosta: well, I see what you mean, but note for instance that builddependencies must be loaded before the config step, not only during the build step

in any case, the ordering is of course not strict, it's more about readability and where most people expect things to be

migueldiascosta:
**SUCCESS**
See https://gist.github.com/148de0ff0edc5fc75b2e5657240411b5 for a full test report.

migueldiascosta:
Going in, thanks @albertogfernan!

PR #10024, opened 2020-03-09
Kraken2 updated to version 2.0.8, together with dependencies

sassy-crick:
(created using `eb --new-pr`)


smoors:
@sassy-crick a few comments:
- the patch you provide seems to be exactly the same, only the version has changed. this is not needed at all as we can reuse patches from previous versions as long as they apply cleanly.
- I prefer to use the system provided version of `wget`, as it is more likely to get security patches (you are of course free to use an EB provided version).

in short, I don't think it is a good idea to merge this PR. If you make a separate PR for `wget` (or just remove the Kraken2 bits), I'll be happy to review that.

sassy-crick:
@smoors No problems. I have switched the version of the patch back to 2.0.7 and removed wget. 
Do you think it is a good idea to submit a seperate PR for wget? I agree that the system provided one should be more up to date.

smoors:
Test report by @smoors
**SUCCESS**
node367.hydra.os - Linux centos linux 7.7.1908, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/4d4efef742d88f2b92a3bb4766a7e0a3 for a full test report.

smoors:

Issue #9951, opened 2020-03-01
Test dataset for PlasFlow v1.1.0 not working


lexming:
The easyconfig `PlasFlow-1.1.0-foss-2019b-Python-3.7.4.eb` (#9941) seems broken and I think that it was already broken before upgrading its dependency on `TensorFlow` to v2 (#9947). 

Performing the Test Dataset (https://github.com/smaegol/PlasFlow#test-dataset) fails with
```
$ PlasFlow.py --input Citrobacter_freundii_strain_CAV1321_scaffolds.fasta --output test.plasflow_predictions.tsv --threshold 0.7
Importing sequences
Imported  [1] 78
   sequences 
Traceback (most recent call last):
   File "/user/brussel/101/vsc10122/.local/easybuild/software/PlasFlow/1.1.0-foss-2019b-Python-3.7.4/bin/PlasFlow.py", line 92, in <module>
     pd_accessions = pandas2ri.ri2py(accessions) 
AttributeError: module 'rpy2.robjects.pandas2ri' has no attribute 'ri2py'
```
which means that `PlasFlow` is not compatible with `rpy2` v3. The dependency on `rpy2` is indeed listed as for version v2 (`>=2.8`) in the project's github.

boegel:
@Darkless012 Can you take a look at this?

Darkless012:
Will do :)

lexming:
Fix in #10028

PR #10028, opened 2020-03-10
remove PlasFlow easyconfig, doesn't work with TensorFlow 1.5 or rpy2 3.x (at least not without extra work)

boegel:
(created using `eb --new-pr`)
cfr. #9951 (cc @Darkless012)


Darkless012:
rpy2 fix: https://github.com/easybuilders/easybuild-easyconfigs/pull/10038

The only version which works is TF-0.10.x - paths and others gimmicks
But this requires Python 3.5 - first TF on PYPI with Python 3.6 is TF-0.12.x

It throws same error with TF-0.12.x as with TF-1.x so concluding 
it could work with TF-1.x.x but this requires to train tensorflow before using it (models)

It doesn't work with TF-2.x since contrib/learn subdir has been deprecated and removed in 2.x

Training can be done via: https://github.com/smaegol/plasflow_processing/blob/master/README.md
I'm attempting to train it via the script, but it downloads a lot of stuff from FTPs.

lexming:
Merging this to fix issue #9951 before next release.
Further work on `PlasFlow` can continue on #10038

PR #10038, opened 2020-03-10
{bio}[foss/2019b] PlasFlow v1.1.0 - update

### LINKED ISSUE #9951: Test dataset for PlasFlow v1.1.0 not working

Darkless012:
(created using `eb --new-pr`)


lexming:
The situation is definitely better now and the test command in https://github.com/smaegol/PlasFlow#test-dataset starts and goes beyond the error in issue #9951 . Unfortunately, after a few minutes I hit the following error

> tensorflow.python.framework.errors_impl.NotFoundError: **Tensor name "dnn/hiddenlayer_0/biases" not found** in checkpoint files /theia/scratch/brussel/101/vsc10122/easybuild-local/software/PlasFlow/1.1.0-foss-2019b-TensorFlow-1.15.0-Python-3.7.4/models/kmer5_split_20_20_neurons_relu/model.ckpt-50000-?????-of-00001

which seems the same issue as https://github.com/smaegol/PlasFlow/issues/6

Honestly, I do not know how critical are those old tensors from `TensorFlow-0.10` in `PlasFlow`. But I do not see any option allowing to bypass them or choose alternatives. If those missing tensors could be disabled, then we could at least add a comment in the easyconfig informing about the missing functionality.

boegel:
@Darkless012 Any updates on this, you mentioned a re-training effort with the correct `TensorFlow` version?

Darkless012:
Got stuck on broken Perl training script, need to take deeper look on it.

boegel:
This effort was by our (HPC-UGent) request, and has been abandonded since Oct'20, so no need to spend time on this, closing...

PR #10040, opened 2020-03-10
Adding CREST-2.9.eb

dplacencia:
CREST software is a dependency of the xtb software

### DIFF
--- a/easybuild/easyconfigs/c/CREST/CREST-2.9.eb
+++ b/easybuild/easyconfigs/c/CREST/CREST-2.9.eb
@@ -11,7 +11,8 @@ functions for calculations with the GFNn–xTB methods.  """
 
 toolchain = SYSTEM
 
-sources = ['%(name)s-200219.tgz']
+local_verdate = '200219'
+sources = [{'download_filename': '%%(name)s-%s.tgz' % local_verdate, 'filename': SOURCE_TGZ}]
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/c/CREST/CREST-2.9.eb
```diff
@@ -0,0 +1,23 @@
+easyblock = 'Binary'
+
+name = 'CREST'
+version = '2.9'
```
smoors: @dplacencia please add the release date as local var:
```
local_verdate = '200219'
```

dplacencia:
> @dplacencia do you have a specific reason why you're using a pre-release version instead of a release?
Hi, I am working together with the developers of xTB to publish a release soon. At this time the latest version of this software is available only in pre-release. Could this change be postponed?

awvwgk:
@dplacencia Thanks for your effort, we are always happy to make our programs widely available. *But* since this is a binary redistribution of the `crest` program, which is not governed under an open source license right now, so you should ask the authors of the program first.

I guess there will be no problem with adding `crest` here, but I am not the one who can actually give you allowance for this. I kindly ask you suspend this PR for the moment and contact the authors of `crest` first.

Also pinging @pprcht since he is one of the authors of `crest`.

dplacencia:
closing pull request momentarily until authorization is obtained

PR #10084, opened 2020-03-17
temporarily limit test configurations during 2020a sprint

boegel:
I'll open a follow-up PR so we don't forget to roll this back...


akesandgren:
Going in, thanks @boegel!

boegel:
PR to revert this after the sprint: #10085

PR #10125, opened 2020-03-18
2020a

maxim-masterov:
Some dependencies required by Qt5. The Python-2.7.17 contains just a version upgrade, all sub-packages are left as they were. Python-2.7.17 is simply needed to build SCons.


akesandgren:
Note that we didn't decide on Python 2.7.17 or 2.7.18 yet.

maxim-masterov:
Yes, I'm canceling this PR anyway.

PR #10236, opened 2020-03-23
10234 fix LS-PrePost

tsubic:
Fixed older versions of easyconfig files for LS-PrePost and added a new version (4.7).


easybuild/easyconfigs/l/LS-PrePost/LS-PrePost-4.7.eb
```diff
@@ -0,0 +1,22 @@
+easyblock = "Tarball"
+
+name = 'LS-PrePost'
+version = '4.7'
```
Micket: Looks like they have started to use all 3 numbers for versions, which would solve problem with different date suffixes for the same "version". I think we should use 4.7.8 going forward.

easybuild/easyconfigs/l/LS-PrePost/LS-PrePost-4.7.eb
```diff
@@ -0,0 +1,22 @@
...
+
+homepage = 'http://lstc.com/products/ls-prepost'
+description = """LS-PrePost is an advanced pre and post-processor that is delivered free with LS-DYNA."""
+
+toolchain = SYSTEM
+
+source_urls = ['http://ftp.lstc.com/anonymous/outgoing/lsprepost/%(version)s/linux64/']
```
Micket: and we should switch to %(version_major_minor)s here

easybuild/easyconfigs/l/LS-PrePost/LS-PrePost-4.7.eb
tsubic: ok, will fix it, let's hope that they wont change dates anymore

easybuild/easyconfigs/l/LS-PrePost/LS-PrePost-4.7.eb
tsubic: so, having one variable "version" which woudl in this case be 4.7.8, and one variable "version_major_minor" which would be just 4.7?

The problem is that in the url they use just 4.7

tsubic: thanks for the comment, fixed

Micket:
Test report by @Micket
**SUCCESS**
hebbe-c1 - Linux centos linux 7.7.1908, Intel Core Processor (Haswell, no TSX), Python 2.7.5
See https://gist.github.com/7b87d67629d400320a98184921bfd974 for a full test report.

Micket:
Going in, thanks @tsubic!

PR #10286, opened 2020-03-30
delete oneTBB and use tbb instead

branfosj:
As per discussion on Slack, we'll stick with using the `tbb` name. So this deletes `oneTBB-2020.1-GCCcore-9.3.0.eb'.  There are no easyconfigs using oneTBB.


lexming:
Going in, thanks @branfosj !

PR #10296, opened 2020-03-31
Restore libaio

edmondac:
This file was introduced in https://github.com/bear-rsg/easybuild-easyconfigs/pull/102, but then accidentally removed again in https://github.com/bear-rsg/easybuild-easyconfigs/pull/103


edmondac:
Sigh - wrong repo.

PR #10341, opened 2020-04-06
use apt-spy2 before using apt in GitHub Actions

boegel:
This should help in dealing with the "`Could not connect to azure.archive.ubuntu.com`" issues that have been popping up...


lexming:
Going in, thanks @boegel !

PR #10374, opened 2020-04-11
stop using apt-spy2 to pick an alternative mirror, it's expensive and may result in picking a faulty mirror...

boegel:
Tests are failing now because a "faulty" mirror is being selected by `apt-spy2`:

```
The closest mirror is: http://mirrors.codec-cluster.org/ubuntu/
Updating /etc/apt/sources.list
Updated '/etc/apt/sources.list' with http://mirrors.codec-cluster.org/ubuntu/
...
E: Failed to fetch http://mirrors.codec-cluster.org/ubuntu/pool/universe/l/lua-bit32/lua-bit32_5.3.0-3_amd64.deb  403  Forbidden [IP: 65.49.71.107 80]
```

According to https://github.com/actions/virtual-environments/issues/675#issuecomment-610914822 the problems with `azure.archive.ubuntu.com` that made us start using `apt-spy2` should be largely resolved now...

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -30,14 +30,9 @@ jobs:
 
     - name: install OS & Python packages
       run: |
-        # use apt-spy2 to select closest apt mirror,
-        # which helps avoid connectivity issues in Azure;
-        # see https://github.com/actions/virtual-environments/issues/675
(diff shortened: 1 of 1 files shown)


zao:
Going in, thanks @boegel !

PR #10335, opened 2020-04-05
{math}[foss/2018b] numpy v1.18.2 w/ Python 3.6.6

kelseymh:
(created using `eb --new-pr`)
Update to latest numpy module.  Requires use of Cython 0.29.14 or later, so depends upon PR #10334.

PR #10334, opened 2020-04-05
{lang}[foss/2018b] Cython v0.29.16 w/ Python 3.6.6

kelseymh:
(created using `eb --new-pr`)
Explicit new module with updated version of Cython.  This will be needed for new numpy module, which requires Cython 0.29.14 or later for build.

Issue #6493, opened 2018-06-22
Possible problem with the new "No multi-variant deps found for" Assertion check in Travis


akesandgren:
The above assertion check might be too restrictive.
At the moment it checks that all versions of some software is at the same level for a toolcahin of version 201[89][ab]. (btw, should fix that to handle 2020 and onwards too)

But, there are cases when software A of version 1 has a dependency on software B version 2
and A ver 2 has a dep on B ver 3, while at the same time A ver 1 can not use B ver 3.

The current Assert will error out if both ver 1 and 2 of A is built with foss/2018a.
And there are lots of software where you might need both versions of A built with the same toolchain version.

Doing the verification only on the dependencies of A ver x instead of for all versions of A to be the same would probably eliminate the problem.

akesandgren:
There is a basic problem here...
intel/foss 2017b, and 2018a (and a few other non-a|b versions) all uses GCCcore 6.4.0
But we can't expect versions for software A to be the same for 2017b and 2018a just because they use GCCcore 6.4.0.
The current check tries to enforce that.

boegel:
@akesandgren The check doesn't enforce that only `GCCcore/6.4.0` easyconfig files for one version of software A exist.

It enforces that only one variant of software A is used as a dependency in easyconfigs that use `intel/2018a` (and it's fine if that's different from versions of A for dependencies in other easyconfigs).

So the latter is not a problem, the check only enforces things i) for dependencies, ii) for easyconfigs using the same (common) toolchain generation (e.g. `2018a`).

akesandgren:
Ok, I might have misunderstood the complaints from travis for my p4vasp PR.
I didn't investigate it fully due to kernel updates...

boegel:
@akesandgren The errors in #6527 are due to conflicts, not the version check this issue refers to...

Use `eb --check-conflicts` on the `p4vasp` easyconfig.

branfosj:
The multi-variant checks have been enhanced significantly (see  #12687, #23133, #23134, #23177). So I'll close this.

PR #10377, opened 2020-04-12
{math}[foss/2018b] numpy v1.15.0 w/ Python 3.6.6

kelseymh:
(copied title from closed PR #10335 )

Update to latest version of Numpy for users tied to foss-2018b toolchain.  This version of Numpy requires Cython 0.29.14 or later, so this branch also includes a Cython-0.29.16 build with the same toolchain and Python.

Resubmitted per comments from @boegel in closed PR #10334 .


kelseymh:
I still don't understand how to fix the failing test.  The reported error doesn't even seem to be local to this .eb file:

> AssertionError: ['pytest'] is not false : No multi-variant deps found for '^.*-(?P<tc_gen>201[89][ab]).*\.eb$' easyconfigs:
found 3 variants of 'pytest' dependency in easyconfigs using '2018b' toolchain generation
* version: 3.8.2; versionsuffix: -Python-2.7.15 as dep for {'PSI4-1.2.1-intel-2018b-Python-2.7.15-maxam8.eb', 'PSI4-1.2.1-intel-2018b-Python-2.7.15.eb'}
* version: 3.8.2; versionsuffix: -Python-3.6.6 as dep for {'CITE-seq-Count-1.4.3-foss-2018b-Python-3.6.6.eb', 'ProjectQ-0.4.2-intel-2018b-Python-3.6.6.eb'}
* version: 4.4.0; versionsuffix: -Python-3.6.6 as dep for {'numpy-1.18.2-foss-2018b-Python-3.6.6.eb'}

I'd be happy to back up to version 3.8.2 (which is available for foss-2018b-Python-3.6.6), but I think I'd still get the error above, just without the third bullet.

Advice?

kelseymh:
Related issue: #6493 .  It seems like a _lot_ of pull requests have failed due to this test -- 87 pull requests matched a search for "multi-variant dep", in addition to this one.  I'm going to try pulling my numpy back to pytest 3.8.2 (I don't have any reason not to do so).  

It really would help if either the documentation, or the test output, could provide some guidance to writers on what they should do to address this.  This test doesn't appear to get run by `--check-contrib`, so we can't even experiment locally before submitting.

kelseymh:
Woo hoo!  So backing up to the same pytest 3.8.2 worked.  Now we wait :-)

smoors:
@kelseymh the reason for the failed check is that, in the official repo, only one version of a given software can be used as dependency for other softwares (with some exceptions). pytest-3.8.2 is already used as a dep, so 4.4.0 is not allowed.
ideally the same checks should indeed be performed by `--check-contrib`.

kelseymh:
@smoors wrote:
> ideally the same checks should indeed be performed by `--check-contrib`.

Agreed!  But they aren't, at least not with the EB installation at Texas A&M HPRC (which is currently using EasyBuild/2020a_sprint).  Should I create an issue ticket requesting this feature?

lexming:
@kelseymh thanks for updating the PR but `numpy` v1.15.0 is already provided by `Python-3.6.6-foss-2018b.eb`. Sorry for the confusion. The way we handle it has changed a lot in the last toolchains, from having its own easyconfig to be bundled into Python to now be part of `Scipy-bundle`. I think that we can close this PR and focus on #10424 .

PR #10388, opened 2020-04-14
Check sdist with different python versions

Flamefire:
Fixes regression introduced in https://github.com/easybuilders/easybuild-easyconfigs/pull/10326/commits/1ad052ec99a97a563ee65257cad710df321bf122

@boegel


boegel:
Thanks for the fix @Flamefire!

PR #10394, opened 2020-04-14
sync develop with master after EasyBuild v4.2.0 release + bump version to 4.2.1dev

boegel:


lexming:
Going in, thanks @boegel !

PR #10400, opened 2020-04-15
2020a synch with devel

ocaisa:


boegel:
I did this sync of `2020a` with current `develop` manually just now, so I'll close this...

PR #10488, opened 2020-04-26
clean up easyconfigs in 2020a branch that are no longer relevant

boegel:
(created using `eb --new-pr`)

This easyconfigs were added before jumping to GCCcore 9.3.0, they don't need to be collapsed into `develop`...


akesandgren:
Going in, thanks @boegel!

PR #10280, opened 2020-03-29
impi v2019.6: add dependency on UCX

lexming:
(created using `eb --new-pr`)

Intel MPI update 6 has introduced the new `MLX` libfabric provider that needs `UCX` to work. https://software.intel.com/en-us/articles/improve-performance-and-stability-with-intel-mpi-library-on-infiniband

As an additional note, the documentation says that `UCX v1.4 or higher` is required, but in our extensive testing `MLX` is unusable with `UCX-1.5.1` whereas it works well with `UCX-1.7.0`.

PR #10489, opened 2020-04-26
add URL in comment to explain why UCX is included as a dep for impi 2019.7.217

boegel:
@lexming should have been part of #10280 ;)


lexming:
Going in, thanks @boegel !

PR #10602, opened 2020-05-08
{bio}[GCC/8.3.0] Control-FREEC and dependencies

jordiblasco:
(created using `eb --new-pr`)


easybuild/easyconfigs/c/Control-FREEC/Control-FREEC-11.5-GCC-8.3.0.eb
```diff
@@ -0,0 +1,59 @@
...
+toolchain = {'name': 'GCC', 'version': '8.3.0'}
+
+builddependencies = [
+    ('Autoconf', '2.69', '', ('GCCcore', '8.3.0')),
+    ('Automake', '1.16.1', '', ('GCCcore', '8.3.0')),
+    ('Autotools', '20180311', '', ('GCCcore', '8.3.0')),
+]
+
+dependencies = [
+    ('R', '3.6.2', '', ('foss', '2019b')),
```
Micket, line 30: This breaks HMNS. This config simply needs to be on the foss level.

jordiblasco: done!

akesandgren:
Closing since this is an old toolchain and we have newer versions available already.
If the newer ones are missing dependencies please open a new PR using a newer version and a newer toolchain

PR #10441, opened 2020-04-20
Create tmux-3.0.eb

levlafayette:

PR #10658, opened 2020-05-21
Create libevent-2.1.11.eb

levlafayette:


easybuild/easyconfigs/l/libevent/libevent-2.1.11.eb
```diff
@@ -0,0 +1,29 @@
...
+
+description = """
+ The libevent API provides a mechanism to execute a callback function when
+ a specific event occurs on a file descriptor or after a timeout has been
+ reached.  Furthermore, libevent also support callbacks due to signals or
+ regular timeouts.
+"""
+
+toolchain = SYSTEM
+toolchainopts = {'optarch': True}
```
smoors, line 16: @levlafayette please add 'pic' to be in line with the GCCcore based version, and remove 'optarch' because it is already True by default:
```suggestion
toolchainopts = {'pic': True}
```

easybuild/easyconfigs/l/libevent/libevent-2.1.11.eb
```diff
@@ -0,0 +1,29 @@
...
+
+source_urls = ['https://github.com/%(name)s/%(name)s/releases/download/release-%(version)s-stable/']
+sources = ['%(name)s-%(version)s-stable.tar.gz']
+checksums = ['a65bac6202ea8c5609fd5c7e480e6d25de467ea1917c08290c521752f147283d']
+
+
```
smoors, line 22: @levlafayette please add binutils builddep:
```
builddependencies = [('binutils', '2.34')]
```

akesandgren:
We already have libevent 2.1.11 at GCCcore level from GCCcore 7.2.0 onwards. Why would you need  it at SYSTEM level?
What's the use case?

levlafayette:
> We already have libevent 2.1.11 at GCCcore level from GCCcore 7.2.0 onwards. Why would you need it at SYSTEM level?
> What's the use case?

Dependency for SYSTEM Tmux 3.0

akesandgren:
Being curious, why would you want that, Tmux, built with easybuild? I would just use the one from the OS anytime...

boegel:
Quite old by now, and no longer needed since #10441 is closed, so closing this PR

PR #10711, opened 2020-05-28
R-3.6.0-foss-2019a: add intrinsicDimension extension

asoete:
Add intrinsicDimension and its dependency yaImpute to R-6.3.0 extensions. This
extension is part of R-4.0.0 so I figured it wasn't a problem to include it
into R-6.3.0.

Is there a policy as to which extensions are included by default?


boegel:
No activity since May 2020, `intrinsicDimension` is included in more recent `R` installations (>= 3.6.2), so I don't think this is worth pursuing...

PR #10751, opened 2020-06-04
Qt5 5.12.8

bartoldeman:
@mboisson this one builds, can you please review if I did not miss anything obvious?

I updated sip from 4.19.17 to .18 for Py 3.8 compatibility and made everything a bit more like the upstream recipes for PyQt and Qt.

For py2.7 we just need enum34 (which is not needed for py3), so the pip install is conditional there, it's a little hackish but it works.

In the end it seems the vectorizer can remain enabled I think (still waiting for the avx512 test build to finish..)


bartoldeman:
wrong place...

PR #10823, opened 2020-06-15
{bio}[GCC/9.3.0] Pysam v0.15.4 w/ Python 2.7.18 + 3.8.2

verdurin:
(created using `eb --new-pr`)


FokkeDijkstra:
I was already working on version 0.16.0. I ran into the same problem with sanity_pip_check you have. But sanity_pip_check has been explicitly disabled in the previous version. This is still stated in the comments.
Unfortunately I can't find more information on why it was disabled. The previous version was simply accepted as far as I can see.

verdurin:
@FokkeDijkstra Thanks for pointing that out. One way around this would be to remove the Python 2 version, which isn't of particular interest for us at least.

FokkeDijkstra:
I did a small test for 0.16.0 with one installations having sanity_pip_check enabled and the other disabled. I can't find meaningful differences between both installation directories. The only differences are in .pyc and .so files, which is to be expected.
So I will switch to setting sanity_pip_check_enabled to True.

FokkeDijkstra:
Is there a reason you are using 0.15.4 instead of moving to 0.16.0? Personally I don't care, but it is best to not have more versions than necessary in the repository for a given GCC version.
If you have a good reason for sticking with 0.15.4, I'll close my PR for 0.16.0.

verdurin:
It's because the changelog says there is a behaviour change in `0.16.0` and I was only really interested in `Pysam` as a dependency of something else. Being perhaps over-conservative.

FokkeDijkstra:
I've seen that too. But I wasn't sure if it would affect deepTools, for which I required Pysam.

smoors:
> It's because the changelog says there is a behaviour change in `0.16.0` and I was only really interested in `Pysam` as a dependency of something else. Being perhaps over-conservative.

@verdurin which dependency are you referring to exactly?

verdurin:
On Mon, Jun 22, 2020 at 06:00:10AM -0700, Sam Moors wrote:
> > It's because the changelog says there is a behaviour change in `0.16.0` and I was only really interested in `Pysam` as a dependency of something else. Being perhaps over-conservative.
> 
> @verdurin which dependency are you referring to exactly?
> 

https://github.com/easybuilders/easybuild-easyconfigs/pull/10825

smoors:
@verdurin when installing UMI-tools with `pip` and `pip3`, pysam-0.16.0.1 is installed as a dependency, so I think it's safe to upgrade to `0.16.0.1`?

verdurin:
@smoors thanks for testing. I'm happy to close this one in favour of @FokkeDijkstra if he updates it to the latest, then I'll alter #10825

verdurin:
Closing as now obsolete.

PR #10839, opened 2020-06-22
Create MrBayes-3.2.7-foss-2018b.eb

levlafayette:
About time for an update :)


akesandgren:
This toolchain is old, we have newer versions available.
Closing.

PR #10924, opened 2020-07-06
{vis}[foss/2019b] update Graphviz v2.40.1

fizwit:
(created using `eb --new-pr`)


easybuild/easyconfigs/g/Graphviz/Graphviz-2.40.1-foss-2019b.eb
```diff
@@ -0,0 +1,71 @@
...
+description = """Graphviz is open source graph visualization software. Graph
+ visualization is a way of representing structural information as diagrams of
+ abstract graphs and networks. It has important applications in networking,
+ bioinformatics,  software engineering, database and web design, machine
+ learning, and in visual interfaces for other technical domains."""
+
+toolchain = {'name': 'foss', 'version': '2019b'}
+
+source_urls = ['https://graphviz.gitlab.io/pub/graphviz/stable/SOURCES']
+sources = [SOURCELOWER_TAR_GZ]
```
lexming, line 21: This does not work. The URL for the official download is http://graphviz.gitlab.io/pub/graphviz/stable/SOURCES/graphviz.tar.gz. The problem is that it has always lacked any versioning, so there is no way to keep track of changes to it.

fizwit:
Sorry, I did not see it!


John Dey
john@fuzzdog.com

lexming:
@fizwit no problem, thanks for reporting the broken download for Graphviz v2.40.1. I opened PR #10944 to fix it in the affected easyconfigs.

PR #10935, opened 2020-07-07
(WIP) added CUDA 11.0.2 and related recipes

mboisson:


easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb
```diff
@@ -0,0 +1,17 @@
+easyblock = 'Bundle'
+name = 'CUDA'
+version = '11.0'
```
Micket, line 3: Is there really such a benefit from dropping the last version? 
If the purpose is to treat this like we do for Java, we it would either need to be
1. CUDAcore/11.0 a wrapper at the core level that toolchains use (but I don't think we ever want to change the CUDA behind the back of a toolchain)
2. Both CUDA/11.0.2  and the "wrapper" CUDA/11.0 for all the CUDA's installed under iccifort.

There are some important nuances here, especially when it comes to HMNS

easybuild/easyconfigs/c/CUDAcore/CUDAcore-11.0.2.eb
```diff
@@ -0,0 +1,19 @@
...
+
+toolchain = SYSTEM
+
+#http://developer.download.nvidia.com/compute/cuda/11.0.2/local_installers/cuda_11.0.2_450.51.05_linux.run
+
+source_urls = ['https://developer.download.nvidia.com/compute/cuda/%(version)s/local_installers/']
+sources = ['cuda_%(version)s_450.51.05_linux.run']
+checksums = ['48247ada0e3f106051029ae8f70fbd0c238040f58b0880e55026374a959a69c1']
+
+moduleclass = 'system'
```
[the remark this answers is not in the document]
mboisson: mmmm, where does the `%%(cudaarch)s` come from ?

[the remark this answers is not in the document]
Micket: The cuda easyblock itself;
https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/c/cuda.py#L74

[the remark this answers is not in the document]
mboisson: ok, done

easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb
mboisson: I think it adds site-specific flexibility. 

If a site installs CUDA, but wants to install a revision that is more recent than what EasyBuild upstream provides for, that site can still use all of the EasyConfig which use `('CUDA', '11.0')` as dependencies without having to change a lot of easyconfigs

mboisson:
I'm not sure what this failure means 
```
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 773, in test_changed_files_pull_request
    self.check_sanity_check_paths(changed_ecs)
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 650, in check_sanity_check_paths
    self.assertFalse(failing_checks, '\n'.join(failing_checks))
AssertionError: ['No custom sanity_check_paths found in CUDA-11.0-GCC-9.3.0.eb', 'No custom sanity_check_paths found in CUDA-11.0-iccifort-2020.1.217.eb'] is not false : No custom sanity_check_paths found in CUDA-11.0-GCC-9.3.0.eb
No custom sanity_check_paths found in CUDA-11.0-iccifort-2020.1.217.eb
``` 

there are no `sanity_check_paths` in any of the other CUDA recipes ?

mboisson:
Ah, it's because it's using a `Bundle` EasyBlock, which is not in here : 
``` 628         whitelist = ['BuildEnv', 'CrayToolchain', 'GoPackage', 'ModuleRC', 'PythonBundle', 'PythonPackage',
 629                      'Toolchain']``` 

Why are `PythonBundle` excluded, but not `Bundle` ?

**Comment by Micket** [comment]
Test report by @Micket
**SUCCESS**
hebbe-c1 - Linux centos linux 7.8.2003, x86_64, Intel Core Processor (Haswell, no TSX), Python 2.7.5
See https://gist.github.com/4c6265827ff87a3f08452518f6b763c0 for a full test report.

**Comment by Micket** [comment]
> Why are `PythonBundle` excluded, but not `Bundle` ?

PythonPackage and PythonBundle are excluded because they contain default implementations for sanity checks. It checks for the `site-packages` directory
Bundles typically always contain stuff though, this is really a.. uh. well I don't know really. The only cases that are module-only are actually `ModuleRC` (which wouldn't work) and `Toolchain` as far as I know.

**Comment by mboisson** [comment]
> > Why are `PythonBundle` excluded, but not `Bundle` ?
> 
> PythonPackage and PythonBundle are excluded because they contain default implementations for sanity checks. It checks for the `site-packages` directory
> Bundles typically always contain stuff though, this is really a.. uh. well I don't know really. The only cases that are module-only are actually `ModuleRC` (which wouldn't work) and `Toolchain` as far as I know. (oh, and BuildEnv, but that's not right either)

I can try to create an empty `sanity_check_paths` ?

**Comment by mboisson** [comment]
Ah, no, I need to add to 
https://github.com/easybuilders/easybuild-easyconfigs/blob/48c67860efc91fab6c41620a3364b1d7b86422d6/test/easyconfigs/easyconfigs.py#L635

File: easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb (hunk 1 above)
Comment (Micket) [reply easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb:3]: OK, hm, well, I'm not saying I'm against it but just some thoughts:
1. This is just the CUDA part so, you're still stuck when it comes to the compiler and such.
2. Wouldn't be necessary for the fosscuda level, so that limits the amount of software affected anyway
3. It would come back to bite us (slightly) if we ever decided to also add `CUDA-11.0.3-iccifort-2020.1.217`
Both the for- and against arguments seem to be somewhat theoretical though, so I'm not sure it matters much.

Just to be clear, as it stands, this is very different implementation (with a different motivation/rationale) from what @boegel mentioned in the chat:
> We do this for Java too. And we should consider doing it for more stuff, like Python, and do in-place updates.

This can not do in-place updates, and, I'm not sure it ever can, considering it's part of a toolchain hierarchy thing.

**Comment by mboisson** [comment]
> Ah, no, I need to add to
> https://github.com/easybuilders/easybuild-easyconfigs/blob/48c67860efc91fab6c41620a3364b1d7b86422d6/test/easyconfigs/easyconfigs.py#L635

That is done in 
https://github.com/easybuilders/easybuild-easyconfigs/pull/10936

File: easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb (hunk 1 above)
Comment (mboisson) [reply easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb:3]: The idea is that the toolchains also use `('CUDA', '11.0')`. See the `gcccuda-2020a.eb` and `iccifortcuda-2020a.eb` attached to this PR. 

A site who wants to install `CUDAcore-11.0.3.eb` and use that as their base only has to change the two files`CUDA-11.0-iccifort-2020.1.217.eb` and `CUDA-11.0-GCC-9.3.0.eb` files, since toolchains depend on those, and not directly on `CUDAcore-11.0.2.eb`

**Comment by Micket** [comment]
(this is still the most reliable way to re-test a PR)

**Comment by Micket** [comment]
Re-running tests

File: easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb (hunk 1 above)
Comment (mboisson) [reply easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb:3]: Maybe @bartoldeman and @boegel can pitch in on this

**Comment by mboisson** [comment]
I'm not sure why the tests fail with 
```
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 147, in test_conflicts
    self.process_all_easyconfigs()
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 120, in process_all_easyconfigs
    EasyConfigTest.ordered_specs = resolve_dependencies(EasyConfigTest.parsed_easyconfigs, modules_tool(), retain_all_deps=True)
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/easybuild/tools/robot.py", line 460, in resolve_dependencies
    raise_error_missing_deps(totally_missing, extra_msg="no easyconfig file or existing module found")
  File "/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/easybuild/tools/robot.py", line 323, in raise_error_missing_deps
    raise EasyBuildError(error_msg)
easybuild.tools.build_log.EasyBuildError: 'Missing dependencies: CUDAcore/11.0.2-GCC-9.3.0, CUDAcore/11.0.2-iccifort-2020.1.217 (no easyconfig file or existing module found)'
``` 

The recipes do not call for `CUDAcore/11.0.2-GCC-9.3.0`. Shouldn't EasyBuild automatically resolve the dependencies are at the SYSTEM level ?

**Comment by mboisson** [comment]
Looks like the build is simply too long ?

File: easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb (hunk 1 above)
Comment (bartoldeman) [reply easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb:3]: There was some discussion in the conf call (I wasn't there, it's the early one): https://github.com/easybuilders/easybuild/wiki/Conference-call-notes-20200708

I'd stay with 11.0.3, it makes more sense if you look at the other toolchains (from an upstream perspective, from a CC perspective 11.0 is ok but that also has to do in how we expose cuda to users).

**Comment by bartoldeman** [comment]
Travis has issues from time to time. I'll try again.

File: easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb (hunk 1 above)
Comment (mboisson) [reply easybuild/easyconfigs/c/CUDA/CUDA-11.0-iccifort-2020.1.217.eb]: ok. I restored the revision number.

**Comment by bartoldeman** [comment]
UCX should be built with CUDA support too, Open MPI in here is using the non-CUDA enabled UCX. But that would need the gcccorecuda toolchain as explained in slack.

**Comment by mboisson** [comment]
Ah.... so we need to do a PR for easyblocks ? Is it just the new toolchain, or something else ?
  

  
  

  
  
>

**Comment by bartoldeman** [comment]
No new easyblock is needed, but a gcccorecuda toolchain needs to be added to the framework.

Framework PR filed here: https://github.com/easybuilders/easybuild-framework/pull/3385

File: easybuild/easyconfigs/o/OpenMPI/OpenMPI-4.0.3-iccifortcuda-2020a.eb [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+toolchain = {'name': 'iccifortcuda', 'version': '2020a'}
+
+source_urls = ['https://www.open-mpi.org/software/ompi/v%(version_major_minor)s/downloads']
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['6346bf976001ad274c7e018d6cc35c92bbb9426d8f7754fac00a17ea5ac8eebc']
+
+dependencies = [
+    ('zlib', '1.2.11'),
+    ('hwloc', '2.2.0'),
+    ('UCX', '1.8.0', '', ('gcccorecuda', '2020a')),
```
boegel, line 16: There should be a comment here to explain why we do this.

boegel (changes requested):

mboisson:
This is no longer relevant I guess.

PR #10936, opened 2020-07-07
Whitelist CUDA bundles from sanity_check_paths requirement

mboisson:


mboisson:
This needs review from @Micket. It goes with the PR for cuda 11.0.2

PR #10952, opened 2020-07-09
Update EB 4.2.2 easyconfig, `dummy` is deprecated

ocaisa:


ocaisa:
> this was left like this in order to allow people to upgrade from EB 3

Whoops, sorry for the noise

PR #11008, opened 2020-07-22
use travis_wait to extend 10m timeout when no output is produced

boegel:
I consider this a last resort for keeping Travis...
If it still causes trouble with this included, we should stop using it for this repo at least.

I'm already working on getting @boegelbot to submit comments based on the test suite results from GitHub Actions rather than Travis (for easyconfigs)...


verdurin:
Going in, thanks @boegel!

PR #11020, opened 2020-07-23
Use is_generic_easyblock from filetools

Flamefire:
This avoids the deprecation warnings on CI.

Seems like my autopep8 also found 2 overlong lines in that file. I'm inclined to keep that change.


Flamefire:
Deprecation introduced in https://github.com/easybuilders/easybuild-framework/pull/3189 by @migueldiascosta

PR #10954, opened 2020-07-09
(WIP) {g,i}omklc toolchains and relatives, version 2020a

mboisson:

PR #11035, opened 2020-07-29
added g-i-omkl-2020a toolchains

mboisson:
This is to split the cuda part from the PR https://github.com/easybuilders/easybuild-easyconfigs/pull/10954

@Micket


mboisson:
This PR puts MKL at the SYSTEM level. This means that toolchains such as `gmkl` and `iimkl`, which include MKL but does not include MPI, are possible. 

This allows - for example -  building R with MKL, without having to place it at the MPI level. In fact, a `gcccoremkl` is even possible (although not include in this PR), allowing to install software which depend on MKL, but not on vectorization at the core level. This avoids having to recompile those pieces of software multiple times.

boegel:
@mboisson This needs a sync with `develop` now #11036 is merged...

mboisson:
@boegel done

PR #11055, opened 2020-08-03
stop testing easyconfig PRs with Travis, only use GitHub Actions from now on

boegel:
The tests in Travis CI have been a PITA recently: next to the frequent fluke failures (which were mitigated through my bot fairly efficiently), we've also been hitting timeouts where the test suite doesn't generate any output for too long, causing Travis to cancel the test job.

We tried using `travis_wait` (cfr. #11008) to make Travis be more patient, but that doesn't seem to be enough, see for example:

```
$ travis_wait python -O -m test.easyconfigs.suite

Still running (20 of 20): python -O -m test.easyconfigs.suite

Timeout (20 minutes) reached. Terminating "python -O -m test.easyconfigs.suite"

The command python -O -m test.easyconfigs.suite exited with 137.
```

(from https://travis-ci.org/github/easybuilders/easybuild-easyconfigs/jobs/714454534)

Since we're only testing with Python 2.7 and 3.6 in Travis, there's nothing we're testing in Travis that is not also being tested in GitHub Actions.

So, I'm calling it: no more testing with Travis for easyconfig PRs, only GitHub Actions from now on!


casparvl:
Ok, I have to admit that I don't know exactly where the Travis integration links in, but I'm guessing on the `easyconfigs` side of things, it is limited to these two? Are there other places that need to be adapted together with this PR on the `easybuild-framework` side?

boegel:
@casparvl The presence of the `.travis.yml` file determines whether Travis will test PRs (along with some configuration in Travis itself).

I'm only disabling Travis for the easyconfigs repository here. We can keep it for the other repos for now (and certainly for framework I'd like to for now, otherwise we won't be testing against Python 2.6 anymore).

casparvl:
RIP Travis... Going in, thanks @boegel

boegel:
@casparvl docs update in https://github.com/easybuilders/easybuild/pull/641

PR #11084, opened 2020-08-10
rename sdsl-lite to SDSL (to use one single name)

mboisson:

PR #11103, opened 2020-08-12
{bio}[GCC/9.3.0] BioBloomTools v2.3.2-20200731

mboisson:
(created using `eb --new-pr`)


mboisson:
This recipe can't provide checksums. Also, the missing dependencies are other PRs except for Boost-1.72.0-GCC-9.3.0.eb.  

Should I open a PR for Boost-1.72.0-GCC-9.3.0.eb, or is it frowned upon to install Boost without MPI support ? If I should not, then I won't open more PRs that depend on Boost at compiler level.

@boegel @bartoldeman

mboisson:
In my opinion, installing Boost only at the MPI level of the hierarchy bumps way too many recipes up to the MPI level.

mboisson:
This PR requires https://github.com/easybuilders/easybuild-easyconfigs/pull/11145

mboisson:
This PR also requires https://github.com/easybuilders/easybuild-easyconfigs/pull/11084

akesandgren:
Boost was moved to GCC level some time ago, feel free to updated this to a newer toolchain, this toolchain has been deprecated.

PR #11120, opened 2020-08-15
Adding support for FPM 1.4.0

alirezaghavaminia:
Trying to install FPM on CentOS 8, It appears that the GCC version on the CentOS 8 is very high so the Ruby and FPM easyconfigs must be updated accordingly.


boegel:
@alirezaghavaminia Thanks a lot for your contribution!

Do let us know if you need help with making the test suite pass in CI...

You basically should add SHA256 checksums to your easyconfigs (for which we recommend using `eb --inject-checksums`), and there are some minor style issues to fix:

```
$ eb --from-pr 11120 --check-style

Running style check on 2 easyconfig(s)...

/tmp/eb-SWASHW/files_pr11120/f/FPM/FPM-1.4.0-Ruby-2.7.1.eb:26:1: W293 blank line contains whitespace
/tmp/eb-SWASHW/files_pr11120/f/FPM/FPM-1.4.0-Ruby-2.7.1.eb:26:1: W293 blank line contains whitespace
/tmp/eb-SWASHW/files_pr11120/f/FPM/FPM-1.4.0-Ruby-2.7.1.eb:26:1: W391 blank line at end of file
[FAIL] /tmp/eb-SWASHW/files_pr11120/f/FPM/FPM-1.4.0-Ruby-2.7.1.eb
[PASS] /tmp/eb-SWASHW/files_pr11120/r/Ruby/Ruby-2.7.1.eb

File: easybuild/easyconfigs/f/FPM/FPM-1.4.0-Ruby-2.7.1.eb [hunk 1]
```diff
@@ -0,0 +1,26 @@
+easyblock = 'RubyGem'
+
+name = 'FPM'
+version = '1.4.0'
```
boegel, line 4: @alirezaghavaminia Why not update to the latest available version (1.11.0)? See https://github.com/jordansissel/fpm/releases

alirezaghavaminia: Oh, I see. I updated based on [RubyGems.org](https://rubygems.org/gems/fpm/versions/1.4.0) website.

alirezaghavaminia:
Style check will pass now. 

```bash
$ eb --from-pr 11120 --check-style
== temporary log file in case of crash /tmp/eb-6AK3FH/easybuild-RhqZKB.log

Running style check on 2 easyconfig(s)...

[PASS] /tmp/eb-6AK3FH/files_pr11120/f/FPM/FPM-1.4.0-Ruby-2.7.1.eb
[PASS] /tmp/eb-6AK3FH/files_pr11120/r/Ruby/Ruby-2.7.1.eb

>> All style checks PASSed!
```

easybuild/easyconfigs/f/FPM/FPM-1.4.0-Ruby-2.7.1.eb
alirezaghavaminia: I think I had dependency conflict issues. I'm not sure...

alirezaghavaminia:
> http://cache.ruby-lang.org/pub/ruby/

I have checked. The `HTTP` address is correct. I don't know what did I done wrong.

lexming:
@alirezaghavaminia the URL http://cache.ruby-lang.org/pub/ruby/ does work, but https://cache.ruby-lang.org/pub/ruby/ is also available. In such a case we prefer to use the secure URL

PR #11132, opened 2020-08-18
 MPICH-3.3.2-GCC-9.3.0-2.34.eb

DinaMah:


easybuild/easyconfigs/m/MPICH/MPICH-3.3.2-GCC-9.3.0-2.34.eb
```diff
@@ -0,0 +1,14 @@
+name = 'MPICH'
+version = '3.3.2'
+
+homepage = 'http://www.mpich.org/'
```
boegel, line 4: @DinaMah Please use `https` here:

```suggestion
homepage = 'https://www.mpich.org/'

File: easybuild/easyconfigs/m/MPICH/MPICH-3.3.2-GCC-9.3.0-2.34.eb [hunk 2]
```diff
@@ -0,0 +1,14 @@
...
+description = """MPICH v3.x is an open source high-performance MPI 3.0 implementation.
+It does not support InfiniBand (use MVAPICH2 with InfiniBand devices)."""
+
+toolchain = {'name': 'GCC', 'version': '9.3.0-2.34'}
```
boegel, line 8: We recently dropped the binutils version from the `GCC` toolchain, since we stick to a single binutils version per GCC version anyway:

```suggestion
toolchain = {'name': 'GCC', 'version': '9.3.0'}

boegel (changes requested):

PR #11153, opened 2020-08-20
Easyconfig for OpenSCAD v2019.05

rtripath89:


mboisson:
@rtripath89, if it is a Binary, it should probably be installed with the SYSTEM toolchain, rather than foss-2018 ?

mboisson:
Note that this does not work on our system. This app requires FUSE, which may or may not be installed on a given system.

rtripath89:
> Note that this does not work on our system. This app requires FUSE, which may or may not be installed on a given system.

Aahhh okay ...... I'll try to add FUSE as a dependency ....... it does require foss toolchain to work properly.

mboisson:
>  it does require foss toolchain to work properly.

Really ? It seems to be a single binary with very few dependencies : 
```
[mboisson@beluga4 ~]$ readelf -d /home/mboisson/.local/easybuild/software/2017/Core/openscad/2019.05/openscad

Dynamic section at offset 0x2b7c0 contains 23 entries:
  Étiquettes Type                         Nom/Valeur
 0x0000000000000001 (NEEDED)             Bibliothèque partagée: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Bibliothèque partagée: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Bibliothèque partagée: [libz.so.1]
 0x0000000000000001 (NEEDED)             Bibliothèque partagée: [libc.so.6]
 0x000000000000000c (INIT)               0x4040d8
 0x000000000000000d (FINI)               0x420d98
``` 

What requires foss in that ?

rtripath89:
image

It gives an error if the gcc version on your system is old.

./OpenSCAD-2019.05-x86_64.AppImage: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./OpenSCAD-2019.05-x86_64.AppImage)

mboisson:
Ah, ok. That would be fixed by a recent enough `GCCcore`, rather than `foss`. But I guess for the 2018 toolchains, we might not have a lot of use for GCCcore yet.

rtripath89:
I think I can just use GCC, as a toolchain as well, and that should work too.

PR #11155, opened 2020-08-20
{data}[GCC/9.3.0,iccifort/2020.1.217] HDF5.Serial v1.10.6

mboisson:
(created using `eb --new-pr`)


lexming (changes requested):
Serial builds of HDF5 should use `HDF5` as package name. They can differentiate from other HDF5 builds by adding `serial` as `versionsuffix`. As an example, we have `HDF5-1.8.17-GCC-5.4.0-2.26-serial.eb`

mboisson:
@lexming no. That is contrary to what was discussed in the latest EB conference call. paging @boegel

lexming:
@mboisson we have always handled MPI and serial builds with `versionsuffix`. What is the advantage in creating multiple package names for the same software? I'll wait for the feedback from @boegel

boegel:
@lexming We discussed this extensively during the last EasyBuild conf call, and @mboisson pointed out.

@mboisson raised the problem (rightfully so) that packages like Boost, HDF5, netCDF which have a small (and rarely actually used) MPI component cause other packages to be installed with an MPI toolchain, which doesn't make much sense. One example is `MariaDB`, which is installed with the `gompi` toolchain only because `Boost` is built with it.

After a fairly extensive discussion during the call we agreed to add easyconfigs like `HDF5.Serial` (although I'd go with `HDF5.serial`, but that's a detail) for the non-MPI version of HDF5, so we can keep things that depend on it (and don't use the MPI features) at a compiler-only toolchain.
That way, we avoid a name clash (and conflict) with `HDF5`, so this can be done without too causing problems in an existing easyconfigs generation (`2020a` in this case). That's *not* the case when changing `HDF5` in place to not support the MPI features (and adding an `HDF5.parallel`), or by using variants of `HDF5` that have a `-serial` or `-parallel` version suffix.

For future easyconfig versions, we can also rename the current `HDF5` to `HDF5.parallel` to indicate it has MPI features.
That way software that depends on `HDF5.serial` can be used together with  software that depends `HDF5.parallel` without trouble. A similar approach is used by `hdf5` modules that come with Cray systems.

lexming:
@boegel thanks for the explanation. If part of the goal is to be able to load both the serial and parallel builds, then I agree with the multiple package names.

lexming:
Test report by @lexming
**SUCCESS**
node376.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/d69b4c2c77be68eb3834f78a0adec4e0 for a full test report.

lexming:
Test report by @lexming
**SUCCESS**
node124.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2680 v2 @ 2.80GHz, Python 2.7.5
See https://gist.github.com/a461e3793ed3708a26a7a35a5399cb8b for a full test report.

lexming (approved):
The resulting build is identical to `HDF5` v1.10.6 except for the MPI support.

LGTM

smoors:
> we agreed to add easyconfigs like `HDF5.Serial` (although I'd go with `HDF5.serial`, but that's a detail)

@mboisson @boegel what's the plan here: keep the name or change to the lowercase variant?

zao:
Test report by @zao
**FAILED**
freja - Linux Ubuntu 20.04, x86_64, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz, Python 3.8.2
See https://gist.github.com/e3c322c10ca85b2c0f2c87d9f63bcc1b for a full test report.

zao:
It appears that my failure for the Intel build of HDF5.Serial also manifests with regular HDF5 on this OS, so not particular to this PR.

zao:
Test report by @zao
**SUCCESS**
freja - Linux Ubuntu 20.04, x86_64, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz, Python 3.8.2
See https://gist.github.com/72d634f7ee3590a70d9b94b9a6372ca4 for a full test report.

mboisson:
> > we agreed to add easyconfigs like `HDF5.Serial` (although I'd go with `HDF5.serial`, but that's a detail)
> 
> @mboisson @boegel what's the plan here: keep the name or change to the lowercase variant?

I am personally agnostic to `S` or `s`. It makes no difference to me.

smoors:
@mboisson I'm personally in favor of lowercase too, so let's change it?

lexming:
@mboisson @smoors We discussed about the introduction of serial packages in today's [EB conf call](https://github.com/easybuilders/easybuild/wiki/Conference-call-notes-20200902#qa) and we will make a specific branch to gather all these PRs and properly test them.

lexming:
Changed target to [easybuilders:dual_serial_mpi](https://github.com/easybuilders/easybuild-easyconfigs/tree/dual_serial_mpi)

lexming:
Now that we have a specific branch, we can change package names later on.

Going in, thanks @mboisson !

PR #11288, opened 2020-09-16
{tools}[gompi/2020a,iimpi/2020a] NCO v4.9.3

henkela:
(created using `eb --new-pr`)


trz42:
Closing this PR, since it (only) uses deprecated toolchain versions (gompi/2020a + iimpi/2020a),

Also ...
- Easyconfigs didn't build due to missing dependencies.
- Newer versions of NCO with newer toolchain versions for both foss and intel are available.

PR #11297, opened 2020-09-17
{bio}[foss/2019a,fosscuda/2019a] Taiyaki v5.1.0-20200617

boegel:
(created using `eb --new-pr`)

PR #10573, opened 2020-05-04
{bio}[fosscuda/2019b] Taiyaki v5.0.1-20200316, Biopython v1.75 w/ Python 3.7.4

boegel:
(created using `eb --new-pr`)

PR #10572, opened 2020-05-04
{bio}[foss/2019b] Taiyaki v5.0.1-20200316 w/ Python 3.7.4

boegel:
(created using `eb --new-pr`)

PR #11301, opened 2020-09-17
remove easyconfigs for Taiyaki that depend on PyTorch 1.3.1, since latest version still requires PyTorch 1.2.0

boegel:
The `Taiyaki` easyconfigs added in #10572 and #10573 are useless, because Taiyaki currently doesn't support anything newer than PyTorch 1.2.0.

I figured it was safe to ignore the hardcoded `torch == 1.2.0` in `setup.py`, but we learned the hard way that that's not a good idea...

Working `Taiyaki` easyconfigs are added in #11297


lexming (approved):
LGTM

PR #11310, opened 2020-09-18
{compiler,devel}[system/system] GHC v8.6.5, Stack v2.3.3

arielzn:
(created using `eb --new-pr`)


arielzn:
These are the binary installs providing the `ghc` haskell compiler and its `stack` tool.

Installing both from source is a bit annoying as they require a present version of both to compile themselves.
I checked time ago there was this PR https://github.com/easybuilders/easybuild-easyconfigs/pull/32  where it was pushed some kind of bootstrap to get a source version built.

As it's a compiler i guess is not worth to go for these kind of options, as long as the binary works should be fine.

But the big "annoyance" with current versions, is that at some point they stopped shipping a generic `linux-x86_64` binary build as it was the case for 6.12.3 .

And now they provide builds against system libs of some distros they decided to choose: https://www.haskell.org/ghc/download_ghc_8_6_5.html

I tested the pushed easyconfig on `centos7`. It works and it allowed to build some other stuff requested to us that requires GHC >= 8.2 

But I'm not sure which would be the most proper way to have this upstream, if we can somehow.
So far there's a `local_distro` variable that will point to a given binary bundle.

smoors:
Test report by @smoors
**SUCCESS**
node370.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz (skylake_avx512), Python 2.7.5
See https://gist.github.com/60a46aaee3ae659631fbc7534a8f4cd5 for a full test report.

lexming:
Test report by @lexming
**SUCCESS**
node378.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/3d08535bfa38995e16d7a2887679e3a0 for a full test report.

lexming:
Test report by @lexming
**SUCCESS**
node115.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2680 v2 @ 2.80GHz, Python 2.7.5
See https://gist.github.com/a4c464d44325c21b786b0a4cd13088c6 for a full test report.

arielzn:
To clarify the point that i don't see clear if we could have this upstream.

Can we just name this as `GHC/8.6.5` as in the current version pushed? 
or given that's a binary install specific for certain distros, we should have the suffix explicitly making this clear ? 

Doing the latter we'll get modules named as `GHC/8.6.5-centos7`,  `GHC/8.6.5-deb9`  and so on .

But then after, if we want to push easyconfigs for other apps that need `ghc` for the build. 
Can we somehow specify on those easyconfigs something as ?

    builddependencies = [ ('GHC', '8.6.5', '*', True) ]

The more I think, the less i see it possible, according to the approach followed.

We should somehow find a generic linux-x86_64, maybe I can contact the devs to ask why they stopped doing that.

lexming (commented):
I checked the installation of GHC and I think that we can use any of those versions that they distribute. We just have to pick one and guarantee the dynamic linking of those libraries that are distro dependent.
Fortunately, I think that the only tricky one is GMP, all the other libraries picked from the system are standard, such as `libc`, `libm`, `libpthread`... This is the `ldd` output for `ghc` on CentOS 7

```
$ ldd bin/ghc
        linux-vdso.so.1 =>  (0x00007fffe1be0000)
        /usr/$LIB/libsnoopy.so => /usr/lib64/libsnoopy.so (0x00007ff8ee8c4000)
        libm.so.6 => /usr/lib64/libm.so.6 (0x00007ff8ee5c2000)
        libHShaskeline-0.7.4.3-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../haskeline-0.7.4.3/libHShaskeline-0.7.4.3-ghc8.6.5.so (0x00007ff8eeb33000)
        libHSstm-2.5.0.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../stm-2.5.0.0/libHSstm-2.5.0.0-ghc8.6.5.so (0x00007ff8eeb12000)
        libHSghc-8.6.5-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../ghc-8.6.5/libHSghc-8.6.5-ghc8.6.5.so (0x00007ff8e9d91000)
        libHSterminfo-0.4.1.2-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../terminfo-0.4.1.2/libHSterminfo-0.4.1.2-ghc8.6.5.so (0x00007ff8e9d55000)
        libHSprocess-1.6.5.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../process-1.6.5.0/libHSprocess-1.6.5.0-ghc8.6.5.so (0x00007ff8e9d15000)
        libHShpc-0.6.0.3-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../hpc-0.6.0.3/libHShpc-0.6.0.3-ghc8.6.5.so (0x00007ff8e9ce1000)
        libHSghci-8.6.5-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../ghci-8.6.5/libHSghci-8.6.5-ghc8.6.5.so (0x00007ff8e9abe000)
        libHStransformers-0.5.6.2-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../transformers-0.5.6.2/libHStransformers-0.5.6.2-ghc8.6.5.so (0x00007ff8e9979000)
        libHStemplate-haskell-2.14.0.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../template-haskell-2.14.0.0/libHStemplate-haskell-2.14.0.0-ghc8.6.5.so (0x00007ff8e962a000)
        libHSpretty-1.1.3.6-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../pretty-1.1.3.6/libHSpretty-1.1.3.6-ghc8.6.5.so (0x00007ff8e95bb000)
        libHSghc-heap-8.6.5-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../ghc-heap-8.6.5/libHSghc-heap-8.6.5-ghc8.6.5.so (0x00007ff8e9564000)
        libHSghc-boot-8.6.5-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../ghc-boot-8.6.5/libHSghc-boot-8.6.5-ghc8.6.5.so (0x00007ff8e94d6000)
        libHSghc-boot-th-8.6.5-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../ghc-boot-th-8.6.5/libHSghc-boot-th-8.6.5-ghc8.6.5.so (0x00007ff8e9495000)
        libHSbinary-0.8.6.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../binary-0.8.6.0/libHSbinary-0.8.6.0-ghc8.6.5.so (0x00007ff8e93f4000)
        libHSdirectory-1.3.3.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../directory-1.3.3.0/libHSdirectory-1.3.3.0-ghc8.6.5.so (0x00007ff8e938e000)
        libHSunix-2.7.2.2-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../unix-2.7.2.2/libHSunix-2.7.2.2-ghc8.6.5.so (0x00007ff8e92ab000)
        libHStime-1.8.0.2-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../time-1.8.0.2/libHStime-1.8.0.2-ghc8.6.5.so (0x00007ff8e91c5000)
        libHSfilepath-1.4.2.1-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../filepath-1.4.2.1/libHSfilepath-1.4.2.1-ghc8.6.5.so (0x00007ff8e91a2000)
        libHScontainers-0.6.0.1-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../containers-0.6.0.1/libHScontainers-0.6.0.1-ghc8.6.5.so (0x00007ff8e8e3c000)
        libHSbytestring-0.10.8.2-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../bytestring-0.10.8.2/libHSbytestring-0.10.8.2-ghc8.6.5.so (0x00007ff8e8d52000)
        libHSdeepseq-1.4.4.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../deepseq-1.4.4.0/libHSdeepseq-1.4.4.0-ghc8.6.5.so (0x00007ff8e8d38000)
        libHSarray-0.5.3.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../array-0.5.3.0/libHSarray-0.5.3.0-ghc8.6.5.so (0x00007ff8e8cb6000)
        libHSbase-4.12.0.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../base-4.12.0.0/libHSbase-4.12.0.0-ghc8.6.5.so (0x00007ff8e832c000)
        libHSinteger-gmp-1.0.2.0-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../integer-gmp-1.0.2.0/libHSinteger-gmp-1.0.2.0-ghc8.6.5.so (0x00007ff8e82ec000)
        libHSghc-prim-0.5.3-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../ghc-prim-0.5.3/libHSghc-prim-0.5.3-ghc8.6.5.so (0x00007ff8e7e5b000)
        libHSrts_thr-ghc8.6.5.so => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../rts/libHSrts_thr-ghc8.6.5.so (0x00007ff8e7de2000)
        libffi.so.7 => /theia/scratch/brussel/101/vsc10122/easybuild-skl/software/GHC/8.6.5/lib/ghc-8.6.5/bin/../rts/libffi.so.7 (0x00007ff8e7bd4000)
        libtinfo.so.5 => /usr/lib64/libtinfo.so.5 (0x00007ff8e79aa000)
        librt.so.1 => /usr/lib64/librt.so.1 (0x00007ff8e77a2000)
        libutil.so.1 => /usr/lib64/libutil.so.1 (0x00007ff8e759f000)
        libdl.so.2 => /usr/lib64/libdl.so.2 (0x00007ff8e739b000)
        libpthread.so.0 => /usr/lib64/libpthread.so.0 (0x00007ff8e717f000)
        libgmp.so.10 => /usr/lib64/libgmp.so.10 (0x00007ff8e6f07000)
        libc.so.6 => /usr/lib64/libc.so.6 (0x00007ff8e6b39000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff8eead4000)
```

I also checked the shared objects under `lib/` and the only special dependency is again GMP. 

So I would keep the name of the easyconfig (without any suffix), pick the tarball for CentOS (for instance) and add GMP as a dependency of GHC, probably the most recent one. This also implies adding `GCCcore` as toolchain of GHC.

smoors:
@lexming I did that initially, but after discussing with @arielzn we considered it makes more sense to use `libgmp` from the system, as that is what the binary expects.
I checked that `libgmp` is automatically installed on the system, at least on centos7 and ubuntu20.04.
keeping it at the SYSTEM level means it can be reused in all toolchains and versions.

arielzn:
As GMP seems to be a lib present on default linux installs, it would be cleaner to use it to avoid a toolchain dependency for GHC binaries.

According to [this](https://abi-laboratory.pro/index.php?view=timeline&l=gmp)  GMP seems to be very backwards compatible, more than [glibc](https://abi-laboratory.pro/index.php?view=timeline&l=glibc), so it doesn't seem critical as to make GHC binaries dependent from a easyconfig GMP build.  
If we use the build for the OS with oldest versions of those, which seems to be Centos7 (glibc 2.17, gmp 6.0.0 ) your proposal might work using just relying on standard system libs.

Debian8, already EOL, was (2.19, 6.1.0) so any other _live_ Debian/Ubuntu version must be higher.
But this would require at least a test from someone on those distros.

Something introducing some noise, might be the dependency of ncurses

    libtinfo.so.5 => /usr/lib64/libtinfo.so.5

which bumped soname to .so.6 [in 2015](https://abi-laboratory.pro/index.php?view=timeline&l=ncurses), Debian 9 is on that.
If I try the install of the deb9 binary source on Centos 7, `ghc` seems to work, but every time is run prints some annoying warnings related with this.

But probably the other way around, using C7 binaries on anything newer is fine.

lexming (approved):
@smoors I agree, if you really want to stay in `SYSTEM` this is a good approach.

LGTM

arielzn:
If ncurses is a blocker, there's already an ec for `ncurses-5.9.eb` that provides the same by default on Centos7.

And eventually the same could be done with GMP, but from the links above, the one by default on each current linux should not give issues.

boegel:

boegel:
Test report by @boegel
**SUCCESS**
node3404.kirlia.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) Gold 6240 CPU @ 2.60GHz (cascadelake), Python 2.7.5
See https://gist.github.com/190f1dee8eac599a0ae67cd9ecdebf3b for a full test report.

boegel:
Thanks a lot for the great discussion here!

I agree that not using a versionsuffix makes sense here, as long as we can try and make sure the installation works on different operating systems.

zao:
Test report by @zao
**FAILED**
freja - Linux Ubuntu 20.04, x86_64, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz, Python 3.8.2
See https://gist.github.com/41ee17ebdc46dad5f9d1c2a3620d4efe for a full test report.

lexming:
Ok, tests passed in CentOS 8 with no issues.

The failed test from @zao in Ubuntu is the same issue as reported by @arielzn in https://github.com/easybuilders/easybuild-easyconfigs/pull/11310#issuecomment-697029284.

@arielzn can you add the dependency on `ncurses` as you suggested. I think that once that is added all tests will be fine.

Micket:
So, I'm a bit torn on this myself, but, should this GHC include `x86_64` as a versionsuffix? We do so for other stuff like this like Julia (and Stack, in this very PR).
But, at the same time, perhaps that isn't a good idea, if we plan to add new configs that depend on Haskell as the compiler, it would be convenient to be able to specify "GHC" as a dep regardless of whether it is for ARM or x86_64.
We just don't have a lot of easyconfigs that do this yet, so, and most of us are using x86_64 so we just haven't considered this issue before.

lexming:
I think that having a versionsuffix per architecture is not a big problem. We could still have single packages depending on GHC by using the following syntax 
```
dependencies = [
    ('GHC', {'arch=x86_64': '%(version)s.x86_64',
             'arch=ARM': '%(version)s.arm'}),
]
```

arielzn:
@lexming I tested the modification of adding `('ncurses', '5.9')`  as dependency and it worked.

But then, i tested again using the `GHC/8.6.5` module built like that, for an easyconfing of a soft that requires a GHC compiler but also Perl.

Perl easyconfigs >= 5.30  have another `ncurses` dep from their corresponding toolchain, so this breaks their usage together with a GHC having a different `ncurses` dep.

This bring us again to make GHC to depend on a toolchain `ncurses` and then making it toolchain-fixed even if it's binary.

Or a possible workaround could be a GHC easyblock that checks if is present on the system `libtinfo.so.5` then we pull the centos7 binaries if it's `libtinfo.so.6` we pull the `deb9` one.
Would that be too hacky/unclean ?

lexming:
@arielzn Since it needs ncurses 5 which is quite old, selecting a toolchain does not completely solve the problem because that would be restricted to GCC v4.9 and earlier. Then I suggest to stick to SYSTEM and add an `osdependencies` on all known packages that provide `libtinfo.so.5` . This is what I found
* CentOS 7: ncurses-libs
* CentOS 8: ncurses-compat-libs
* Debian: libtinfo5

lexming (approved):
@arielzn thanks for the update!

LGTM

lexming:
Test report by @lexming
**SUCCESS**
node377.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/33bed22e4a157d2fd6804e91ac1fc325 for a full test report.

arielzn:

lexming:
Test report by @lexming
**SUCCESS**
node377.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/12439a60dd9f2c073bf761bcea76a9d2 for a full test report.

lexming:
Test report by @lexming
**SUCCESS**
node127.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2680 v2 @ 2.80GHz, Python 2.7.5
See https://gist.github.com/4fd3e281fcf785bd6829cf9c4191d721 for a full test report.

arielzn:
tested to ping boegelbot but i don't have the right i guess :P

@zao could you try again on the Ubuntu 20.04 box

zao:
Without polluting my system with the legacy package:
```
== temporary log file in case of crash /tmp/eb-e0qwy6o0/easybuild-1k5xxs15.log
ERROR: Failed to process easyconfig /tmp/eb-e0qwy6o0/files_pr11310/g/GHC/GHC-8.6.5-x86_64.eb: One or more OS dependencies were not found: [('ncurses-libs', 'ncurses-compat-libs', 'libncurses5', 'libtinfo5')]
```

zao:
Test report by @zao
**SUCCESS**
freja - Linux Ubuntu 20.04, x86_64, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz, Python 3.8.2
See https://gist.github.com/e46332c8ca4c68fe799349cb6d1acf18 for a full test report.

zao:
Notable is that the package `libncurses5` cited as being for "OpenSUSE" collides with an Ubuntu package so there would technically be a false positive, but that package depends on `libtinfo5` in turn anyway.

zao:
Unfortunately, the GHC installation is not fully functional. GHCi fails to launch with:
```
 zao  freja  SSH  hp  $  ghci
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
<command line>: can't load .so/.DLL for: libgmp.so (libgmp.so: cannot open shared object file: No such file or directory)
```

zao:
Even worse, ghc itself can't link programs:
```
$ cat foo.hs
main = putStrLn "hi!"
$ ghc foo.hs
Linking foo ...
/usr/bin/ld.gold: error: cannot find -lgmp
collect2: error: ld returned 1 exit status
`gcc' failed in phase `Linker'. (Exit code: 1)
$
```

zao:
While `libgmp10` is present on most Ubuntu 20.04 installations and provides `libgmp.so.10`, the package for `libgmp.so` is `libgmp-dev` which isn't installed.

arielzn:
@zao the information for libncurses5 on OpenSUSE i got it from [pkgs.org](https://opensuse.pkgs.org/15.1/opensuse-oss-x86_64/libncurses5-6.1-lp151.5.41.x86_64.rpm.html) and there's the same package on [debian family indeed](https://debian.pkgs.org/9/debian-main-amd64/libncurses5_6.0+20161126-1+deb9u2_amd64.deb.html)

But yeah, as you point for Debian having that or `libtinfo5` will be fine to assure `libtinfo.so.5` is there.

I thought it was a good idea to have the osdeps for the main "server" distros.
I was actually surprised to see that all distros are *dragging* a package for this legacy library even on their latest release.

Regarding gmp, your finding shows that we actually need `gmp-devel` on SUSE/RHEL and not just `gmp`.
`gmp-devel` was also present on the Centos we tested and missed that.
 
Does having `libgmp-dev` installed makes that `ghc` install work on Ubuntu?

smoors:
perhaps, in order to have this package as portable as possible, both across OS distributions and toolchains and sites (with different non-default os packages installed), and also have it as reusable as possible as a dep in other packages, maybe we should bundle the installation of GMP and/or ncurses together with GHC?

zao:
> The same goes for debian, it will pull libgmp10 as dep. Does having `libgmp-dev` installed makes the `ghc` install work on Ubuntu?

Yes, with the package installed I can compile with `ghc` and run `ghci`.
The programs compiled with `ghc` depend on `libgmp.so.10` and doesn't need the extra package to run.

arielzn:
@smoors  this latest push according to @zao feedback, now requesting the correct gmp devel libraries should work across OS just being on the SYSTEM toolchain, as we intended.

The question is now, if we want the easyconfig requiring these OS packages as deps, gmp devel and the legacy ncurses 5.
GHC will be in principle always used as builddep, so for apps built with it, it won't be necessary on 'worker' nodes.

smoors:
> The question is now, if we want this easyconfig requiring these OS packages as deps to work, `gmp devel` and the legacy `ncurses 5`. Both so far seems to be provided on all major distros, just not by default.

indeed, that's why I suggested bundling these deps with the GHC package, as I can imagine not all sites have or want to have these os packages installed.

verdurin:
@arielzn any update here?

arielzn:
@verdurin still not.

In principle the current binary installation will now work everywhere, with the constraint of requesting these OS deps:

```
osdependencies = [
    ('gmp-devel',        # provides libgmp.so.10 on RHEL family
     'libgmp-dev',   # provides libgmp.so.10 on OpenSUSE and Debian/Ubuntu family
     ),
    ('ncurses-libs',         # provides libtinfo.so.5 on RHEL 7
     'ncurses-compat-libs',  # provides libtinfo.so.5 on RHEL 8
     'libncurses5',          # provides libtinfo.so.5 on OpenSUSE and Debian/Ubuntu
     ),
] 
```

I didn't try a bundle still.

arielzn:
This should be working now with GMP and ncurses5 deps for `ghc`  bundled together to avoid the OS deps as asked by @smoors 

As  refresh, `ghc` requires a `ghc`  compiler to be built from source. So to avoid a crazy bootstrap, as to build a given source target it requires a working `ghc` with version somehow close to it, we go for pulling a binary tarball.

This `ghc` binary requires the GMP headers during runtime, which was fulfilled before asking the corresponding `*gmp-dev*`  OS deps. 
But binaries produced by it don't need that, the shared lib used by built binaries `libgmp.so` seems to be installed by default in every distro.

For ncurses5 is because we are pulling the `ghc` binary for Centos7 which is linked to this version provided there.
Newer distros are on ncurses6 , but they keep anyway the ncurses5 as this seems to be required by other packages in the repos. In this case, is the shared libs required by `ghc` just at runtime, produced binaries are not linked to them.

So we can get rid of those OS deps, by bundling them together.

This way we can keep the bundle at the SYSTEM level and use it just as builddep on easyconfigs, together with deps from any toolchain, when a haskell compiler is required.

We pull the `ghc Centos7 x86_64` binaries as is, to date, the still supported "server" distro with oldest glibc, so this binary should be upwards compatible with all newer distros. 

@zao would you mind to test again on your Ubuntu 20.04 in a clean env ? without the OS deps requested in previous proposal.
and see if you get  working `ghc`

easybuild/easyconfigs/g/GHC/ncurses-5.9_configure_darwin.patch
```diff
@@ -0,0 +1,13 @@
+# in relation to -no-cpp-precomp bug: https://lists.gnu.org/archive/html/bug-ncurses/2012-12/msg00001.html
+# tried/proposed by Fotis, meant for MacOSX compatible ncurses builds...
```
smoors, line 2: @arielzn can you add the full author credentials here?

smoors:
Test report by @smoors
**SUCCESS**
node376.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz (skylake_avx512), Python 2.7.5
See https://gist.github.com/baa31a10a24d8c833d8c4d6a2419e16d for a full test report.

smoors:

zao:
Test report by @zao
**FAILED**
Build succeeded for 1 out of 2 (2 easyconfigs in total)
eb-u2004 - Linux Ubuntu 20.04, x86_64, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz, Python 3.8.5
See https://gist.github.com/a5173980d26cb4d086fdc246b9116e36 for a full test report.

zao:
```
lrwxrwxrwx 1 easybuild folk     13 Dec 17 08:24 /eb/software/GHC/8.6.5-x86_64/lib/libtinfo.so -> libtinfo.so.5
lrwxrwxrwx 1 easybuild folk     15 Dec 17 08:24 /eb/software/GHC/8.6.5-x86_64/lib/libtinfo.so.5 -> libtinfo.so.5.9
-rwxr-xr-x 1 easybuild folk 240744 Dec 17 08:24 /eb/software/GHC/8.6.5-x86_64/lib/libtinfo.so.5.9
```

smoors:
the libtinfo5 package is not installed by default in ubuntu 20.04, so I this probably still needs and osdependency
```
/eb/software/GHC/8.6.5-x86_64/lib/ghc-8.6.5/bin/ghc-pkg: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```

easybuild/easyconfigs/g/GHC/ncurses-5.9_configure_darwin.patch
arielzn: ok, but i actually just pull it from ncurses tree, https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/n/ncurses/ncurses-5.9_configure_darwin.patch

i just put original pusher (Fotis) creds ?

arielzn:
@zao I don't get why you get the linking error for GHC 

```
"/eb/software/GHC/8.6.5-x86_64/lib/ghc-8.6.5/bin/ghc-pkg" --force --global-package-db "/eb/software/GHC/8.6.5-x86_64/lib/ghc-8.6.5/package.conf.d" update rts/dist/package.conf.install
/eb/software/GHC/8.6.5-x86_64/lib/ghc-8.6.5/bin/ghc-pkg: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```

if `libtinfo.so.5` was correctly built and installed in the previous step, as you mention in your comment.

do we need to do something as adding the  ` %(installdir)s/lib` to `LD_LIBRARY_PATH`  before running the linking of `ghc`  ?

i thought it was natural in a bundle that each component needs all the stuff built before,  as e.g. with R extensions.

easybuild/easyconfigs/g/GHC/ncurses-5.9_configure_darwin.patch
smoors: ah so `ncurses-5.9_configure_darwin.patch` is exactly the same as the one already in the repo?
then just remove it here, normally EB should be able to find it.

smoors:
> @zao I don't get why you get the linking error for GHC
> 
> ```
> "/eb/software/GHC/8.6.5-x86_64/lib/ghc-8.6.5/bin/ghc-pkg" --force --global-package-db "/eb/software/GHC/8.6.5-x86_64/lib/ghc-8.6.5/package.conf.d" update rts/dist/package.conf.install
> /eb/software/GHC/8.6.5-x86_64/lib/ghc-8.6.5/bin/ghc-pkg: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
> ```
> 
> if `libtinfo.so.5` was correctly built and installed in the previous step, as you mention in your comment.
> 
> do we need to do something as adding the ` %(installdir)s/lib` to `LD_LIBRARY_PATH` before running the linking of last component `ghc` ?
> 
> i thought it was natural in a bundle that each component needs all the stuff built before, as it happens e.g. with R extensions.

I think that, indeed, you have to add the lib path explicitly, probably because extensions behave differently from components.
see for example PyQt5-5.15.1-GCCcore-10.2.0.eb, where PATH and PYTHONPATH are also added explicitly to one of the components.

easybuild/easyconfigs/g/GHC/ncurses-5.9_configure_darwin.patch
arielzn: it doesn't seem to be the case, all automatic tests failed badly when it was not there

see the boegelbot above

smoors: hm, you're right. installing also fails unless you specifically add the path the the original patch.
you can ignore my stupid comments then and leave as is.

arielzn:
ok, should be good now, @zao do you mind running the test again ?

zao:
Test report by @zao
**SUCCESS**
eb-u2004 - Linux Ubuntu 20.04, x86_64, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz, Python 3.8.5
See https://gist.github.com/8fe14e60c15b10303c2c7ab4833f10f6 for a full test report.

zao:
`ghci` works, `ghc` can build and link programs.

smoors:

arielzn:
@smoors looks like generoso is not listening anymore ? it didn't trigger yesterday neither

boegel:
@smoors You have to use `@boegelbot: please test @ generoso` (the `:` is missing).

We can make that less strict :)

smoors:

Micket:
Since when do we require a : ? I never use that, and it doesn't seem to work for @smoors yet?

Micket:

boegel:
I'm trying to figure out why boegelbot isn't responding... I think part of the problem is the number of comments in this PR.

smoors:
@arielzn can you make a new PR, so we can hopefully get boegelbot working on it?

smoors:
Test report by @smoors
**SUCCESS**
node375.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz (skylake_avx512), Python 2.7.5
See https://gist.github.com/bb3788cab18a894df883d0fc2c746eb7 for a full test report.

smoors (approved):
lgtm

smoors:
Going in, thanks @arielzn!

PR #11314, opened 2020-09-21
Make CI sanity check messages less confusing ("is" -> "should be")

zarybnicky:
As per Slack discussion, this PR aims to make GitHub Action unit test failures
easier to read, eg. `AssertionError: 'download_dep_fail' is set in [...]` implies that
is shouldn't be set despite demanding the opposite, whereas `AssertionError:
'download_dep_fail' should be set in [...]` makes that clear.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -575,30 +575,30 @@ class EasyConfigTest(TestCase):
             # download_dep_fail should be set when using PythonPackage
             if easyblock == 'PythonPackage':
                 if download_dep_fail is None:
-                    failing_checks.append("'download_dep_fail' set in %s" % ec_fn)
+                    failing_checks.append("'download_dep_fail' should be set in %s" % ec_fn)
 
             # use_pip should be set when using PythonPackage or PythonBundle (except for whitelisted easyconfigs)
             if easyblock in ['PythonBundle', 'PythonPackage']:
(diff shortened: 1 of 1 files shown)


test/easyconfigs/easyconfigs.py
```diff
@@ -575,30 +575,30 @@ def check_python_packages(self, changed_ecs, added_ecs_filenames):
...
                 if download_dep_fail or exts_download_dep_fail:
-                    fail = "'*download_dep_fail' set in %s (shouldn't, since PythonBundle easyblock is used)" % ec_fn
+                    fail = "'*download_dep_fail' should not be set in %s since PythonBundle easyblock is used" % ec_fn
                     failing_checks.append(fail)
 
             elif exts_defaultclass == 'PythonPackage':
                 # bundle of Python packages should use PythonBundle
                 if easyblock == 'Bundle':
-                    fail = "'PythonBundle' easyblock is used for bundle of Python packages in %s" % ec_fn
+                    fail = "'PythonBundle' easyblock should be is used for bundle of Python packages in %s" % ec_fn
```
boegel, line 594: @zarybnicky

```suggestion
                    fail = "'PythonBundle' easyblock should be used for bundle of Python packages in %s" % ec_fn
```

boegel (changes requested):

boegel (approved):

PR #11391, opened 2020-09-28
{compiler}[GCC/9.3.0] NVHPC v20.7

AndiH:
PGI beyond 20.4 is not called PGI any more but _NVIDIA HPC SDK_ (NVHPC), with 20.7 being the first version; https://developer.nvidia.com/hpc-sdk.

This pull request is part of a set of three pull requests to introduce NVHPC to EasyBuild ([1](https://github.com/easybuilders/easybuild-framework/pull/3454), [2](https://github.com/easybuilders/easybuild-easyblocks/pull/2190), [3](https://github.com/easybuilders/easybuild-easyconfigs/pull/11391)). The stack has been successfully deployed at JSC and is under use by our HPC users.  
The EasyBuild files are created after discussions with the NVHPC product owners, taking into account HPC practices and EasyBuild conventions.

This request provides an example easyconfig to make use of toolchain definitions and the NVHPC EasyBlock.

edit (@boegel): requires ~~https://github.com/easybuilders/easybuild-easyblocks/pull/2190~~ (NVHPC easyblock)


easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,43 @@
...
+version = '20.7'
+local_gccver = '9.3.0'
+versionsuffix = '-GCC-%s' % local_gccver
+
+homepage = 'https://developer.nvidia.com/hpc-sdk/'
+description = """C, C++ and Fortran compilers included with the NVIDIA HPC SDK (previously: PGI)"""
+
+toolchain = SYSTEM
+
+sources = ['nvhpc_2020_%(version_major)s%(version_minor)s_Linux_x86_64_cuda_multi.tar.gz']
```
boegel, line 11: No public download for this?
Can we use the `%(arch)s` template rather than hardcoding `x86_64`?

Also need a SHA256 checksum

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,43 @@
...
+
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.34', '', ('GCCcore', local_gccver)),
```
boegel, line 15: Why not simply make `GCC/9.3.0` the `toolchain`?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,43 @@
...
+    ('CUDA', '11.0', '', SYSTEM),
+    # This is necessary to avoid cases where just libnuma.so.1 is present in the system and -lnuma fails
+    ('numactl', '2.0.13', '', SYSTEM)
```
boegel, line 18: Why not use `numactl` built with `GCC/9.3.0`?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,43 @@
...
+]
+
+default_cuda_version = "11.0"
```
boegel, line 21: You can use the `%(cudaver)s` template here:

```python
default_cuda_version = "%(cudaver)s"

File: easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb [hunk 5]
```diff
@@ -0,0 +1,43 @@
...
+module_add_math_libs = False  # Add NVHPC's math libraries (which should be there from CUDA anyway)
+module_add_profilers = False  # Add NVHPC's NVIDIA Profilers
+module_add_nccl = False       # Add NVHPC's NCCL library
+module_add_nvshmem = False    # Add NVHPC's NVSHMEM library
+module_add_cuda = False       # Add NVHPC's bundled CUDA
+
+# We use a HMNS, so let's enforce a unique compiler
+modluafooter = '''
+family("compiler")
+add_property("arch","gpu")
```
boegel, line 36: Hmm, this is probably site-specific, what if these `arch` and `gpu` properties are not defined, then the installation would fail?

Same goes for the `compiler` family.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,43 @@
...
+'''
+
+# Always do a recursive unload on compilers
+recursive_module_unload = True
```
boegel, line 40: Any specific reason for hardcoding this?

This seems like site-specific to me (and you don't actually need this as long as you configure EasyBuild with `--recursive-module-unload`)

boegel (changes requested):

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: I took that from the old`PGI`. @damianam, do you have anything to add?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Paging @damianam on this one as well…

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Well, if I remove it here, it will not be fitting for JSC. @damianam? What should I do?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Also here, I don't know! @damianam

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
damianam: `NVHPC` is a compiler, the same as `iccifort` so this seems appropriate to me ;-)

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
damianam: We (JSC) need `numactl` in the `SYSTEM` toolchain due to other low-level dependencies that require it. So we took it from there as the lowest common point. But in this case, having it in `SYSTEM` or in `GCCcore` does not make a different IMO, so we can change it for the PR if you feel it fits better the overall stack.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
damianam: Remove them, that's part of the reason why we have a separate repo. The packages in this PR are not supposed to be used directly by us without further "tweaking". Think about `site_contacts` for instance.

Medium term, we should get a hook done that injects these things, to maximize the overlap between upstream and our repo, but it is nothing too critical.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
damianam: In a hierarchical setup makes sense I think, but upstream probably doesn't, so I would say that this line should be removed here.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Ok, but that means I can never _really_ test the final files. But I guess we can manage with these simple changes.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: CUDA is not a hard dependency (see the discussion in the EasyBlock); so I'd rather be explicit here.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: I think it's better not to reply on stuff built with the system compiler if it can be avoided, so please change:

```python
('numactl', '2.0.13', '', ('GCCcore', '9.3.0')),
```

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: This should be changed to `default_cuda_version = None` to rely on what the NVHPC easyblock does by default (use the CUDA version listed as a dep unless told otherwise)?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,43 @@
...
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.34', '', ('GCCcore', local_gccver)),
+    ('CUDA', '11.0', '', SYSTEM),
+    # This is necessary to avoid cases where just libnuma.so.1 is present in the system and -lnuma fails
+    ('numactl', '2.0.13', '', SYSTEM)
+]
+
+default_cuda_version = "11.0"
+compute_capability = "70"
```
boegel, line 22: This should be removed entirely, and let the easyblock produce a proper error message if nothing is specified via `--cuda-compute-capabilities`?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,43 @@
...
+
+# Options to add/remove things to/from environment module
+module_byo_compilers = False  # Remove compilers from PATH (Bring-your-own compilers)
+module_nvhpc_own_mpi = False  # Add NVHPC's own pre-compiled OpenMPI
+module_add_math_libs = False  # Add NVHPC's math libraries (which should be there from CUDA anyway)
+module_add_profilers = False  # Add NVHPC's NVIDIA Profilers
+module_add_nccl = False       # Add NVHPC's NCCL library
+module_add_nvshmem = False    # Add NVHPC's NVSHMEM library
+module_add_cuda = False       # Add NVHPC's bundled CUDA
```
boegel, line 31: Same as defaults in NVHPC easyblock, so drop these lines?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: Agreed, should be removed.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: This is just a public download, so please add:

```python
source_urls = ['https://developer.download.nvidia.com/hpc-sdk/%(version)s/']
sources = ['nvhpc_2020_%(version_major)s%(version_minor)s_Linux_x86_64_cuda_multi.tar.gz']
checksums = ['a5c5c8726d2210f2310a852c6d6e03c9ef8c75e3643e9c94e24909f5e9c2ea7a']

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb [hunk 9]
```diff
@@ -0,0 +1,43 @@
...
+description = """C, C++ and Fortran compilers included with the NVIDIA HPC SDK (previously: PGI)"""
+
+toolchain = SYSTEM
+
+sources = ['nvhpc_2020_%(version_major)s%(version_minor)s_Linux_x86_64_cuda_multi.tar.gz']
+
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.34', '', ('GCCcore', local_gccver)),
+    ('CUDA', '11.0', '', SYSTEM),
```
boegel, line 16: There's no `CUDA-11.0.eb` easyconfig file in the central repository, so this will have to be tweaked too...

boegel (changes requested):

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Ah, that has changed. But one still needs to accept the Software License Agreement. Are you sure we can just do that?

AndiH: (I've already added the SHA-sum but not yet uploaded the new file. It's a bit tricky to handle this abstract file and the specific JSC file.)

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Yep, that's in the current file.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: I find it quite good to keep here a reference easyconfig which presents the available options in a concise and static manner.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: The CUDA dependency actually is optional as one _can_ use the NVHPC-shipped CUDA version. But we want to use our system CUDA module.

Maybe it's easiest to add a CUDA 11 module in the central repo?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Changed.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Changed.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: I've uploaded an alternative in a22f90649b. It moves this documentation of the available features to comments, but elaborates a bit on the behavior.
Would that be ok?

AndiH:
I think for now I've implemented all I could and would further feedback now from you, @boegel 

* What to do about the CUDA 11 dependency?
* What to do about the automatic download? Can we just accept the software license agreement for the users?
* Should we use the easyconfig to document available EasyBlock options? Should we do it explicitly by calling the variables, or as comments? Or not at all (which I'm not in favor for)?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: I think it's fine to add the source URL and auto-download it, since the download URL is public.

Please use a SHA256 checksum (use `eb --inject-checksums` for example, or the `sha256sum` command on Linux).

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: We already have easyconfigs for CUDA 11.0.x (see `CUDA-10.1.243.eb` at https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/c/CUDA), but this looks like some kind of a wrapper to allow in-place updating of CUDA?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,47 @@
...
+    # This is necessary to avoid cases where just libnuma.so.1 is present in the system and -lnuma fails
+    ('numactl', '2.0.13', '', ('GCCcore', local_gccver))
+]
+
+default_cuda_version = None  # Use strings like "11.0" or use the command line: --try-amend=default_cuda_version="10.2"
+
+# this bundle serves as a compiler-only toolchain, so it should be marked as compiler (important for HMNS)
+moduleclass = 'compiler'
```
boegel, line 25: Please move this line (and the comment) to the end of the easyconfig file, and put the block of comments explaining the different custom easyconfig parameter above it.

boegel (changes requested):

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: Adding the block of comments explaining the different custom easyconfig parameters is fine.

We don't do this often, since you can also check `eb -a -e EB_NVHPC`, but it's helpful so no problem to have it there.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Really? The download URL can only be seen after confirming to the License Agreement. We circumvent this check if we just download it.

I'll update the checksum to SHA256.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: @damianam, can you help?

damianam: I think we should simply change the PR here to use whatever CUDA 11 version is available in this repo. We have `CUDA-11.0.eb` for certain constrains and design decisions that do not apply here, so I would suggest to replace `('CUDA', '11.0', '', SYSTEM),` with `('CUDA', '11.0.2'),`

AndiH: Changed as suggested in 0a1a1a3e88.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
damianam: On one hand, nvidia's fault for not protecting the tarball better. On the other hand, we do know they want us to accept the license agreement. So maybe we should not have auto-download 🤷

boegel: I'm fine with not making this auto-download, but we should add a comment above `sources` with some simple instructions where to download, and mention "requires accepting license agreement".

AndiH: Added!
Should I include something to the likes of the following:
```
# If you accept the License Agreement, you may also uncomment the following line to automatically download the sources
# source_urls = ['https://developer.download.nvidia.com/hpc-sdk/%(version)s/']
sources = ['nvhpc_2020_%(version_major)s%(version_minor)s_Linux_x86_64_cuda_multi.tar.gz']
checksums = ['a5c5c8726d2210f2310a852c6d6e03c9ef8c75e3643e9c94e24909f5e9c2ea7a']
```

boegel (changes requested):

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: @AndiH That makes a lot of sense to me, yes.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,47 @@
...
+
+toolchain = SYSTEM
+
+sources = ['nvhpc_2020_%(version_major)s%(version_minor)s_Linux_x86_64_cuda_multi.tar.gz']
+checksums = ['a5c5c8726d2210f2310a852c6d6e03c9ef8c75e3643e9c94e24909f5e9c2ea7a']
+
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.34', '', ('GCCcore', local_gccver)),
+    ('CUDA', '11.0.2'),
```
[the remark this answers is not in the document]
AndiH: I've added that in 2f59e94265!

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
AndiH: Added in 485bf1d622!

boegel:
Test report by @boegel
Using easyblocks from https://github.com/easybuilders/easybuild-easyblocks/pull/2190
**SUCCESS**
node3118.skitty.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/c545458254849b5b36c0d355d03024df for a full test report.

boegel:
EB_ARGS="--cuda-compute-capabilities=7.0"

boegel (approved):

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,51 @@
...
+# It requires accepting the HPC SDK Software License Agreement (https://docs.nvidia.com/hpc-sdk/eula/index.html)
+# If you accept the License Agreement, you may also uncomment the following line to automatically download the sources
+# source_urls = ['https://developer.download.nvidia.com/hpc-sdk/%(version)s/']
+sources = ['nvhpc_2020_%(version_major)s%(version_minor)s_Linux_x86_64_cuda_multi.tar.gz']
+checksums = ['a5c5c8726d2210f2310a852c6d6e03c9ef8c75e3643e9c94e24909f5e9c2ea7a']
+
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.34', '', ('GCCcore', local_gccver)),
+    ('CUDA', '11.0.2', '', ('GCC', local_gccver)),
```
bartoldeman, line 21: I think it would be better to use CUDAcore here. The GCC dep is going to mess up in an HMNS I think.

`('CUDAcore', '11.0.2'),`

bartoldeman (changes requested):

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: Hmm, yeah, good point actually... I overlooked `CUDAcore`.

@AndiH @damianam Thoughts?

AndiH: I don't mind.

From the easyconfig I see that `CUDACore` also uses the `CUDA` EasyBlock, right? In that case, `$CUDA_HOME` should also be set, I assume. That's the only hard necessity here.

One caveat, though: CUDA is optional for th NVHPC package. NVHPC is shipped with its own CUDA, which is taken when no external CUDA module (setting `$CUDA_HOME`) is there. If one would want to rather use the NVHPC-shipped CUDA, they would need to move from `CUDAcore` to `GCCcore`. That might be a bit … indirect? What do you think?

boegel: With that in mind, I guess it makes more sense to just not include a specific `CUDA(core)` as a dep here? @bartoldeman?

AndiH: That would possibly be in line with what NVIDIA recommends doing.

For JSC, we (well… I…) decided that we want to use our dedicated CUDA module (which should be compatible as long as a proper version is picked and `$CUDA_HOME` is set), to have no hidden dependencies and use the other stack which is dependent on CUDA – and be more flexible in updating the CUDA version.

boegel: @AndiH Let's go forward with not including a specific `CUDA` dependency then, so we can get this PR merged (and included in the upcoming EasyBuild release).

We've made sure in the easyblock that things work properly when you do include a particular CUDA as dependency, but the default should probably to go with what's included with NVHPC itself.

AndiH: Ok.
I'd add this to the block of comments!

AndiH: How do you like f7cc517a9b?

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,59 @@
...
+
+# NVHPC EasyBlock supports some features, which can be set via CLI or this easyconfig.
+# The following list gives examples for the easyconfig
+#
+# NVHPC needs CUDA to work. Two options are available: 1) Use NVHPC-bundled CUDA, 2) use system CUDA
+# 1) Bundled CUDA
+#    If no easybuild dependency to CUDA is present, the bundled CUDA is taken. A version needs to be specified with
+#      default_cuda_version = "11.0"
+#    in this easyconfig file; alternatively, it can be specified through the command line during installation with
+#      --try-amend=default_cuda_version="10.2"
```
boegel, line 35: @AndiH Having to tweak the `default_cuda_version` value while we use the bundled CUDA is a bit weird...

There's only one CUDA version bundled, right? Shouldn't we set `default_cuda_version` accordingly, and make a note that this has to be tweaked when `CUDA` is included as a dep (setting to `None` would work in that case, since it'll be auto-derived)?

Also, what happens if there's a mismatch?

AndiH: @boegel Nope, if users choose the "nvhpc[…]_Linux_x86_64_cuda_multi.tar.gz" tar ball, it come with the current and two past CUDA versions. There is a tar ball which has only one CUDA bundled, but since that is more specific than this one, I decided to rather focus on the _mutli_ tar ball.

If there is a mismatch, i.e. when the _requested_ CUDA version is not available in the tar ball, it's up to NVHPC's `install.sh` script to error out. We only pass the requested CUDA version to the script via environment variable – everything else is handled by the script.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
bartoldeman: agreed, if NVHPC (like PGI before which also included some CUDA) can be flexibly be combined with any CUDA module and is usable for compiling CPU-only software without CUDA_HOME set, than it should not be a runtime dependency.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
bartoldeman: I still have a feeling that some of this is backwards: with the usual HMNS a user would do
`module load NVHPC CUDA`
and then the NVHPC module is loaded first, and the CUDA module *after*, and NVHPC extends MODULEPATH.

But that means that the NVHPC installation procedure has no knowledge which CUDA is eventually used with it! Sure it could default to the builtin but we need to make sure that it's runtime overridable.

bartoldeman: i.e we need an easyconfig as follows (just to illustrate here what I mean):
```
$ cat CUDA-11.0.2-NVHPC-20.7.eb
easyblock = 'Bundle'
name = 'CUDA'
version = '11.0.2'

homepage = 'https://developer.nvidia.com/cuda-toolkit'
description = """CUDA (formerly Compute Unified Device Architecture) is a parallel
 computing platform and programming model created by NVIDIA and implemented by the
 graphics processing units (GPUs) that they produce. CUDA gives developers access
 to the virtual instruction set and memory of the parallel computational elements in CUDA GPUs."""

toolchain = {'name': 'NVHPC', 'version': '20.7'}

dependencies = [('CUDAcore', '11.0.2', '', True)]

altroot = 'CUDAcore'

moduleclass = 'system'
```

AndiH: The compile-time specified CUDA version goes into a `cudarc` file in the NVHPC directory structure. If an external CUDA is loaded, this is – as far as I know – not taken, but rather deduced by looking at `$CUDA_HOME`. So specifying a different compile-time and run-time CUDA _should_ not be that bad.

I don't have anything against providing a CUDA easyconfig inside the NVHPC toolchain, that would be a good addition for the CUDA-dependency-free NVHPC easyconfig.

For JSC's case, in which we want to more strongly link CUDA and NVHPC, I'd still prefer having a CUDA dependency inside the easyconfig file of NVHPC.

bartoldeman (commented):
I did some experiments with NVHPC 20.7 and looked around a bit: its included CUDA 11.0 is indentical to what the CUDA easyconfig gives us too so there is some overlap here.. on the other hand it seems to ignore CUDA_HOME. I'll dig a bit further.

It's all a little bit of a contrast between NVHPC shipping the world and the way EasyBuild divides full toolchains into components. This is all similar to the Intel cluster toolkit which also includes everything including Intel MPI and MKL, but EB installs those separately. You can argue for both approaches -- it's just more consistent with the rest of EB to reduce NVHPC strictly to the compilers, if at all possible.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
bartoldeman: I did some experiments with NVHPC 20.7 and looked around a bit: its included CUDA 11.0 is indentical to what the CUDA easyconfig gives us too so there is some overlap here.. on the other hand it seems to ignore CUDA_HOME. I'll dig a bit further.

It's all a little bit of a contrast between NVHPC shipping the world and the way EasyBuild divides full toolchains into components. This is all similar to the Intel cluster toolkit which also includes everything including Intel MPI and MKL, but EB installs those separately. You can argue for both approaches -- it's just more consistent with the rest of EB to reduce NVHPC strictly to the compilers, if at all possible.

AndiH: Exactly, NVIDIA wants to position this as the all-in-one solution for HPC, which is fine, but doesn't really align with the way we currently handle modules in a fine-grained manner.

bartoldeman: If you use `nvc++ -stdpar` it ignores `$CUDA_HOME` and insists using the included CUDA. This can all be followed in the various rc files (e.g. `$EBROOTNVHPC/20.7/Linux_x86_64/20.7/compilers/bin/rcfiles/paralgorc`) but these files have a scary notice at the top which doesn't invite fixing that issue. I'll get back to you again tomorrow to get a bit more of a clear view.

bartoldeman:
Test report by @bartoldeman
**FAILED**
build-node.computecanada.ca - Linux centos linux 7.8.2003, x86_64, Intel Xeon Processor (Skylake, IBRS), Python 3.7.7
See https://gist.github.com/0d072c6652435c92f9e3904474865a3e for a full test report.

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
bartoldeman: In the end I believe that because currently the easyconfig does not install out-of-the-box without `--try-amend=default_cuda_version=<version>` it should use the version that the installer picks by default (11.0).

One can make that even more explicit using a versionsuffix, i.e.
versionsuffix = '-CUDA-11.0.2'
and then rename to:
NVHPC-20.7-GCC-9.3.0-CUDA-11.0.2.eb

Then using:
```dependencies = [('CUDAcore', '11.0.2')]```
locks the dependency into place and makes the behaviour of NVHPC consistent with or without a `CUDA` module loaded.. this consistency is needed since e.g. `nvc++ -stdpar` ignores CUDA_HOME (and then the only purpose of the `CUDA` module, unlike `CUDAcore`, is to just extend MODULEPATH).

Also note that if NVHPC is installed with the bundled CUDA, then CUDA_HOME is NOT set which in turn makes other software believe that CUDA is not installed, so I would keep the CUDAcore installation in any case; setting CUDA_HOME in the NVHPC module to the bundled CUDA would be an alternative but quite unusual.

but please delete the `dependencies = [('CUDA', '11.0.2', '', ('GCC', local_gccver))]` example since that won't work in an HMNS.

So here's my proposal:
1. add `versionsuffix = '-CUDA-11.0.2'`
2. rename to `NVHPC-20.7-GCC-9.3.0-CUDA-11.0.2.eb`
3. use `dependencies = [('CUDAcore', '11.0.2')]`
4. remove comment about `dependencies = [('CUDA', '11.0.2', '', ('GCC', local_gccver))]`

Comment (bartoldeman) [reply easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb]: I'll let @boegel vote on whether or not to use versionsuffix -- it's conceivable but not super likely to also have
`NVHPC-20.7-GCC-9.3.0-CUDA-10.2.89.eb`

Comment (AndiH) [reply easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb]: I'm fine with all of that and would implement it as soon as @boegel also gives his ok!

Should I create a second easyconfig, `NVHPC-20.7-GCC-9.3.0.eb` which does not contain the `CUDAcore` dependency and is fully _tuned_ to use the NVHPC-built-in CUDA?

File: easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb (hunk 13 above)
Comment (AndiH) [reply easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb:35]: I think there is a good reason for that: the NVHPC-`stdpar` is (/may be) slightly newer than then CUDA `stdpar`. They probably want to make sure that there's no interference.

Comment (boegel) [reply easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb:35]: I would like to settle this today, so we can include this easyconfig in the EasyBuild v4.3.1 release which we're pushing out tomorrow.

@AndiH @bartoldeman Any chance you would have time for a quick call to settle this around 4pm CET today (I still need to read up on the discussion)? Or maybe @AndiH can join the EasyBuild Slack (see https://easybuild.readthedocs.io/en/latest/#getting-help) for a more interactive discussion.

Comment (AndiH) [reply easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb:35]: I've messaged you on Slack.

File: easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb [hunk 14]
```diff
@@ -0,0 +1,59 @@
...
+# 2) System CUDA
+#    Use CUDA or CUDAcore as a dependency, for example
+#      dependencies = [('CUDA', '11.0.2', '', ('GCC', local_gccver))]
```
boegel, line 38: @AndiH As discussed with @bartoldeman, this line should be removed (indirectly depending on `GCC` is not a good idea)

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,59 @@
+name = 'NVHPC'
+version = '20.7'
+local_gccver = '9.3.0'
+versionsuffix = '-GCC-%s' % local_gccver
```
boegel, line 4: @AndiH As discussed, we shouldn't use a `versionsuffix` at all (which implies renaming the file accordingly)

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
```diff
@@ -0,0 +1,59 @@
...
+checksums = ['a5c5c8726d2210f2310a852c6d6e03c9ef8c75e3643e9c94e24909f5e9c2ea7a']
+
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.34', '', ('GCCcore', local_gccver)),
+    # This is necessary to avoid cases where just libnuma.so.1 is present in the system and -lnuma fails
+    ('numactl', '2.0.13', '', ('GCCcore', local_gccver))
+]
+
+default_cuda_version = None  # Use strings like "11.0" or use the command line: --try-amend=default_cuda_version="10.2"
```
boegel, line 25: this should specify the most recent CUDA version included in the multi tarball

boegel (changes requested):

easybuild/easyconfigs/n/NVHPC/NVHPC-20.7-GCC-9.3.0.eb
boegel: @AndiH I've made the changes we discussed in https://github.com/AndiH/easybuild-easyconfigs/pull/1 .

If you can merge this soon (at the very latest tomorrow morning CET), then we can still squeeze this in for EasyBuild v4.3.1 .
I've changed the target branch for this PR to `4.3.x` to ensure it gets included in the release (cc @migueldiascosta).
Shortly after the release we'll collapse the `4.3.x` branch into `develop` too to sync things up.

AndiH: @boegel's changes are merged!

I'm happy this is shipping with the next EasyBuild release!

boegel: :+1:

boegel (approved):

bartoldeman (approved):
LGTM

boegel:
Test report by @boegel
**SUCCESS**
node3101.skitty.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/6e170d9316a931f92e197d3ffe0e745d for a full test report.

bartoldeman:
Test report by @bartoldeman
**FAILED**
build-node.computecanada.ca - Linux centos linux 7.8.2003, x86_64, Intel Xeon Processor (Skylake, IBRS), Python 3.7.7
See https://gist.github.com/16b6ceac4b35282ca210b44ad36f7678 for a full test report.

boegel:
@bartoldeman `--cuda-compute-capabilities` is a strict requirement

bartoldeman:
we still need cuda_compute_capabilities set.. or it will not install by default :(

bartoldeman:
ah ok, will use that

bartoldeman:
Test report by @bartoldeman
**SUCCESS**
build-node.computecanada.ca - Linux centos linux 7.8.2003, x86_64, Intel Xeon Processor (Skylake, IBRS), Python 3.7.7
See https://gist.github.com/5d5d8cc8d60a9de9ad3e35cd3c906616 for a full test report.

boegel:
Going in, thanks @AndiH!

bartoldeman (approved):
LGTM

PR #11567, opened 2020-10-23
{chem}[GCCcore/9.3.0] XCFun 2.1.0

stigrj:
There exists an old version under the name `XCfun`, but the name really should be [XCFun](https://xcfun.readthedocs.io/en/latest/index.html), so where do you want me to put this, @boegel?

### DIFF
--- a/easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
+++ b/easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
@@ -13,7 +13,7 @@ sources = ['v%(version)s.tar.gz']
 checksums = ['64aac8c933cc129ce6326f3827e342abfd10b94ea4a302aaca9f10d342ad7656']
 
 builddependencies = [
-    ('CMake', '3.16.4', '', ('GCCcore', '9.3.0'))
+    ('CMake', '3.16.4')
 ]
 
 separate_build_dir = True
--- a/easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
+++ b/easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
@@ -1,0 +1,29 @@
+easyblock = 'CMakeMake'
+
+name = 'XCFun'
+version = '2.1.0'
+
+homepage = 'xcfun.readthedocs.io'
+description = """Arbitrary order exchange-correlation functional library"""
+
+toolchain = {'name': 'foss', 'version': '2020a'}
+
+source_urls = ['https://github.com/dftlibs/xcfun/archive/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['64aac8c933cc129ce6326f3827e342abfd10b94ea4a302aaca9f10d342ad7656']
+
+builddependencies = [
+    ('CMake', '3.16.4', '', ('GCCcore', '9.3.0'))
+]
+
+separate_build_dir = True
+build_type = 'release'
+
+modextravars = {'XCFun_DIR': '%(installdir)s/share/cmake/XCFun/'}
+
+sanity_check_paths = {
+    'files': ['lib/libxcfun.%s' % SHLIB_EXT],
+    'dirs': ['include/XCFun']
+}
+
+moduleclass = 'chem'
--- a/easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-GCCcore-9.3.0.eb
+++ b/easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-GCCcore-9.3.0.eb
@@ -6,7 +6,7 @@ version = '2.1.0'
 homepage = 'xcfun.readthedocs.io'
 description = """Arbitrary order exchange-correlation functional library"""
 
-toolchain = {'name': 'foss', 'version': '2020a'}
+toolchain = {'name': 'GCCcore', 'version': '9.3.0'}
 
 source_urls = ['https://github.com/dftlibs/xcfun/archive/']
 sources = ['v%(version)s.tar.gz']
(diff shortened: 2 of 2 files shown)


easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
```diff
@@ -0,0 +1,29 @@
...
+description = """Arbitrary order exchange-correlation functional library"""
+
+toolchain = {'name': 'foss', 'version': '2020a'}
+
+source_urls = ['https://github.com/dftlibs/xcfun/archive/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['64aac8c933cc129ce6326f3827e342abfd10b94ea4a302aaca9f10d342ad7656']
+
+builddependencies = [
+    ('CMake', '3.16.4', '', ('GCCcore', '9.3.0'))
```
Micket, line 16:
```suggestion
    ('CMake', '3.16.4')
```

easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
```diff
@@ -0,0 +1,29 @@
...
+]
+
```
Micket, line 18: No more python dep?

Micket (changes requested):
So, the name question; Since it's not a widely used dependency, i suppose we can consider renaming it. If we do, we need to update the dependency name in PySCF as well.

easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
stigrj: Well, XCFun is a C++ library with an optional Python interface, and this particular build would be without Python support. I tried first to make the general `XCFun-2.1.0-foss-2020a-Python-3.8.2.eb` version, but it pulled in some more dependencies which I don't really need (but PySCF does). Do you think it's worth having a separate Python-free version like this, or should we only provide the general version _with_ Python support?

lexming:
Test report by @lexming
**SUCCESS**
node101.hydra.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) CPU E5-2680 v2 @ 2.80GHz, Python 2.7.5
See https://gist.github.com/c65e6bd6c209af646dccb525e38b6621 for a full test report.

lexming:
Test report by @lexming
**SUCCESS**
node369.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/d3fed647f0fe654949ea854220bfa1f0 for a full test report.

easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
Micket: Typically I would say, do opt in, unless there is some reason not to. 

If we don't want to opt it, then it might be possible to move this up the hierarchy.

stigrj: Not sure what you mean with moving up the hierarchy?

I discussed briefly with XCFun maintainer @robertodr. The Python interface is not widely used, so we think a separate recipe without Python support is worth while.

lexming: I agree that we can merge this easyconfig without Python as the name of the easyconfig will be different in any case. If anybody needs support for Python it can be added later on.

@stigrj  @Micket was talking about changing the toolchain of the easyconfig. Since XCFun does not use BLAS/LAPACK, FFTW or MPI, then it does not need the `foss` toolchain. It can be moved to `GCC-10.2.0`, making it more compatible with other easyconfigs.

easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
```diff
@@ -0,0 +1,29 @@
+easyblock = 'CMakeMake'
+
+name = 'XCFun'
+version = '2.1.0'
+
+homepage = 'xcfun.readthedocs.io'
```
lexming, line 6:
```suggestion
https://xcfun.readthedocs.io
```

lexming (changes requested):
As discussed, I think that this can be merged once it is moved to `GCC` and the homepage URL is fixed (see below)

easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
stigrj: Thanks for the input @lexming! I switched to `GCCcore-9.3.0`, I don't think 10 is available yet on my clusters :slightly_smiling_face:

easybuild/easyconfigs/x/XCFun/XCFun-2.1.0-foss-2020a.eb
stigrj: done

boegel:
Can we make sure an accompanying (separate) PR is opened to rename the existing `XCfun` to `XCFun` (which is a bugfix) when this is merged?

stigrj:
See #11603

lexming:
Test report by @lexming
**SUCCESS**
node379.hydra.os - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/923fd50182968f26d32ba26c6ce132e1 for a full test report.

lexming:
Test report by @lexming
**SUCCESS**
node101.hydra.os - Linux centos linux 7.8.2003, x86_64, Intel(R) Xeon(R) CPU E5-2680 v2 @ 2.80GHz, Python 2.7.5
See https://gist.github.com/5ea22375b4ce0e0d303917a50a1caefe for a full test report.

lexming (approved):
LGTM

lexming:
Going in, thanks @stigrj !

PR #11576, opened 2020-10-26
{chem}[intel/2019b] LAMMPS patch_20Nov2019

hajgato:


hajgato:
@boegel I really do not see here what is the style error

lexming (commented):
A couple of questions:
* We have an easyblock for LAMMPS, could this easyconfig use it?
* I see that this is a patch release of LAMMPS, does it provide a full LAMMPS installation or just the patched files?

hajgato:
@lexming We did not use the easyblock here, because this is a "bare" `LAMMPS` installation. It is a full LAMMPS installation, but no whistles and bells. The version is `patch_20Nov2019`, so this is a release between the "stable" releases. (You can call it beta, RC, unstable, etc.). It was a special request from our users, but in general it is better to use a stable version. See more about LAMMPS versions: https://lammps.sandia.gov/bug.html

akesandgren:

PR #11731, opened 2020-11-17
Create pugixml-1.10-GCCcore-8.3.0.eb

levlafayette:
Useful dependency for other software (e.g., ASAP).


Micket (changes requested):
missing subdirectory

Micket (changes requested):
Missing a subdirectory for the software. Should be `p/pugixml/pugixml-1.10-GCCcore-8.3.0.eb`

akesandgren:

We also have newer versions available

PR #11751, opened 2020-11-23
{chem}[foss/2020a] CP2K {5,7}.1

hajgato:


easybuild/easyconfigs/c/CP2K/CP2K-5.1-foss-2020a.eb
```diff
@@ -0,0 +1,45 @@
...
+]
+
+# values to use here are +1 those used for building Libint
+# see https://github.com/cp2k/cp2k/blob/master/cp2k/tools/hfx_tools/libint_tools/README_LIBINT
+extradflags = "-D__LIBINT_MAX_AM=6 -D__LIBDERIV_MAX_AM1=5"
+
+# regression test reports failures
+ignore_regtest_fails = True
+
+modextravars = {'OMP_NUM_THREADS': '1'}
```
smoors, line 43: I agree it's a good idea to set this, but other sites may not want it here, so better remove it.

easybuild/easyconfigs/c/CP2K/CP2K-7.1-foss-2020a.eb
```diff
@@ -0,0 +1,35 @@
...
+builddependencies = [
+    ('flex', '2.6.4'),
+    ('Bison', '3.5.3'),
+]
+
+# regression test reports handful of failures,
+# we're assuming those are OK to ignore...
+ignore_regtest_fails = True
+
+modextravars = {'OMP_NUM_THREADS': '1'}
```
smoors, line 33: same here

smoors (changes requested):
@hajgato any specific reason for the old CP2K version?

easybuild/easyconfigs/c/CP2K/CP2K-5.1-foss-2020a.eb
hajgato: I cannot really think a use case when `OMP_NUM_THREADS` is not set by "hand" or it has a meaningful value other than 1

akesandgren: Setting OMP_NUM_THREADS in a userlevel code like this is bad.
Especially when doing it without checking if it is already set.
I strongly oppose having this set in CP2K!!

hajgato: This is the POPT version, so it is MPI only. In case of PSMP which is OMP+MPI hybrid yes, it makes sense not to set OMP_NUM_THREADS. Anyway, I give it up and I will not set, but do not be surprised when users will ask why the performance of this version is terribly bad. (because it will be terribly bad, if user runs it N cpus, N^2 threads will be started)

akesandgren: Hmm, I thought cp2k had been fixed to build all versions at once as it should do.
In that case it's less bad.

smoors:

smoors:

smoors:
@boegelbot: please test @ generoso EB_ARGS="CP2K-5.1-foss-2020a.eb"

smoors (approved):
lgtm

smoors:
Going in, thanks @hajgato!

PR #11812, opened 2020-12-03
adding easyconfigs: HDF5-1.10.1-foss-2020a.eb

deniskristak:
upgrade of HDF5


deniskristak:
closing, need another  file, cannot add it currently due to github token issues...

PR #11813, opened 2020-12-03
adding easyconfigs: HDF5-1.10.1-foss-2020a.eb and patches: HDF5-1.10.…

deniskristak:
…1-mpi_deprecated_methods.patch

upgrade for HDF5

### DIFF
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.10.1-foss-2020a.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.10.1-foss-2020a.eb
@@ -1,0 +1,29 @@
+name = 'HDF5'
+version = '1.10.1'
+
+homepage = 'https://portal.hdfgroup.org/display/support'
+description = """HDF5 is a data model, library, and file format for storing and managing data.
+ It supports an unlimited variety of datatypes, and is designed for flexible
+ and efficient I/O and for high volume and complex data."""
+
+toolchain = {'name': 'foss', 'version': '2020a'}
+toolchainopts = {'pic': True, 'usempi': True}
+
+source_urls = [
+    'https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-%(version_major_minor)s/hdf5-%(version)s/src']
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/h/HDF5/HDF5-1.10.1-foss-2020a.eb
```diff
@@ -0,0 +1,29 @@
...
+    '3d564dca389eea5ce09e88476fa4c048ca0c79690859a133b09a46db58a60c88',
+]
+
+dependencies = [
+    ('OpenSSL', '1.1.1e'),
+    ('zlib', '1.2.11'),
+    ('Szip', '2.1.1'),
+    ('libxml2', '2.9.10'),
+]
+
```
akesandgren, line 28: We should probably turn on the testing, i.e.
runtest = "test"
on all new HDF5 easyconfigs.
Depends on having https://github.com/easybuilders/easybuild-easyblocks/pull/2250 though.

@boegel thoughts on doing this?

easybuild/easyconfigs/h/HDF5/HDF5-1.10.1-foss-2020a.eb
```diff
@@ -0,0 +1,29 @@
+name = 'HDF5'
+version = '1.10.1'
```
Micket, line 2: Why this old bugfix version?
I know there are ABI differences between 1.12 and 1.10 series, but there is already a 1.10.6 version in 2020a.

Also, if this is truly needed, it should be moved down to gompi level.

Micket (changes requested):

deniskristak:
closing as it is not relevant anymore..found an alternative

easybuild/easyconfigs/h/HDF5/HDF5-1.10.1-foss-2020a.eb
boegel: @akesandgren Please open a dedicated issue for that, and we can do it in the most recent HDF5 easyconfigs?

PR #11907, opened 2020-12-17
{lib}[fosscuda/2020a,fosscuda/2020b] libgpuarray v0.7.6 w/ Python 3.8.2 + 3.8.6

Micket:
(created using `eb --new-pr`)

depends on https://github.com/easybuilders/easybuild-framework/pull/3541

PR #11898, opened 2020-12-16
{numlib}[system/system] cuDNN v8.0.5.39

lexming:
(created using `eb --new-pr`)

Nvidia has stopped shipping the version for PPC and instead it now ships a version for ARM. The title of the download section still says "_x86_64 & PPC architecture_" though. So I'm not sure if this is really intended, but right now this is what's available.

PR #11908, opened 2020-12-17
Remove cuDNN-8.0.4.30-gcccuda-2020a

Micket:
(created using `eb --new-pr`)


Micket:
As mentioned in #11898 I don't believe we want this. With the introduction of CUDAcore, we use the same CUDA everywhere, and we are able to avoid other duplicate NCCL and cuDNN stuff.  See e.g. #11907

lexming:
Basically copypasting my comment in the other PRs, but I'll expand it here for the sake of discussion.
So the plan is using these dependencies from `SYSTEM` in every easyconfig, regardless of their toolchain. In such a case, the role of the `fosscuda` or `gcccuda` toolchain is quite confusing
*  why have these cuda toolchains if the CUDA version has to be defined for each dependency needing CUDA? This will lead to conflicts if not using the same version of CUDA defined in the toolchain, but that is not obvious. 
* we could just use non-cuda toolchains and add the CUDA stack from `SYSTEM` to easyconfigs in `foss` or `GCC`?

My point is, we will get neither the benefits of using the CUDA definition from the toolchain (so, automatic version control), nor the benefits of using a common toolchain for cuda and non-cuda stuff. So I'm confused.

Micket:
> So the plan is using these dependencies from `SYSTEM` in every easyconfig, regardless of their toolchain. In such a case, the role of the `fosscuda` or `gcccuda` toolchain is quite confusing
>     * we could just use non-cuda toolchains and add the CUDA stack from `SYSTEM` to easyconfigs in `foss` or `GCC`?

The fact that OpenMPI (optionally) is compiled with CUDA support is the blocker for this, since we don't want to wait for it to support the latest GCC before be can create `gompi/202Xx`. Otherwise, yeah, I think a CUDA-versionssuffix approach under `foss` would work, and it definitely would have benefits (compatibility with all the non-cuda stuff). 

> My point is, with the current plan we will get neither the benefits of using the CUDA definition from the toolchain (automatic version control), nor the benefits of using a common toolchain for cuda and non-cuda stuff. So I'm confused.

I didn't imagine this to be the supposed benefit (though it is a step in the same direction), but rather not having to have multiple identical binary installations of (specifically) cuDNN/NCCL at system + gcccuda + iccifortcuda
Having to specify the versionssuffix was the price to pay. 

I actually wanted to use `-CUDA-%(cudaver)s`  in the versionsuffix which would certainly remedy some of that cost, but apparently that variable is only defined if you have direct dependencies. @boegel Perhaps we can extend that so it's available in for everything that at least uses the fosscuda/gompic/gcccuda toolchains?
https://github.com/easybuilders/easybuild-framework/blob/0fb8d5e0e410521191f4f0004c1b5083e0356d4e/easybuild/framework/easyconfig/templates.py#L245
Just adding in the toolchains (recursive deps) or something here maybe?

lexming:
I see the advantage of just having a single package for these binaries. I guess this has never been a concern for us because we only installed the versions in `gcccuda` already. But then I think that we definitely need something like `-CUDA-%(cudaver)s` to work. Otherwise it defeats the purpose of having a toolchain with CUDA.

Micket:
Well, I.M.H.O. I don't think the word "purpose" fits, but rather an "unfortunate necessity" that we have to have a separate gompic and gompi. (*I think it would be perfectly fine to have a `TensorFlow-2.3.1-foss-2020b-CUDA-11.1.1.eb` if some user didn't care about OpenMPI with GPUDirect-support and needed only to use TensorFlow together a ton of other non-CUDA stuff we already have under `foss`*).

But I do believe this is an different question though; the case of cuDNN and NCCL was just the fact that they are binary packages that could just as well go under system level, just like we do with Java. They just happen to also have direct dependency on a specific CUDA version (which means we could only start doing this now when we introduced the CUDAcore+CUDA split).
fosscuda would have to remain even if cuDNN and NCCL didn't even exist.

I'll see if I can get a patch to include `%(cudaver)s` for the CUDA-toolchains.

lexming (approved):
As discussed in https://github.com/easybuilders/easybuild/wiki/Conference-call-notes-20201223, we will use dependencies on `SYSTEM` for these CUDA binary packages in `2020a`.

LGTM

PR #11902, opened 2020-12-17
{tools} Added git as a dependency to GitPython version 3.x

jhein32:
fixes #11901

PR #11916, opened 2020-12-21
add CESM-deps to whitelist in check for custom sanity_check_paths

lexming:
Needed by https://github.com/easybuilders/easybuild-easyconfigs/pull/11902


boegel (approved):
lgtm

PR #11868, opened 2020-12-12
{compiler}[GCCcore/9.3.0] AOCC v2.3.0

SebastianAchilles:
(created using `eb --new-pr`)

- [ ] Requires easybuilders/easybuild-easyblocks#2295

PR #12018, opened 2021-01-18
{compiler}[AOMP/11.12-0] Add AOMP compiler module

nordmoen:
This change adds an [AOMP](https://github.com/ROCm-Developer-Tools/aomp) compiler module to EasyBuild.

The setup has been tested and is working on both AMD and Nvidia GPUs.

The patch is necessary to remove some hard coded behavior on RHEL8 and to get the build process to respect EasyBuild `--parallel` option. The reason for the older `GCC` version is to conform with what AOMP expects when building, it could probably work with a newer version, but I have not tried. The same goes for `gcccuda`, AOMP recommends `CUDA` version `10`.


migueldiascosta:
@nordmoen thank you for your contribution!

quick note while starting to test, this requires `libelf` if `elfutils-libelf-devel` or equivalent is not installed - you can point to the one in https://github.com/easybuilders/easybuild-easyconfigs/pull/12048 once that's merged or add the easyconfig to this PR

nordmoen:
I'll add the dependency once it is merged =]

migueldiascosta:

migueldiascosta:
**FAILED**
sms - Linux centos linux 7.6.1810, x86_64, AMD EPYC 7601 32-Core Processor (zen), Python 2.7.5
See https://gist.github.com/29ae97e128926e9b446fe1983d3687fa for a full test report.

migueldiascosta:
@nordmoen w.r.t. to my failed test above, it seems that at least `libomptarget` is including the system header files instead of the ones in `GCCcore`, and I suppose this isn't a problem for you or in `generoso` because those have more recent system headers (e.g., centos 8 vs centos 7) 

```
In file included from /dev/shm/build/AOMP/11.12-0/GCCcore-9.3.0/aomp11/amd-llvm-project/openmp/libomptarget/src/device.cpp:13:
/dev/shm/build/AOMP/11.12-0/GCCcore-9.3.0/aomp11/amd-llvm-project/openmp/libomptarget/src/device.h:95:43: error: too few template arguments for class template 'less'
typedef std::set<HostDataToTargetTy, std::less<>> HostDataToTargetListTy;
                                          ^
/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/bits/stl_function.h:231:12: note: template is declared here
    struct less : public binary_function<_Tp, _Tp, bool>
```

nordmoen:
> @nordmoen w.r.t. to my failed test above, it seems that at least `libomptarget` is including the system header files instead of the ones in `GCCcore`, and I suppose this isn't a problem for you or in `generoso` because those have more recent system headers (e.g., centos 8 vs centos 7)
> 
> ```
> In file included from /dev/shm/build/AOMP/11.12-0/GCCcore-9.3.0/aomp11/amd-llvm-project/openmp/libomptarget/src/device.cpp:13:
> /dev/shm/build/AOMP/11.12-0/GCCcore-9.3.0/aomp11/amd-llvm-project/openmp/libomptarget/src/device.h:95:43: error: too few template arguments for class template 'less'
> typedef std::set<HostDataToTargetTy, std::less<>> HostDataToTargetListTy;
>                                           ^
> /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/bits/stl_function.h:231:12: note: template is declared here
>     struct less : public binary_function<_Tp, _Tp, bool>
> ```

I'm looking around in the build setup, but I can't really find anything where it should be hardcoded and not set correctly by loading `GCCcore`.

nordmoen:
@migueldiascosta I discovered that additional build tools were needed, with one of them being `pkg-config` which could account for the wrong link (?). Could you test the new changes with your setup?

migueldiascosta:
@nordmoen `libomptarget` is being built with the supplied `clang`, so the problem is likely this: https://github.com/easybuilders/easybuild-easyconfigs/pull/11868#issuecomment-744066721...

Since we are not rebuilding clang (et al), I suppose we need a way of always passing ` --gcc-toolchain=$EBROOTGCCCORE` to clang (et al) (@boegel, @SebastianAchilles ?)

SebastianAchilles:
Yes, as far as I understood you have to use `--gcc-toolchain=$EBROOTGCCCORE` in order to use the correct `gcccore` from the easybuild installation. Otherwise the `gcc` from the system will be used (if found).

nordmoen:
Looking through the build sources of `AOMP` it seems that several places need to change to have this take effect. We could of course do this in a patch, but maybe there is a better way?

migueldiascosta:
maybe a compiler wrapper... @boegel ?

boegel:
@nordmoen My apologies for never replying to your hail for help/input, this got totally snowed under...

We can definitely put lightweight wrappers in place to make sure the right GCC installation is picked up, if there's no better alternative, but then we'll need to implement a small easyblock for this I guess.
It's a bit frustrating that Clang doesn't just pick up the first `gcc` it finds in `$PATH`...

That said, is there any particular reason why we're not building AOMP from source, and configure the build with `-DGCC_INSTALL_PREFIX` (cfr. @SebastianAchilles' comment in https://github.com/easybuilders/easybuild-easyconfigs/pull/11868#issuecomment-744066721).

Maybe it's better to discuss this interactively in EasyBuild Slack?

easybuild/easyconfigs/a/AOMP/AOMP-11.12-0-gcccuda-2019b.eb
```diff
@@ -0,0 +1,63 @@
...
+preinstallopts += 'AOMP_REPOS="%(builddir)s/aomp%(version_major)s" '
+preinstallopts += 'AOMP_CHECK_GIT_BRANCH=0 '
+preinstallopts += 'AOMP_APPLY_ROCM_PATCHES=0 '
+preinstallopts += 'AOMP_STANDALONE_BUILD=1 '
+preinstallopts += 'NUM_THREADS=%(parallel)s'
+# 'NVPTXGPUS' should be set to the NVPTX architecture you would like to
+# support, e.g. '50,75'.
+preinstallopts += ' AOMP_BUILD_CUDA=1 '
+preinstallopts += 'CUDA="$CUDA_ROOT" '
+preinstallopts += 'NVPTXGPUS="70"'
```
Flamefire, line 45: I think this should be configurable. Does it support 7.0 or sm_70 format? Then you can use the templates cuda_compute_capabilities etc.

nordmoen: Based on the [pre-configured](https://github.com/ROCm-Developer-Tools/aomp/blob/master/bin/aomp_common_vars#L110) targets, I assumed one had to use whole numbers only.

Flamefire: Hm, then I'd say an EasyBlock is better here. I'm afraid short of patching the EC (e.g. via hooks) this would limit the GPU usability unusually, especially after we introduced the `--cuda-compute-capabilities` option
This has the additional benefit that those hard-coded, repeated preinstallopts can be unified and handled better (e.g. detecting presence of CUDA)

nordmoen:
It seems that `AOMP` is deprecated in favor of built in support in `ROCm` (as per the [official documentation](https://rocmdocs.amd.com/en/latest/Programming_Guides/openmp_support.html)).

Because of this I don't see the value in adding this package, especially when it seems like there is quite a lot work to get it set up properly.

nordmoen:
Just had a meeting with AMD where they confirmed that `AOMP` is the official way to use OpenMP directives with AMD GPUs.

nordmoen:
Closing in favor of #12909

PR #12062, opened 2021-01-29
Speedup test_dep_versions_per_toolchain_generation

Flamefire:
The list comprehension in get_deps_for was called about 357k times and iterates over all ~20k(?) ECs to find one by name.
This makes the test VERY long.

The new approach uses a dict and dynamic programming to memorize found dependency mappings making the test VERY fast.

For comparison check the runtime in any other PR: Runtime went from 16-28min down to 6-9min.


smoors (approved):
lgtm

PR #12134, opened 2021-02-10
{bio}[GCC/10.2.0] PRIMUS v1.9.0

guimaluf:


easybuild/easyconfigs/p/PRIMUS/PRIMUS-1.9.0.eb
```diff
@@ -0,0 +1,25 @@
...
+"""""
+
+toolchain = SYSTEM
+
+source_urls = ['https://primus.gs.washington.edu/docroot/versions/']
+sources = ['PRIMUS_v%(version)s.tgz']
+checksums = ['89f8c76c43534ed9bdcaebc8c32bce4049120cacc661f2e215d223344b9f3b58']
+
+sanity_check_paths = {
+    'files': ["bin/run_PRIMUS.pl"],
```
Micket, line 23: So, I think it's best to move this into GCCcore and depend on a perl version here.

easybuild/easyconfigs/p/PRIMUS/PRIMUS-1.9.0.eb
```diff
@@ -0,0 +1,25 @@
...
+
+easyblock = 'Tarball'
+
+name = 'PRIMUS'
+version = '1.9.0'
+
+homepage = 'https://primus.gs.washington.edu/primusweb/res/form.html'
+description = """PRIMUS: Pedigree Reconstruction and Identification of a
+    Maximum Unrelated Set
+"""""
```
Micket, line 14: You can one-line this (we allow for 120 columns in easyconfigs) (and i think the extra tab/space would be part of the description, which might look ugly in the actual module file).

Micket (changes requested):

easybuild/easyconfigs/p/PRIMUS/PRIMUS-1.9.0.eb
guimaluf: thanks @Micket for all the feedback.
this is my first time contributing to easyconfigs. 

by moving this into GCCcore do you mean setting this to the latest as well, right?

boegel: @guimaluf Not that using GCCcore as toolchain also implies renaming the easyconfig file (that's why the tests fail).

Filename should be `PRIMUS-1.9.0-GCCcore-10.2.0.eb`, and you should also include `Perl` as a dependency (to prevent relying on `perl` provided by the OS).

boegel:

PR #12216, opened 2021-02-20
fix checksum for patch file in PyTorch 1.7.1 easyconfig using foss/2020b

boegel:
(created using `eb --new-pr`)

This patch was tweaked in #12147, but that was not taken into account in #12008.

I'll look into verifying the checksums of patch files in the easyconfigs test suite that we run in CI, we should have caught this earlier...

PR #12221, opened 2021-02-20
verify checksum of all patch files in easyconfigs test suite

boegel:
To avoid issues like #12216 going unnoticed...

~~WIP, needs some more work to cover special cases.~~

requires ~~#12216~~ + ~~#12223~~


verdurin (approved):
Looks fine.

PR #12249, opened 2021-02-23
use 'main' rather than 'master' branch in test suite configuration

boegel:
We're making the necessary changes to rename the `master` branch to `main`.

Currently both are available (and they're identical), but we'll remove `master` as soon as we're done with making the necessary changes.

See also https://github.com/easybuilders/easybuild-framework/pull/3589


migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @boegel!

PR #12256, opened 2021-02-23
sync with main branch + bump version to 4.3.4dev

boegel:


test/easyconfigs/easyconfigs.py
```diff
@@ -794,7 +794,7 @@ def get_eb_files_from_diff(diff_filter):
                 merge_base = target_branch
 
             # determine list of changed files using 'git diff' and merge base determined above
-            cmd = "git diff --name-only --diff-filter=%s %s..HEAD" % (diff_filter, merge_base)
+            cmd = "git diff --name-only --diff-filter=%s %s..HEAD --" % (diff_filter, merge_base)
```
akesandgren, line 797: What's with the added "--"??

boegel: That was an attempt to fix an issue that popped up during the release, but the actual fix was to replace a `'master'` entry I overlooked.

This `--` doesn't hurt though, it's actually more correct this way. It is used to separate options to `git diff` from the list of paths on which it should operate (which is empty here), see also https://phrase.com/blog/posts/git-5-tricks-for-git-you-may-not-use-yet-part-2/#Trick_2_Double_Dash_To_Scope_On_A_Folder

akesandgren: Ok, just wanted to know why it was added.

akesandgren (approved):
LGTM

akesandgren:
Going in, thanks @boegel!

PR #12257, opened 2021-02-24
Create imake-1.0.8-foss-2020b.eb

Ghepardo:


Micket (changes requested):
This belongs to be at the GCCcore level instead, so please move it up the hierarchy to GCCcore 10.2.0 (and add binutils as a build dep)
(when updating really old easyconfigs for newer toolchains, moving things up the hierarchy as well is often the best option).

Ghepardo:
> This belongs to be at the GCCcore level instead, so please move it up the hierarchy to GCCcore 10.2.0 (and add binutils as a build dep)
> (when updating really old easyconfigs for newer toolchains, moving things up the hierarchy as well is often the best option).

Hi @Micket , I'm not sure I understand you here.  Do you mean I should simply replace foss/2020b with GCCcore/10.2.0 in the toolchain specification?

Micket:
Yes (and rename the file `imake-1.0.8-GCCcore-10.2.0.eb`)

Ghepardo:
> 
> 
> Yes (and rename the file `imake-1.0.8-GCCcore-10.2.0.eb`)

New pull request created for imake-1.0.8-GCCcore-10.2.0.eb.

PR #12269, opened 2021-02-24
github_account in public repo?

surak:
Should this be there at all? There is a 

`github_account = 'lanpa'` on the easyconfig


surak:
Why does it need it in order to find the file?

akesandgren:
Just for the record in case someone finds this, github_account is used to resolve the GITHUB_SOURCE template, i.e. it is inserted in https://github.com/%(github_account)s/%(name)s/archive

PR #12391, opened 2021-03-12
{lang}[foss/2020b] Cython v0.29.22 w/ Python 3.8.6

robert-mijakovic:
(created using `eb --new-pr`)


Micket:
We have actually moved Cython into the Python bundle nowdays, so we already have version 0.29.21 inside Python 3.8.6 already. Do you have something that requires this slightly newer version? Else, i would suggest holding off until the next python in 2021a toolchains.

robert-mijakovic:
> We have actually moved Cython into the Python bundle nowdays, so we already have version 0.29.21 inside Python 3.8.6 already. Do you have something that requires this slightly newer version? Else, i would suggest holding off until the next python in 2021a toolchains.

Not really. I wasn't aware of that when I created this easyconfig. I created a list of stuff that I wanted to have in my software stack and just saw that Cython is missing there and decided to add the newest version of that time.

Of course, it makes more sense to keep as you suggested for the moment. I will delete the PR.

PR #12399, opened 2021-03-15
{bio}[foss/2019b] RevBayes + deps

YensyGomez:


Micket (changes requested):
Does this require Boost 1.74? If not, please use 1.71.0 that already exists for 2019b, since it would make this module more compatible with other 2019b modules.
If it does require it, then there is already a newer bzip2 version for GCCcore 8.3.0 so a foss version should be added.

easybuild/easyconfigs/r/RevBayes/RevBayes-1.1.1-foss-2019b.eb
```diff
@@ -0,0 +1,46 @@
...
+toolchain = {'name': 'foss', 'version': '2019b'}
+
+source_urls = ['https://github.com/revbayes/revbayes/archive']
+sources = ['%(version)s.tar.gz']
+checksums = ['d61293fceac817d8203ed1e828661d76c73fa16bf04458a50a37057e99fd40c0']
+
+
+builddependencies = [
+    ('CMake', '3.15.3'),
+    ('Boost', '1.71.0'),
```
boegel, line 28: @YensyGomez It doesn't make sense to include `Boost` as a dependency twice, only via `dependencies` is fine (unless `Boost` is only a build dependency, then only via `builddependencies` of course)

boegel (changes requested):
@YensyGomez Would it be possible to include a simple test command like `rb --help` in `sanity_check_commands`?

Micket:
We have 2 RevBayes PRs right now, and they differ slightly.
#12411 includes Java, and has slightl configure flags. I don't know which is more correct

boegel:
Closing this PR since there's no reply at all after the review...

PR #12429, opened 2021-03-18
Sphinx v3.5.2 for GCCcore/10.2.0

SimonPinches:


Micket (commented):
Do we need an updated sphinx? Sphinx is bunled with Python 3.8.6 though a slightly older version than this

SimonPinches:
> Do we need an updated sphinx? Sphinx is bundled with Python 3.8.6 though a slightly older version than this

Good point!  I'm happy for this PR to be dropped/put on hold until a compelling case can be made.  (Which escapes my mind at the moment).

Micket:
I often just pose this question for some python packages we have nowdays included in Python already (since it's not obvious that they are no longer needed). I doubt that newest sphinx is really necessary, so I think we will prefer to just drop this PR for now. (but feel free to reopen if you there is a need).

PR #12430, opened 2021-03-18
{bio}[foss/2019b] BRAKER v2.1.5, GeneMark-ET v4.65, spaln v2.4.03

stevecdavis:
(created using `eb --new-pr`)


boegel:

easybuild/easyconfigs/b/BRAKER/BRAKER-2.1.5-foss-2019b-Python-3.7.4.eb
```diff
@@ -0,0 +1,41 @@
...
+toolchain = {'name': 'foss', 'version': '2019b'}
+
+source_urls = ['https://github.com/Gaius-Augustus/BRAKER/archive/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['528507c4fe3335865ead5421341f6e77959d2d327183b6c59d0858e6869d7ace']
+
+dependencies = [
+    ('Perl', '5.30.0'),
+    ('AUGUSTUS', '3.3.3'),
+    ('GeneMark-ET', '4.65'),
```
boegel, line 20: @stevecdavis Within the same generation of easyconfigs (`2019b`, in this case), we stick to a single dependency version to maximize the modules in that generation that are compatible with each other.

So, a couple of ways forward here are (more or less in order of preference):

* sticking to GeneMark-ET 4.57 as a dependency here
* bumping to a newer toolchain (for example `2020b`)
* adding an exception that allows diverging from our policy in this particular case (we typically only do that if there's a very good reason to use a different version)

boegel (changes requested):

easybuild/easyconfigs/b/BRAKER/BRAKER-2.1.5-foss-2019b-Python-3.7.4.eb
boegel: There's another option: extending the `versionsuffix` above to include the different `GeneMark-ET` version:

```python
local_pysuff = '-Python-3.7.4'
local_genemark_ver = '4.65'
versionsuffix = local_pysuff + '-GeneMark-ET-' + local_genemark_ver
```

and then here use:

```python
   ('GeneMark-ET', local_genemark_ver),
```

and below use:

```python
    ('Biopython', '1.75', local_pysuff),

boegel: Oh, and this implies renaming the `BRAKER` easyconfig file, which you can do via `--update-pr` by deleting the current BRAKER eaysconfig via `eb --update-pr :BRAKER-2.1.5-foss-2019b-Python-3.7.4.eb` (note the `:`, see https://docs.easybuild.io/en/latest/Integration_with_GitHub.html#deleting-easyconfig-files-or-patches), and adding `BRAKER-2.1.5-foss-2019b-Python-3.7.4-GeneMark-ET-4.65.eb`.

stevecdavis: made these changes and resubmitted the file

jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 12430` and update the PR to use `SYSTEM` instead.

boegel:
closing this since `foss/2019b` is no longer supported, see https://docs.easybuild.io/policies/toolchains

PR #12436, opened 2021-03-19
Make sure that path to eb is always found by tests

ocaisa:
Addresses https://github.com/easybuilders/easybuild/issues/684 for the easyconfig repo


boegel (approved):
lgtm

PR #12438, opened 2021-03-19
{math}[foss/2020b] Keras v2.4.3, Theano v1.1.2 w/ Python 3.8.6

dithwick:
(created using `eb --new-pr`)


verdurin:

verdurin:
Test report by @verdurin
**SUCCESS**
Build succeeded for 4 out of 4 (3 easyconfigs in total)
easybuild-c7.novalocal - Linux centos linux 7.9.2009, x86_64, Intel Xeon Processor (Skylake, IBRS), Python 3.6.8
See https://gist.github.com/41f916eee34335d353a5bf7c142a2301 for a full test report.

verdurin:
Test report by @verdurin
**SUCCESS**
Build succeeded for 4 out of 4 (3 easyconfigs in total)
nuc.lan - Linux Fedora 33, x86_64, Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz, Python 3.9.2
See https://gist.github.com/06f7eb5da5f0a63e8399fcd32654c537 for a full test report.

easybuild/easyconfigs/m/mkl-service/mkl-service-2.3.0-foss-2020b.eb
```diff
@@ -0,0 +1,32 @@
+easyblock = 'PythonPackage'
+
+name = 'mkl-service'
+version = '2.3.0'
+
+homepage = 'https://github.com/IntelPython/mkl-service'
+description = "Python hooks for Intel(R) Math Kernel Library runtime control settings."
+
+toolchain = {'name': 'foss', 'version': '2020b'}
```
verdurin, line 9: @dithwick I'm half-wondering whether this should in fact be using the `gomkl` toolchain?

dithwick: Yeah it probably could do come to think of it. I think I just put it under foss to keep it consistent with the intel chain but I'll try moving it.

dithwick: Actually thinking about it, wouldn't that then force Keras on to gomkl as well or have I misunderstood the hierarchy? I want to avoid doing that as otherwise Tensorflow etc. will also need to be built on gomkl.

dithwick:
The other thought that occurred to me is what happens for ppc64le or aarch64. Is there a way to only use a dependency (namely mkl-service in keras) for certain architectures?

dithwick:
@boegel What do you think is the best way forward with this?

smoors:
combining two BLAS libraries (MKL and OpenBLAS) as is done here is probably not a good idea, as you don't know which one will be used.
we're investigating the use of FlexiBLAS in the future, which would fix this issue. we could then have a toolchain with multiple BLAS implementations, and select the BLAS lib at runtime.

easybuild/easyconfigs/t/Theano/Theano-1.1.2-foss-2020b-PyMC.eb
```diff
@@ -0,0 +1,41 @@
...
+
+sanity_check_paths = {
+    'files': ['bin/theano-cache'],
+    'dirs': ['lib/python%(pyshortver)s/site-packages'],
+}
+
+local_blas_warning = "Using NumPy C-API based implementation for BLAS functions"
+sanity_check_commands = [
+    # make sure that warning about falling back to (slow) NumPy C-API for BLAS functions is *not* printed
+    # mkl-service dependency is required to ensure this doesn't happen
```
smoors, line 37: this comment is actually not correct. if the `mkl-service` dependency is not present, the warning is not shown because it will happily use the BLAS that numpy was linked to (OpenBLAS in this case).
see here for more info:
https://theano-pymc.readthedocs.io/en/latest/troubleshooting.html#how-do-i-configure-test-my-blas-library

dithwick: Thanks @smoors for both comments. I had based the easyconfig on the intel one, incorrectly assuming there would be a benefit. I will update accordingly (remove the mkl dependency), it'll make it easier to produce a fosscuda version as well.

dithwick:
I think it failed because a PYPI_SOURCE was added as a default to the pythonpackage easyblock (commit ee1777b). If I remove this then presumably I need to add a minimum easybuild version but presumably I can't set this to 4.3.4 since it hasn't been released yet (the change isn't in 4.3.3). What should I do, update the PR after 4.3.4 is released?

smoors:
> I think it failed because a PYPI_SOURCE was added as a default to the pythonpackage easyblock (commit ee1777b). If I remove this then presumably I need to add a minimum easybuild version but presumably I can't set this to 4.3.4 since it hasn't been released yet (the change isn't in 4.3.3). What should I do, update the PR after 4.3.4 is released?

PYPI_SOURCE is now indeed the default source_url. you can just remove that line and re-upload the easyconfig (or do it directly in the github UI), don't worry about the EB version. someone with a develop version of EB will run a test report.

dithwick:
> > I think it failed because a PYPI_SOURCE was added as a default to the pythonpackage easyblock (commit ee1777b). If I remove this then presumably I need to add a minimum easybuild version but presumably I can't set this to 4.3.4 since it hasn't been released yet (the change isn't in 4.3.3). What should I do, update the PR after 4.3.4 is released?
> 
> PYPI_SOURCE is now indeed the default source_url. you can just remove that line and re-upload the easyconfig (or do it directly in the github UI), don't worry about the EB version. someone with a develop version of EB will run a test report.

Thanks @smoors I've made the change.

smoors:

smoors:
@dithwick sorry for the confusion. what I meant is to remove the mkl-service dependency and the comment, but not the actual `sanity_check_commands` themselves.

dithwick:
@smoors No worries, it was probably me not paying enough attention. I'll add the sanity check commands back in.

smoors:
Test report by @smoors
**SUCCESS**
node379.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz (skylake_avx512), Python 2.7.5
See https://gist.github.com/c9990a4c949bd30856f6abe14d2be220 for a full test report.

smoors (approved):
lgtm

smoors:
Going in, thanks @dithwick!

PR #12445, opened 2021-03-22
made .gitignore to exclude everything other than a few exceptions, rather than allow everything other than exceptions

mboisson:


Flamefire:
C&P my arguments against ignoring everything from the Slack discussion for future reference:

TBH: I find ignoring specific files better. As ignored files won't even show up in git status this might lead to confusion later on or for devs not familiar with that ignore stuff.
So I'd rather ignore specific, known to not be committed stuff (e.g. tar.gz) and optionally run something like your script on CI instead.
.gitignore is more for things that might be created during regular work with the repo, e.g. build artefacts, not so much as a safeguard for randomly added stuff. Maybe netCDF-4.3.0_nc-config.in.cmake  is required? What now? Add another unignore rule for that specific file? IMO that becomes really messy at some point
Also I'm wondering why the test folder is unignored. I could imagine some temporary files to be created there, like `__pycache__` or similar which should be ignored

My usual workflow is using git status to check what I need to commit. If a file doesn't appear there I'll forget it.
I see nothing bad with having users add stuff as long as we ignore the "usual" things (pycache, pytest cache...) to allow that. The rest is up to review and CI to catch IMO. I mean a newly added file will show up in the files -view on GH

Putting it in the repos .gitignore changes behavior on the dev side, which I find quite intrusive for anything which isn't a "usual build artifact to be ignored"

I mean there are 2 sides: You definitely never want to commit a `__pycache__`  folder, but you may want to commit a tgz. That is what .gitignore IMO should do.
The other part is catching peoples mistake where they end up committing stuff that should have never been in the repo in the first place. I think that is quite rare to make a change for all devs.
That's what I meant with not "predicting" unimportant things but "knowing". Protection of the upstream repo against unwanted files/changes is job of the CI

mboisson:
Other side of the arguments : 

The problem with ignore specific things is workflows get creative, and file names along with it. 
someone can always add a file that is supposed to be ignored with `-f` 
people have the bad habit of doing things like `git add *` or `git commit -a` and this picks up whatever is not ignored explicitly.

Extra files in a commit must be caught by a human. Missing files will be caught by tests. It’s easy to not catch a file that shouldn’t be there. it’s hard not to catch a file that should be there but isn’t
Predicting what’s important is easier than predicting what’s not important, because there are a hell of a lot more things that are not important than things that are important

boegel:
Since we haven't seen many (or even any) cases where people are accidentally adding files that shouldn't be there, I'm leaning towards @Flamefire's view here, so I'll go ahead and close this (also because it has bit-rotted, and most people use `--new-pr` anyway for creating easyconfig PRs)

PR #12454, opened 2021-03-23
Stop tests changing the EasyBuild easyconfigs

branfosj:
When the tests are run they changes the EasyBuild easyconfigs, to replace
```
```
with
```
toolchain = SYSTEM
```

This changes the tests to just not do the check for the `dummy` toolchain on the EasyBuild easyconfigs.


boegel (approved):
lgtm

boegel:
Related to this: we should avoid that EasyBuild produces a warning when it's installing an easyconfig file for `EasyBuild` which still has `dummy` a toolchain, since we know it's OK there...

PR #12453, opened 2021-03-23
remove PYPI_SOURCE source URL from easyconfigs using GCC(core) 9.3.0 or 10.2.0 as toolchain

Flamefire:
This is no longer required as it became the default with easybuilders/easybuild-easyblocks#2364

Tests should (and mine were) be done with e.g. `eb --from-pr 12453 --rebuild --fetch --force-download --sourcepath=/dev/shm/sources --upload-test-report` to only test the download and of course with latest develop easyblocks

PR #12452, opened 2021-03-23
remove PYPI_SOURCE source URL from easyconfigs using 2020a toolchain

Flamefire:
This is no longer required as it became the default with easybuilders/easybuild-easyblocks#2364

Tests should (and mine were) be done with e.g. `eb --from-pr 12452 --rebuild --fetch --force-download --sourcepath=/dev/shm/sources --upload-test-report` to only test the download and of course with latest develop easyblocks

PR #12451, opened 2021-03-23
remove PYPI_SOURCE source URL from easyconfigs using 2020b toolchain

Flamefire:
This is no longer required as it became the default with https://github.com/easybuilders/easybuild-easyblocks/pull/2364

Tests should (and mine were) be done with e.g. `eb --from-pr 12451 --rebuild --fetch --force-download --sourcepath=/dev/shm/sources --upload-test-report` to only test the download and of course with latest develop easyblocks

PR #12456, opened 2021-03-23
Add CI test checking if the Python default source_urls are used and Fix CI check where use_pip=False was ignored

Flamefire:
This is for #12451, #12452, #12453 to avoid reintroducing that

The 2nd commit fixes the failure seen in #12453: https://github.com/easybuilders/easybuild-easyconfigs/runs/2174370525

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -629,7 +629,7 @@ class EasyConfigTest(TestCase):
 
                 download_dep_fail = ec.get('download_dep_fail')
                 exts_download_dep_fail = ec.get('exts_download_dep_fail')
-                use_pip = ec.get('use_pip') or exts_default_options.get('use_pip')
+                use_pip = ec.get('use_pip', default=exts_default_options.get('use_pip'))
 
             # download_dep_fail should be set when using PythonPackage
             if easyblock == 'PythonPackage':
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -629,14 +629,17 @@ class EasyConfigTest(TestCase):
 
                 download_dep_fail = ec.get('download_dep_fail')
                 exts_download_dep_fail = ec.get('exts_download_dep_fail')
-                use_pip = ec.get('use_pip', default=exts_default_options.get('use_pip'))
+                use_pip = ec.get('use_pip')
+                if use_pip is None:
+                    use_pip = exts_default_options.get('use_pip')
 
             # download_dep_fail should be set when using PythonPackage
             if easyblock == 'PythonPackage':
                 if download_dep_fail is None:
                     failing_checks.append("'download_dep_fail' should be set in %s" % ec_fn)
                 if ec.get('source_urls', resolve=False) == python_default_urls:
-                    failing_checks.append("'source_urls' can be omitted when using the default value in %s" % ec_fn)
+                    failing_checks.append("'source_urls' should not be defined when using the default value "
+                                          "in %s" % ec_fn)
 
             # use_pip should be set when using PythonPackage or PythonBundle (except for whitelisted easyconfigs)
             if easyblock in ['PythonBundle', 'PythonPackage']:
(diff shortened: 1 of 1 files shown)


test/easyconfigs/easyconfigs.py
```diff
@@ -616,21 +617,26 @@ def check_python_packages(self, changed_ecs, added_ecs_filenames):
...
+                easyblock = ec.get('easyblock')
+                exts_defaultclass = ec.get('exts_defaultclass')
+                exts_default_options = ec.get('exts_default_options', {})
 
-            download_dep_fail = ec.get('download_dep_fail')
-            exts_download_dep_fail = ec.get('exts_download_dep_fail')
-            use_pip = ec.get('use_pip') or exts_default_options.get('use_pip')
+                download_dep_fail = ec.get('download_dep_fail')
+                exts_download_dep_fail = ec.get('exts_download_dep_fail')
+                use_pip = ec.get('use_pip', default=exts_default_options.get('use_pip'))
```
boegel, line 632: I don't think this is correct...

`ec.get('use_pip')` will always work for `PythonBundle`, even if the easyconfig file does not define `use_pip` to `False` or `True`, because `use_pip` is a known easyconfig parameter for `PythonBundle`:

```
$ eb -a -e PythonBundle | grep use_pip | head -1
use_pip*                   Install using 'pip install --prefix=%(prefix)s %(installopts)s %(loc)s' [default: None]
```

So you'll never fall back to `exts_default_options.get('use_pip')`.

The existing line is incorrect too though, we should fall back to `exts_default_options.get('use_pip')` when `use_pip` is still set to `None` (the default) rather than to `False` or `True` explicitly. It's currently wrong when `use_pip = False` is used, but this isn't the correct way of fixing that problem...
```

File: test/easyconfigs/easyconfigs.py [hunk 2]
```diff
@@ -616,21 +617,26 @@ def check_python_packages(self, changed_ecs, added_ecs_filenames):
...
 
             # download_dep_fail should be set when using PythonPackage
             if easyblock == 'PythonPackage':
                 if download_dep_fail is None:
                     failing_checks.append("'download_dep_fail' should be set in %s" % ec_fn)
+                if ec.get('source_urls', resolve=False) == python_default_urls:
+                    failing_checks.append("'source_urls' can be omitted when using the default value in %s" % ec_fn)
```
boegel, line 639: `cam be omitted` -> `should be omitted` or `should not be defined` (latter is clearer imho)

test/easyconfigs/easyconfigs.py
```diff
@@ -642,6 +648,9 @@ def check_python_packages(self, changed_ecs, added_ecs_filenames):
                 if download_dep_fail or exts_download_dep_fail:
                     fail = "'*download_dep_fail' should not be set in %s since PythonBundle easyblock is used" % ec_fn
                     failing_checks.append(fail)
+                if exts_default_options.get('source_urls') == python_default_urls:
+                    failing_checks.append("'source_urls' can be omitted in exts_default_options when using the default"
```
boegel, line 652: `cam be omitted` -> `should be omitted` or `should not be defined` (latter is clearer imho)

boegel (changes requested):

test/easyconfigs/easyconfigs.py
Flamefire: Ah, true. I forgot that those are then set by default

Flamefire:
@boegel On a related issue: Why isn't `sanity_pip_check` required for packages installed without pip? IMO it should always be required as this could/does(?) catch the recent issue where installation via setup.py resulted in a version of 0.0.0

If there is nothing I missed, I'd add that to this PR as it is closely related

boegel (approved):
lgtm

boegel:
> @boegel On a related issue: Why isn't `sanity_pip_check` required for packages installed without pip? IMO it should always be required as this could/does(?) catch the recent issue where installation via setup.py resulted in a version of 0.0.0
> 
> If there is nothing I missed, I'd add that to this PR as it is closely related

I'm not sure it makes sense to run `pip check` when `pip` was not used to install the package...

How would running `pip check` catch the problem with `0.0.0` versions? That would require enhancements in the generic `PythonPackage` PR first?

I would keep that for a separate PR, since the default source URLs and using `pip` are already two separate (though related) things, and `pip check` is yet another...

PR #12478, opened 2021-03-25
Create alf_2.1_make.patch

levlafayette:


smoors (changes requested):
@levlafayette can you add an easyconfig that uses this patch, so we can test it?
also, please add a description and the author to the patch?

PR #12493, opened 2021-03-31
{base,bio}[fosscuda/2020b] CuPy v8.5.0, cuTENSOR v1.2.2.5, Cython v0.29.22, ... w/ Python 3.8.6

ThomasHoffmann77:
(created using `eb --new-pr`)


Micket:
@ThomasHoffmann77 I think this PR is to big to review in one go. Could you please split it up into a few smaller PRs?

* Also, we can move cuTENSOR into SYSTEM level since it's a binary. I have been meaning to open a PR like that but never got around to it: https://github.com/easybuilders/easybuild-easyconfigs/pull/12495

* Cython is bundled with Python nowdays, we probably don't need to have this package unless it's really necessary to get an updated version compared to Python bunle in GCCcore 10.2.0.

boegel:
Follow-up in #12509 (CuPy) and #12504 (PyTom) + the PRs it requires for the tests to pass, thanks for splitting this up @ThomasHoffmann77!

PR #12494, opened 2021-03-31
Fix/interproscan data

jshleap:
Interproscan data need a postprocessing step that was recently added


jshleap:
Ignore this, made a mistake

PR #12505, opened 2021-04-01
Fix/interproscan

jshleap:
Trying to add a pull request for a fix in interproscan-data. A post processing is needed and have been added to the recipe


boegel:
@jshleap Wrong target repo? :)

jshleap:
@boegel Yes, sorry about that!

PR #12517, opened 2021-04-02
fix sanity check error for OpenCV v4.5.1

### LINKED ISSUE #12491: OpenCV_4.5.1-foss-2020b-contrib sanity check error on libopencv_freetype.so
Hi!
Trying to build OpenCV_4.5.1-foss-2020b-contrib, I get this error:
`Sanity check failed: no file found at 'lib64/libopencv_freetype.so' in /opt/apps/easybuild/software/OpenCV/4.5.1-foss-2020b-contrib`
I checked in the folder, and it seems that libopencv_freetype is indeed not built. I checked the log, but it does not seem to complain about anything, only the sanity check is failing.
Should it be removed from sanity check, or am I missing something?
I'm tagging you, @boegel because you're the one who commited this.
Thanks,
Guillaume.

guimou:
fixes #12491 
Freetype and Harfbuzz are required to build freetype: https://github.com/opencv/opencv_contrib/tree/master/modules/freetype
=> Added those modules as dependencies

OpenCV contrib bug in ocv_check_modules prevents detection of Freetype and HarfBuzz, required to build freetype module, ref:
- https://github.com/opencv/opencv_contrib/issues/1497
- https://github.com/opencv/opencv_contrib/issues/2239

=> Applied configopts to "force detection" of the module as per described workaround in issue 2239.

Module requires Python 3:
=> Added configopt to disable Python2 compilation (removes error messages)

### DIFF
--- a/easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
+++ b/easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
@@ -37,6 +37,8 @@ dependencies = [
     ('SciPy-bundle', '2020.11'),  # for numpy
     ('zlib', '1.2.11'),
     ('FFmpeg', '4.3.1'),
+    ('freetype', '2.10.3'),
+    ('HarfBuzz', '2.6.7'),
     ('libjpeg-turbo', '2.0.5'),
     ('libpng', '1.6.37'),
     ('LibTIFF', '4.1.0'),
@@ -50,7 +52,20 @@ dependencies = [
     ('HDF5', '1.10.7'),  # needed by hdf from contrib
 ]
 
-configopts = '-DOPENCV_EXTRA_MODULES_PATH=%(builddir)s/%(namelower)s_contrib-%(version)s/modules'
+# XXXX in configurations is a bug fix in OpenCV because ocv_check_modules is not able to recognize freetype and harfbuzz
+# ref: https://github.com/opencv/opencv/blob/6e8daaec0f46aaba9ea22e2afce47307b1dbff9f/cmake/OpenCVUtils.cmake#L861
+configopts = '-DOPENCV_EXTRA_MODULES_PATH=%(builddir)s/%(namelower)s_contrib-%(version)s/modules '
+configopts += '-DFREETYPE_FOUND=ON '
+configopts += '-DFREETYPE_INCLUDE_DIRS=$EBROOTFREETYPE/include/freetype2/ '
+configopts += '-DFREETYPE_LIBRARIES=$EBROOTFREETYPE/lib64/libfreetype.so '
+configopts += '-DFREETYPE_LINK_LIBRARIES=$EBROOTFREETYPE/lib64/libfreetype.so '
+configopts += '-DFREETYPE_LINK_LIBRARIES_XXXXX=ON '
+configopts += '-DHARFBUZZ_FOUND=ON '
+configopts += '-DHARFBUZZ_INCLUDE_DIRS=$EBROOTHARFBUZZ/include/harfbuzz '
+configopts += '-DHARFBUZZ_LIBRARIES=$EBROOTHARFBUZZ/lib64/libharfbuzz.so '
+configopts += '-DHARFBUZZ_LINK_LIBRARIES=$EBROOTHARFBUZZ/lib64/libharfbuzz.so '
+configopts += '-DHARFBUZZ_LINK_LIBRARIES_XXXXX=ON '
+configopts += '-DBUILD_opencv_python2=OFF '
 
 enhance_sanity_check = True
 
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
```diff
@@ -37,6 +37,8 @@ dependencies = [
     ('SciPy-bundle', '2020.11'),  # for numpy
     ('zlib', '1.2.11'),
     ('FFmpeg', '4.3.1'),
+    ('freetype', '2.10.4'),
```
boegel, line 40: @guimou We have an easyconfig file for freetype 2.10.3 for this generation of easyconfigs (2020b, so `freetype-2.10.3-GCCcore-10.2.0.eb` will be picked up), so please change this to `'2.10.3'`, which should fix the failing tests

boegel (changes requested):

easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
guimou: Oh, yes, my bad! It was a remnant from a test when I was trying to figure it our, I created a package with the latest version in my dev environment. Corrected.

easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
```diff
@@ -50,7 +52,18 @@ dependencies = [
...
+configopts += '-DFREETYPE_FOUND=ON '
+configopts += '-DFREETYPE_INCLUDE_DIRS=$EBROOTFREETYPE/include/freetype2/ '
+configopts += '-DFREETYPE_LIBRARIES=$EBROOTFREETYPE/lib64/libfreetype.so '
+configopts += '-DFREETYPE_LINK_LIBRARIES=$EBROOTFREETYPE/lib64/libfreetype.so '
+configopts += '-DFREETYPE_LINK_LIBRARIES_XXXXX=ON '
+configopts += '-DHARFBUZZ_FOUND=ON '
+configopts += '-DHARFBUZZ_INCLUDE_DIRS=$EBROOTHARFBUZZ/include/harfbuzz '
+configopts += '-DHARFBUZZ_LIBRARIES=$EBROOTHARFBUZZ/lib64/libharfbuzz.so '
+configopts += '-DHARFBUZZ_LINK_LIBRARIES=$EBROOTHARFBUZZ/lib64/libharfbuzz.so '
+configopts += '-DHARFBUZZ_LINK_LIBRARIES_XXXXX=ON '
```
boegel, line 65: What's this `XXXXX` configure option about?

boegel:
@guimou It would be better to fix this in the [OpenCV easyblock](https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/o/opencv.py), or you up for looking into that?

easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
guimou: That's totally weird code in OpenCV to prevent configure from wiping <PKG>_LINK_LIBRARIES: https://github.com/opencv/opencv/blob/6e8daaec0f46aaba9ea22e2afce47307b1dbff9f/cmake/OpenCVUtils.cmake#L861
It seems they put it there as a fix to prevent exactly what's happening, ocv_check_modules not being able to recognize freetype and harfbuzz...

guimou:
> @guimou It would be better to fix this in the [OpenCV easyblock](https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/o/opencv.py), or you up for looking into that?

I could have a look. But unless I'm mistaken, this easyblock does not include the contrib libraries? Which is why I guess @Micket created this package in the first place. I don't know, I just tried to correct this package so that I could install latest OpenCV version.

easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
boegel: I think it's worth adding a comment in the easyconfig file above this line to clarify that, because it looks really weird...

boegel (changes requested):

easybuild/easyconfigs/o/OpenCV/OpenCV-4.5.1-foss-2020b-contrib.eb
guimou: Comment added (finally!)

boegel:
CORE_CNT=16

boegel:
Test report by @boegel
**SUCCESS**
node3134.skitty.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/eff7ca71136493962d15e1836cd87aea for a full test report.

boegel (approved):

boegel:
Going in, thanks @guimou!

PR #12548, opened 2021-04-08
added GPAW-21.1.0-foss-2020a-ASE-3.21.1.eb and the dependencies

xwuupb:
added GPAW-21.1.0-foss-2020a-ASE-3.21.1.eb and the dependencies. It should work.


xwuupb:
Yes, I will take a look and solve it.

PR #12576, opened 2021-04-13
{devel}[GCCcore/10.2.0] glew v2.2.0

ThomasHoffmann77:
(created using `eb --new-pr`)


Micket:
There is actually a new maintainer of this package, and it seems the old webpage isn't updated to reflect the fact that there is a newer release already.

Might be better to point towards the new github page instead.
https://github.com/nigels-com/glew
and we can use the 2.2.0 release (needs to use the tgz file, the tar.gz file is autogenerated by github and it's broken for this repo)

I wanted to build this myself, in order to use it with some other software with off-screen rendering. It turns out GLEW does support both GLX (onscreen or requiring VirtualGL), EGL (technically can do both onscreen and offscreen), OSMESA (offscreen software only).

So,the build variants are mutually exclusive as far as I know. Specify nothing and you get GLX, or you can opt in to either build type by `SYSTEM=linux-egl` and `SYSTEM=linux-osmesa` respectively.

So in my opinion
1. We should probably update to 2.2.0 (and maybe change the homepage?)
2. We should use a versionsuffix for each backend type. Perhaps just a plain -glx, -egl, -osmesa.

I'd probably just skip out on OSMesa, it's a bit dated and EGL is the future which should someday replace both OSMESA and GLX. In fact, maybe one can only use EGL (but that requires that applications using GLEW supports also EGL for rendering).

easybuild/easyconfigs/g/glew/glew-2.1.0-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,36 @@
+# Contribution from the Crick HPC team
+# uploaded by J. Sassmannshausen
+# th77embl updated to GCCcore 10.2.0
+easyblock = 'ConfigureMake'
+
+name = 'glew'
+version = '2.1.0'
+
+homepage = 'http://glew.sourceforge.net/'
```
Micket, line 9:
```suggestion
homepage = 'https://github.com/nigels-com/glew'
```

easybuild/easyconfigs/g/glew/glew-2.1.0-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,36 @@
+# Contribution from the Crick HPC team
+# uploaded by J. Sassmannshausen
+# th77embl updated to GCCcore 10.2.0
+easyblock = 'ConfigureMake'
+
+name = 'glew'
+version = '2.1.0'
```
Micket, line 7: 2.2.0

easybuild/easyconfigs/g/glew/glew-2.1.0-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,36 @@
...
+
+toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
+
+source_urls = [SOURCEFORGE_SOURCE]
+sources = ['%(name)s-%(version)s.tgz']
+checksums = ['04de91e7e6763039bc11940095cd9c7f880baba82196a7765f727ac05a993c95']
+
+builddependencies = [('binutils', '2.35')]
+
+dependencies = [('Mesa', '20.2.1')]
```
Micket, line 22: This actually has a dependency on X11 as well (i see it link to libXau, libXdmcp, libxcb)

Micket (changes requested):

easybuild/easyconfigs/g/glew/glew-2.2.0-GCCcore-10.2.0-egl.eb
```diff
@@ -0,0 +1,44 @@
...
+C/C++ extension loading library. GLEW provides efficient run-time mechanisms
+for determining which OpenGL extensions are supported on the target platform."""
+
+toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
+
+source_urls = ['https://github.com/nigels-com/glew/releases/download/%(name)s-%(version)s/']
+sources = ['%(name)s-%(version)s.tgz']
+checksums = ['d4fc82893cfb00109578d0a1a2337fb8ca335b3ceccf97b97e5cc7f08e4353e1']
+
+github_account = 'nigels-com'
```
Micket, line 24: Since we can't use GITHUB_SOURCE template for the source_urls (since this repo doesn't use the actual archive straight up)  then we can just drop this github_account variable (that's all it's good for)

Micket (changes requested):

easybuild/easyconfigs/g/glew/glew-2.2.0-GCCcore-10.2.0-egl.eb
```diff
@@ -0,0 +1,44 @@
...
+skipsteps = ['configure']
+
+preinstallopts = 'GLEW_PREFIX=%(installdir)s GLEW_DEST=%(installdir)s '
+install_cmd = 'make install.all'
+
+sanity_check_paths = {
+    'files': ['lib/libGLEW.a', 'lib/libGLEW.%s' % SHLIB_EXT] +
+             ['bin/glewinfo', 'bin/visualinfo'] +
+             ['include/GL/%s.h' % h for h in ['glew', 'glxew', 'wglew']],
+    'dirs': ['', ]
```
Micket, line 41: while we're at it, ['', ] doesn't seem useful. Can just use an empty list here instead

Micket (changes requested):

Micket (approved):
lgtm

Micket:
Test report by @Micket
**SUCCESS**
vera-c1 - Linux centos linux 7.9.2009, x86_64, Intel Xeon Processor (Skylake), Python 2.7.5
See https://gist.github.com/8422b4cbc962678415a708dec510cbfe for a full test report.

Micket:
Going in, thanks @ThomasHoffmann77!

PR #12577, opened 2021-04-13
{bio}[foss/2020b] APBS v3.0.0 w/ Python 3.8.6

ThomasHoffmann77:
(created using `eb --new-pr`)
requires #12587 (pdb2pqr)


Micket:
Any reason why we don't compile this? We typically strive to avoid binary installs of software when we can avoid it.

ThomasHoffmann77:
> Any reason why we don't compile this? We typically strive to avoid binary installs of software when we can avoid it.
 I assumed it would be better to at least have the binary install instead of nothing. I have been working on an installation from source. However, it is not ready, yet.

easybuild/easyconfigs/a/APBS/APBS-3.0.0_Linux.eb
```diff
@@ -0,0 +1,29 @@
+# This file is an EasyBuild reciPY as per https://github.com/easybuilders/easybuild
+# Author: Pablo Escobar Lopez
+# Swiss Institute of Bioinformatics
+# Biozentrum - University of Basel
+# TH77EMBL update do 3.0.0, 2021/04
+name = 'APBS'
+version = '3.0.0'
+versionsuffix = '_Linux'
```
boegel, line 8: Maybe tag this as `x86-64`? Linux is kind of a given...

But +1 on building this from source instead...

boegel (changes requested):

boegel:

PR #12583, opened 2021-04-13
add patch to fix hardcoded num_cores in DMCfun extension included with R 4.0.x

### LINKED ISSUE #12569: R's DMCFun might detects number of (available) cores for parallel run incorrectly in PBS
It seems that DMCFun hard codes the number of threads here:
https://github.com/igmmgi/DMCfun/blob/2fa097966150af1c80665f78abd2a122f71ea17f/R/dmcFit.R#L162
it seems that it even thinks that the cores are hyperthreaded (purely speculation, due to the /2)

We should make a patch something like:
```
    chk <- tolower(Sys.getenv("_R_CHECK_LIMIT_CORES_", ""))
    if (nzchar(chk) && (chk == "true")) {
      num_cores <- 2L
    } else {
      num_cores <- as.numeric(Sys.getenv("PBS_NP", parallel::detectCores() / 2))
    }
```

hajgato:
fixes #12569
This only solves PBS cases so other queue systems might need different solution(s)

### DIFF
--- a/easybuild/easyconfigs/r/R/DMCfun-1.3.0_fix-parallel-detect.patch
+++ b/easybuild/easyconfigs/r/R/DMCfun-1.3.0_fix-parallel-detect.patch
@@ -10,7 +10,7 @@ diff -ru DMCfun.orig/MD5 DMCfun/MD5
  034299149a3219bf581d440b19d94fbb *R/RcppExports.R
  4e07b13ccaf91457dfc167540129c61f *R/dmcData.R
 -0bce9adf54182001c9eadbca3c6af0f8 *R/dmcFit.R
-+1f8c6ded3317c4efee6cdd8538907acc *R/dmcFit.R
++8f8759267c76167e07ac0f015edfbd45 *R/dmcFit.R
  655fea3d9c03b02d6f2539ec478ab9b0 *R/dmcSim.R
  f301185baf6007e9ade686f9964ecf23 *R/methods.R
  1b5964cab6608330647f601921f15a5c *R/plot.R
@@ -22,7 +22,7 @@ diff -ru DMCfun.orig/R/dmcFit.R DMCfun/R/dmcFit.R
        num_cores <- 2L
      } else {
 -      num_cores <- parallel::detectCores() / 2
-+      num_cores <- as.numeric(Sys.getenv("PBS_NP", parallel::detectCores() / 2))
++      num_cores <- as.numeric(Sys.getenv("SLURM_NTASKS", Sys.getenv("PBS_NP", parallel::detectCores() / 2))
      }
      
      # Older code (pre June 2020) used doSNOW progress bar 
@@ -31,7 +31,7 @@ diff -ru DMCfun.orig/R/dmcFit.R DMCfun/R/dmcFit.R
      num_cores <- 2L
    } else {
 -    num_cores <- parallel::detectCores() / 2
-+    num_cores <- as.numeric(Sys.getenv("PBS_NP", parallel::detectCores() / 2))
++    num_cores <- as.numeric(Sys.getenv("SLURM_NTASKS", Sys.getenv("PBS_NP", parallel::detectCores() / 2))
    }
    cl <- parallel::makeCluster(num_cores)
    invisible(parallel::clusterExport(cl = cl, varlist=c("dmcSim", "calculateCostValue")))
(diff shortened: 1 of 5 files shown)


verdurin:
I think it would be good to see testing of this on different schedulers before merging.

easybuild/easyconfigs/r/R/DMCfun-1.3.0_fix-parallel-detect.patch
```diff
@@ -0,0 +1,37 @@
...
+ 1b5964cab6608330647f601921f15a5c *R/plot.R
+diff -ru DMCfun.orig/R/dmcFit.R DMCfun/R/dmcFit.R
+--- DMCfun.orig/R/dmcFit.R	2020-12-03 18:53:03.000000000 +0100
++++ DMCfun/R/dmcFit.R	2021-04-13 09:28:47.506520017 +0200
+@@ -159,7 +159,7 @@
+     if (nzchar(chk) && (chk == "true")) {
+       num_cores <- 2L
+     } else {
+-      num_cores <- parallel::detectCores() / 2
++      num_cores <- as.numeric(Sys.getenv("PBS_NP", parallel::detectCores() / 2))
```
akesandgren, line 25: Why not extend this to 
Sys.getenv("PBS_NP", Sys.getenv("SLURM_CPUS_ON_NODE", parallel::detectCores() / 2))

akesandgren: Or possibly the other way around seing that Slurm is more common than PBS these days.

hajgato: I have added `num_cores <- as.numeric(Sys.getenv("SLURM_NTASKS", Sys.getenv("PBS_NP", parallel::detectCores() / 2))`
I think SLURM_NTASKS are closer to the meaning of PBS_NP, which is the number of processors (nodes=x:ppn=y, so PBS_NP=X*Y)

hajgato:
@verdurin I am open for suggestions. I can only reach slurm with pbs compatibility enabled.

easybuild/easyconfigs/r/R/DMCfun-1.3.0_fix-parallel-detect.patch
akesandgren: Ah, so it is MPI aware?

hajgato: no.

akesandgren: Then SLURM_CPUS_ON_NODE is probably more correct.

hajgato: Honestly, I find this hairsplitting, but be it. All in all, if the user incorrectly assumes from an only OpenMP software that it is MPI, then both versions are wrong. You can argue which one is worst, though, but it does not change the fact that both are incorrect. If it is used in a correct way, both versions are just fine. 
Just telling in advance, I will not compare  SLURM_CPUS_ON_NODE  and  SLURM_NTASKS with each other and if they differ, then program will not stop nor write a friendly message that please do not use multiple nodes, because this software is only for one node.

boegel:
@hajgato Patch introduces broken code it seems, `DMCfun` fails to reinstall:

```
Error in parse(outFile) :
  /tmp/boegelbot/R/4.0.3/foss-2020b/DMCfun/DMCfun/R/dmcFit.R:163:5: unexpected '}'
162:       num_cores <- as.numeric(Sys.getenv("SLURM_CPUS_ON_NODE", Sys.getenv("PBS_NUM_PPN", parallel::detectCores() / 2))
163:     }
         ^
ERROR: unable to collate and parse R files for package DMCfun
```

boegel:
EB_ARGS="--skip R-4.0.3-foss-2020b.eb  R-4.0.4-foss-2020b.eb"

boegel:
I've tested the patched installation of `DMCfun` on a Slurm system, and it indeed behaves as expected now: it only starts as many threads as there are cores available in the job :+1:

boegel (approved):
lgtm

boegel:
Going in, thanks @hajgato!

PR #12646, opened 2021-04-15
{bio}[fosscuda/2020a] OpenMM v7.5.0 w/ Python 3.8.2

akesandgren:
(created using `eb --new-pr`)

PR #12666, opened 2021-04-16
{bio}[fosscuda/2020b] OpenMM v7.5.0

krometis:


boegel:
@krometis There are some minor differences with the OpenMM easyconfig that was merged in #12646 which are worth looking into.

You can easily tell what's different using `eb --review-pr 12666`...

krometis:
@boegel I honestly don't know what is the right way to go - it's possible that both approaches work. My approach of setting `OPENMM_BUILD_CUDA_LIB` is from the OpenMM [docs](http://docs.openmm.org/latest/userguide/library.html#all-platforms). And setting `OPENMM_CUDA_COMPILER` is also from the [docs](http://docs.openmm.org/latest/userguide/library.html#cuda-platform). 

But maybe OpenMM figures it out anyway because of the patch added in https://github.com/easybuilders/easybuild-easyconfigs/pull/12646?
```
      char* compiler = getenv("OPENMM_CUDA_COMPILER");
 -    string nvcc = (compiler == NULL ? "/usr/local/cuda/bin/nvcc" : string(compiler));
 +    string nvcc = (compiler == NULL ? "nvcc" : string(compiler));
```

Feel free to use whichever approach seems fit.

boegel:
@akesandgren Thoughts on this? Is there any reason why you didn't use `OPENMM_CUDA_COMPILER`?

boegel:
A `fosscuda/2020b` version of OpenMM 7.5.0 (using the patch from #12646) is available in #12689

akesandgren:
I think I misread that part of the code a bit and deemed it broken, which it isn't. Besides, it's broken in it self to have hard coded things like the path to nvcc...

boegel:
Is `OPENMM_CUDA_COMPILER` something that should just be set at compilation time, or also at runtime?

Looks like it's a runtime thing?

boegel (approved):

boegel:
Going in, thanks @krometis!

PR #12761, opened 2021-04-27
Make sure Qt does not require kernel >3.17

ocaisa:
See https://github.com/EESSI/software-layer/issues/99 for context


ocaisa:
Closing this, it is already supported in the easyblock but will need tweaking for EESSI

PR #12810, opened 2021-05-04
{chem}[GCC/9.3.0] networkTools v2

kErica:
(created using `eb --new-pr`)


easybuild/easyconfigs/n/networkTools/networkTools-2-foss-2019b.eb
```diff
@@ -0,0 +1,30 @@
...
+easyblock = 'MakeCp'
+
+name = 'networkTools'
+version = '2'
+
+homepage = 'http://faculty.scs.illinois.edu/schulten/software/networkTools/'
+description = """Dynamical Network Analysis is a method of characterizing allosteric
+signalling through biomolecular complexes."""
+
+toolchain = {'name': 'foss', 'version': '2019b'}
```
Micket, line 17: I think this can (and should) move to GCC level instead of using foss.

Micket (changes requested):

boegel:
@kerica Thanks a lot for your contribution, and welcome to EasyBuild!

One more suggestion on top of @Micket's review: please consider adding a simple sanity check command (via the `sanity_check_commands` easyconfig parameter), something like `gncommunities --help` for example.

And don't hesitate to reach out if you have any question (either in this PR, or via Slack).

easybuild/easyconfigs/n/networkTools/networkTools-2-foss-2019b.eb
```diff
@@ -0,0 +1,30 @@
...
+
+source_urls = ['http://faculty.scs.illinois.edu/schulten/software/networkTools']
+sources = ['%(name)s.tgz']
+checksums = ['fb98a414b56ab8b5dc57d97c67961a4cf311fc99f296bab44420513ecae70863']
+
+files_to_copy = [(['%(builddir)s/%(name)s/gncommunities'], 'bin'), (['%(builddir)s/%(name)s/subopt'], 'bin')]
+
+sanity_check_paths = {
+    'files': ['bin/gncommunities', 'bin/subopt'],
+    'dirs': ['']
```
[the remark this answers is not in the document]
kerica: Thank you @Micket for the suggestions, I actually changes both the dirs and the toolchain

kerica:
@boegel thanks for the suggestion and for the welcoming, it is my first PR, I am not sure how to introduce it in the slack channel so I answer here.

I have tried both a test and the sanity check you suggested, but it gives an error. The problem is that running the binaries with the help flag or without flags returns an exit error 1, that's why we decided to avoid it.

Since I changed the toolchain, should I make a new PR or I can just update this one?

Micket:
A quick glance at the source code leads me to believe that this software doesn't have any suitable commands for sanity checking. No -h or -v flags at all.

kerica:
I checked it just by running the command and it provides the USAGE, but with the exit code 1, which results in an error when testing

easybuild/easyconfigs/n/networkTools/networkTools-2-foss-2019b.eb
```diff
@@ -0,0 +1,30 @@
...
+# Authors::     Erica Bianco <erica.bianco@hpcnow.com>, Helena Gomez <helena.gomez@hpcnow.com> 
+# License::     GPL-v3.0
+# #
+
+easyblock = 'MakeCp'
+
+name = 'networkTools'
+version = '2'
+
+homepage = 'http://faculty.scs.illinois.edu/schulten/software/networkTools/'
```
Micket, line 13:
```suggestion
homepage = 'https://faculty.scs.illinois.edu/schulten/software/networkTools/'
```

easybuild/easyconfigs/n/networkTools/networkTools-2-foss-2019b.eb
```diff
@@ -0,0 +1,30 @@
...
+description = """Dynamical Network Analysis is a method of characterizing allosteric
+signalling through biomolecular complexes."""
+
+toolchain = {'name': 'foss', 'version': '2019b'}
+
+source_urls = ['http://faculty.scs.illinois.edu/schulten/software/networkTools']
```
Micket, line 19:
```suggestion
source_urls = ['https://faculty.scs.illinois.edu/schulten/software/networkTools']
```

Micket (changes requested):
You can just leave out the sanity check command since there are no good options.

easybuild/easyconfigs/n/networkTools/networkTools-2-foss-2019b.eb
kerica: Thanks

easybuild/easyconfigs/n/networkTools/networkTools-2-GCC-9.3.0.eb
```diff
@@ -0,0 +1,30 @@
...
+name = 'networkTools'
+version = '2'
+
+homepage = 'https://faculty.scs.illinois.edu/schulten/software/networkTools/'
+description = """Dynamical Network Analysis is a method of characterizing allosteric
+signalling through biomolecular complexes."""
+
+toolchain = {'name': 'GCC', 'version': '9.3.0'}
+
+source_urls = ['http://faculty.scs.illinois.edu/schulten/software/networkTools']
```
Micket, line 19:
```suggestion
source_urls = ['https://faculty.scs.illinois.edu/schulten/software/networkTools']
```

Micket (changes requested):

Micket (approved):

Micket (approved):
lgtm

Micket (approved):

Micket:

kErica:
@Micket and @boegel any news? I don't know how it works the testing stage. There is anything else I can do to pushing forward my first recipy?

Micket:
Sorry @kErica, so many PRs it's hard to keep track on things.

You could just whitelist the URLs in the test-suite and switch back to http:// since their webserver is incorrectly configured
https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/test/easyconfigs/easyconfigs.py#L865-L871
(unfortunately, you can't use the eb --update-pr commands for this, you'll need to check these changes into your PR branch manually)

Micket:
Actually, looking at this closer, perhaps we don't need to whitelist it. You can start by just switching over to http again and lets see if the test suite fails.

kErica:
I'm sending an update with just the https to http changed. Let's see if this pass @ generoso like this

Micket:
And here is the problem, the automatic test suite won't allow http urls when https exists. I'm a bit puzzled though, as `urlopen` does (correctly) fail on my machines due to the outdated SSL configuration on this host.

But, this PR needs to add a whitelist on the tests for this URL ot package name (in the code i linked above)

Micket:

Micket (approved):
lgtm

Micket:
Going in, thanks @kErica!

PR #12833, opened 2021-05-07
{phys}[gompi/2020b,iimpi/2020b] libGridXC v0.9.6

micaeljtoliveira:


Micket (changes requested):
So compared to the old libGridXC version we have, this switches moduleclass to 'phys'. Is this more accurate than 'chem'? 

The previous version was also at iimpi level. If this library doesn't use BLAS/LAPACK at all, then i will be beneficial to move them to gompi and iiimpi respectively.

micaeljtoliveira:
Thanks for the quick review and sorry for the not so quick follow-up.

> So compared to the old libGridXC version we have, this switches moduleclass to 'phys'. Is this more accurate than 'chem'?

Honestly I didn't look at the old libGridXC easyconfig, as my original easyconfig actually predates it (just took a while to decide to create a PR for it), so I just choose the class that seemed more appropriate. In this case this is a bit subjective, as all the electronic structure packages like this one and others like libxc, can be used by both condensed matter physicists and  quantum chemists. For this package in particular, I think it will be mostly be used by physicists, so probably better to change the class. 

> The previous version was also at iimpi level. If this library doesn't use BLAS/LAPACK at all, then i will be beneficial to move them to gompi and iiimpi respectively.

Good point. Should be fixed now.

Issue #12830, opened 2021-05-06
Update ROOT configopts


rarensu:
The configopts for ROOT 6.14 and 6.20 are wildly out of date. It looks like they were copied from 6.10 without checking. I created an updated set that seems to work for 6.24 (that's what I've been working on) so probably some of what I did will apply to 6.14 and 6.20 as well. This stuff is confusing, OK? 

```
# NOTE: Ensure that each configopts string begins with a blank
# disable some components
configopts = " -Dbuiltin_llvm=OFF "
configopts += " -Dxrootd=OFF -Dmysql=OFF -Doracle=OFF -Dpgsql=OFF "

# make sure some components are enabled
configopts += " -Dpyroot=ON -Dunuran=ON -Dminuit2=ON -Droofit=ON "

# Add component-specific settings based on dependencies
configopts += ' -Dfftw3=ON -Dgsl_shared=ON -DOpenGL_GL_PREFERENCE=GLVND'
```

rarensu:
It turns out we still need the builtin LLVM; ROOT keeps patching their internal LLVM even though they promised not to.

Zafran-31205:
49062619

PR #12106, opened 2021-02-05
{astro,lib}[GCCcore/10.2.0] CFITSIO v3.49, HEALPix v3.70

jose-d:
Healpix 3.70 with patch solving this bug: https://sourceforge.net/p/healpix/bugs/113/ .

(created using `eb --new-pr`)

PR #12860, opened 2021-05-13
{data}[foss/2020b] update ROOT 6.24.00

rarensu:
Update ROOT and dep GL2PS to foss-2020b toolchain. Fix previous ROOT config opts which were incorrect.

Refers to issue #12830 

Depends on PR #12106 and PR ~~#12576~~

### DIFF
--- a/easybuild/easyconfigs/g/GL2PS/GL2PS-1.4.2-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/g/GL2PS/GL2PS-1.4.2-GCCcore-10.2.0.eb
@@ -1,0 +1,34 @@
+easyblock = 'CMakeMake'
+
+name = 'GL2PS'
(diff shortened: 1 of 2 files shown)


boegel:
@rarensu Sorry for the delay in picking up on this...

I think it's worth opening a separate PR for tweaking the configure options for the existing ROOT easyconfigs (which would close #12830), do you think that makes sense?

It would make sense to include a brief description in the PR body to clarify the changes, like why it's OK to drop `-Dpython=ON -Dpython3=ON -Dpcre=ON -Dzlib=ON`, etc.

rarensu:
I have overridden the head of this branch with a commit that does not contain the config opt fixes. Those fixes have been put on another branch which is now also a pull request.

Red3amer (commented):
``

boegel:

PR #12862, opened 2021-05-13
{data}[foss/2018b,foss/2019b] fix configure options for ROOT v6.14.06 and v6.20.04

rarensu:
Refers to issue #12830 

Options
`-Dpython=ON -Dpython3=ON -Dpcre=ON -Dzlib=ON -Dgsl=ON`
Were removed for 6.14.06, they no longer have any function. Those features are now mandatory. In addition, the builtins are disabled by default, which is correct since easybuild provides the external dependency for them. I suppose if you wanted, you could explicitly disable the builtins to keep a record in the file. 

Option
`-Dgsl_shared=ON`
Was added for 6.14.06, but as far as I can tell, it does nothing. 

Option
`-Dexplicitlink=ON`
Was removed for 6.20.04, it no longer has any function. That feature is now mandatory. 

Option
`-Dpyroot=ON`
Was added for 6.20.04, but its default value is ON so this definition actually does nothing.

### DIFF
--- a/easybuild/easyconfigs/r/ROOT/ROOT-6.14.06-foss-2018b-Python-3.6.6.eb
+++ b/easybuild/easyconfigs/r/ROOT/ROOT-6.14.06-foss-2018b-Python-3.6.6.eb
@@ -44,10 +44,9 @@ dependencies = [
 configopts = " -Dxrootd=OFF -Dmysql=OFF -Dkrb5=OFF -Dodbc=OFF -Doracle=OFF -Dpgsql=OFF -Dqt=OFF"
 
 # make sure some components are enabled
-configopts += " -Dpython=ON -Dpython3=ON -Dpcre=ON -Dzlib=ON"
 configopts += " -Dunuran=ON -Dtable=ON -Dexplicitlink=ON -Dminuit2=ON -Droofit=ON "
 
 # Add component-specific settings based on dependencies
-configopts += ' -Dfftw3=ON -Dgsl=ON -DOpenGL_GL_PREFERENCE=GLVND'
+configopts += ' -Dfftw3=ON -Dgsl_shared=ON -DOpenGL_GL_PREFERENCE=GLVND'
 
 moduleclass = 'data'
--- a/easybuild/easyconfigs/r/ROOT/ROOT-6.14.06-foss-2018b-Python-3.6.6.eb
+++ b/easybuild/easyconfigs/r/ROOT/ROOT-6.14.06-foss-2018b-Python-3.6.6.eb
@@ -44,9 +44,9 @@ dependencies = [
 configopts = " -Dxrootd=OFF -Dmysql=OFF -Dkrb5=OFF -Dodbc=OFF -Doracle=OFF -Dpgsql=OFF -Dqt=OFF"
 
 # make sure some components are enabled
-configopts += " -Dunuran=ON -Dtable=ON -Dexplicitlink=ON -Dminuit2=ON -Droofit=ON "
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/r/ROOT/ROOT-6.14.06-foss-2018b-Python-3.6.6.eb
```diff
@@ -44,10 +44,9 @@ dependencies = [
 configopts = " -Dxrootd=OFF -Dmysql=OFF -Dkrb5=OFF -Dodbc=OFF -Doracle=OFF -Dpgsql=OFF -Dqt=OFF"
 
 # make sure some components are enabled
-configopts += " -Dpython=ON -Dpython3=ON -Dpcre=ON -Dzlib=ON"
 configopts += " -Dunuran=ON -Dtable=ON -Dexplicitlink=ON -Dminuit2=ON -Droofit=ON "
 
 # Add component-specific settings based on dependencies
-configopts += ' -Dfftw3=ON -Dgsl=ON -DOpenGL_GL_PREFERENCE=GLVND'
+configopts += ' -Dfftw3=ON -Dgsl_shared=ON -DOpenGL_GL_PREFERENCE=GLVND'
```
boegel, line 50: @rarensu It looks like this part is missing in `ROOT-6.14.06-foss-2018b-Python-2.7.15.eb`, do you mind adding it there too while you're at it?

boegel (changes requested):

easybuild/easyconfigs/r/ROOT/ROOT-6.14.06-foss-2018b-Python-3.6.6.eb
rarensu: Did I do that correctly?

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 4 out of 7 (3 easyconfigs in total)
node2605.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/4b6fa9e892c00b131e5ef191e6fe95b5 for a full test report.

easybuild/easyconfigs/r/ROOT/ROOT-6.14.06-foss-2018b-Python-3.6.6.eb
boegel: Yes, just adding another commit like you did is OK, but I would also make sure that the comments are the same in both 6.14.06 easyconfigs?

rarensu:
The test report looks like ROOT failed do download some things it needed to build its internals. Is this expected? Because if it's not, then its likely because CERN changed the URL of the items to be downloaded, and therefore we need to be using tag ROOT-*-patches where the URL is fixed.

rarensu:
Boegel bot is complaining that the url in the description is http instead of https. This is the case is almost every ROOT eb file.

boegel:
> Boegel bot is complaining that the url in the description is http instead of https. This is the case is almost every ROOT eb file.

We installed a check a while ago to transition to `https` URLs. Can you please fix that aspect in the easyconfig files you're touching in this PR?

boegel:
> The test report looks like ROOT failed do download some things it needed to build its internals. Is this expected? Because if it's not, then its likely because CERN changed the URL of the items to be downloaded, and therefore we need to be converting some ROOT pull requests where the URL is fixed into patches.

No, it's not expected, and I think your analysis is correct. Are you up for trying to resolve those issue?

boegel:
closing this since `foss/2019b` and older toolchains are no longer supported, see https://docs.easybuild.io/policies/toolchains

PR #12863, opened 2021-05-14
{toolchain}[system/system] FCC v21.05

migueldiascosta:
(created using `eb --new-pr`)

requires ~~https://github.com/easybuilders/easybuild-framework/pull/3677~~ and ~https://github.com/easybuilders/easybuild-easyblocks/pull/2425~

being tested at Fugaku


migueldiascosta:
**SUCCESS**
g28-4009c - Linux RHEL 8.3, AArch64, UNKNOWN (a64fx), Python 3.6.8
See https://gist.github.com/d8469dba5fe3549142f62285b32da3b5 for a full test report.

easybuild/easyconfigs/f/FFTW3-sve/FFTW3-sve-1.0.0-ffmpi-2105.eb
```diff
@@ -0,0 +1,26 @@
...
+ with 512-bit Scalable Vector Extension (SVE) support."""
+
+toolchain = {'name': 'ffmpi', 'version': '2105'}
+toolchainopts = {'pic': True}
+
+source_urls = ['https://github.com/fujitsu/fftw3/archive/refs/tags/']
+sources = ['sve-v%(version)s.tar.gz']
+checksums = ['b5931e352355d8d1ffeb215922f4b96de11b8585c423fceeaffbf3d5436f6f2f']
+
+preconfigopts = "touch ChangeLog && autoreconf --verbose --install --symlink --force && "
```
Micket, line 17: Probably needs Autotools?

Micket (dismissed):

easybuild/easyconfigs/f/FFTW3-sve/FFTW3-sve-1.0.0-ffmpi-2105.eb
migueldiascosta: yeah, but for now I would only add an `osdep`?  (this toolchain is at least for now only meant to use on top of what is already available at Fugaku)

migueldiascosta:
Some requests for comments here:

* hierarchy: there are basically two levels, the one which consists of what's provided by the 'lang' external module, which is compiler + mpi + linalg, and then the full toolchain is simply the first level plus fftw3-sve. I called the first level ffmpi, but maybe that's confusing?

* versioning of external module: should we pin the version of the 'lang' module? At Fugaku the latest version is the default, but that does change the compiler version...

* versioning of the toolchain: I chose `2105` following the year/month of the cray toolchains, and thinking to map those to the default module version at fugaku at the time, but we could also map to the common toolchains, e.g. `2021a`, referring to using the same versions of everything else (once I get to those)

What do people think?

easybuild/easyconfigs/f/FFTW3-sve/FFTW3-sve-1.0.0-ffmpi-2105.eb
migueldiascosta: I suppose an `osdep` on `binutils` would also make sense (in `ffmpi`), since the `lang` external module wouldn't work without it...

boegel:
> Some requests for comments here:
> 
>     * hierarchy: there are basically two levels, the one which consists of what's provided by the 'lang' external module, which is compiler + mpi + linalg, and then the full toolchain is simply the first level plus fftw3-sve. I called the first level ffmpi, but maybe that's confusing?

`ffmpi` suggests only compiler + mpi, so perhaps that's indeed a bit confusing.
If you take `f` for Fujitsu compilers, `m` for MPI, and `ssl` for the SSL part (which is confusing by itself I guess), I get `fmssl`. Not sure if that's better though, but at least there it'll make people think "I better check what this is" rather than assuming "`ffmpi` must mean compiler + MPI" which is my initial reaction. ;)

>     * versioning of external module: should we pin the version of the 'lang' module? At Fugaku the latest version is the default, but that does change the compiler version...

I would say pin it, but I guess that depends on the Fujitsu ecosystem. Are older versions of the `lang` module retained when system updates are done, or do they disappear?

>     * versioning of the toolchain: I chose `2105` following the year/month of the cray toolchains, and thinking to map those to the default module version at fugaku at the time, but we could also map to the common toolchains, e.g. `2021a`, referring to using the same versions of everything else (once I get to those)

I would stay away from the common toolchain versioning mechanism here, and go with something like `21.05` instead, like CSCS is also doing for the recent Cray toolchains (see for example https://github.com/eth-cscs/production/tree/master/easybuild/easyconfigs/c/cpeGNU).

migueldiascosta:
> I would say pin it, but I guess that depends on the Fujitsu ecosystem. Are older versions of the `lang` module retained when system updates are done, or do they disappear?

There are modules for two older versions, but I see even older versions in the filesystem - don't know if there were modules for them and disappeared or if there were never modules for them in the first place

migueldiascosta:
**SUCCESS**
Build succeeded for 17 out of 17 (17 easyconfigs in total)
i31-4208c - Linux RHEL 8.3, AArch64, UNKNOWN (a64fx), Python 3.6.8
See https://gist.github.com/e0260c27d881328628a8153d5a01f866 for a full test report.

easybuild/easyconfigs/p/Perl/Perl-5.32.0-FCC-21.05.eb
```diff
@@ -0,0 +1,392 @@
...
+    ('openssl-devel', 'libssl-dev', 'libopenssl-devel'),
+]
+
+exts_default_options = {'buildopts': 'CC="$CC"'}
+
+
+# !! order of extensions is important !!
+# extensions updated on Aug 21st 2020
+exts_list = [
+    ('Config::General', '2.63', {'source_tmpl': 'Config-General-%(version)s.tar.gz', 'source_urls': ['https://cpan.metacpan.org/authors/id/T/TL/TLINDEN'], 'checksums': ['0a9bf977b8aabe76343e88095d2296c8a422410fd2a05a1901f2b20e2e1f6fad']}),
```
migueldiascosta, line 37: not sure what happened in `exts_lists` here (also if `Python` easyconfigs), all extensions are in a single (too long) line, was it `--try-toolchain --copy-ec`?

easybuild/easyconfigs/f/FFTW3-sve/FFTW3-sve-1.0.0-ffmpi-21.05.eb
```diff
@@ -0,0 +1,25 @@
+easyblock = 'EB_FFTW'
+
+name = 'FFTW3-sve'
```
boegel, line 3: Wouldn't it make more sense to use `-fujitsu` versionsuffix here, since this is the Fujitsu fork of FFTW, rather than the upstream?

boegel:
@migueldiascosta This PR is pretty big, I think it would make sense to break it up into smaller bits that are easier to review/test?

Start with `FCC`, then tackle `Perl`, then `Autotools`, etc.?

easybuild/easyconfigs/f/FFTW3-sve/FFTW3-sve-1.0.0-ffmpi-21.05.eb
migueldiascosta: yes, but won't `FFTW-1.0.0-ffmpi-21.05-fujitsu.eb` be confusing?

migueldiascosta:
**SUCCESS**
j28-4101c - Linux RHEL 8.3, AArch64, UNKNOWN (a64fx), Python 3.6.8
See https://gist.github.com/d47e5ea66d7d3ee8520e5c7d10fdb75a for a full test report.

migueldiascosta:
**SUCCESS**
e31-1001c - Linux RHEL 8.3, AArch64, UNKNOWN (a64fx), Python 3.6.8
See https://gist.github.com/7bec4051c24e91758072aedbfdafaf18 for a full test report.

easybuild/easyconfigs/f/FFTW3-sve/FFTW3-sve-1.0.0-ffmpi-21.05.eb
boegel: No, I think that's the right way of naming it. The `version` and `versionsuffix` are a bit disconnected, sure, but I doubt we want to go and tweak the naming scheme for easyconfigs just because of this example. ;)

boegel (approved):
lgtm

PR #12905, opened 2021-05-19
Fix BWA hardcoded Makefile issues and remove unneeded dependency

jordiblasco:


jordiblasco:
I realized that BWA Makefile has some key hardcoded variables (CC, LDFLAGS, and CFLAGS).
I introduced fixes for the latest version 0.7.17 and I also removed the dependency with zlib as it is loaded with the basic toolchain.

easybuild/easyconfigs/b/BWA/BWA-0.7.17-GCC-10.2.0.eb
```diff
@@ -31,9 +32,11 @@ checksums = ['980b9591b61c60042c4a39b9e31ccaad8d17ff179d44d347997825da3fdf47fd']
 
 dependencies = [
     ('Perl', '5.32.0'),
-    ('zlib', '1.2.11'),
```
boegel, line 34: @jordiblasco Wouldn't you agree it's better to keep this dependency if it's required by BWA, rather than relying that it gets loaded via the toolchain?

boegel (changes requested):

easybuild/easyconfigs/b/BWA/BWA-0.7.17-GCC-10.2.0.eb
jordiblasco: @boegel Sometimes I find that there are several duplicated dependencies with a minor release update. I was thinking that removing that dependency could mitigate that issue. See some examples below. 

What do you think?

```
[easybuild@slurm-simulator ~]$ eb foss-2019b.eb -r -D | grep Bison
 * [ ] $CFGS/b/Bison/Bison-3.3.2.eb (module: Bison/3.3.2)
 * [x] $CFGS/b/Bison/Bison-3.5.3.eb (module: Bison/3.5.3)
 * [ ] $CFGS/b/Bison/Bison-3.3.2-GCCcore-8.3.0.eb (module: Bison/3.3.2-GCCcore-8.3.0)
 ```

```
[easybuild@slurm-simulator ~]$ eb intel-2019b.eb -r -D | grep Bison
 * [ ] $CFGS/b/Bison/Bison-3.3.2.eb (module: Bison/3.3.2)
 * [x] $CFGS/b/Bison/Bison-3.5.3.eb (module: Bison/3.5.3)
 * [ ] $CFGS/b/Bison/Bison-3.3.2-GCCcore-8.3.0.eb (module: Bison/3.3.2-GCCcore-8.3.0)
```

boegel: Removing the `zlib` dependency won't affect that at all, especially not if it's an indirect dependency anyway.

Those `Bison` dependencies installed with the `system` toolchain are essentially build dependencies for bootstrapping `GCCcore`, so they'll always be there (and it's actually fine to remove them once `GCCcore` is installed). The latter `Bison` is the one being used as a build dependency for various easyconfigs using `GCCcore/8.3.0` as toolchain.

boegel:
closing this since this only touches easyconfigs that use a toolchain that is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

PR #12925, opened 2021-05-21
dump easyconfig before initializing easyblock in order to compare it with original easyconfig

boegel:
This change is required because of the fix to `get_config_dict` in https://github.com/easybuilders/easybuild-framework/pull/3692 .

The result of the fix is that `ec.parser.get_config_dict()` now accurately reflects the original easyconfig file, whereas before the fix the value of mutable easyconfig parameters could have been tweaked by the easyblock constructor (like `PythonBundle` does for `exts_default_options`).

The easyconfig file that is dumped via `ec.dump()` will still include the modified easyconfig parameter though, and hence the comparison between the original easyconfig file and the dumped easyconfig file failed for easyconfigs that use an easyblock like `PythonBundle` that fiddles with easyconfig parameters at initialization.

Note that the behavior of `dump()` wasn't affected by the change in https://github.com/easybuilders/easybuild-framework/pull/3692, but we're now (accurately) comparing the parsed dumped easyconfig with the actual original easyconfig parameter as defined in the easyconfig file.

The fix here is to just dump the easyconfig file before the easyblock instance is created, so any tweaks done by the easyblock constructor to parameter values don't affect the value comparison for easyconfig parameters we want to do in the tests...

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -1040,6 +1040,16 @@ def template_easyconfig_test(self, spec):
     error_msg = "%s relies on automagic fallback to ConfigureMake, should use easyblock = 'ConfigureMake' instead" % fn
     self.assertTrue(easyblock or app_class is not ConfigureMake, error_msg)
 
+    # dump the easyconfig file;
+    # this should be done before creating the easyblock instance (done below via app_class),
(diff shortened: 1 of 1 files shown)


branfosj (approved):
lgtm

PR #12962, opened 2021-05-26
in easyconfig tests, check version of dependencies named Python, not if dependencies with certain versions are named Python

migueldiascosta:
I think this is what was intended?

This showed up when testing https://github.com/easybuilders/easybuild-easyconfigs/pull/12863, which has an external module dependency with no version information, but I suppose `test_changed_files_pull_request` should also filter out external modules, like `process_all_easyconfigs` does?


boegel (approved):

boegel:
This certainly looks backwards to me, thanks for the fix @migueldiascosta!

If you also want to filter out external modules from `dependencies` where relevant, please open a separate PR for that.

PR #13043, opened 2021-06-04
{bio}[GCC/8.3.0] Gautomatch v0.56

kErica:
(created using `eb --new-pr`)


easybuild/easyconfigs/g/Gautomatch/Gautomatch-0.56-GCC-8.3.0-CUDA-10.1.243-GCC-8.3.0.eb
```diff
@@ -0,0 +1,38 @@
...
+# Copyright::   HPCNow!
+# Authors::     Erica Bianco <erica.bianco@hpcnow.com>
+# License::     GPL-v3.0
+# #
+easyblock = "BinariesTarball"
+
+name = "Gautomatch"
+version = "0.56"
+
+cuda_ver = '10.1.243-GCC-8.3.0'
```
boegel, line 12: @kErica You should rename this to `local_cuda_ver` to fix the failing tests, see also https://docs.easybuild.io/en/latest/Easyconfig-files-local-variables.html

boegel (changes requested):

easybuild/easyconfigs/g/Gautomatch/Gautomatch-0.56-GCC-8.3.0-CUDA-10.1.243-GCC-8.3.0.eb
kErica: Hi @boegel,
this binary is tricky and it works only for the supposed Cuda version. I'll modify it to use a generic cuda-10 and hope it works 
thanks for the suggestion!

boegel:
closing this since `GCC/8.3.0` is no longer supported, see https://docs.easybuild.io/policies/toolchains

@kErica If `Gautomatch` is still relevant, please open a new PR with a more recent toolchain

PR #13074, opened 2021-06-08
Create gmpich-2021.06.eb

Ghepardo:
Build updated for GCC 10.2.0 and MPICH 3.3.2.


Ghepardo:
Hi, is there any issue with this one?  Normally you guys turn pull requests round very quickly.  I have a number of other pull requests that I'd like to raise, but they are dependent on this one being committed.  (Unless there's a way I can raise them before this one has been committed?)

PR #10258, opened 2020-03-26
{lib}[fosscuda/2019b] Racon v1.4.7

Darkless012:
(created using `eb --new-pr`)

PR #13083, opened 2021-06-09
{bio}[fosscuda/2020b] Bonito v0.4.0 + GenomeWorks v2021.02.2

deniskristak:
(created using `eb --new-pr`)


easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021-02-2_dep_versions_n_file_extension.patch
```diff
@@ -0,0 +1,46 @@
...
++Cython==0.29.21
+ intervaltree==3.1.0
+-matplotlib==3.3.2
+-networkx==2.4
+-numpy~=1.17.5
+-pytest==4.4.1
++matplotlib==3.3.3
++networkx==2.5.1
++numpy==1.19.4
++pytest==6.2.4
```
boegel, line 20: This is going to be annoying to maintain, why not use a `sed` command in the easyconfig to transform the `==` to `>=` instead (less maintenance going forward)

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021-02-2_dep_versions_n_file_extension.patch
```diff
@@ -0,0 +1,46 @@
...
++tqdm==4.56.2
+diff -ruN GenomeWorks-dev_orig/pygenomeworks/setup.py GenomeWorks-dev/pygenomeworks/setup.py
+--- GenomeWorks-dev_orig/pygenomeworks/setup.py	2021-06-08 14:59:57.511155059 +0200
++++ GenomeWorks-dev/pygenomeworks/setup.py	2021-06-08 15:00:39.418965132 +0200
+@@ -57,7 +57,7 @@
+     return requirements_file_content
+ 
+ 
+-def copy_all_files_in_directory(src, dest, file_ext="*.so"):
++def copy_all_files_in_directory(src, dest, file_ext="*.a"):
```
boegel, line 34: Hmm, doesn't this mean we're missing an `--enable-shared` configuration option in the easyconfig file, or something like that?

It's possible that having static libraries (`*.a`) is even a real problem in combination with Python, which does a lot of dynamic loading (which required shared libraries (`*.so`) I think)

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
```diff
@@ -0,0 +1,135 @@
...
+
+homepage = 'https://github.com/clara-parabricks/GenomeWorks'
+description = "SDK for GPU accelerated genome assembly and analysis"
+
+toolchain = {'name': 'fosscuda', 'version': '2020b'}
+toolchainopts = {'pic': True}
+
+sources = [
+    {
+        'source_urls': ['https://github.com/clara-parabricks/GenomeWorks/archive/'],
```
boegel, line 16: Wouldn't it be *a lot* easier to download via `git_config` here, with `recursive` enabled?

See https://docs.easybuild.io/en/latest/Writing_easyconfig_files.html#common-easyconfig-param-sources-git-config

We recently switched to this for `PyTorch` (see https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/p/PyTorch/PyTorch-1.8.1-foss-2020b.eb#L10), and it simplifies the easyconfig file quite a bit (and it's way easier to maintain).

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
```diff
@@ -0,0 +1,135 @@
...
+        'download_filename': 'd0bafdb9fdde2ee99932d42f1deec74d53e19514.tar.gz',
+    },
+    {
+        'source_urls': ['https://github.com/rvaser/spoa/archive'],
+        'filename': 'spoa-20190416.tar.gz',
+        'download_filename': '0b9e56da3241d6a887d64adcfd1ebfae29810c04.tar.gz',
+    },
+]
+checksums = [
+    'f036f066b3ae22e603786f79d374cde28942c22064a169f123fb085fd6b4ccaf',  # v2021.02.2.tar.gz
```
boegel, line 56: If you use `git_config`, use `checksums = [None]`, since the source tarball is created locally on download then, and the checksum won't be identical across different systems...

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
```diff
@@ -0,0 +1,135 @@
...
+# googletest
+preconfigopts += "mv %%(builddir)s/googletest-*/* %s/googletest/ && " % local_vendor_dir
+# htslib
+preconfigopts += "mv %%(builddir)s/htslib-*/* %s/htslib/ && " % local_vendor_dir
+# kseqpp
+preconfigopts += "mv %%(builddir)s/kseqpp-*/* %s/kseqpp/ && " % local_vendor_dir
+# libcudacxx
+preconfigopts += "mv %%(builddir)s/libcudacxx-*/* %s/libcudacxx/ && " % local_vendor_dir
+# spoa
+preconfigopts += "mv %%(builddir)s/spoa-*/* %s/spoa/ && " % local_vendor_dir
```
boegel, line 89: All of this isn't needed when using `git_config` with `recursive: True`...

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
```diff
@@ -0,0 +1,135 @@
...
+
+# add missing include to a file
+preconfigopts += "sed -i'' 's/#include <exception>/#include <exception>\\n#include <stdexcept>/g'"
+preconfigopts += " %(builddir)s/GenomeWorks-*/3rdparty/spoa/src/alignment_engine.cpp && """
+
+configopts = "-Dgw_cuda_gen_all_arch=OFF"
```
boegel, line 95: Maybe add a comment to explain what this does?

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
```diff
@@ -0,0 +1,135 @@
...
+
+skipsteps = ['test']
```
boegel, line 97: Also explain why this is needed with a comment?

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
```diff
@@ -0,0 +1,135 @@
...
+    'use_pip': True,
+}
+
+exts_list = [
+    ('genomeworks', version, {
+        'patches': ['GenomeWorks-2021-02-2_dep_versions_n_file_extension.patch'],
+        'postinstallcmds': ["""
+                            sed -i'' 's/==/>=/g' %(builddir)s/GenomeWorks-%(version)s/pygenomeworks/requirements.txt
+                            """],
+        'preinstallopts': """export GW_INSTALL_DIR=%(installdir)s &&
```
boegel, line 113: define `local_genomeworks_preinstallopts` above `exts_list` to avoid triple quoting here:

```python
local_genomeworks_preinstallopts = "export GW_INSTALL_DIR=%(installdir)s && "
local_genomeworks_preinstallopts += "export GW_VERSION=%(version)s && "
local_genomeworks_preinstallopts += "export GW_ROOT_DIR=%(builddir)s/GenomeWorks-%(version)s"

exts_list = [
...
        'preinstallopts': local_genomeworks_preinstallopts,

File: easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb [hunk 9]
```diff
@@ -0,0 +1,135 @@
...
+            """,
+        'sources': ['v%(version)s.tar.gz'],
+        'start_dir': 'pygenomeworks',
+        'checksums': [
+            'f036f066b3ae22e603786f79d374cde28942c22064a169f123fb085fd6b4ccaf',  # v2021.02.2.tar.gz
+            # GenomeWorks-2021-02-2_dep_versions_n_file_extension.patch
+            '3e33e4ecc03aa43699748a623a331ed9be45a9e8a9c519c0fca6deff17144baf',
+        ],
+    }),
+    ('sortedcollections', '2.1.0', {
```
boegel, line 125: Since these are required for `genomeworks`, list these first in `exts_list`?

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
```diff
@@ -0,0 +1,135 @@
...
+        'source_urls': ['https://pypi.python.org/packages/source/s/sortedcollections/'],
+        'checksums': ['d8e9609d6c580a16a1224a3dc8965789e03ebc4c3e5ffd05ada54a2fed5dcacd'],
+    }),
+    ('pytest', '6.2.4', {
```
boegel, line 129: Is this really needed here? `pytest` 6.1.1 is included with Python 3.8.6...

easybuild/easyconfigs/p/pyspoa/pyspoa-0.0.8-GCC-10.2.0.eb
```diff
@@ -24,27 +24,27 @@ patches = ['pyspoa-%(version)s_use-spoa-dep.patch']
...
+    '504f6df0c8789093ff91eed0340c2ad9532700ed03cd2ca202d10c434bca34b2',  # pyspoa-0.0.8_use-spoa-dep.patch
 ]
 
-builddependencies = [('CMake', '3.18.4')]
-
 dependencies = [
     ('Python', '3.8.6'),
     ('pybind11', '2.6.0'),
     ('spoa', '4.0.7'),
+    ('CMake', '3.18.4'),
```
boegel, line 34: Are you sure `CMake` is a *runtime* dependency here? That looks wrong to me...

easybuild/easyconfigs/p/pyspoa/pyspoa-0.0.8_use-spoa-dep.patch
```diff
@@ -16,7 +16,7 @@ author: Kenneth Hoste (HPC-UGent)
          language='c++',
          extra_objects=[
 -            'src/build/lib/libspoa.a'
-+             os.path.join(os.getenv('EBROOTSPOA'), 'lib64/libspoa.a'),
++            os.path.join(os.getenv('EBROOTSPOA'), 'lib64/libspoa.a'),
```
boegel, line 19: Hmm, this looks like a non-necessary accidental change to the patch?

easybuild/easyconfigs/p/python-parasail/python-parasail-1.2.4-fosscuda-2020b.eb
```diff
@@ -12,12 +12,10 @@ source_urls = ['https://pypi.python.org/packages/source/%(nameletter)s/parasail'
 sources = ['parasail-%(version)s.tar.gz']
 checksums = ['8e6e0762924a461bb437cbb71122765c6912dace880e338bd19c7f9d12efead2']
 
-builddependencies = [
-    ('parasail', '2.4.3'),
-]
 dependencies = [
     ('Python', '3.8.6'),
     ('SciPy-bundle', '2020.11'),
+    ('parasail', '2.4.3'),
```
boegel, line 18: Are you sure `parasail` is a runtime dep here?

If so, maybe it's better to make this a separate PR (so it stands out in the release notes...)

easybuild/easyconfigs/s/spoa/spoa-4.0.7-GCC-10.2.0.eb
```diff
@@ -14,7 +14,7 @@ source_urls = ['https://github.com/rvaser/spoa/releases/download/%(version)s/']
 sources = ['spoa-v%(version)s.tar.gz']
 checksums = ['f7a8cd039e4aabea1481e123387fedaa8f87e55418a3649408a615f5c6b1b9a4']
 
-builddependencies = [('CMake', '3.18.4')]
+dependencies = [('CMake', '3.18.4')]
```
boegel, line 17: Again, this looks wrong to me... In what sense is `CMake` a runtime dependency here?

boegel (changes requested):

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
Darkless012: This was my idea, since I did similar thing in Racon https://github.com/easybuilders/easybuild-easyconfigs/pull/10258/files
Does the `git_config` + `recursive` keeps things reproducible? I remember that there was the issue with the checksums (downloading + making a zip again)

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021-02-2_dep_versions_n_file_extension.patch
deniskristak: @boegel whether i put the sed in preconfigopts of the main file or in the first extension, it does not work, maybe pip is doing some magic idk about (copying files before anything else happens) but i've tried what you're proposing multiple times and it did now work

easybuild/easyconfigs/p/python-parasail/python-parasail-1.2.4-fosscuda-2020b.eb
deniskristak: yeah, it's needed..okay, ill make a new PR

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021.02.2-fosscuda-2020b.eb
boegel: We can't include a checksum, but it is reproducible, since we're cloning a specific git commit.

boegel: Switch to `git_config` done in 142f9f7, which leads to a much cleaner easyconfig...

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021-02-2_dep_versions_n_file_extension.patch
boegel: Fixed by adding `-Dgw_build_shared=ON` to `configopts` so shared libraries are built, and then this patch file is no longer needed \o/

easybuild/easyconfigs/g/GenomeWorks/GenomeWorks-2021-02-2_dep_versions_n_file_extension.patch
boegel: You need to take into account that when the `genomeworks` extensions (Python package) is installed, that's done in a fresh unpacked source directory, so you need to make sure the `sed` commands are used there.

I've fixed that in 142f9f7 by adding the necessary `sed` command to `local_genomeworks_preinstallopts`

boegel:
Test report by @boegel
**SUCCESS**
node3308.joltik.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6242 CPU @ 2.80GHz (cascadelake), Python 3.6.8
See https://gist.github.com/387264354ffb1260c2f40a83ddb45574 for a full test report.

boegel:
Going in, thanks @deniskristak!

Issue #12484, opened 2021-03-29
Merging CUDA and non CUDA toolchains into one


Micket:
As has been discussed multiple times in zoom and in chat, merging fosscuda+foss, and just using versionsuffixes for CUDA-variants of the (relatively few) easyconfigs that does have CUDA bindings.

One thing holding us back has been the CUDA support in MPI, which prevents us from moving into foss since it's part of the toolchain definitions itself, but now with UCX, it might be possible to get the best of both worlds (I'm going to just dismiss the legacy CUDA stuff in openmpi and just focus on UCX).

We want something that
1. works regardless of RPATH is used or not
2. can be opt int after foss(without CUDA) is already in place
3. supports all the RDMA goodies we have today in fosscuda.

Can it be done? Maybe; UCX has an environment variable for all the plugins it uses (`ucx_info -f` lists all variables):
We could introduce a a UCX-package (`UCX-CUDA` maybe?) which shadows non-CUDA UCX, and we start setting the environment variable;
```bash
UCX_MODULE_DIR='%(installdir)s/lib/ucx'
```
and, well, that should be it?

Example UCX-CUDA how i envision it:
```python
easyblock = 'ConfigureMake'

name = 'UCX-CUDA'
version = '1.9.0'
local_cudaversion = '11.1.1'
versionsuffix = '-CUDA-%s' % local_cudaversion

homepage = 'http://www.openucx.org/'
description = """Unified Communication X
An open-source production grade communication framework for data centric
and high-performance applications
"""

toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
toolchainopts = {'pic': True}

source_urls = ['https://github.com/openucx/ucx/releases/download/v%(version)s']
sources = ['%(namelower)s-%(version)s.tar.gz']
checksums = ['a7a2c8841dc0d5444088a4373dc9b9cc68dbffcd917c1eba92ca8ed8e5e635fb']

builddependencies = [
    ('binutils', '2.35'),
    ('Autotools', '20200321'),
    ('pkg-config', '0.29.2'),
]

osdependencies = [OS_PKG_IBVERBS_DEV]

dependencies = [
    ('UCX', version),
    ('numactl', '2.0.13'),
    ('CUDAcore', local_cudaversion, '', True),
    ('GDRCopy', '2.1', versionsuffix),
]

configure_cmd = "contrib/configure-release"
configopts = '--enable-optimizations --enable-cma --enable-mt --with-verbs '
[...]

PR #13107, opened 2021-06-10
{toolchain} nvompic v2021a

SebastianAchilles:
(created using `eb --new-pr`)

Depends on
- [x] https://github.com/easybuilders/easybuild-framework/pull/3735
~- [ ] CUDA for 2021a (no PR yet)`~


SebastianAchilles:

SebastianAchilles:
**SUCCESS**
rocky8-eb - Linux rocky linux 8.4, x86_64, Intel(R) Core(TM) i7-6900K CPU @ 3.20GHz (broadwell), Python 3.6.8
See https://gist.github.com/0c883508f1f2cd69a3e26171c5ad7e5a for a full test report.

SebastianAchilles:
**SUCCESS**
Build succeeded for 4 out of 4 (2 easyconfigs in total)
centos8-eb - Linux centos linux 8.3.2011, x86_64, Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz (skylake), Python 3.6.8
See https://gist.github.com/c21c48a9300d461c88bddaa92edb15ea for a full test report.

akesandgren:
@boegelbot Please test @ generoso

akesandgren:
This should eventually use UCX-CUDA when that stuff is completely in place (which should be soon)

easybuild/easyconfigs/o/OpenMPI/OpenMPI-4.1.1-NVHPC-21.5.eb
```diff
@@ -0,0 +1,38 @@
...
+
+builddependencies = [
+    ('pkg-config', '0.29.2'),
+]
+
+dependencies = [
+    ('zlib', '1.2.11'),
+    ('hwloc', '2.4.1'),
+    ('libevent', '2.1.12'),
+    ('UCX', '1.10.0'),
```
akesandgren, line 27: Let's replace this with the now available UCX-CUDA of the same version.

akesandgren (changes requested):

akesandgren:
And add a HPL and OSU-Micro-Benchmarks built on top of this so we can test it.

easybuild/easyconfigs/o/OSU-Micro-Benchmarks/OSU-Micro-Benchmarks-5.7.1-nvompic-2021a.eb
```diff
@@ -0,0 +1,26 @@
...
+description = """OSU Micro-Benchmarks"""
+
+toolchain = {'name': 'nvompic', 'version': '2021a'}
+
+source_urls = ['https://mvapich.cse.ohio-state.edu/download/mvapich/']
+sources = [SOURCELOWER_TGZ]
+checksums = ['cb5ce4e2e68ed012d9952e96ef880a802058c87a1d840a2093b19bddc7faa165']
+
+configopts = 'CC="$MPICC" CXX="$MPICC"'
+configopts += ' --enable-cuda --with-cuda=$EBROOTCUDA'
```
akesandgren, line 16: I think that we only have CUDAcore loaded so it might need to be EBROOTCUDACORE

akesandgren:
This is now tripping up in tools/module_naming_scheme/hierarchical_mns.py:det_modpath_extensions line 231

akesandgren:
@boegel I think we might need some guidance here...
When it is trying to build CUDA-11.3.1-NVHPC-21.5.eb the hmns code sees CUDA and in det_modpath_extensions we have
non_system_cuda = true, entering that if stmt, "if ec['name'] in extend_comps:" is true since CUDA is in at least one of the COMP_NAME_VERSION_TEMPLATES keys, same for "if ec['name'] in comp_names:", etc but eventually comp_name_ver will still be None since there is no 'CUDA,NVHPC' key. But if we're trying to move away from the compiler-CUDA modulepath extension, https://github.com/easybuilders/easybuild-easyconfigs/issues/12484 option 1, we need to do this differently. Either in this PR or in framework.

What is the right way forward here?

Micket:
Considering the recent foss/fosscuda merge changes (including renaming CUDAcore -> CUDA), we need to do similar things here;

1. We use the system level CUDA package as a dependency. This allows us to use the UCX-CUDA plugin directly.
2. Presumably, NVHPC shouldn't provide it's own CUDA on top then? Because it does right now by default.
3. There won't be a second CUDA package, it's just NVHPC as the compiler level, and it already implies CUDA. There is no non-cuda variant like there used to be with GCC vs GCC-CUDA. I don't think we need to do anything special with modulepaths here. It's just NVHPC + OpenMPI to consider.

SebastianAchilles:
Closing since we added a `nvofbf/2022.07` toolchain in https://github.com/easybuilders/easybuild-easyconfigs/pull/16724

PR #13165, opened 2021-06-15
[GCCcore/10.3.0] groff requires makeinfo as a build dep, which requires Perl, which requires groff. Break cyclic redundancy by adding Perl-minimal and makeinfo-minimal.

akesandgren:
(created using `eb --new-pr`)

Depends on https://github.com/easybuilders/easybuild-easyblocks/pull/2474

PR #13164, opened 2021-06-15
[GCCcore/10.2.0] groff requires makeinfo as a build dep, which requires Perl, which requires groff. Break cyclic redundancy by adding Perl-minimal and makeinfo-minimal.

akesandgren:
(created using `eb --new-pr`)

Depends on https://github.com/easybuilders/easybuild-easyblocks/pull/2474

PR #13243, opened 2021-06-23
add GCCcore-1[0-9].x to test_dep_versions_per_toolchain_generation

SebastianAchilles:
This PR adds `GCCcore-1[0-9]` to `test_dep_versions_per_toolchain_generation`. Before only `GCCcore-[7-9]` was included, that is why the check was not executed in https://github.com/easybuilders/easybuild-easyconfigs/pull/13164 and https://github.com/easybuilders/easybuild-easyconfigs/pull/13165


verdurin (approved):
Looks fine.

PR #13257, opened 2021-06-25
code cleanup + improvements for easyconfigs test suite

Flamefire:
Major improvements to the test suite

- Remove a forgotten debug print
- Checkout full history on CI to avoid failure to get merge base
- Remove workaround to cleanup test TMPDIR in favor of class teardown method
- Process ECs on demand (preparation for next change, but also means for non-prs the ecs-in-pr-changed test is a no-op instead of parsing all ECs and never using them)
- Refactor the ECs-changed-in-PR test to run the subtests as separate test cases so failures in one of them don't abort the others
- Ignore cmdline args passed to the test suite, those are handled by unittest and should not be handled by easybuild. This allows to run subsets of tests like: `python test/easyconfigs/easyconfigs.py -v -k test_pr` to run all PR tests only
- Remove unused setting of `EASYBUILD_DEPRECATED` which, if in effect, would cause many failures due to deprecated ECs (using old versions) or even using LMod 6 (on CI) is already deprecated
- Honor setting of `EASYBUILD_TMP_LOGDIR`
- Allow filtering over all tests via: `python test/easyconfigs/suite.py -k foo` which includes the EC parsing tests. This removes the custom output of "ERROR: Not all tests were successful." but still has the output from unittest

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -21,6 +21,8 @@ jobs:
       fail-fast: false
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0 # Required for git merge-base to work
 
     - name: Cache source files in /tmp/sources
       id: cache-sources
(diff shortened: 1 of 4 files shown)


boegel (approved):
lgtm

PR #13277, opened 2021-06-29
{lib}[GCCcore/10.2.0] cget v0.1.9 w/ Python 3.8.6

verdurin:
(created using `eb --new-pr`)


easybuild/easyconfigs/c/cget/cget-0.1.9-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,54 @@
...
+sanity_pip_check = True
+
+sanity_check_paths = {
+    'files': ['bin/cget'],
+    'dirs': ['lib/python%(pyshortver)s/site-packages']
+}
+
+sanity_check_commands = ["cget --help"]
+
+modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
```
Micket, line 52: I don't think we need this for a pythonbundle right?

Micket (changes requested):

easybuild/easyconfigs/c/cget/cget-0.1.9-GCCcore-10.2.0.eb
verdurin: I think that's right yes. Can I also remove the `dirs` from `sanity_check_paths`, or is that still required?

Micket: Still required i think.

branfosj: You can remove all of `sanity_check_paths`. The `dirs` part is checked by default for a `PythonBundle` and the binary in the `files` section is checked via the `sanity_check_commands`.

Micket (approved):
lgtm

easybuild/easyconfigs/c/cget/cget-0.1.9-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,47 @@
...
+
+exts_list = [
+    ('click', '8.0.1', {
+        'source_urls': ['https://pypi.python.org/packages/source/c/click/'],
+        'checksums': ['8c04c11192119b1ef78ea049e0a6f0463e4c48ef00a30160c704337586f3ad7a'],
+    }),
+    ('six', '1.16.0', {
+        'source_urls': ['https://pypi.python.org/packages/source/s/six/'],
+        'checksums': ['1e61c37477a1626458e36f7b1d82aa5c9b094fa4802892072e49de9c60c4c926'],
+    }),
```
Micket, line 36: Actually, i just noticed these dependencies. It's fine if we want to keep these newer versions, but I suspect maybe the exact version of these isn't very important, if so, they are nowdays included in the Python bundle.

boegel: @verdurin ping?

verdurin: Yes, according to https://github.com/pfultz2/cget/blob/master/requirements.txt the Python bundle versions of those extensions are new enough.

Would you prefer that I remove them from this bundle, then?

boegel: Absolutely, yes, if the `click` and `six` that come with the Python dependency suffice, there's no reason to include specific version here.

boegel:

PR #13253, opened 2021-06-24
cleanup of easyconfigs for (bundles) of Python packages: remove default PYPI_SOURCE source URLs + use PythonBundle rather than Bundle easyblock

Flamefire:
- Removed PyPi URLs from PythonPackage/PythonBundle ECs
- Convert Bundle -> PythonBundle (checked by CI, reduces amount of code in EC)
- Remove (now) redundant sanity-checks for python-site-packages folder
- Remove redundant (and checked by CI) download_dep_fail (sometimes even duplicated)
- Replace http urls by https as required by CI
- **DISABLE** sanity_pip_check and use_pip CI checks, to be **REMOVED** right after this was merged. Using this would require actually building those ECs
- [x] Requires https://github.com/easybuilders/easybuild-easyblocks/pull/2493
- [x] Requires https://github.com/easybuilders/easybuild-easyblocks/pull/2494

Test plan: Run with `--fetch`

PR #13295, opened 2021-06-29
re-enable checksum and Python checks in easyconfigs test suite

boegel:
These hacks were included in #13253 to make the PR mergeable, but should be re-activated now ASAP...


Micket (approved):
lgtm

PR #13331, opened 2021-07-02
{bio}[system] PLINK2 20210701

fizwit:
(created using `eb --new-pr`)

Please accept this new version of PLINK. I've changed the name from PLINK to PLINK2. The project is forked and PLINK2  the official name and matches the binary. PLINK2 uses YYYMMDD for the version.  The version suffix can be used to choose the platform.


boegel:
Looks like this version can no longer be downloaded, and there's no test report to rely on, so I'll close this...

@fizwit If `PLINK2` is still relevant, please consider opening a new pull request with a more recent version that is still available for download...

PR #13447, opened 2021-07-16
{bio}[GCC/10.2.0] OMA v2.5.0

sdx23:
(created using `eb --new-pr`)
New version that is not broken on Debian systems. For some reason it needs Python as a build dependency. I'm not sure I've set the toolchain correctly to achieve this, could you please comment on that?


verdurin:
Test report by @verdurin
**SUCCESS**
nuc.lan - Linux Fedora 33, x86_64, Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz, Python 3.9.5
See https://gist.github.com/d792637e9a3453bd0570a7e6fce2012a for a full test report.

verdurin:
@sdx23 welcome to EasyBuild and thanks for the contribution.

That is indeed a bit strange - a binary that nevertheless requires Python. I suppose that given it's only a build dep, there's no need to include Python in the `versionsuffix`. Normally a binary would use the `SYSTEM` toolchain. `GCCcore` might make more sense than `GCC` and make it more compatible with other toolchains.

easybuild/easyconfigs/o/OMA/OMA-2.5.0-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,38 @@
...
+source_urls = ['http://omabrowser.org/standalone/']
+sources = ['%(name)s.%(version)s.tgz']
+checksums = ['30f5f4e73152c07a61a811ce982bdba2a5e1d236312cc16faa0b388b759e937e']
+
+unpack_options = '--strip-components=1'
+
+install_cmd = 'sh ./install.sh %(installdir)s'
+
+builddependencies = [
+    ('Python', '3.8.6'),
```
verdurin, line 26: @sdx23 with the toolchain change, you will need to add `binutils` as a dependency.

```suggestion
    ('Python', '3.8.6'),
    ('binutils', '2.35'),
```

verdurin (changes requested):

boegel:

PR #13475, opened 2021-07-22
{bio}[GCC/10.2.0] snippy v4.6.0, wgsim v20111017, any2fasta v0.4.2, samclip v0.4.0, snp-sites v2.5.1

deniskristak:
(created using `eb --new-pr`)


deniskristak:
@boegel ok, let's continue discussion from [#13477](https://github.com/easybuilders/easybuild-easyconfigs/pull/13477)
what good does the `snpEff` dep make then? 
ill go check where it's used in `snippy` in the first place...

lexming (changes requested):
@deniskristak thanks for the update. I reviewed this PR and it would need a few more changes to be merged:
* the wrapper script `snpEff` is provided by the easyconfig of snpEff. It is located in the `scripts` folder. We can use an easyconfigs such as `snpEff-5.0e-GCCcore-10.2.0-Java-13.eb` but updated for `GCCcore/10.2.0`.  Therefore, it is not necessary to add the `binaries/noarch` folder to `$PATH`.
* use the easyblock `CmdCp` for wgsim. This avoids needing to patch a Makefile and simplifies the easyconfig of wgsim
* add the script `wgsim_eval.pl` to the installation of wgsm
* keep the original attribution in the easyconfig of `any2fasta` from the previous toolchain (`GCCcore/9.3.0` in this case)

I made a PR to this one with these changes in https://github.com/deniskristak/easybuild-easyconfigs/pull/2. If you agree with my suggestions please merge it to update this PR.

deniskristak:
@lexming thank you very much for the changes...i've looked and they look good..
one thing to mention is that snpEff easyconfig. @boegel mentioned that we should use the one packed with `snippy`?
here: https://github.com/easybuilders/easybuild-easyconfigs/pull/13477#discussion_r683319681
@boegel please review, if you don't see any problems with this, i see that @lexming used another source?
i'll merge right after @boegel reviews this, thank you again

boegel:
@deniskristak Can you check if there's a difference between the `snpEff` script provided with `snpEff` itself, and the one included with `snippy`? I suspect the latter is just a copy of the former, and then it makes more sense to stick to the one that comes with `snpEff` itself...

The changes proposed by @lexming look fine to me, so OK to merge already imho. If needed, we can still add additional commits.

deniskristak:
ok, merged, i'll take a look at snpEff's code

lexming:
Test report by @lexming
**SUCCESS**
Build succeeded for 35 out of 35 (6 easyconfigs in total)
node377.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/700414e8fbb33daec2c002e82f1fc380 for a full test report.

lexming (approved):
LGTM

deniskristak:
@lexming @boegel 
diff for exact snpEff files we talked about - https://www.diffchecker.com/0ew3uMJj
RIGHT - included with snippy
LEFT - original snpEff
im not rly sure which one we want though

boegel:
@deniskristak The `snpEff` that is shipped with `snippy` was last updated 5 years ago, while the one that comes with `snpEff` itself has been updated more recently, so I would go with the latter...

lexming:
Going in, thanks @deniskristak !

PR #13474, opened 2021-07-22
{bio}[GCC/10.2.0] freebayes v1.3.5, SeqLib v1.2.0, fermi-lite v20190320, SSW v1.1

deniskristak:
(created using `eb --new-pr`)

PR #13477, opened 2021-07-22
bump HTSlib dependency to 1.12 for VCFtools 0.1.16

deniskristak:
we need to tweak existing easyconfigs a little to satisfy requirements of https://github.com/hpcugent/eb_inuits/issues/198

### DIFF
--- a/easybuild/easyconfigs/s/snpEff/snpEff-5.0-Java-13.eb
+++ b/easybuild/easyconfigs/s/snpEff/snpEff-5.0-Java-13.eb
@@ -16,6 +16,7 @@ checksums = ['2b177440eb4a2c3bbfc8f50d4adf9e405f3dcfc05a31a55c96f773f07832af34']
 
 dependencies = [('Java', '13')]
 
+# we need this to appear as an executable (i.e. in snippy easyconfig)
 postinstallcmds = ['ln -s %(installdir)s/%(name)s.jar %(installdir)s/%(name)s && chmod +x %(installdir)s/%(name)s']
 modextrapaths = {'PATH': ['']}
 
--- a/easybuild/easyconfigs/s/snpEff/snpEff-5.0-Java-13.eb
+++ b/easybuild/easyconfigs/s/snpEff/snpEff-5.0-Java-13.eb
@@ -16,6 +16,9 @@ checksums = ['2b177440eb4a2c3bbfc8f50d4adf9e405f3dcfc05a31a55c96f773f07832af34']
 
 dependencies = [('Java', '13')]
 
+postinstallcmds = ['ln -s %(installdir)s/%(name)s.jar %(installdir)s/%(name)s && chmod +x %(installdir)s/%(name)s']
+modextrapaths = {'PATH': ['']}
+
 sanity_check_paths = {
     'files': ['%(name)s.jar', 'SnpSift.jar'],
     'dirs': [],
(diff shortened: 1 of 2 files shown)


verdurin:
@deniskristak could you describe the problems in this PR, given that looks like a private repo?

deniskristak:
okay, we're solving two things here...firstly, in the snpEff file,

deniskristak:
we are solving two problems in this PR...
firstly, there is snpEff software, where I added symlink (name of the software without .jar extensions), made it executable and added it to the path. This is because our new easyconfig - https://github.com/easybuilders/easybuild-easyconfigs/pull/13475 - especially it's test command (`snippy --check`) relies on this - no extension SW name accessible on path.

secondly, I changed VCFtools' `HTSlib` version to 1.12, because `snippy` requires `BCFtools`, for which there's an easyconfig for `GCC 10.2.0` already, so we naturally use that one. This version uses `HTSlib` of version 1.12.
There's a clash though, because additionally, we need to include VCFtools, which also already exists for `GCC 10.2.0`, but uses `HTSlib` of version 1.11.
This results in a clash of `HTSlib`'s versions, so we had to change one of them. Picked the newer since I suppose it's generally a good idea.
Also, I made sure that the easyconfig I'm changing isn't used as dependency anywhere else with different `HTSlib` version, so that I don't create more instances of the same problem that I'm solving.

easybuild/easyconfigs/s/snpEff/snpEff-5.0-Java-13.eb
```diff
@@ -16,6 +16,10 @@ checksums = ['2b177440eb4a2c3bbfc8f50d4adf9e405f3dcfc05a31a55c96f773f07832af34']
 
 dependencies = [('Java', '13')]
 
+# we need this to appear as an executable (i.e. in snippy easyconfig)
+postinstallcmds = ['ln -s %(installdir)s/%(name)s.jar %(installdir)s/%(name)s && chmod +x %(installdir)s/%(name)s']
+modextrapaths = {'PATH': ['']}
```
boegel, line 21: Isn't the `snpEff` that's required by `Snippy` the script in https://github.com/tseemann/snippy/blob/master/binaries/noarch/snpEff?

You can't run a JAR file directly, so making it executable and stripping the `.jar` doesn't make much sense...

boegel (changes requested):

easybuild/easyconfigs/s/snpEff/snpEff-5.0-Java-13.eb
deniskristak: looks like you're right, I'll try to add that executable to PATH for snippy then

boegel: @deniskristak So we should change this PR to only bump the HTSlib dep in VCFtools 0.1.16?

deniskristak: correct, snippy works with that `binaries/noarch/snpEff` you mentioned...
removing my change then.

deniskristak: @boegel on the second thought, isn't it a bit against what we want? :D using attached prebuilt binaries

deniskristak: ive removed my change, so it's only HTSlib now

boegel: @deniskristak Well, it depends... I think we can still add `snpEff` as a dependency to `Snippy`, but keep using the `snpEff` script that comes with `Snippy` (but this discussion belongs in #13475)

lexming:
@deniskristak @boegel the change for `VCFTools` is already merged in #13474  and the discussion about `snpEff` continues in #13475. Closing this PR.

PR #13499, opened 2021-07-24
{lib}[fosscuda/2020b] TensorRT v8.0.1.6 w/ Python 3.8.6

robert-mijakovic:
(created using `eb --new-pr`)


lexming:
Test report by @lexming
**SUCCESS**
Build succeeded for 2 out of 2 (1 easyconfigs in total)
node226.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz, Python 2.7.5
See https://gist.github.com/20ca5d7300c9ac1a3eccc336a00acf42 for a full test report.

lexming (approved):
Even though the name of the sources say `cuda-11.3`, this tarball for `x86_64` is compatible with CUDA v11.1 as well. This is not the case with the tarballs for ARM and Power though, those are *only* compatible with CUDA v11.3.

LGTM

LGTM

lexming (commented):
After further inspection, it seems that there is a soft requirement of TensorRT 8.0 on cuDNN v8.2 and we might be going into uncharted territory by using cuDNN v8.0.4. So I think that we should downgrade to TensorRT 7.2.2, which more closely matches the available dependencies in `fosscuda/2020b`, namely Python 3.8 and cuDNN 8.0.

If you agree with this change, I made a PR with the downgrade in https://github.com/robert-mijakovic/easybuild-easyconfigs/pull/2

PR #11552, opened 2020-10-21
{bio}[golf/2020a] PLINK v1.9b5

akesandgren:
(created using `eb --new-pr`)

PR #13535, opened 2021-08-01
{chem}[foss/2020a,foss/2020b] GULP v6.0

sassy-crick:
(created using `eb --new-pr`)


easybuild/easyconfigs/g/GULP/gulp-6.0-mkgulp.patch
```diff
@@ -0,0 +1,57 @@
+diff --git a/gulp-6.0.org/Src/mkgulp b/gulp-6.0/Src/mkgulp
```
Micket, line 1: Please add a description of what this patch does, and who the author is (you presumably). I think the patch could be named something with `fftw_blas_linking` or something instead of `mkgulp`.

easybuild/easyconfigs/g/GULP/gulp-6.0-mkgulp.patch
```diff
@@ -0,0 +1,57 @@
...
++echo 'INC=-I.. -I${EBROOTFFTW}/include -I/usr/local/include/' >> makefile
+ #--USER--End
+ #  If using Macports then use the following instead
+ #--USER--Start
+@@ -202,19 +202,19 @@ case $compiler in
+                echo 'SLIBS= ' >> makefile
+ #  If there are optimised maths libraries available replace the routines below
+ #--USER--Start
+-               echo 'MLIBS=lapack.o blas.o' >> makefile
++               echo 'MLIBS=-L${LD_LIBRARY_APTH} -lscalapack -lopenblas ' >> makefile
```
Micket, line 19: Typo "APTH", you can't use LD_LIBRARY_PATH for this since -L doesn't understand : lists, and it should be LIBRARY_PATH if anything but it would still not work. We have a better flag that doesn't hardcode "openblas".
`SCALAPACK_LIB_DIR`, `LIBSCALAPACK` , `LIBBLAS` should suffice here.

easybuild/easyconfigs/g/GULP/gulp-6.0-mkgulp.patch
```diff
@@ -0,0 +1,57 @@
...
++               echo 'RUNCC=mpif90 -fallow-argument-mismatch ' >> makefile
++#               echo 'RUNF90=mpif90 ' >> makefile
++#               echo 'RUNCC=mpif90 ' >> makefile
+ #  Provide the location and name of Scalapack, Blacs, Pblas, Lapack and Blas libraries.
+ #  On the Mac the following is sufficient when using Macports:
+ #--USER--Start
+-               echo 'SLIBS=-L/opt/local/lib -lscalapack ' >> makefile
+-               echo 'MLIBS= ' >> makefile
++               echo 'SLIBS=-L${LD_LIBRARY_PATH} -lscalapack ' >> makefile
++               echo 'MLIBS=-L${LD_LIBRARY_APTH} -lscalapack -lopenblas ' >> makefile
```
Micket, line 37: Same with these flags

easybuild/easyconfigs/g/GULP/gulp-6.0-mkgulp.patch
```diff
@@ -0,0 +1,57 @@
...
+            fi
+            ;;
+@@ -321,12 +321,13 @@ then
+ else
+   echo 'FDEFS=-DFFTW3 ' >> makefile
+   if [ "$parallel" = "" ]
++  then
+ #--USER--Start
+-    echo 'FLIBS=-L/usr/local/lib/ -lfftw3.a' >> makefile
++    echo 'FLIBS=-L${LD_LIBRARY_PATH} -lfftw3' >> makefile
```
Micket, line 48: Drop the LD_LIBRARY_PATH. Use `LIBFFT`

easybuild/easyconfigs/g/GULP/gulp-6.0-mkgulp.patch
```diff
@@ -0,0 +1,57 @@
...
+ #--USER--End
+-  then
++  else
+ #--USER--Start
+-    echo 'FLIBS=-L/usr/local/lib/ -lfftw3 -lfftw3_mpi' >> makefile
++    echo 'FLIBS=-L${LD_LIBRARY_PATH} -lfftw3 -lfftw3_mpi' >> makefile
```
Micket, line 54: and here as well

Micket (changes requested):

easybuild/easyconfigs/g/GULP/gulp-6.0-mkgulp.patch
sassy-crick: I tried to follow your suggestions. I removed the `LD_LIBRARY_PATH` etc but I was unable to follow what you suggested regarding using `LIBFFT` etc. 
I have added the author and description in the patch file and noticed the `foss-2020a` should not have worked as the patch file was wrong for `gcc-9.3.0`. So I added a 2nd patch file for that version. I also noticed a missing flag for `foss` so I added that one to have a parallel build. 
Let me know how to deal with the patch files if you are not happy with them.  :-)

easybuild/easyconfigs/g/GULP/gulp-6.0-fftw-blas-linking-gcc-9.3.patch
```diff
@@ -0,0 +1,52 @@
...
+diff --git a/gulp-6.0.org/Src/mkgulp b/gulp-6.0/Src/mkgulp
+index 28afb80..84549b0 100755
+--- a/gulp-6.0.org/Src/mkgulp
++++ b/gulp-6.0/Src/mkgulp
+@@ -170,7 +170,7 @@ fi
+ #  Include file directories
+ #
+ #--USER--Start
+-echo 'INC=-I.. -I/usr/local/include/' >> makefile
++echo 'INC=-I.. -I${EBROOTFFTW}/include -I/usr/local/include/' >> makefile
```
Micket, line 12: Remove `-I/usr/local/include` and switch to use the `FFTW_INC_DIR` instead

easybuild/easyconfigs/g/GULP/gulp-6.0-fftw-blas-linking-gcc-9.3.patch
```diff
@@ -0,0 +1,52 @@
...
+ #--USER--End
+            else
+ # NB: For gfortran v10 then uncomment these lines and comment the ones immediately below!
+@@ -213,8 +213,8 @@ case $compiler in
+ #  Provide the location and name of Scalapack, Blacs, Pblas, Lapack and Blas libraries.
+ #  On the Mac the following is sufficient when using Macports:
+ #--USER--Start
+-               echo 'SLIBS=-L/opt/local/lib -lscalapack ' >> makefile
+-               echo 'MLIBS= ' >> makefile
++               echo 'SLIBS=-L${EBROOTSCALAPACK}/lib -L${EBROOTOPENBLAS}/lib -lscalapack -lopenblas' >> makefile
```
Micket, line 31: It is still hardcoding -lscalapack and -lopenblas: you shouldn't. Also, don't use the EBROOT* environment variables for blas/lapack/fftw, instead use the *LIB_DIR variables that are defined by EB.
`-L${BLAS_LAPACK_LIB_DIR} -L${SCALAPACK_LIB_DIR} ${LIBSCALAPACK} ${LIBBLAS} `

easybuild/easyconfigs/g/GULP/gulp-6.0-fftw-blas-linking-gcc-9.3.patch
```diff
@@ -0,0 +1,52 @@
...
+            fi
+            ;;
+@@ -321,12 +321,13 @@ then
+ else
+   echo 'FDEFS=-DFFTW3 ' >> makefile
+   if [ "$parallel" = "" ]
++  then
+ #--USER--Start
+-    echo 'FLIBS=-L/usr/local/lib/ -lfftw3.a' >> makefile
++    echo 'FLIBS=-L${EBROOTFFTW}/lib -lfftw3' >> makefile
```
Micket, line 43: `-L${FFTW_LIB_DIR} ${LIBFFT}`

easybuild/easyconfigs/g/GULP/gulp-6.0-fftw-blas-linking-gcc-9.3.patch
```diff
@@ -0,0 +1,52 @@
...
+ #--USER--End
+-  then
++  else
+ #--USER--Start
+-    echo 'FLIBS=-L/usr/local/lib/ -lfftw3 -lfftw3_mpi' >> makefile
++    echo 'FLIBS=-L${EBROOTFFTW}/lib -lfftw3 -lfftw3_mpi' >> makefile
```
Micket, line 49: `-L${FFTW_LIB_DIR} ${LIBFFT}` but.. what of MPI? I mean you are building against a serial FFTW?

Micket (changes requested):

easybuild/easyconfigs/g/GULP/gulp-6.0-fftw-blas-linking-gcc-9.3.patch
sassy-crick: This is where I am stuck: In the patch file I have:
```diff
-echo 'INC=-I.. -I/usr/local/include/' >> makefile
+echo 'INC=-I.. -I${FFTW_INC_DIR}' >> makefile
```

Which results in the makefile to:
`INC=-I.. -I${FFTW_INC_DIR}`
which translates in the build to:
```
gfortran  -O3 -ffpe-summary=invalid,zero,overflow  -I.. -I    -DFLUSH -DNOFOX     -DFFTW3  -c ../datatypes.F90
f951: Warning: Nonexistent include directory -DFLUSH [-Wmissing-include-dirs]
```
and the build fails. I was already playing around with that which is why I fell back to the environment variables and the hard-coded lib-name. 
I perfectly agree, that is not desirable, but I either miss out something here, or that code does things differently. 
There is no `configure` involved here! You run the `mkgulp` file which is creating the `makefile` .

Micket: I can only assume that environment variables at all are expended in this context, since you were using them I assumed they were. Now I assume your previous `-I${EBROOTFFTW}/include` etc. just got expanded to `-I/include` and things just happened to work anyway because we set CPATH and LIBRARY_PATH which the compiler can see. If so, I would confirm if this is the case, and then investigate why all our (important) environments are dropped.

Flamefire:
I'd suggest to not include GULP-GCC and the FFTW-serial. Otherwise this will be a major pitfall for users:
Loading the GULP-GCC module and any module from the foss level will mean that 2 different FFTWs are loaded.

Similar in HMNS: GULP now exists at the GCC and the foss level. Not sure how this gets presented to the user, but I guess at least loading foss at some point will switch the GULP version.

sassy-crick:
@Flamefire The serial version of GULP is what most users in the community are using, so that is needed. I don't know if the serial code is different from the parallel one in the case of GULP. Either way, we need the serial version. The parallel one is just the 'freebee' so to speak.
I don't know anybody who will load GULP with anything else when running a calculation. So a clash here is more hypothetical. 
The switch to the golf toolchain means I get the variables I need to for the patchfiles. For the serial code we don't need MPI so using foss will be an overkill.
Besides, if somebody really needs to use the non-serial version of FFTW3, they still can load the parallel version of GULP and run it on a single core. So I don't see much, if any, problems here. 
Not everything is working better when run in parallel. Most GULP runs are actually rather short, compared to other Chemistry code. 
@Micket I have updated the patch-files as discussed. In order to have on file for the serial and parallel code, I switched from the GCC to the golf toolchain, so I get the variables I need.

Flamefire:
> The serial version of GULP is what most users in the community are using, so that is needed.
> Not everything is working better when run in parallel.

I am not proposing to force anyone to use GULP running in parallel. GULP needs FFTW and we have FFTW (only) at a level where OpenMPI happens to be there too. Including another FFTW in the hierarchy creates actual problems so unless the GULP build at the gompi or foss level has any real downside which cannot be avoided in other ways I would rather say the adding yet another FFTW is what is overkill. There is no real downside in reusing what we already have.

I see that you build the foss GULP with `./mkgulp -f -m -c gfortran` while the other is missing the `-m` flag. So I guess the `-m` enables the parallel version. So if users really only want the serial version, drop that flag and we are all good, aren't we?

We have other cases where we have a serial and parallel version of a software and we usually suffix the name there. So we would have `GULP/6.0-...` and `GULP/6.0-...-parallel`. This means users can load either the serial or the parallel version. With the current approach this is not really possible because in HMNS LMod will (IIRC) auto-switch from one to the other.

>  I don't know anybody who will load GULP with anything else when running a calculation. So a clash here is more hypothetical.

Fully disagree here. Users do what users do and that often is not what we expect. We usually take great care in EasyBuild to avoid such situations as much as possible and here I haven't seen a reason why we can't avoid it.
But I'll let the maintainers decide on that one.

sassy-crick:
> GULP needs FFTW and we have FFTW (only) at a level where OpenMPI happens to be there too

Not quite. There is still the `golf` toolchain for exactly that purpose. That is the reason why I switched from `GCC`over to `golf`
Your comment any my upload happened at more or less the same time.

Flamefire:
Ok I see. Got to admit that I'm not familiar with the golf toolchain and don't really see its purpose:
I've just rebuild and checked the GCC-FFTW and the gompi-FFTW and besides the `libfftw3_mpi*` libs they are virtually the same. So for "serial" builds one could just use the gompi-FFTW and not (explicitly) link against fftw-mpi as GULP already does without the `-m` build flag

My suggestion:
- Use a different name for serial and parallel GULP, e.g. GULP and GULP-parallel (you said the serial version is what people use by default) or at least use a version-suffix so users know what they get (IIRC in HMNS you don't see the toolchain and e.g. FFTW uses the same scheme)
- Maybe put GULP (serial) on foss level to avoid having to add the other 2 ECs as the golf toolchain seemingly isn't really used. Just to keep things simple.

Maybe @Micket or @boegel could help clear things up here especially regarding the golf toolchain. Also ccing @akesandgren as he moved PLINK from foss to golf in #11552 which is basically the only golf EC we have (the other is Python-3.6.4-golf-2018a.eb)
I'm not sure what the advantage is here.

@sassy-crick Please don't get my nagging wrong here. I think this is a prime example (serial+parallel version of the same software switchable via configure flag) to discuss how to do such things in general and in the future. So I want to make sure we choose the right path here so that future contributors have something as a reference. :)

sassy-crick:
Thanks for the useful discussion earlier on. 
As discussed: 
- I have added the `usempi: True` to both of the foss EasyConfig files 
- I have removed all the serial builds as we can use the MPI build of `gulp` in serial as well
- consequently I have also removed the `fftw` serial EC files
- updated both patch files accordingly by removing the `-lfftw3_mpi` entry as that is now provided by EasyBuild. 
I think that looks much better now. Thanks for the help!

easybuild/easyconfigs/g/GULP/GULP-6.0-foss-2020a.eb
```diff
@@ -0,0 +1,51 @@
...
+homepage = 'https://gulp.curtin.edu.au/gulp/'
+description = """GULP is a program for performing a variety of types of simulation on materials
+ using boundary conditions of 0-D (molecules and clusters), 1-D (polymers), 2-D (surfaces, slabs
+ and grain boundaries), or 3-D (periodic solids)Band Unfolding code for Plane-wave based calculations"""
+
+toolchain = {'name': 'foss', 'version': '2020a'}
+toolchainopts = {
+    'openmp': True,
+    'usempi': True,
+}
```
Flamefire, line 20: As far as I can tell openmp is not required here (this option makes it link e.g. fftw3_omp.so), at least I can't find anything in the gulp sources. Did I miss anything? If not please use the below, else put both on the same line (see the toolchain dict) for consistency with other ECs

```suggestion
toolchainopts = {'usempi': True}
```

easybuild/easyconfigs/g/GULP/GULP-6.0-foss-2020a.eb
```diff
@@ -0,0 +1,51 @@
...
+sources = [SOURCELOWER_TGZ]
+patches = ['gulp-6.0-fftw-blas-linking-gcc-9.3.patch']
+checksums = [
+    'b63cff96300356728ee641e7bac8f065a01f108a2289711c3d3d11d4050527f6',  # gulp-6.0.tgz
+    'c778df9d6b8d31594be81a45271387ed885753e54e4c5009ee0f55382aa6a4d6',  # gulp-6.0-fftw-blas-linking-gcc-9.3.patch
+]
+
+start_dir = 'Src'
+
+build_cmd = "./mkgulp -f -m -c gfortran"
```
Flamefire, line 32: As GULP isn't semi-closed source, it might make sense to include a short comment about the used options. IIRC `-c` is the compiler to use and `-m` is for parallel.
Not really required but I think it makes it easier for anyone checking on this later.

Flamefire (changes requested):
Looks good to me, just 2 minor things to polish.

For future reference: `-I'${FFTW_INC_DIR} `, `-L'${BLAS_LAPACK_LIB_DIR}` etc. is optional as those or in CPATH/LIBRARY_PATH already and could be removed. Shouldn't hurt to keep them though.

easybuild/easyconfigs/g/GULP/GULP-6.0-foss-2020a.eb
sassy-crick: I was actually wondering about the `openmp` as well but as my magic is not that advanced, I thought I leave it in there. This came from the `GULP-5.1-intel-2019a.eb` file which I used as a template. I have removed it.

easybuild/easyconfigs/g/GULP/GULP-6.0-foss-2020a.eb
sassy-crick: I have added a comment regarding the meaning of the various flags plus how to use the serial and parallel run for GULP.
This way, it should be possible for at least the person who installs it to see a pure serial build is not required.

Flamefire (approved):
LGTM!

Flamefire:
Test report by @Flamefire
**FAILED**
taurusa5 - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2603 v4 @ 1.70GHz (broadwell), Python 2.7.5
See https://gist.github.com/b4d97ae522a006dc2d45f0d00b9b5a0d for a full test report.

sassy-crick:
Why does your build fail with the wrong checksum? I get:
>$ sha256sum gulp-6.0.tgz
b63cff96300356728ee641e7bac8f065a01f108a2289711c3d3d11d4050527f6  gulp-6.0.tgz

I hope that is not one of these annoying checksum problems.

Flamefire:
Because I had the wrong file there. Rerunning...

Flamefire:
Test report by @Flamefire
**FAILED**
taurusi8017 - Linux centos linux 7.9.2009, x86_64, AMD EPYC 7352 24-Core Processor (zen2), Python 2.7.5
See https://gist.github.com/3a73b59db3f00872db60483582b65b17 for a full test report.

Flamefire:
Test report by @Flamefire
**FAILED**
taurusa5 - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2603 v4 @ 1.70GHz (broadwell), Python 2.7.5
See https://gist.github.com/bd18b8306bc58c4990bf56b3a6444a23 for a full test report.

Flamefire:
Test report by @Flamefire
**SUCCESS**
taurusi8017 - Linux centos linux 7.9.2009, x86_64, AMD EPYC 7352 24-Core Processor (zen2), Python 2.7.5
See https://gist.github.com/af6655573807039311ff87e66ee3c46d for a full test report.

Flamefire:
Test report by @Flamefire
**SUCCESS**
taurusa4 - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2603 v4 @ 1.70GHz (broadwell), Python 2.7.5
See https://gist.github.com/56ff2e5c9d15dfdd0ef73839ccf74981 for a full test report.

sassy-crick:
Just to catch up on that: it is all now ready to get merged, isn't it?

jfgrimm:
Test report by @jfgrimm
**FAILED**
himem06.pri.viking.alces.network - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) Platinum 8160 CPU @ 2.10GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/41542f68a66cc3e2e298a3db9fc9a6a8 for a full test report.

jfgrimm:
Test report by @jfgrimm
**FAILED**
himem06.pri.viking.alces.network - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) Platinum 8160 CPU @ 2.10GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/69fd3f016145f856389875ab6cd16e7d for a full test report.

jfgrimm:
@sassy-crick I'm getting a different checksum for gulp 6.0: `7977fdd427430008097edcd17852e0de7ff37877b9717aaea2110127744913b0`

sassy-crick:
We both are right. As that dates some time back, they done changes to the tarball. 
```
$ sha256sum gulp-6.0.tgz
b63cff96300356728ee641e7bac8f065a01f108a2289711c3d3d11d4050527f6  gulp-6.0.tgz

new:
$ sha256sum gulp-6.0.tgz
7977fdd427430008097edcd17852e0de7ff37877b9717aaea2110127744913b0  gulp-6.0.tgz
```
Attached is the diff-file between the two versions. Another case of how not to do things. :-( 

I will update the EC. Thanks for finding that, I did not bother to download it again as I had a copy of the source code already from way back. 

[diff.txt](https://github.com/easybuilders/easybuild-easyconfigs/files/9004387/diff.txt)

boegel:
GULP source tarballs keep changing, see also #15984 

Time to reach out to the developers and ask to step up their game and use [semver](https://semver.org/)?

sassy-crick:
I have updated the checksum to the source tarball I just have downloaded. 
See also [15984](https://github.com/easybuilders/easybuild-easyconfigs/issues/15984)

sassy-crick:
I am closing this PR as a newer, already merged version is out and even more newer versions are available on their website.

PR #13622, opened 2021-08-05
{tools}[foss/2020b] jax v0.2.19

deniskristak:
edit (by @boegel): requires ~https://github.com/easybuilders/easybuild-easyblocks/pull/2534~


easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
```diff
@@ -0,0 +1,63 @@
...
+     ],
++    patch_cmds=['sed -i -e "s|arguments = \[args\],|arguments = \[args\],\\n\\tuse_default_shell_env = True,|" third_party/mlir/tblgen.bzl']
+ )
+ 
+-# For development, one can use a local TF repository instead.
++# # For development, one can use a local TF repository instead.
+ # local_repository(
+ #    name = "org_tensorflow",
+-#    path = "tensorflow",
++#    path = "/apps/gent/CO7/skylake-ib/software/TensorFlow/2.5.0-foss-2020b",
```
boegel, line 54: This stuff is commented out, so has no effect at all?

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
```diff
@@ -0,0 +1,63 @@
+diff -ruN jax_orig/build/build.py jax/build/build.py
```
boegel, line 1: please add a short patch description on top

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,35 @@
...
+
+preinstallopts = 'python3 build/build.py --output_user_root=%(builddir)s --bazel_path="$EBROOTBAZEL/bin/bazel" && '
+preinstallopts += 'pip install dist/*.whl && '
+
+dependencies = [
+    ('Python', '3.8.6'),
+    ('Bazel', '3.7.2'),
+    ('SciPy-bundle', '2020.11'),
+    ('Cython', '0.29.22'),
+    # ('TensorFlow', '2.5.0'), do we need this?
```
boegel, line 29: Based on the feedback in https://github.com/google/jax/discussions/7467, including TensorFlow as a dependency doesn't make sense.

When building `jax` from source, I think we should try and control which TensorFlow sources are used, by pre-downloading them and making sure that Bazel picks them up (rather than downloading stuff itself).

It seems like that's doable by:

* adding the TensorFlow 2.5.0 tarball as an additional source
* patching `jax/WORKSPACE` so it uses that rather than the TensorFlow commit that is specified now; I guess you could try to replace the `https://` URL with a `file://` URL?

You'll need to do the latter at runtime (with `sed`, not a patch file), since the location where the pre-downloaded TensorFlow source tarball is located is not known up front.

Through @Flamefire ?

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,35 @@
...
+# sources = [SOURCE_TAR_GZ]
+
+source_urls = ['https://github.com/google/jax/archive/']
+sources = ['%(name)s-v%(version)s.tar.gz']
+patches = [
+    'jax-0.2.18-correct_libraries.patch',
+]
+
+preinstallopts = 'python3 build/build.py --output_user_root=%(builddir)s --bazel_path="$EBROOTBAZEL/bin/bazel" && '
+preinstallopts += 'pip install dist/*.whl && '
```
boegel, line 22: @deniskristak This takes part of building `jaxlib` from source, and hence there's no need to include it as a dependency/component, is that right?

cc @sassy-crick

boegel (changes requested):

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
Flamefire: Correct

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
Flamefire: I guess you can basically use the local_repository, see https://github.com/easybuilders/easybuild-easyconfigs/pull/13622/files#r683278973 for that.
This would at least allow to use a patch for the TF sources as required (the default_shell_env stuff) and only need the sed for the path (the patch can insert a nice marker where the sed can apply)

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
```diff
@@ -0,0 +1,63 @@
...
+--- jax_orig/build/build.py	2021-07-30 10:18:43.599233082 +0200
++++ jax/build/build.py	2021-08-02 15:58:09.654165000 +0200
+@@ -41,7 +41,7 @@
+ else:
+   from distutils.spawn import find_executable as which
+ # pylint: enable=g-import-not-at-top
+-
++use_default_shell_env = True
```
Flamefire, line 9: Not required here, this is for Bazel files only, isn't it?

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
deniskristak: yep, just keeping it here in case we need it

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
deniskristak: true

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
deniskristak: removing, it's in git's history now

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
deniskristak: done

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: please elaborate, I dont understand :D

Flamefire: We don't need an extra dependency in EC on jaxlib as this EC installs jaxlib AND jax. That was @boegel question

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,33 @@
...
+preinstallopts = 'python3 build/build.py '
+preinstallopts += '--bazel_startup_options="--output_user_root=%(builddir)s" '
+preinstallopts += ' --bazel_path="$EBROOTBAZEL/bin/bazel" && '
+preinstallopts += 'pip install dist/*.whl && '
+
+dependencies = [
+    ('Python', '3.8.6'),
+    ('Bazel', '3.7.2'),
+    ('SciPy-bundle', '2020.11'),
+    ('Cython', '0.29.22'),
```
Flamefire, line 27: Can you fix the deps please?
- Bazel is a build dep
- Is Cython really required?
- This whole thing needs to be a PythonBundle as absl-py and opt-einsum are required
- flatbuffers-python is also missing

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,33 @@
...
+
+toolchain = {'name': 'foss', 'version': '2020b'}
+
+source_urls = ['https://github.com/google/jax/archive/']
+sources = ['%(name)s-v%(version)s.tar.gz']
+patches = [
+    'jax-0.2.18-correct_libraries.patch',
+]
+
+preinstallopts = 'python3 build/build.py '
```
Flamefire, line 18: Minor:
```suggestion
preinstallopts = 'python build/build.py '
```

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,33 @@
...
+preinstallopts += 'pip install dist/*.whl && '
+
+dependencies = [
+    ('Python', '3.8.6'),
+    ('Bazel', '3.7.2'),
+    ('SciPy-bundle', '2020.11'),
+    ('Cython', '0.29.22'),
+]
+
+use_pip = True
```
Flamefire, line 30:
```suggestion
use_pip = True
download_dep_fail = True
sanity_pip_check = True
```

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,33 @@
...
+
+source_urls = ['https://github.com/google/jax/archive/']
+sources = ['%(name)s-v%(version)s.tar.gz']
+patches = [
+    'jax-0.2.18-correct_libraries.patch',
+]
+
+preinstallopts = 'python3 build/build.py '
+preinstallopts += '--bazel_startup_options="--output_user_root=%(builddir)s" '
+preinstallopts += ' --bazel_path="$EBROOTBAZEL/bin/bazel" && '
```
Flamefire, line 20: Bazel is in PATH and already checked. So maybe this instead:
```suggestion
preinstallopts += '--bazel_options="--subcommands --jobs=%(parallel)s" && '
```

Flamefire:
Need another 2 changes in build.py:
- `shell([bazel_path, "shutdown"])` needs to be removed or changed to `shell([bazel_path] + args.bazel_startup_options + ["shutdown"])`
- in the function check_bazel-version change to use --version. See https://github.com/google/jax/issues/7507

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,33 @@
...
+preinstallopts += 'pip install dist/*.whl && '
```
Flamefire, line 21:
```suggestion
preinstallopts += 'pip install --prefix %(installdir)s dist/*.whl && '
```

Flamefire: BTW: @boegel I think it makes sense to enhance PythonPackage to be able to specify the location for pip. Because we add quite some useful stuff to the pip command such as `--no-deps --ignore-installed` etc. which is lost here. But we can't use PythonPackage to install this whl file yet

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: so no changes on this line required, right? :D

Flamefire: Basically yes. Although I'd like to see jaxlib being separate (but in the same PythonBundle EC) I'll open a PR for PythonPackage to make that possible

Flamefire: PythonPackage PR: https://github.com/easybuilders/easybuild-easyblocks/pull/2534
@boegel I think this would be great, especially as we can easily test this with this PR once adapted

@deniskristak Once you converted this into a PythonBundle (see https://github.com/easybuilders/easybuild-easyconfigs/pull/13622#discussion_r683396454 and the TF EC for an example and suggested versions) I can help you get this up with the new change if the description is not enough

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
```diff
@@ -0,0 +1,82 @@
...
+-  try:
+-    version_output = shell([bazel_path, "--bazelrc=/dev/null", "version"])
+-  except subprocess.CalledProcessError:
+-    return False
+-  match = re.search("Build label: *([0-9\\.]+)[^0-9\\.]", version_output)
+-  if match is None:
+-    return False
+-  actual_ints = [int(x) for x in match.group(1).split(".")]
+-  return actual_ints >= [2, 0, 0]
++    # try:
```
Flamefire, line 18: Why is this commented? the commented code is actually correct

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: done

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: done

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
Flamefire: Yeah with that you need to use `actual_ints = [int(x) for x in match.group(2).split(".")]` as you introduced a new group
However what about simply using the 2 commits/changes I linked above?

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,75 @@
...
+        'source_tmpl': 'opt_einsum-%(version)s.tar.gz',
+        'modulename': 'opt_einsum',
+    }),
+    ('absl-py', '0.13.0', {
+        'modulename': 'absl'
+    })]
+
+components = [
+    (name, version, {
+        'easyblock': 'PythonPackage',
```
Flamefire, line 51: I'd suggest to make an extra component for jaxlib:
- Also use PythonPackage (basically C&P what you have here but adjust name and version [jaxlib has a different versioning])
- put the stuff from preinstallopts into buildcmd
- use `install_src = 'dist/*.whl'`

See https://github.com/easybuilders/easybuild-easyblocks/pull/2534 which then needs to be used.
The pip install command is actually a bit more involved and it makes sense to install jax and jaxlib separately as they are 2 packages which both would need sanity checks etc.

The jax part can then be trivial: Download from PyPI (default -> no source_urls) with IIRC default SOURCE_TAR_GZ template, no patches, no preinstallopts

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
Flamefire: Or better: Use jaxlib 0.1.70 which already includes that change and does not require this patch :)

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
boegel: @Flamefire Can you clarify why you prefer separate components for `jaxlib` and `jax`?

Apparently both come from the same repository, but have different versioning schemes...

So if we split them up, rather than just using the `jaxlib` subdirectory in the `jax` release, we may be installing a combination that nobody else uses? Or am I missing something?

Also, why haven't I come up with "use one big repository for different tools, release with different versioning schemes" in my funny talk? It certainly ticks off the "cause confusion" requirement...

branfosj: I'm for separate, known versions of `jaxlib` and `jax` - so that we are using released versions of both - rather than whatever state `jaxlib` was in when for the version of `jax` we are using.

I also think we should be naming `jax-0.2.18-foss-2020b-jaxlib-0.1.70.eb`, or something similar, that makes it clear what versions of each we are installing.

For the combination we install - that should be versions of both released at about the same time - unless there is a known broken combination.

Flamefire: >I'm for separate, known versions of jaxlib and jax - so that we are using released versions of both - rather than whatever state jaxlib was in when for the version of jax we are using.

Exactly. Also simply for the sanity checks. jaxlib should succeed without jax installed

And no I wouldn't suffix that. Same as with e.g. TF which as an absl-py as an extension and mentions that in the module file.

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,87 @@
...
+        'source_urls': [
+            'https://github.com/google/jax/archive/',
+            'https://github.com/tensorflow/tensorflow/archive/'
+        ],
+        'patches': [
+            ('jaxlib-%s-correct_libraries.patch' % local_jaxlib_ver, 1),
+            ('jaxlib-%s-controlled_tf_tarball.patch' % local_jaxlib_ver, '../tensorflow-%s' % local_tf_tarball_version),
+        ],
+        'start_dir': 'jax-jaxlib-v%s' % local_jaxlib_ver,
+        'preinstallopts': local_jax_preinstallopts,
```
Flamefire, line 75: This should rather be prebuildopts shouldn't it?

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,87 @@
...
+local_jax_build_cmd += '--bazel_options=--jobs=1 --bazel_options=--action_env=PYTHONPATH '
+local_jax_build_cmd += '--bazel_options=--action_env=EBPYTHONPREFIXES'
+
+
+exts_list = [
+    ('opt-einsum', '3.3.0', {
+        'source_tmpl': 'opt_einsum-%(version)s.tar.gz',
+        'modulename': 'opt_einsum',
+    }),
+    ('absl-py', '0.13.0', {
```
Flamefire, line 47: jaxlib requires absl-py, so this needs to be in components before jaxlib

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,87 @@
...
+        ],
+        'start_dir': 'jax-jaxlib-v%s' % local_jaxlib_ver,
+        'preinstallopts': local_jax_preinstallopts,
+        'use_pip': True,
+        'sanity_pip_check': True,
+        'download_dep_fail': True,
+        'buildcmd': local_jax_build_cmd,
+        'install_src': 'dist/*.whl',
+    }),
+    ('jax', '0.2.18', {
```
Flamefire, line 82: I'd suggest moving that to exts_list, then you can keep opt-einsum there

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
boegel: We should only have a versionsuffix when we diverge for whatever reason. The default should be latest `jaxlib` with latest `jax`. If we diverge from that, we can discriminate with a versionsuffix?

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
boegel: Isn't it better to have this as a proper dependency then?

Flamefire: I wouldn't mind having absl-py and opt-einsum as ECs although it does slow down the loading of python imports (due to more paths needing to be checked). However then I'd strongly suggest to use the same versions as used in the TF ECs and reuse the ECs there instead of the exts

Having said that: The slow-down due to many PYTHONPATH entries is an issue. So there is a reason to keep such bundles. But as mentioned: It needs to be installed before jaxlib.
Best would of course be if exts_list could be used for jaxlib. @boegel Can you open an issue in framework what needs to be done so we can support multiple sources there and anything that comes to mind why it may not work?

boegel:
@deniskristak https://github.com/easybuilders/easybuild-easyblocks/pull/2534 merged

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
deniskristak: yeah, each time it generates it gets deleted :D ill add it when it's ready

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: yes, done

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: moved `absl-py` to components, moved `jax` to exts

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: done

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
deniskristak: done, using 0.1.70

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: this should be resolved

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,90 @@
...
+
+builddependencies = [
+    ('Bazel', '3.7.2'),
+]
+
+dependencies = [
+    ('Python', '3.8.6'),
+    ('SciPy-bundle', '2020.11'),
+    ('Cython', '0.29.22'),
+    ('opt-einsum', '3.3.0'),
```
boegel, line 25: @deniskristak The easyconfig for `opt-einsum` is missing?

I guess this should be removed, since you have `opt-einsum` included in `exts_list` below...

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,90 @@
...
+    ('flatbuffers-python', '1.12'),
+]
+
+local_tf_builddir = "%%(builddir)s/tensorflow-%s" % local_tf_tarball_version
+
+local_jax_prebuildopts = "sed -i -e 's$pathToSed$%s$g' WORKSPACE && " % local_tf_builddir
+
+# local_jax_build_cmd = "sed -i -e 's$pathToSed$%s$g' WORKSPACE && " % local_tf_builddir
```
boegel, line 33: Please remove this commented out line?

boegel (changes requested):

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: done

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
deniskristak: done

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
```diff
@@ -0,0 +1,88 @@
...
+
+local_tf_builddir = "%%(builddir)s/tensorflow-%s" % local_tf_tarball_version
+
+local_jax_prebuildopts = "sed -i -e 's$pathToSed$%s$g' WORKSPACE && " % local_tf_builddir
+
+local_jax_build_cmd = 'python build/build.py '
+local_jax_build_cmd += '--bazel_startup_options="--output_user_root=%(builddir)s" '
+local_jax_build_cmd += '--bazel_path="$EBROOTBAZEL/bin/bazel" '
+local_jax_build_cmd += '--bazel_options=--subcommands '
+local_jax_build_cmd += '--bazel_options=--jobs=1 --bazel_options=--action_env=PYTHONPATH '
```
boegel, line 36: It would be really nice if we could use `--jobs=%(parallel)s` here to avoid building `jax` with just a single core, but it seems like this may be breaking the build...

Here's the build error I was hitting when using `--jobs=%(parallel)s`:

```
INFO: Build completed successfully, 5551 total actions
INFO: Running command line: bazel-bin/build/build_wheel '--output_path=/tmp/vsc40023/easybuild_build/jax/0.2.18/foss-2020b/jax-jaxlib-v0.1.70/dist' '--cpu=x86_64'
INFO: Build completed successfully, 5551 total actions
  File "/tmp/vsc40023/easybuild_build/jax/0.2.18/foss-2020b/134f18b726e1fd90f125772c3f16cba2/execroot/__main__/bazel-out/k8-opt/bin/build/build_wheel.runfiles/__main__/build/build_wheel.py", line 252, in <module>
    prepare_wheel(sources_path)
  File "/tmp/vsc40023/easybuild_build/jax/0.2.18/foss-2020b/134f18b726e1fd90f125772c3f16cba2/execroot/__main__/bazel-out/k8-opt/bin/build/build_wheel.runfiles/__main__/build/build_wheel.py", line 210, in prepare_wheel
    patch_copy_xla_extension_stubs(jaxlib_dir)
  File "/tmp/vsc40023/easybuild_build/jax/0.2.18/foss-2020b/134f18b726e1fd90f125772c3f16cba2/execroot/__main__/bazel-out/k8-opt/bin/build/build_wheel.runfiles/__main__/build/build_wheel.py", line 106, in patch_copy_xla_extension_stubs
    with open(r.Rlocation(
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I'm testing again now (with the changes from https://github.com/deniskristak/easybuild-easyconfigs/pull/3 included) to confirm that problem is indeed due to the parallel build. If that's the case, I'll open an issue in the `jax` GitHub repo to ask whether building in parallel is supposed to be working...

boegel (changes requested):
@deniskristak The name of the patch files as they are included in the PR are incorrect.

I fixed that in https://github.com/deniskristak/easybuild-easyconfigs/pull/3, along with a couple of other things, like including checksums.

easybuild/easyconfigs/j/jax/jax-0.2.18-foss-2020b.eb
boegel: fixed in 1210ab0, helps a lot with speeding up the build :)

easybuild/easyconfigs/j/jax/jax-0.2.18-correct_libraries.patch
boegel: fixed in 54fb8efadb

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 3 (3 easyconfigs in total)
node3111.skitty.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/8ece804dca9e7ea990fb2dbfa0714095 for a full test report.

boegel:
CORE_CNT=16

boegel:
Test report by @boegel
**SUCCESS**
node3111.skitty.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/5891da8c2282beb78f56f429cf1e49f6 for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3525.doduo.os - Linux RHEL 8.2, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/dbe92c04b4452fa1055631a989bc9a0a for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node2607.swalot.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/ef3d0914628552932371ffaca619f0a2 for a full test report.

boegel (approved):

boegel:
This can/should probably be cleaned up by implementing a custom easyblock for `jaxlib`.

I'll merge this as is though, since it's working, and a major hurdle towards AlphaFold, and keep that as an exercise for an upcoming PR...

boegel (approved):
lgtm

boegel:
Going in, thanks @deniskristak!

Flamefire:
Test report by @Flamefire
**FAILED**
Build succeeded for 2 out of 3 (3 easyconfigs in total)
taurusi8002 - Linux centos linux 7.9.2009, x86_64, AMD EPYC 7352 24-Core Processor (zen2), Python 2.7.5
See https://gist.github.com/de777494f99609e7c0c32a739adaf447 for a full test report.

Flamefire:
Test report by @Flamefire
**FAILED**
Build succeeded for 2 out of 3 (3 easyconfigs in total)
taurusi6606.taurus.hrsk.tu-dresden.de - Linux RHEL 7.9, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz (haswell), Python 2.7.5
See https://gist.github.com/13f062a38935f5541d68de71b5113a15 for a full test report.

PR #13753, opened 2021-08-16
update jax easyconfig to use custom easyblock for jaxlib

deniskristak:
must be used with https://github.com/easybuilders/easybuild-easyblocks/pull/2542

### DIFF
--- a/easybuild/easyconfigs/j/jax/jax-0.2.19-foss-2020b.eb
+++ b/easybuild/easyconfigs/j/jax/jax-0.2.19-foss-2020b.eb
@@ -29,16 +29,7 @@ dependencies = [
 local_tf_commit = '4039feeb743bc42cd0a3d8146ce63fc05d23eb8d'
 
 local_tf_builddir = "%%(builddir)s/tensorflow-%s" % local_tf_commit
-
 local_jax_prebuildopts = "sed -i -e 's$pathToSed$%s$g' WORKSPACE && " % local_tf_builddir
-
-local_jax_build_cmd = 'python build/build.py '
-local_jax_build_cmd += '--target_cpu_features=native '
-local_jax_build_cmd += '--bazel_startup_options="--output_user_root=%(builddir)s" '
-local_jax_build_cmd += '--bazel_path="$EBROOTBAZEL/bin/bazel" '
-local_jax_build_cmd += '--bazel_options=--subcommands '
-local_jax_build_cmd += '--bazel_options=--jobs=%(parallel)s --bazel_options=--action_env=PYTHONPATH '
-local_jax_build_cmd += '--bazel_options=--action_env=EBPYTHONPREFIXES'
 
 use_pip = True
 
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/j/jax/jax-0.2.19-foss-2020b.eb
```diff
@@ -56,7 +47,7 @@ components = [
         'use_pip': True,
     }),
     ('jaxlib', local_jaxlib_ver, {
-        'easyblock': 'PythonPackage',
+        'easyblock': 'EB_jaxlib',
```
boegel, line 50: When https://github.com/easybuilders/easybuild-easyblocks/pull/2543 is merged, this can be removed too.

easybuild/easyconfigs/j/jax/jax-0.2.19-foss-2020b.eb
```diff
@@ -84,10 +75,8 @@ components = [
         ],
         'start_dir': 'jax-jaxlib-v%s' % local_jaxlib_ver,
         'prebuildopts': local_jax_prebuildopts,
-        'use_pip': True,
         'sanity_pip_check': True,
```
boegel, line 78: This is futile, there's no sanity check for each bundle component individually (like there is for extensions).

easybuild/easyconfigs/j/jax/jax-0.2.19-foss-2020b.eb
```diff
@@ -84,10 +75,8 @@ components = [
...
         'download_dep_fail': True,
-        'buildcmd': local_jax_build_cmd,
         'install_src': 'dist/*.whl',
```
boegel, line 80: This should be stripped out, it's handled by the custom easyblock for `jaxlib`.

easybuild/easyconfigs/j/jax/jax-0.2.19-foss-2020b.eb
```diff
@@ -101,13 +90,6 @@ exts_list = [
     }),
 ]
 
-# can't run tests in sanity check step, because build dependencies must be available
-postinstallcmds = [
-    "cd %%(builddir)s/jax-jaxlib-v%s && "
-    "PYTHONPATH=%%(installdir)s/lib/python%%(pyshortver)s/site-packages:$PYTHONPATH "
-    "pytest -n auto tests" % local_jaxlib_ver,
```
boegel, line 108: We still need to run the tests through this easyconfig, can't be done in the `jaxlib` easyblock since `jax` is also required.

I think @Flamefire is looking into a better way of doing this, via `runtest` for `jax`?

boegel (changes requested):

Flamefire:
Updated version at https://github.com/easybuilders/easybuild-easyconfigs/pull/13760

boegel:
Let's proceed with #13760, which is inspired by this PR, thanks for the effort @deniskristak!

PR #13767, opened 2021-08-18
{cae}[foss/2021a] OpenFOAM v9

boegel:
(created using `eb --new-pr`)

requires https://github.com/easybuilders/easybuild-easyblocks/pull/2544 + ~#13266~ (CGAL)

PR #13770, opened 2021-08-18
make the testsuite print error and still return a non-zero exitcode

branfosj:
see https://github.com/easybuilders/easybuild-easyconfigs/pull/13767#discussion_r691452271

If the tests fail then the `|| echo "ERROR: Not all tests were successful"` means that the non-zero exit code is taken since the `echo` succeeds.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -121,7 +121,8 @@ jobs:
           eb --show-system-info
 
           # run test suite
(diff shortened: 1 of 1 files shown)


Micket (approved):
lgtm.. how can this not have been caught earlier?

boegel:
> lgtm.. how can this not have been caught earlier?

@Micket This was introduced very recently, see #13745

boegel (approved):
lgtm

PR #13950, opened 2021-09-09
Have GROMACS log files indicate EasyBuild was used

mabraham:
Knowing that EasyBuild has patched the code and organized the build is potentially valuable information for GROMACS users and developers troubleshooting their runs.

### DIFF
--- a/easybuild/easyconfigs/g/GROMACS/GROMACS-2021.3-foss-2021a.eb
+++ b/easybuild/easyconfigs/g/GROMACS/GROMACS-2021.3-foss-2021a.eb
@@ -11,6 +11,7 @@
 # * Ake Sandgren <ake.sandgren@hpc2n.umu.se>
 # * J. Sassmannshausen <Crick HPC team>
 # * Dugan Witherick <dugan.witherick@warwick.ac.uk>
+# * Mark Abraham <mark.j.abraham@gmail.com>
 # License::   MIT/GPL
 
 name = 'GROMACS'
@@ -53,6 +54,10 @@ checksums = [
     # GROMACS-2020.5_fix_threads_gpu_Gmxapitests.patch
     '89fbb7e2754de45573632c74f53563bb979df9758c949238a35865391d6b53fb',
 ]
+
+# Ensure that the GROMACS log files report how the code was patched
+# during the build, so that any problems are easier to diagnose.
+configopts = '-DGMX_VERSION_STRING_OF_FORK=EasyBuild'
 
 builddependencies = [
     ('CMake', '3.20.1'),


mabraham:
This capability goes back a few GROMACS versions. I'm new to EasyBuild - is it OK if I update other .eb files similarly?

boegel:
@mabraham Thanks a lot for the contribution!

This makes total sense, but it would be better to handle this in the `configure_step` method of the [custom easyblock we have from GROMACS](https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/g/gromacs.py), to ensure that this is set for every GROMACS installation done with EasyBuild (see [here](https://easybuilders.github.io/easybuild-tutorial/2021-isc21/adding_support_additional_software/#easyconfigs-vs-easyblocks) for more info on easyconfigs vs easyblocks).

Added benefit is that the value can be more fine-grained than just `'EasyBuild'`, it can also include the EasyBuild version, using something like this (untested, but should work as expected):

```python
from easybuild.tools.version import VERBOSE_VERSION
...
def configure_step(self):
    ...
    if '-DGMX_VERSION_STRING_OF_FORK=' not in self.cfg['configopts']:
        self.cfg.update('configopts', '-DGMX_VERSION_STRING_OF_FORK=%s' % VERBOSE_VERSION)
```

An additional guard with a version check can be used in case this is only supported for sufficiently recent GROMACS versions.

I noticed in https://github.com/spack/spack/pull/25869 that it makes sense to also indicate whether PLUMED was used, this can also be done quite easily in the GROMACS easyblock (see [where we run `plumed-patch`](https://github.com/easybuilders/easybuild-easyblocks/blob/e9c2ea266874d39d59d4ecfffd279543216769f0/easybuild/easyblocks/g/gromacs.py#L197-L209))

Does this make sense?
If so, are you up for contributing this yourself to the GROMACS easyblock?
If that feels like too much, we're happy to tackle it ourselves...

Additionally: I can't help but wondering if there was a particular situation that came up where you wish that it was clear from the start that you were dealing with a GROMACS installation that was done using EasyBuild?

mabraham:
Thanks! I suspected that there was an easier way to do it generally, but didn't find the right docs or repo. I'll do that tomorrow.

There's been no incident specific to EasyBuild, but we have had confusion over log files produced with PLUMED looking like normal ones. We have had a feature to see if there'd been code changes without this CMake variable set and to generically mark the version as modified, but that proved hard to make work well. It also views the patches that packaging projects make as modifications, which is not always a good thing. However, if we've got packaging projects helping us out with acknowledging their involvement, then the case for keeping that feature is reduced, while still giving users and devs accurate information about what version is being used.

boegel:
I fully support "watermarking" the GROMACS installations produced with EasyBuild if that makes things easier upstream. :+1:

Do let us know if you need any help with tackling this through the GROMACS easyblock!
A good way to get in touch more interactive is through Slack, see https://easybuild.io/join-slack .

mabraham:
Superseded by https://github.com/easybuilders/easybuild-easyblocks/pull/2571

PR #13384, opened 2021-07-08
remove superflous -DCMAKE_BUILD_TYPE=Release, use of 'build_type = Release', or enabling separate_build_dir from easyconfigs using CMakeMake easyblock

Flamefire:
This is the default for a few releases now and may lead to confusion or bugs.

Examples:
- `RELEASE` used instead of `Release` -> confusing
- `ON` used in 1 place -> bug

Bonus: CMakeMake reacts on toolchainopts.debug to set the build type correctly which works now after this change

Dependend on:

- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13613
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13614
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13615
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13616
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13617
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13619
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13620
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13621
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13623
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13624
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13456
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/13434

PR #14008, opened 2021-09-16
Add CI check for CMAKE_BUILD_TYPE

Flamefire:
Checks that `-DCMAKE_BUILD_TYPE` is no longer used in configopts, see https://github.com/easybuilders/easybuild-easyconfigs/pull/13384
and that `Release` is not used as the `build_type` as it is already the default

Also a change to only parse changed ECs when running the PR tests only instead of all ECs (which takes ages). This leverages the caching behavior of EasyConfigTest._parsed_easyconfigs so when later parsing all ECs the already parsed ones are not reparsed.
This makes running the PR tests only very fast


Flamefire:
Rebased onto develop resolving the conflicts.

test/easyconfigs/easyconfigs.py
```diff
@@ -1153,6 +1147,23 @@ def check_https_url(http_url):
...
+        for ec in self.changed_ecs:
+            ec_fn = os.path.basename(ec.path)
+            configopts = ec.get('configopts')
+            build_type = ec.get('build_type')
+
+            if configopts and '-DCMAKE_BUILD_TYPE' in configopts:
+                failing_checks.append("Found -DCMAKE_BUILD_TYPE in configopts. Use build_type instead: %s" % ec_fn)
+            if build_type == 'Release':
+                failing_checks.append("build_type was set to the default of 'Release'. "
+                                      "Ommit this to base it on toolchain_opts.debug: %s" % ec_fn)
```
branfosj, line 1163:
```suggestion
                                      "Omit this to base it on toolchain_opts.debug: %s" % ec_fn)
```

branfosj (approved):

PR #14046, opened 2021-09-21
allow old ITK

deniskristak:
added old ITK to exceptions for one-version-per-toolchain-gen tests


boegel:
@deniskristak It's easier to just include this commit in #14023, since that's where it's actually required...

deniskristak:
fair enough, closing

PR #14083, opened 2021-10-01
add path to Intel libraries to $LIBRARY_PATH for iccifort 2020.1.217

jordiblasco:
When compiling HDF5 with Intel toolchain, it reports the following error in the "configure" stage:
```
configure: error: Fortran compiler requires either intrinsic functions SIZEOF or STORAGE_SIZE
```
This issue can be fixed by exposing the libraries available in ```compilers_and_libraries_%(version)s/linux/compiler/lib/intel64_lin```

### DIFF
--- a/easybuild/easyconfigs/i/iccifort/iccifort-2020.1.217.eb
+++ b/easybuild/easyconfigs/i/iccifort/iccifort-2020.1.217.eb
@@ -35,6 +35,8 @@ components = [
 
 dontcreateinstalldir = True
 
+modextrapaths  = {'LIBRARY_PATH': 'compilers_and_libraries_%(version)s/linux/compiler/lib/intel64_lin'}
+
 license_file = HOME + '/licenses/intel/license.lic'
 
 moduleclass = 'compiler'
--- a/easybuild/easyconfigs/i/iccifort/iccifort-2020.1.217.eb
+++ b/easybuild/easyconfigs/i/iccifort/iccifort-2020.1.217.eb
@@ -35,7 +35,7 @@ components = [
 
 dontcreateinstalldir = True
 
-modextrapaths  = {'LIBRARY_PATH': 'compilers_and_libraries_%(version)s/linux/compiler/lib/intel64_lin'}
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/i/iccifort/iccifort-2020.1.217.eb
```diff
@@ -35,6 +35,8 @@ components = [
 
 dontcreateinstalldir = True
 
+modextrapaths  = {'LIBRARY_PATH': 'compilers_and_libraries_%(version)s/linux/compiler/lib/intel64_lin'}
```
migueldiascosta, line 38: `iccifort-2020.1.217.eb:38:14: E221 multiple spaces before operator` (two spaces before `=`)

akesandgren:
Exactly which HDF5 where you trying to build? The ones we have available already for iimpi toolchains haven't shown any problems.

jordiblasco:
> Exactly which HDF5 where you trying to build? The ones we have available already for iimpi toolchains haven't shown any problems.

I'm working on a new recipe for PSI4, which requires the intel toolchain. HDF5 is one of the dependencies.

akesandgren:
So I repeat my question, exactly which HDF5 did you have problems with?

akesandgren:
Note also that this is the wrong place to do this if it is really needed.

akesandgren:
For some background,  our HDF5-1.10.6-iimpi-2020a.eb shows this in the configure output:
```
checking if Fortran compiler supports intrinsic SIZEOF... yes
checking if Fortran compiler supports intrinsic C_SIZEOF... yes
checking if Fortran compiler supports intrinsic STORAGE_SIZE... yes
checking if Fortran compiler supports intrinsic module ISO_FORTRAN_ENV... yes
```
and so does HDF5-1.10.7-iimpi-2020b.eb

So to be able to investigate this properly we need to know exactly which HDF5 easyconfig you where building when you saw that problem.

boegel:
closing this since `GCCcore/9.30` is deprecated, see https://docs.easybuild.io/policies/toolchains

PR #14131, opened 2021-10-13
{data}[GCC/10.2.0] fdict v0.8.0

alinelena:


ocaisa:
Out of curiosity I tried this with the intel compiler (see details of the last test report). This does indeed use the Intel compiler and flags (which  is good):
```
ifort -E -P -x c  -I./src -I./src -I. ./src/variable_pp.F90 | sed -f ./src/filter.sed > variable.f90
ifort: command line warning #10155: ignoring option '-x'; argument required
ifort: error #10236: File not found:  'c'
ifort -c -o variable.o -O2 -xHost -ftz -fp-speculation=safe -fp-model source  variable.f90
VPATH="." ./src/dictionary.sh
ifort -E -P -x c  -I./src -I./src -I. ./src/dictionary_pp.F90 | sed -f ./src/filter.sed > dictionary.f90
ifort: command line warning #10155: ignoring option '-x'; argument required
ifort: error #10236: File not found:  'c'
ifort -c -o dictionary.o -O2 -xHost -ftz -fp-speculation=safe -fp-model source  dictionary.f90
```
This still installs correctly but it is clear that the preprocessing step is returning an error. @alinelena You might want to raise that with @zerothi if you decide to include an intel version of this.

PR #14133, opened 2021-10-13
{math}[foss/2020a,intel/2020a] ELSI v2.6.4, NTPoly v2.5.1

micaeljtoliveira:


ocaisa:
I see in https://github.com/hpcuantwerpen/UAntwerpen-easyconfigs/blob/master/e/ELSI/ELSI-2.5.0-intel-2020a.eb#L53 that they set specific flags for BSEPACK but I guess these are probably already the defaults.

ocaisa:
@micaeljtoliveira Do you have an updated easyblock for this? I imagine it may need to be extended for the new external `NTPoly`?

micaeljtoliveira:
@ocaisa Is there anything missing here? AFAIKS, all tests are passing for this PR and for the corresponding easyblock.

PR #14150, opened 2021-10-14
{bio}[GCCcore/10.3.0] BGEN v1.1.7

sassy-crick:
(created using `eb --new-pr`)


Micket:
I'm quite sure it's 2 softwares developed independently, which both deal with the bgen file format. Both are actively developed.
One is a command line tool, the other seems to be a library.

sassy-crick:
Thanks for your advice, unfortunately, in the grand scheme of things that is not working.
I switched to the Waf easyblock as soon as you told me about it. However, without the patch file to remove at least some of the 3rd_party software, I still see for example this in the build log:
`[76/85] Linking build/3rd_party/sqlite3/libsqlite3.a`
So that is not what I want. 
Patching it does not help either as with the current patch file I get this error message:
`:vector<unsigned char, std::allocator<unsigned char> >*): error: undefined reference to 'ZSTD_decompress'
collect2: error: ld returned 1 exit status`
That happens despite the `c4che/_cache.py` file contains:
`LDFLAGS = ['-L/home/sassy/apps/software/zstd/1.4.9-GCCcore-10.3.0/lib64',`
So I would have assumed it picks it up, but it does it not unless you patch it with that ugly construct I have. 

Finally: for the files to copy, this piece of software provides the `build/libbgen.a` which is required by [Regenie](https://github.com/easybuilders/easybuild-easyconfigs/pull/14171) which is the reason for doing all of that. 

These are the reasons why I am seeking advice as I am stuck here. I am happy to drop the patching for the 3rd party pieces of software but I am afraid we are stuck a bit with the files to copy part. 

At least some kind of good news: the `postinstallcmds` works better than what I had. However, when I am using the `Waf` easyblock, both tests are running. With the `CmdCp` the second one constantly fails with this:
`Welcome to bgenix
(version: 1.1.7, revision )
(C) 2009-2017 University of Oxford
!! Error opening index file "example/example.16bits.bgen.bgi": Could not open the index file "example/example.16bits.bgen.bgi"
Use "bgenix -g example/example.16bits.bgen -index" to create the index file.`
The files in question do exist in the build folder. I am puzzled.

Micket:
> However, without the patch file to remove at least some of the 3rd_party software, I still see for example this in the build log: `[76/85] Linking build/3rd_party/sqlite3/libsqlite3.a` So that is not what I want.

Yes i completely expected this. It's not perfect, but they are statically linked into the binaries and the libs aren't installed so there is at least no conflict with the rest of the software stack to worry about; the other easyconfigs can use the binaries without any incompatibility issues.

sassy-crick:
As I said, I need the `build/libbgen.a` for Regine. So, I am happy to switch to the `Waf` Easyblock but I still need to copy that static library. In your opinion: does that cause problems in the grant scheme of things? Having a static library with has a different version of for example `zstd` ?

Micket:
If we can replace them with external versions, that's the best option.
If these attemp fails; the best option i can think of it to place them in a directory that's not `lib` or `lib64`. 
Say, for example, `bgen-enkre-lib` subdirectory instead.
That way, they will at least not be picked up by LIBRARY_PATH by default, so they wouldn't inadvertently be picked up by any other builds that (directly or indirectly) has any connection to BGEN-enkre.
It would mean a patch to regenie to tell it where these "special" versions lie (but that might be necessary regardless).

However; it a separate directory doesn't solve the possible issue with symbols when building things like Regenie later; everything it links to (and the stuff that those libraries link to etc.) must be compatible, and here one can be sure when everything is just using 1 version of the library). If we can remove 2 of them, then, that's at least better than nothing.

verdurin:
Would be good to see an update here.

The current maintainer (i.e. the owner of the enkre.net site) works near me, and he confirms that it's the canonical source now.

sassy-crick:
I finally managed to replace `Boost` as well but only with the `1.55.0` version. I was not able to replace it with the one in the `GCC/10.3.0` toolchain. If somebody can do that, I have no objections but here my programming knowledge is not good enough.
So now all three 3rd party modules are replaced by what EasyBuild has. 
Was that what you had in mind @verdurin ?

sassy-crick:
One problem fixed, another one created:
"Conflict found for dependencies of BGEN-enkre-1.1.7-GCC-10.3.0: bzip2-1.0.8-GCCcore-10.3.0 vs bzip2-1.0.6 "
Any suggestions?

sassy-crick:
I had another look at this and this is the 'best' solution I can come up with. I am stuck with the error message though. Once I know how to solve that error, I also can move the `libbgen.a` if that is still required. I have also updated PR #14171 to reflect the change.

verdurin:
Test report by @verdurin
**FAILED**
Build succeeded for 2 out of 3 (1 easyconfigs in total)
centos7.lan - Linux UNKNOWN UNKNOWN, x86_64, Intel Core Processor (Skylake, IBRS), Python 3.6.8
See https://gist.github.com/2a2f70bc8d7553072fda4aae7a33bb64 for a full test report.

verdurin:
@sassy-crick as you can see from the failed build log, there's a problem with the tarball download, so this may need to be updated.

sassy-crick:
@verdurin  The problem is the slow download. Did you use the in the EC file recommended `-download-timeout=1000` ? 
I addressed another problem as well: The dependency graph showed two versions of `bzip2`: `1.0.6`, coming from the ancient Boost version, and `1.0.8` coming from the SQlight. I somewhat 'fixed' that by upgrading the ancient Boost, which was using the SYSTEM compiler, to the same version (sorry, we are bound to that!) but with the GCCcore tool chain. 
I put both files in that PR but I am happy to move the Boost to a new PR if that makes more sense. Just let me know.

verdurin:
Test report by @verdurin
**SUCCESS**
centos7.lan - Linux UNKNOWN UNKNOWN, x86_64, Intel Core Processor (Skylake, IBRS), Python 3.6.8
See https://gist.github.com/b5fea32f1bcd59dec9fbad83e7cbc61b for a full test report.

verdurin:
@sassy-crick thanks, shame on me for not spotting the download recommendation - the test build succeeds now.

Will have a think about the remaining issues.

verdurin:
@sassy-crick could you change to the `waf` easyblock as discussed above?

sassy-crick:
@verdurin : I got two problems in doing that: 
- we need the library to be copied away as well (minor issue I hope)
- we need to make sure that `build/c4che/_cache.py` contains the libraries for the 3rd party programs which we replaced with the EasyBuild one. Unfortunately, here my ẁaf`knowledge is not sufficient to add that to the `wscript` file. I am happy to add that to the patch-file, if somebody could show me what to add. 

What I have done though is I have added the test, to make sure the build is working. I forgot that the last time.

fizwit:
I also need to build Regenie and need BGEN. I have pulled all the files from this PR and adjusted them to GCC-10.2.0 and they built without issue.  Thanks for the work.

sassy-crick:
I will close that PR as it appears there is a newer version out (v.1.1.8) and with the above mentioned issues, I think a fresh start for the newer version might be a better way forward. Thanks for all your suggestions and help.

PR #14238, opened 2021-10-28
Create BamM-1.7.3-foss-2020b-Python-2.7.18.eb

Ghepardo:
New version for foss 2020b + Python 2.7.18.  This is a prerequisite for a new easyconfig GroopM-0.3.4-foss-2020b-Python-2.7.18.eb.

### DIFF
--- a/easybuild/easyconfigs/b/BamM/BamM-1.7.3-foss-2020b-Python-2.7.18.eb
+++ b/easybuild/easyconfigs/b/BamM/BamM-1.7.3-foss-2020b-Python-2.7.18.eb
@@ -1,0 +1,40 @@
+easyblock = 'PythonPackage'
+
+name = 'BamM'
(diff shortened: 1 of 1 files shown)


Ghepardo:
Help please.  I'm afraid I have no idea how to resolve this.  I have specified SAMtools version 1.12 in the easyconfig, yet the automated checks regime seems to be saying this is ambiguous: that SAMtools version 1.11 could match as well.  Why does it think that?

bedroge:
@Ghepardo Thanks for your contribution, and apologies for not getting back to you sooner; your PR slipped between the cracks...
The CI tests are failing here because we try to stick to a single version of dependencies within each toolchain. For `foss 2020b` the default `SAMtools` version used by other applications is 1.11, while you're trying to use 1.12. Would it be possible to use 1.11 here?

PR #14273, opened 2021-11-03
{bio}[fosscuda/2020b] AlphaFold v2.1.1 w/ Python 3.8.6

ThomasHoffmann77:
(created using `eb --new-pr`)


SebastianAchilles:

SebastianAchilles:
**SUCCESS**
rocky8-eb - Linux rocky linux 8.4, x86_64, Intel(R) Core(TM) i7-6900K CPU @ 3.20GHz (broadwell), Python 3.6.8
See https://gist.github.com/d82cb40b903d63efec42ea93ded9b018 for a full test report.

easybuild/easyconfigs/a/AlphaFold/AlphaFold-2.1.0-fosscuda-2020b.eb
```diff
@@ -0,0 +1,189 @@
...
+    ('ml_collections', '0.1.0', {
+        'preinstallopts': "touch requirements.txt && touch requirements-test.txt && ",
+        'checksums': ['59a17fcd1c140153009788517f304caaddd7a94f06690f9f0ed09987beebcf3c'],
+    }),
+]
+
+postinstallcmds = [
+    "mkdir -p %(installdir)s/bin",
+    # run_alphafold.py script is missing a shebang...
+    "echo '#!/usr/bin/env python' > %(installdir)s/bin/run_alphafold.py",
```
terjekv, line 148: Is this something `fix_python_shebang_for` should be doing (from https://docs.easybuild.io/en/latest/version-specific/easyconfig_parameters.html), or it the line simply missing and the shebang fixer doesn't work in that case?

ThomasHoffmann77: @terjekv I did not try to use fix_python_shebang_for. The postinstallcmds were inherited from the previous versions of this EC

ThomasHoffmann77: @SebastianAchilles @terjekv It turned out that the AlphaFold-2.1.0_data-dep-paths.patch is missing some lines to treat the case --model_preset=multimer. Furthermore there is a typo in line 34 (should be uniprot.fasta and not uniprot.data). I'll try to fix this tomorrow.

boegel: I don't think `fix_python_shebang_for` will work here, because there's no shebang line at all...

We could enhance it to also cover that case, but it doesn't currently, it expects a `^#!` line on top...

easybuild/easyconfigs/a/AlphaFold/AlphaFold-2.1.1-fosscuda-2020b.eb
```diff
@@ -0,0 +1,189 @@
...
+    ('SciPy-bundle', '2020.11'),
+    ('PyYAML', '5.3.1'),
+    ('TensorFlow', '2.4.1'),
+    ('Biopython', '1.78'),
+    ('HH-suite', '3.3.0'),
+    ('HMMER', '3.3.2'),
+    ('Kalign', '3.3.1'),
+    ('jax', '0.2.19'),  # also provides absl-py
+]
+configopts = "-DCUDA_CUDA_LIBRARY=${EBROOTCUDA}/lib64/stubs/libcuda.so "
```
boegel, line 31: @ThomasHoffmann77 This was also added to the easyconfig for AlphaFold 2.0.1 (it's not in the AlphaFold 2.0.0 easyconfig), but why is it needed, can you clarify (perhaps with a comment in the easyconfig itself above this line)?

Also, why is it not specific to a particular component? Like this, it applies to all components, is that intentional?

boegel (changes requested):

easybuild/easyconfigs/a/AlphaFold/AlphaFold-2.1.1-fosscuda-2020b.eb
boegel: Actually, that's not entirely correct: the `configopts` value is set to include this for all components, but `configopts` is totally ignored by the `PythonPackage` easyblock, so this actually only applies to the `OpenMM` component.
So, it would be better to specify this via `'configopts': ...` for the `OpenMM` component specifically, to make that clear (along with a comment to clarify why it is required at all).

akesandgren: It is not generically needed. We didn't need it in our install. So if @ThomasHoffmann77 need it they should handled that in a hook.

ThomasHoffmann77: @boegel @akesandgren The line is a leftover from our configuration. I forgot to remove it. We require the line on our system, because we build on nodes missing the Nvidia drivers.

boegel:
Test report by @boegel
**FAILED**
node3144.skitty.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/91010835c5ed434dc0e98cb56c13412f for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
node3301.joltik.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6242 CPU @ 2.80GHz (cascadelake), 1 x NVIDIA NVIDIA Tesla V100-SXM2-32GB, 465.19.01, Python 3.6.8
See https://gist.github.com/58907494f2386d281d5c26c1f3c24dd2 for a full test report.

boegel (approved):

PR #14304, opened 2021-11-08
{lib}[GCCcore/11.2.0] tbb v2020.3

sassy-crick:
(created using `eb --new-pr`)

PR #14303, opened 2021-11-08
{bio}[gompi/2021b] MAFFT v7.490

sassy-crick:
(created using `eb --new-pr`)

PR #14305, opened 2021-11-08
{bio}[gompi/2021b] UShER v0.5.0 w/ Python 3.9.6

sassy-crick:
(created using `eb --new-pr`)


sassy-crick:
Requires:
- [ ] [tbb](https://github.com/easybuilders/easybuild-easyconfigs/pull/14304)
- [ ] [MAFFT](https://github.com/easybuilders/easybuild-easyconfigs/pull/14303)

migueldiascosta:
@sassy-crick it sees you committed a new copy of `easyconfigs.py` at the top of the repo, it should be a change to `test/easyconfigs/easyconfigs.py` (otherwise, the changes look good to me)

sassy-crick:
@migueldiascosta Thanks for your help. It does make more sense indeed. 
I am not sure if I made a mess of it as these two lines were also removed, but not by me:
```
  # Shasta requires spoa 3.x
            'spoa': [(r'3\.4\.0', [r'Shasta-0\.8\.0-'])],
```
From the commit message I saw I think @boegel done something here. I don't want accidentally introduce a bug!

migueldiascosta:
@sassy-crick I suppose you started from a version of easyconfigs.py from before @boegel added those lines, so you are indeed inadvertently removing them here...

sassy-crick:
@migueldiascosta yes, that is what happened. I thought it is better to ask what to do instead of making more of a mess. 
Thanks for helping here. 
However, I am lost why the checks are not working now. It looks like the lines @boegel added are the problem here but then I might be wrong. 
:-(

migueldiascosta:
@sassy-crick https://github.com/sassy-crick/easybuild-easyconfigs/pull/7 should fix it

migueldiascosta:

migueldiascosta (approved):
lgtm

migueldiascosta:

PR #14323, opened 2021-11-11
{bio}[fosscuda/2020b] RELION v3.1.3

aseedb:


aseedb:
Hi, I am not sure if I need to take steps for a review to be performed. Can you please let me know?

easybuild/easyconfigs/r/RELION/RELION-3.1.3-fosscuda-2020b.eb
```diff
@@ -0,0 +1,48 @@
...
+    ('ctffind', '4.1.14'),
+    ('tbb', '2020.3'),
+    ('Ghostscript', '9.53.3')
+]
+
+hiddendependencies = [
+    ('LibTIFF', '4.1.0'),
+]
+
+configopts = '-DCUDA_ARCH=60 '
```
boegel, line 39: @aseedb Should this be hardcoded to this specific CUDA compute capability, or would it be better to have this easily configurable?

EasyBuild already has a global configuration option for this: `--cuda-compute-capabilities`, and corresponding template values like `%(cuda_cc_space_sep)s` are set based on that.

Could that be leveraged here?

easybuild/easyconfigs/r/RELION/RELION-3.1.3-fosscuda-2020b.eb
```diff
@@ -0,0 +1,48 @@
...
+    ('X11', '20201008'),
+    ('LibTIFF', '4.1.0'),
+    ('FLTK', '1.3.5'),
+    ('ctffind', '4.1.14'),
+    ('tbb', '2020.3'),
+    ('Ghostscript', '9.53.3')
+]
+
+hiddendependencies = [
+    ('LibTIFF', '4.1.0'),
```
boegel, line 36: Whether dependencies are hidden or not is a site-specific aspect, it should not be hardcoded in the easyconfig file, so this should be stripped out in order to make this PR eligible for merging.

A better approach is to configure EasyBuild with `--hide-deps=LibTIFF` locally.

boegel (changes requested):
@aseedb Sorry for the long wait, it seems like your PR slipped through the cracks...

I have requests for changes, see below.

Do let us know if something isn't clear!

easybuild/easyconfigs/r/RELION/RELION-3.1.3-fosscuda-2020b.eb
aseedb: Works also fine. Will push the updated eb file

easybuild/easyconfigs/r/RELION/RELION-3.1.3-fosscuda-2020b.eb
boegel: @aseedb You're now not specifying any configure option at all anymore to specify for which CUDA target `RELION` should be built, so RELION is probably falling back to some default?

aseedb: @boegel Oh yes, I see! But I have a problem. RELION wants the '-DCUDA_ARCH=ab ' where a and b are two digits.
So for example if you want to use compute capability 6.0, you need to have '-DCUDA_ARCH=60'.
--cuda-compute-capabilities requires input in this format: 'a.b' 

I have tried to do something like that in the config file
```
cuda_compute_capabilities = ['6.0']
local_caps = [''.join(x.split('.')) for x in %(cuda_compute_capabilities)s]
local_arch = local_caps[0]

configopts = '-DCUDA_ARCH=' + local_arch + ' '

```

This partially works. It compiles with -DCUDA_ARCH=60.
However, whatever I input via the command line --cuda-compute-capabilities has no effect.
And if I comment out `cuda_compute_capabilities = ['6.0']` then an invalid syntax is thrown. 

Can you guide me on how to solve my problem?

boegel:

PR #14361, opened 2021-11-16
Fix for installing it from GPU-less nodes

surak:
The current PyTorch-Geometric will not install any gpu support if the compilation node does not have a GPU. torch-sparse automatically detects a gpu, so it works on nodes with a gpu present, but you end up with this issue if you don't have gpus there: https://github.com/pyg-team/pytorch_geometric/issues/2429

This is a fix for it.

### DIFF
--- a/easybuild/easyconfigs/p/PyTorch-Geometric/PyTorch-Geometric-2.0.1-fosscuda-2020b-PyTorch-1.9.0.eb
+++ b/easybuild/easyconfigs/p/PyTorch-Geometric/PyTorch-Geometric-2.0.1-fosscuda-2020b-PyTorch-1.9.0.eb
@@ -54,6 +54,13 @@ exts_list = [
     }),
 ]
 
+preinstallopts = 'export TORCH_CUDA_ARCH_LIST="%(cuda_cc_semicolon_sep)s" ; '
+preinstallopts += 'export WITH_METIS=1; '
+preinstallopts += 'export FORCE_CUDA=1; '
+preinstallopts += 'export CUDA_HOME=$EBROOTCUDA; '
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/p/PyTorch-Geometric/PyTorch-Geometric-2.0.1-fosscuda-2020b-PyTorch-1.9.0.eb
```diff
@@ -54,6 +54,13 @@ exts_list = [
     }),
 ]
 
+preinstallopts = 'export TORCH_CUDA_ARCH_LIST="%(cuda_cc_semicolon_sep)s" ; '
```
boegel, line 57: What if EasyBuild is not configured with `--cuda-compute-capabilities`, then this will be an empty value...

surak: Then we're back to zero, and there won't gpu support, right?

boegel: `$TORCH_CUDA_ARCH_LIST` being defined but empty or not being defined is not the same thing (but I haven't tested this, I must admit)

boegel (changes requested):
@surak Should we add `python -c 'import torch_sparse'` as a sanity check command, to catch the problem being fixed here?

surak:
I think so, yes.

boegel:
This easyconfig has been archived since the toolchain it uses is no longer supported, see also https://docs.easybuild.io/policies/toolchains

@surak If this is still relevant, please consider opening a new pull request using a more recent toolchain

PR #14417, opened 2021-11-24
TensorFlow 1 with CUDA 11

rarensu:
The only known way to move TensorFlow 1 up past the dreaded CUDA 10 barrier is to switch from the main tensorflow/tensorflow repository to NVIDIA's fork of it nvidia/tensorflow. NVIDIA maintains this repository for the express purpose of keeping TensorFlow 1 compatible with newer versions of CUDA. This is primarily for benchmarking purposes. 

Annoyingly, nvidia tensorflow doesn't use tags to stay organized. Instead, they leave their release branches dangling off into space indefinitely. I have asked them to do some tagging but in the meantime the best I could do is target the head of the branch. 

As usual, TensorFlow 1 requires a specific version of Bazel - nothing surprising about that. 

I had to do some hackery to get this build to work, but looking at the absurd number of patch files already existing in the TensorFlow easyconfigs directory, I am guessing that is just par for the course. 

`cudnn_headers.patch` 
- For some reason I couldn't quite figure out, the configure script is expecting the cudnn version to get passed through in the form '8.0' when in fact it gets passed through as just '8'. I don't know where the real problem is, so I just modified an if statement to be less picky about which form it actually gets. 
- A cudnn header file was missing from the list of expected header files. I don't know why this would be the case. I am guessing there was a slight mismatch in repository version and local cudnn version. In any case, I just added the header file to the list. 

`eigen_def.patch`
- nvidia/tensorflow is in the middle of fixing some bugs related to eigen's definitions. They keep moving them from one place to another. I guess this particular release is slightly broken. I took the code I found here from one of their own commits on a more recent branch, and kept jamming it in places until the errors went away. 

`numpy-1.19.patch`
- tensorflow/tensorflow is known to be incompatible with numpy >= 1.19 . However, the reason for the incompatibility is trivial to fix. I took this code from one of their own issues. https://github.com/tensorflow/tensorflow/issues/40688

### DIFF
--- a/easybuild/easyconfigs/b/Bazel/Bazel-0.26.1-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/b/Bazel/Bazel-0.26.1-GCCcore-10.2.0.eb
@@ -1,0 +1,22 @@
+name = 'Bazel'
+version = '0.26.1'
+
+homepage = 'https://bazel.io/'
+description = """Bazel is a build tool that builds code quickly and reliably.
+It is used to build the majority of Google's software."""
+
+toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
+
+source_urls = ['https://github.com/bazelbuild/bazel/releases/download/%(version)s']
+sources = ['%(namelower)s-%(version)s-dist.zip']
+checksums = ['c0e94f8f818759f3f67af798c38683520c540f469cb41aea8f5e5a0e43f11600']
+
+builddependencies = [
+    ('binutils', '2.35'),
+    ('Python', '3.8.6'),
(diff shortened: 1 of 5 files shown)


jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 14417` and update the PR to use `SYSTEM` instead.

branfosj:
As this PR is WIP or draft and has not been updated in over a year I have closed it. If you work on this again then please reopen the PR.

PR #14468, opened 2021-11-30
To add Grace to GCCcore

damianam:
Note that it is not as feature-complete as the one installed in the
upper toolchains (but should be enough for anybody I guess)

### DIFF
--- a/easybuild/easyconfigs/g/Grace/Grace-5.1.25-GCCcore-11.2.0.eb
+++ b/easybuild/easyconfigs/g/Grace/Grace-5.1.25-GCCcore-11.2.0.eb
@@ -1,0 +1,35 @@
+easyblock = 'ConfigureMake'
+
+name = 'Grace'
+version = '5.1.25'
+
+homepage = 'http://freecode.com/projects/grace'
+description = 'Grace is a WYSIWYG 2D plotting tool for X Windows System and Motif.'
+
+source_urls = ['ftp://plasma-gate.weizmann.ac.il/pub/grace/src/stable']
+sources = [SOURCELOWER_TAR_GZ]
+
+toolchain = {'name': 'GCCcore', 'version': '11.2.0'}
+
+builddependencies = [
+    ('binutils', '2.37'),
+    ('pkg-config',  '0.29.2'),
+]
+
+dependencies = [
+    ('motif', '2.3.8'),
+    ('libpng', '1.6.37'),
+    ('libjpeg-turbo', '2.1.1'),
+    ('X11', '20210802'),
+]
+
+configopts = "--enable-grace-home='$(PREFIX)'"
+
+runtest = 'tests'
+
+sanity_check_paths = {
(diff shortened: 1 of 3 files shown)


easybuild/easyconfigs/g/Grace/Grace-5.1.25-GCCcore-11.2.0.eb
```diff
@@ -0,0 +1,36 @@
+easyblock = 'ConfigureMake'
+
+name = 'Grace'
+version = '5.1.25'
+
+homepage = 'http://freecode.com/projects/grace'
```
SebastianAchilles, line 6:
```suggestion
homepage = 'https://plasma-gate.weizmann.ac.il/Grace'
```

easybuild/easyconfigs/g/Grace/Grace-5.1.25-GCCcore-11.2.0.eb
```diff
@@ -0,0 +1,36 @@
...
+description = 'Grace is a WYSIWYG 2D plotting tool for X Windows System and Motif.'
+
+source_urls = ['ftp://plasma-gate.weizmann.ac.il/pub/grace/src/stable']
```
SebastianAchilles, line 9:
```suggestion
source_urls = ['https://plasma-gate.weizmann.ac.il/pub/grace/src/stable/']
```

SebastianAchilles (changes requested):

migueldiascosta:
@damianam so, w.r.t. being in GCCcore, it fallsback to an internal FFT so that's not really a problem, the main thing one misses is netCDF support?

w.r.t. the version of `libjpeg-turbo`, can't `2.0.6` be used? If not, there must be an exception in the tests

damianam:
> @damianam so, w.r.t. being in GCCcore, without FFTW it fallsback to an internal FFT so that's not really a problem, the main thing one misses is netCDF support?

Correct. I am not a grace expert though, but at least our users do not use that feature

> w.r.t. the version of `libjpeg-turbo`, can't `2.0.6` be used? If not, there must be an exception in the tests

It can but we are using internally 2.1.1 due to the speedup it has WRT 2.0.x (for x <= 6). See https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/ChangeLog.md#2090-21-beta1 I would argue that it makes sense to update to 2.1.X, but changing to 2.0.6 in the MR is trivial.

smoors:
@damianam I would just use `libjpeg-turbo` version `2.0.6`?
as Grace is not performance critical at all, IMHO this does not warrant an exception to the rule "single dep version per toolchain generation".

boegel:
No follow-up and toolchain is deprecated, see https://docs.easybuild.io/policies/toolchains, so I'm closing this.

@damianam If this is still relevant, please consider opening a new pull request using a more recent toolchain

PR #14480, opened 2021-12-01
also install Python bindings for Brotli

deniskristak:
adding python wrapping to Brotli easyconfig

### DIFF
--- a/easybuild/easyconfigs/b/Brotli/Brotli-1.0.9-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/b/Brotli/Brotli-1.0.9-GCCcore-10.2.0.eb
@@ -25,12 +25,34 @@ builddependencies = [
     ('CMake', '3.18.4'),
 ]
 
+dependencies = [
+    ('Python', '3.8.6')
+]
+
+# python wrapping around Brotli
+exts_defaultclass = 'PythonPackage'
+exts_default_options = {
+    'download_dep_fail': True,
+    'use_pip': True,
+    'sanity_pip_check': True,
+    'source_urls': [PYPI_SOURCE],
+    'sources': [SOURCE_ZIP]
+}
+exts_list = [
+    (name, version, {
+        'modulename': 'brotli',
+        'checksums': ['4d1b810aa0ed773f81dceda2cc7b403d01057458730e309856356d4ef4188438'],
+    }),
+]
+
+modextrapaths = {'PYTHONPATH': ['lib/python%(pyshortver)s/site-packages']}
+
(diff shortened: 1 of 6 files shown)


boegel:
@deniskristak Tests are failing because you introduced a conflict in the dependencies of `fontconfig` (via `freetype`, which depends on `Brotli`):

```
$ eb --check-conflicts fontconfig-2.13.93-GCCcore-10.3.0.eb
== Temporary log file in case of crash /tmp/eb-ah33m30h/easybuild-psfr7v5m.log
Conflict found for dependencies of fontconfig-2.13.93-GCCcore-10.3.0: Python-3.9.5-GCCcore-10.3.0-bare vs Python-3.9.5-GCCcore-10.3.0
	Python-3.9.5-GCCcore-10.3.0-bare as dep of: Rust-1.52.1-GCCcore-10.3.0, fontconfig-2.13.93-GCCcore-10.3.0
	Python-3.9.5-GCCcore-10.3.0 as dep of: Brotli-1.0.9-GCCcore-10.3.0, fontconfig-2.13.93-GCCcore-10.3.0, freetype-2.10.4-GCCcore-10.3.0
```

To fix that, we should simply remove the `-bare` from the `Python` build dependency in `fontconfig-2.13.93-GCCcore-10.3.0.eb`, and add a comment to explain why we're not using the bare Python:

```python
builddependencies = [
    # can't use bare Python, because non-bare Python is required via freetype which depends on Brotli
    ('Python', '3.9.5'),
]

boegel (changes requested):
@deniskristak Please make the same changes to `Brotli` (and `fontconfig`) easyconfigs that are using `GCCcore/10.2.0` or `GCCcore/11.2.0` as toolchain (since there you can add `Python` as a runtime dependency without having to introduce a `-Python-*` versionsuffix)

boegel:

deniskristak:
@boegel ready for review, im not sure if I should have added `Python` to `fontconfig-2.13.92-GCCcore-10.2.0.eb` since it wasn't there before...

boegel:
@deniskristak The failing tests show that adding a `Python` dependency to `Brotli` isn't going to work out, since it introduces conflicts (two different Python versions that need to be loaded at the same time) in various places...

Basically, `Brotli` is as a dependency of `fontconfig`, way too deep in the stack to get away with this.

So we'll need to resort to a separate easyconfig file for the Brotli Python bindings after all...

deniskristak:
closing, here's new PR https://github.com/easybuilders/easybuild-easyconfigs/pull/14528

PR #14523, opened 2021-12-09
{chem}[foss/2020b] yaff 1.6.0

hajgato:

PR #14414, opened 2021-11-24
{chem}[fosscuda/2020b] LAMMPS-29Sep with CUDA (via kokkos)

rarensu:
Being a "newer" LAMMPS, this depends on the updated EasyBlock in-progress. Update LAMMPS easyblock for LAMMPS/29Oct20 #2213. 
When I built this myself, I edited the custom easyblock to enable Ampere sm_86. It looks like y'all were going to do that anyways.

PR #2213, opened 2015-12-11
reorder sources in ALADIN easyconfigs because of changed patch file

boegel:
changes made to patch file, that were done because newer versions of the `patch` command are not fond of relative paths in patch files, require that the list of sources in ALADIN easyconfigs are reordered 

This is required because the first source file determines the subdir in which the patch is applied.

PR #14653, opened 2022-01-04
{chem}[foss/2020b] LAMMPS v29Sep2021

arkdavy:
(created using `eb --new-pr`)
requires https://github.com/easybuilders/easybuild-easyblocks/pull/2213


migueldiascosta:
thanks for your first contribution @arkdavy! good to see you're already using eb's github integration!

please note that in the `2020b` toolchains and more recent, we're no longer using the `-Python-%(pyver)s` `versionsuffix` if the python version is the default one for that toolchain, which is the case here

additionally, there has been a package reorganization in LAMMPS, the only remaining USER package is USER-MISC (https://github.com/lammps/lammps/pull/2818), so I'm not sure what's happening to your `user_packages`. See also https://github.com/easybuilders/easybuild-easyconfigs/pull/14414 if you haven't already?

arkdavy:
Dear @migueldiascosta , thanks for viewing my PR, I am happy to contribute! I have many easyconfigs, I just need to learn how to submit PRs properly. 

Regarding the user packages, there is a required by easyblock `user_packages` list, and that was filled with packages which were called USER-* before, but now those have different names, as you pointed out. But indeed, combining it into one list as in #14414 should be a cleaner solution. The LAMMPS package lists are almost identical between my and #14414 PR. 

I had to delete yaff and molmod easyconfigs from this PR, because these were found in the development branch after removing suffixes (which means that molmod will be deleted from #14414 as well). That also caused a deletion of h5py-2.10.0-foss-2020b, which was found inside the `yaff-1.6.0-foss-2020b.eb` easyconfig.

Then, I have managed to compile documentation using the submitted patch and the `Doxygen` builddependency

Finally, #14414 is a fosscuda compilation, I guess it will be useful to have the CPU-LAMMPS as well.

migueldiascosta:
@arkdavy without a doubt, this is a useful contribution, in addition to the `fosscuda` one

the easyblock does still need work, which will be a requirement for merging this PR, but we can still work on this in the meantime

yes, https://github.com/easybuilders/easybuild-easyconfigs/pull/14523 added `yaff`, with a bundled `h5py` version 2, and `molmod` for this toolchain, so that's good

is having separate variants with and without openmp really necessary/valuable in this case?

arkdavy:
Dear Miguel, I was not sure which one to choose. The easyblock has OpenMP in Kokkos as a default. Compiling with the serial backend is suggested by the code when running with one OpenMP thread. My tests were showing that having the serial backend in Kokkos gives a minor performance gain in pure MPI calculations (in simple tests). In spite of this, maybe a better option would be to use the default settings and to leave the instructions on how to compile with serial by either turning on the corresponding flag (as I have done) or by switching off the OpenMP in Kokkos.  Therefore, I have deleted the `-omp.eb` file and added the corresponding instructions.

arkdavy:
I have added checksum, but another error is about missing yaff, which does exist in the `develop` branch. Can/should we test against that one?

migueldiascosta:
@arkdavy your commit https://github.com/easybuilders/easybuild-easyconfigs/pull/14653/commits/f5d8fe87a918a40192d2b7d9e2c9f499a3970fda is actually deleting the existing easyconfig, you just need to add it back

arkdavy:
It's interesting, I thought I was deleting the file I was uploading. This is apparently not the case when such a file exists in the develop branch.

arkdavy:
I guess I am not able to do anything with the last test error, it somehow does not like several `existing` PLUMED versions with the same toolchain (foss-2020b).

migueldiascosta:
w.r.t. `PLUMED`, there are two options:

- use version `2.6.2` as a dependency instead, unless there's a reason not to
- add an exception to `check_dep_vars` in https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/test/easyconfigs/easyconfigs.py (that can be done in this PR) to allow multiple versions of `PLUMED` as a dependency

this is because we try to only use a single version of a dependency per toolchain version (the check is only for the use as a dependency, you're right that there are already different versions of `PLUMED` with this toolchain version), but even then there are exceptions, as you can see

(in this particular case, it may make more sense to say that it's `CP2K` that requires an older version, in `old_dep_versions`)

arkdavy:
Ok, I took a simpler path using 2.6.2 version of PLUMED. In principle, both of them can be regarded to 2020b toolchain generation since 2.6.2 was released on 26 Oct 2020, 2.7.0 was released on 23 Dec 2020. There are many CP2K configs with this PLUMED-2.6.2, therefore, it must be well-tested.

migueldiascosta:
EB_ARGS="--include-easyblocks-from-pr 2213"

migueldiascosta:
this looks good to me, but it requires https://github.com/easybuilders/easybuild-easyblocks/pull/2213, and that will still need to be made backwards compatible, so it may take a while to merge

arkdavy:
Would you like to attach the file to this PR, or we just add VTK as a dependency here?

genio0815:
Attaching here or new PR, both fine, whatever fits best... The VTK 8.2.0 module itself works fine, but does not compile due to https://github.com/lammps/lammps/pull/2998 and https://matsci.org/t/vtk-package-source-files-seem-to-have-had-syntax-errors-introduced/38775 but should be fixed by their latest patches
https://github.com/lammps/lammps/releases/download/stable_29Sep2021/bugfixes_for_stable_29Sep2021.patch.gz
Thus to use the VTK package, pls add mentioned patch

sassy-crick:
I got a request for LAMMPS v29Sep2021 as well. Given some time has passed, and there are a few PRs open for LAMMPS like for exampe PR#[14815](https://github.com/easybuilders/easybuild-easyconfigs/pull/14815), I was wondering if any work has been done here since 5.2. Happy to contribute if that helps. 
Thanks

arkdavy:
Sorry for the late reply. I think I have managed to push a small easyblock update #2213, which will be backwards compatible, but that has to be checked before we can progress here.

sassy-crick:
From what I can see, PR [#2213](https://github.com/easybuilders/easybuild-easyconfigs/pull/2213) has been merged and I only noticed you mentioning this PR there but no new commits. Maybe it is best to open a new PR for the changes you are proposing so there is a clear history?

arkdavy:
Hi Let's keep it alive for a while, didn't have time to fix/test it, but I would like to complete it definitely

PR #14679, opened 2022-01-07
{bio}[foss/2021a] detettore 2.0.2 w/ Python 3.9.5

i-mtz:


i-mtz:
Test report by @i-mtz
**SUCCESS**
shi08.cluster.bc2.ch - Linux centos linux 7.7.1908, x86_64, Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz, Python 2.7.5
See https://gist.github.com/01fabdac6367230dbc6231301ae72dd6 for a full test report.

easybuild/easyconfigs/d/detettore/detettore-2.0.2-foss-2021a-Python-3.9.5.eb
```diff
@@ -0,0 +1,42 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+# Author: Iñaki Mtz de Ilarduya Muñoz
+# sciCORE - University of Basel
+# SIB Swiss Institute of Bioinformatics
+
+easyblock = 'PythonPackage'
+
+name = 'detettore'
+version = '2.0.2'
+versionsuffix = '-Python-%(pyver)s'
```
boegel, line 10: @i-mtz For Python packages that are installed with a recent toolchain, we're not longer using the `-Python-*` versionsuffix as long as Python 3 is used (we're still doing it when using Python 2).

So please remove this line, and rename the easyconfig file accordingly.

easybuild/easyconfigs/d/detettore/detettore-2.0.2-foss-2021a-Python-3.9.5.eb
```diff
@@ -0,0 +1,42 @@
...
+source_urls = ['https://pypi.python.org/packages/source/d/detettore']
+sources = ['%(name)s-%(version)s.tar.gz']
+checksums = ['7f290c99f65a49a27d6f2468b05706ea9191af105e758875c22e432533a94218']
+
+toolchain = {'name': 'foss', 'version': '2021a'}
+
+dependencies = [
+    ('Python', '3.9.5'),
+    ('minimap2', '2.20'),
+    ('Pysam', '0.18.0', '-Python-%(pyver)s'),
```
boegel, line 24: We don't have an easyconfig file for Pysam 0.18.0 yet, and if we would, it (also) wouldn't have a versionsuffix; that's also why the tests are failing currently.

We're already using Pysam 0.16.0.1 as dependency in various other easyconfigs using the `foss/2021a` toolchain, so it would be better to also stick to that here (so the `detettore` can be loaded together with other modules using the `foss/2021a` toolchain without running into version conflicts).

boegel (changes requested):
@i-mtz Thanks a lot for your contribution, and welcome to EasyBuild!

I've made 2 hopefully easy to resolve suggestions for changes, do let us know if something isn't fully clear!

easybuild/easyconfigs/d/detettore/detettore-2.0.2-foss-2021a-Python-3.9.5.eb
i-mtz: @boegel I have refactored the easyconfig based on the feedback from the CI and the output from `sanity_pip_check = True` which I had omitted. 

I agree about the Pysam dependency, but it seems fixed at `0.17.0` on `setup.py`.  I have asked the author of the program if using `0.16.0.1` would be an option and will update the config if we can use the same Pysam as the one default to the toolchain.

i-mtz:
@boegel It seems that the requirements for the dependencies where a bit strict and others might work. I run an example from the project GitHub repo with the current easyconfig with the dependencies patch and the results are the same as with the more strict dependencies. The version hosted in Pypi does not contain the example but the older version on GitHub does. For reference the test I did was:
`detettore  -b reads.bam  -r reference.fasta  -a TE_annotation.gff  -t TE_conensusLib.fasta  -o current_2021a_downgraded  -m tips taps  --require_split  --include_invariant` and compared the resulting files (only changes where dates).

akesandgren:
Closing this since foss/2021a is deprecated, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @i-mtz 

PR #14681, opened 2022-01-07
{lib}[GCCcore/11.2.0] psutil v5.9.0

Ghepardo:
From latest available source.

### DIFF
--- a/easybuild/easyconfigs/p/psutil/psutil-5.9.0-GCCcore-11.2.0.eb
+++ b/easybuild/easyconfigs/p/psutil/psutil-5.9.0-GCCcore-11.2.0.eb
@@ -1,0 +1,21 @@
+easyblock = 'PythonPackage'
+
+name = 'psutil'
+version = '5.9.0'
+
+homepage = 'https://github.com/giampaolo/psutil'
+description = """A Python library for retrieving information on running processes and system utilization"""
+
+toolchain = {'name': 'GCCcore', 'version': '11.2.0'}
+
+sources = [SOURCE_TAR_GZ]
+checksums = ['869842dbd66bb80c3217158e629d6fceaecc3a3166d3d1faee515b05dd26ca25']
+
+use_pip = True
+download_dep_fail = True
+
+dependencies = [('Python', '3.9.6', '-bare'), ('Python', '2.7.18', '-bare')]
+
+builddependencies = [('binutils', '2.37')]
+
+moduleclass = 'lib'


easybuild/easyconfigs/p/psutil/psutil-5.9.0-GCCcore-11.2.0.eb
```diff
@@ -0,0 +1,21 @@
...
+
+toolchain = {'name': 'GCCcore', 'version': '11.2.0'}
+
+sources = [SOURCE_TAR_GZ]
+checksums = ['869842dbd66bb80c3217158e629d6fceaecc3a3166d3d1faee515b05dd26ca25']
+
+use_pip = True
+download_dep_fail = True
+
+dependencies = [('Python', '3.9.6', '-bare'), ('Python', '2.7.18', '-bare')]
```
branfosj, line 17: `dependencies` can not have multiple versions of the same software.

easybuild/easyconfigs/p/psutil/psutil-5.9.0-GCCcore-11.2.0.eb
```diff
@@ -0,0 +1,21 @@
+easyblock = 'PythonPackage'
+
+name = 'psutil'
+version = '5.9.0'
```
branfosj, line 4: There is `psutil` 5.8.0 included in `Python-3.9.6-GCCcore-11.2.0.eb`. That would be the usual route to get `psutil`, unless that version is unsuitable for what you are using it for.

branfosj (changes requested):

easybuild/easyconfigs/p/psutil/psutil-5.9.0-GCCcore-11.2.0.eb
Ghepardo: Thanks @branfosj, that's useful to know.  I'll ditch this pull request - is there a process I should follow for that?  

Just a question about this, for my future reference:

I tried to base this new version on the previous latest easyconfig for psutil, which contained:

```
multi_deps = {'Python': ['3.7.2', '2.7.15']}
```

I wanted to move both specified Python versions along to the latest ones with the latest toolchains, but there were only the "bare" builds available: Python-2.7.18-GCCcore-11.2.0-bare.eb and Python-3.9.6-GCCcore-11.2.0-bare.eb.  I could not find a way to get the multi_deps parameter to pick up these versions.  Does its syntax allow for this?  (No worries if not - as I say I'll not pursue this PR.)

smoors: @Ghepardo to close this pull request, just push the "Close pull request" button at the bottom and add a comment why you're closing it.

I think there is currently no way to specify the -bare modules with multi-deps. In any case, the -bare modules should normally only be used as build dependencies.

Ghepardo:
Advice received that this Python library is actually provided with the Python builds, so a separate build is unnecessary.

PR #14714, opened 2022-01-14
{bio}[GCCcore/11.2.0] TranslatorX v1.1

zemu-unile:
(created using `eb --new-pr`)

PR #14718, opened 2022-01-14
ignore missing binutils dependency for binaries in easyconfigs test suite

zemu-unile:
By default the testsuite checks for a binutils or GCC dependency, when
using the GCCcore toolchain to avoid accidentially using the system
binutils. However, this is not necessary for packages that do not need
to be build like binaries. This commit adds exceptions for the 'Binary'
and the 'PackedBinary' easyblock to the existing exception for the
'Tarball' easyblock.


boegel:
@zemu-unile There are probably cases where the `Binary` or `PackedBinary` easyblocks are used, but still some compilation is being done (and then we really do want `binutils` included as a build dependency), so I'm not sure this is a good idea to do in general (but I do understand that it can be annoying when the tests require including `binutils` when there's no compilation being done at all).

Do you have specific examples of PRs that made you look into this change?

zemu-unile:
#14714 is failing the test as it is just a Perl script.

zemu-unile:
already resolved in  #15932

PR #14752, opened 2022-01-18
{bio}[foss/2021a] GROMACS v2021.4 w/ Python 3.9.5

smoors:
(created using `eb --new-pr`)


akesandgren:
I think 2021.5 was released yesterday while you're at it.

smoors:
@akesandgren this is the pre-patched Colvars release, see https://github.com/Colvars/colvars#gromacs-colvars-releases
2021.5 is not yet available there. I guess it should not be too difficult to do the patching manually, but I prefer to stick to what they provide.

akesandgren:
I meant that you could create a standard 2021.5 .eb for gromacs :-)

easybuild/easyconfigs/g/GROMACS/GROMACS-2021.4-foss-2021a-CUDA-11.3.1-colvars.eb
```diff
@@ -0,0 +1,82 @@
...
+variables).
+"""
+
+toolchain = {'name': 'foss', 'version': '2021a'}
+toolchainopts = {'openmp': True, 'usempi': True}
+
+source_urls = [
+    'https://github.com/Colvars/gromacs/archive/',
+]
+sources = ['v%(version)s-colvars.tar.gz']
```
migueldiascosta, line 28: that filename doesn't seem to be available there (https://github.com/Colvars/gromacs/tags). it is without `-colvars`, but it has a different checksum...

otoh, `v2021.5-colvar` was added 19h ago :)

jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 14752` and update the PR to use `SYSTEM` instead.

boegel:
`foss/2021a` is now a deprecated toolchain (seehttps://docs.easybuild.io/policies/toolchains/#current-situation), and I doubt this is still relevant given the now ancient software version, so closing...

PR #14770, opened 2022-01-19
{bio}[foss/2021b] GRIDSS v2.13.1

zemu-unile:
This only works on x86_64 as it includes a pre-compiled gridsstools.


migueldiascosta:
@zemu-unile it would be better to add a `x86_64` suffix then

jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 14770` and update the PR to use `SYSTEM` instead.

boegel:
Closing this since `foss/2021b` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

@zemu-unile If this is still relevant, please consider opening a new pull request using a more recent toolchain

PR #14825, opened 2022-01-21
{bio}[GCC/10.3.0] fastp 0.23.2

pescobar:


migueldiascosta:

migueldiascosta:

migueldiascosta:
@pescobar I suppose `libdeflate` does fit the criteria of being performance sensitive, but I wonder if anyone would actually want to build it with other compilers, which is the only real reason to use `GCC` instead of `GCCcore` (?)

but if that's the case (no strong feelings either way), then a) the `binutils` dependency is no longer necessary, since it's pulled by `GCC`, and b) should all of the existing easyconfigs be bumped to `GCC`?

pescobar:
@migueldiascosta the only reason why I choosed `GCC` instead of `GCCcore` is because `GCCcore` is confusing for some of my users but you are probably right and it makes sense to use `GCCcore` for libdeflate. I will update the PR.

migueldiascosta:
**SUCCESS**
c2dhpc - Linux CentOS Linux 7.6.1810, x86_64, AMD EPYC 7601 32-Core Processor (zen), Python 2.7.5
See https://gist.github.com/6359e087ed72e1124e427f9478eb816b for a full test report.

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @pescobar!

PR #14861, opened 2022-01-27
{lib}[GCC/10.2.0] librsb v1.2.0.10, v1.2.0.11, v1.3.0.0

### LINKED ISSUE #14860: librsb: updates to version 1.2.0.10,1.2.0.11,1.3.0.0
Hi. You may want to update configs for 1.2.0.10,1.2.0.11,1.3.0.0 tarballs (available on https://sourceforge.net/projects/librsb/files/ ), caring to avoid gcc-11.2 (there is a problem https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103995 affecting librsb-1.3's and its test suite suite -- gcc-11.3 should be fine).

michelemartone:
Fix #14860 by updated librsb-1.2.0.10/librsb-1.2.0.11/1.3.0.0 files.

### DIFF
--- a/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
+++ b/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
@@ -1,0 +1,29 @@
+##
+# Author:    Michele Martone <michelemartone@users.sourceforge.net>
+##
+easyblock = 'ConfigureMake'
+
+name = 'librsb'
+version = '1.3.0.0'
+
+homepage = 'https://sourceforge.net/projects/librsb/'
+description = """A shared memory parallel sparse matrix computations library for the Recursive Sparse Blocks format"""
+
+toolchain = {'name': 'GCC', 'version': '10.2.0'}
+
+source_urls = [SOURCEFORGE_SOURCE]
+sources = [SOURCE_TAR_GZ]
+checksums = ['2ac8725d1f988f57df9383ae6b0bb2ed221ec935187d31ebb62ea95ee868a790']
+
+dependencies = [
+    ('zlib', '1.2.11')
+]
+
+configopts = '--enable-openmp --enable-fortran-module-install'
+
+sanity_check_paths = {
+    'files': ['bin/rsbench', 'lib/librsb.a', 'lib/librsb.%s' % SHLIB_EXT],
+    'dirs': ['include', 'share'],
+}
+
+moduleclass = 'lib'
--- a/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
+++ b/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
@@ -19,8 +19,13 @@ dependencies = [
     ('zlib', '1.2.11')
 ]
 
+# workaround to a configure defect manifesting during --rebuild: avoid mistaken detection of already existing librsb installation
 preconfigopts = "sed -i 's/export LIBRSB_CONFIG=.*$/export LIBRSB_CONFIG=no LIBRSB_CONFIG_VERSION=1.3.0;/g' configure*;"
-preconfigopts += " sed -i 's/^	true/CPPFLAGS+=\" -I.. -DRSBPP_HAS_RSB_H=1\"/g' librsbpp/configure*;"
+
+# workaround to a configure defect manifesting during --rebuild: ensure librsbpp/configure sees and uses root dir rsb.h
+local_cppflags = 'CPPFLAGS="${CPPFLAGS} -I%(builddir)s"'
+prebuildopts = local_cppflags
+preinstallopts = local_cppflags
 
 configopts = '--enable-openmp --enable-fortran-module-install'
 
--- a/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
+++ b/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
@@ -19,6 +19,8 @@ dependencies = [
     ('zlib', '1.2.11')
 ]
 
+preconfigopts = "sed -i 's/export LIBRSB_CONFIG=.*$/export LIBRSB_CONFIG=no;export LIBRSB_CONFIG_VERSION=1.3.0;/g' configure*; sed -i 's/^	true/CPPFLAGS+=\" -I.. -DRSBPP_HAS_RSB_H=1\"/g' librsbpp/configure*;"
+
 configopts = '--enable-openmp --enable-fortran-module-install'
 
 sanity_check_paths = {
--- a/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
+++ b/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
@@ -19,7 +19,8 @@ dependencies = [
     ('zlib', '1.2.11')
 ]
 
-preconfigopts = "sed -i 's/export LIBRSB_CONFIG=.*$/export LIBRSB_CONFIG=no;export LIBRSB_CONFIG_VERSION=1.3.0;/g' configure*; sed -i 's/^	true/CPPFLAGS+=\" -I.. -DRSBPP_HAS_RSB_H=1\"/g' librsbpp/configure*;"
+preconfigopts = "sed -i 's/export LIBRSB_CONFIG=.*$/export LIBRSB_CONFIG=no LIBRSB_CONFIG_VERSION=1.3.0;/g' configure*;"
+preconfigopts += " sed -i 's/^	true/CPPFLAGS+=\" -I.. -DRSBPP_HAS_RSB_H=1\"/g' librsbpp/configure*;"
 
 configopts = '--enable-openmp --enable-fortran-module-install'
 
--- a/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
+++ b/easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
@@ -15,17 +15,17 @@ source_urls = [SOURCEFORGE_SOURCE]
 sources = [SOURCE_TAR_GZ]
 checksums = ['2ac8725d1f988f57df9383ae6b0bb2ed221ec935187d31ebb62ea95ee868a790']
 
+builddependencies = [
+    ('Autotools', '20200321'),
+]
+
 dependencies = [
     ('zlib', '1.2.11')
 ]
 
-# workaround to a configure defect manifesting during --rebuild: avoid mistaken detection of already existing librsb installation
-preconfigopts = "sed -i 's/export LIBRSB_CONFIG=.*$/export LIBRSB_CONFIG=no LIBRSB_CONFIG_VERSION=1.3.0;/g' configure*;"
-
-# workaround to a configure defect manifesting during --rebuild: ensure librsbpp/configure sees and uses root dir rsb.h
-local_cppflags = 'CPPFLAGS="${CPPFLAGS} -I%(builddir)s"'
-prebuildopts = local_cppflags
-preinstallopts = local_cppflags
+# do not rely on `librsb-config --I_opts`, it will inject paths in install dir
+preconfigopts = "sed -i '21,22d' librsbpp/configure.ac && "
+preconfigopts += "autoreconf -fi &&"
 
 configopts = '--enable-openmp --enable-fortran-module-install'
 
--- a/easybuild/easyconfigs/l/librsb/librsb-1.2.0.10-GCC-10.2.0.eb
+++ b/easybuild/easyconfigs/l/librsb/librsb-1.2.0.10-GCC-10.2.0.eb
@@ -1,0 +1,29 @@
+##
+# Author:    Michele Martone <michelemartone@users.sourceforge.net>
+##
+easyblock = 'ConfigureMake'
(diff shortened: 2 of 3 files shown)


robert-mijakovic:
Test report by @robert-mijakovic
**FAILED**
Build succeeded for 12 out of 13 (3 easyconfigs in total)
mel0429 - Linux rocky linux 8.5, x86_64, AMD EPYC 7H12 64-Core Processor (zen2), Python 3.6.8
See https://gist.github.com/2124005799f1e25487c63b8479b29bab for a full test report.

boegel:
@robert-mijakovic test build failed for `librsb-1.3.0.0-GCC-10.2.0.eb` with:

```
mv -f .deps/librsb_base_la-rsb_op.Tpo .deps/librsb_base_la-rsb_op.Plo
f951: Fatal Error: Cannot rename module file blas_sparse.mod0 to blas_sparse.mod: No such file or directory
compilation terminated.
make[2]: *** [Makefile:1447: rsb_blas_sparse.o] Error 1
```

@michelemartone Could it be that `librsb` doesn't like to be built in parallel?
This problem surfaced on a system with 256 cores...

michelemartone:
Hi @boegel  @robert-mijakovic! I tried it on a system with 256 declared cores (nproc giving 256) but the build went well, so I could not replicate this build failure :-/

boegel:
CORE_CNT=16

boegel (approved):
lgtm

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 2 out of 3 (3 easyconfigs in total)
node2640.swalot.os - Linux UNKNOWN UNKNOWN, x86_64, Intel(R) Xeon(R) CPU E5-2660 v3 @ 2.60GHz (haswell), Python 3.6.8
See https://gist.github.com/9e399884b077b1bdffe3e75d937c0686 for a full test report.

boegel:
@michelemartone Installation of `librsb-1.3.0.0-GCC-10.2.0.eb` seems to be failing also on other systems, but differently:

```
rsbpp.hpp:58:10: fatal error: rsb.h: No such file or directory
   58 | #include <rsb.h> // flags etc; TODO: what about constants only ?
      |          ^~~~~~~
compilation terminated.
make[2]: *** [rsbpp.o] Error 1
```

michelemartone:
It seems like the Makefile under `rsblib is getting `CPPFLAGS` without `-I` to the project root dir.
Having Makefile and config.log would be helpful here..

michelemartone:
On a machine of mine, building with ` eb librsb-1.3.0.0-GCC-10.2.0.eb`
it gets through the process correctly, and produces a 
`~/.local/easybuild/software/librsb/1.3.0.0-GCC-10.2.0/easybuild/[easybuild-librsb-1.3.0.0-20220129.232326.log.excerpt.txt](https://github.com/easybuilders/easybuild-easyconfigs/files/7964919/easybuild-librsb-1.3.0.0-20220129.232326.log.excerpt.txt)`
of which I attach an excerpt.

michelemartone:
[easybuild-librsb-1.3.0.0-20220129.232326.log.excerpt.txt](https://github.com/easybuilders/easybuild-easyconfigs/files/7964920/easybuild-librsb-1.3.0.0-20220129.232326.log.excerpt.txt)

robert-mijakovic:
@michelemartone Can you run the test on your system and upload the report with?:
```
eb --from-pr 14861 --rebuild --upload-test-report --robot
```

robert-mijakovic:
> Hi @boegel @robert-mijakovic! I tried `make -j 256` on a system with 256 declared cores (nproc giving 256) but the build went well, so I could not replicate this build failure :-/

Just out of curiosity, is this on BEAST?

michelemartone:
> @michelemartone Can you run the test on your system and upload the report with?:
> 
> ```
> eb --from-pr 14861 --rebuild --upload-test-report --robot
> ```

@robert-mijakovic 
I was on my way to enable --upload-test-report.
When I enter the GitHub token `eb` summons all the callback wrappers of the world, and crashes with an exception stack as here attached:

[eb_github_bum.txt](https://github.com/easybuilders/easybuild-easyconfigs/files/7965796/eb_github_bum.txt)

With incorrect token it just complains "Token validation failed.." and stops.
Isn't there a --with-github-token=... or similar option so to pass by command line (perfectly ok for non-shared systems) ?

michelemartone:
Test report by @michelemartone
**FAILED**
Build succeeded for 0 out of 3 (3 easyconfigs in total)
tiko - Linux UNKNOWN UNKNOWN, x86_64, Intel(R) Core(TM) i7-6500U CPU @ 2.50GHz, Python 3.9.2
See https://gist.github.com/6aff61a480353fc7fbc9e55904a9f2af for a full test report.

michelemartone:
@robert-mijakovic  @boegel 
News: I can reproduce the `fatal error: rsb.h: No such file or directory` issue. 
It happens when I specify `--rebuild`.
I have used `--disable-cleanup-builddir  --disable-cleanup-tmpdir` to keep the logs, which I have inspected and comment here.
The `make -j ..` step runs fine till the end.
Then EasyBuild goes in ('test step', 'Running method stage_install_step part of step install' and others), and then declares `run.py:233 INFO running cmd:  make install`.
Then `make install` fails in a subdirectory  `librsbpp`, where is tricked into rebuilding `rsbpp.o`, but a header previously available is now missing, and the build fails.

I'd need to know what does EasyBuild do once `make -j..` is done, but before launching `make install`.
Is it touching any file in the build directory or its subdirs?

lexming:
Test report by @lexming
**SUCCESS**
node304.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/7e010f57e5f785574338f065304ca1e7 for a full test report.

lexming:
@michelemartone this looks like an issue with the build system of librsb and high parallel setups. I would try to limit  the maximum number of threads in make with `maxparallel = 4`

michelemartone:
> Test report by @lexming **SUCCESS** Build succeeded for 3 out of 3 (3 easyconfigs in total) node304.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5 See https://gist.github.com/7e010f57e5f785574338f065304ca1e7 for a full test report.


Hi @lexming  I guess you're referring to the https://github.com/easybuilders/easybuild-easyconfigs/pull/14861#issuecomment-1024945523 (`blas_sparse.mod0 `) issue mentioned by @robert-mijakovic .
As said, on the `test $(nproc) == 256`  machine I've access to it does not happen; I'd be interesting if you can enable me to reproduce it, and/or to suggest a patch to `Makefile.am`.

lexming:
@michelemartone you are right, I could reproduce the error with a single thread. This problem is not related to EasyBuild though, I had the `fatal error: rsb.h: No such file or directory` occur with a manual `make install`. I'm not sure how to systematically trigger the error though, sometimes `make install` thinks that there is stuff to rebuild and then it fails. Whenever  `make install` just does the copy of files into prefix then it works.

A not-very-pretty workaround is to add the root build directory (containing `rsb.h`) to `CPATH` of the `make install` command 
```
preinstallopts = 'CPATH="$CPATH:%(builddir)s"'
```

easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
```diff
@@ -0,0 +1,31 @@
...
+
+source_urls = [SOURCEFORGE_SOURCE]
+sources = [SOURCE_TAR_GZ]
+checksums = ['2ac8725d1f988f57df9383ae6b0bb2ed221ec935187d31ebb62ea95ee868a790']
+
+dependencies = [
+    ('zlib', '1.2.11')
+]
+
+preconfigopts = "sed -i 's/export LIBRSB_CONFIG=.*$/export LIBRSB_CONFIG=no;export LIBRSB_CONFIG_VERSION=1.3.0;/g' configure*; sed -i 's/^	true/CPPFLAGS+=\" -I.. -DRSBPP_HAS_RSB_H=1\"/g' librsbpp/configure*;"
```
lexming, line 22: This `sed` command is confusing:
* why act on all `configure*` files? I guess that you are only interested in `confgure`. `configure.ac` has no effect unless you manually re-run `autoreconf`.
* the match on `^	true` is prone to error, it can match multiple lines. I guess that you want to replace only line 17038 of `librsbpp/configure`. If that's the case, I suggest 
    ```
    17038s/.*/CPPFLAGS+=\" -I.. -DRSBPP_HAS_RSB_H=1\"/g
    ```

lexming (changes requested):
@michelemartone Thanks for the update. The tests are failing because the `preconfigopts` line is too long. Apart from that I have the following questions

easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
michelemartone: Hi @lexming. The patch was targeted for this (1.3.0.0) release, where the configure produced by me has only one `'^<TAB>true'` line.
If you tell me there is guarantee easybuild will never run `autoreconf`, we can equally use the hardcoded line number as you suggest :-)

michelemartone: (or target `configure` only instead of `configure.ac`, too)

lexming: Easybuild will not run `autoreconf` by default. You can do so in this easyconfig by loading `Autotools` and then run `autoreconf` before configuration with `preconfigopts = "autoreconf -fi &&"`.

If you want to use the approach of either modifying the configure script, or alternatively target `configure.am` and run `autoreconf`, you should use a patch to make those modifications to the code. Those changes are a little bit too much for a `sed` command.

Regarding the issue with `rsb.h`, the end result of your `sed` command is adding the root of the build directory to `CPPFLAGS`.  You would get the same result by executing make with 
```
CPPFLAGS="${CPPFLAGS} -I/path/to/build/root" make
```
which is easier to apply in EasyBuild (and more clear than the changes to configure scripts)
```
local_cppflags =  'CPPFLAGS="${CPPFLAGS} -I%(builddir)s"'
prebuildopts = local_cppflags
preinstallopts = local_cppflags
```

So, I recommend the second solution for v1.3.0.0. If you can fix the configuration scripts, none of this will be needed in future versions.

michelemartone: Indeed, the `CPPFLAGS`-based way is cleaner (and clearer); latest change on  https://github.com/easybuilders/easybuild-easyconfigs/pull/14861/commits/8d3c5c8bfc1b634e1b22724e10aabf7ad0fb521a uses this one as you suggest.

easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
```diff
@@ -0,0 +1,37 @@
...
+    ('zlib', '1.2.11')
+]
+
+# workaround to a configure defect manifesting during --rebuild: avoid mistaken detection of already existing librsb installation
+preconfigopts = "sed -i 's/export LIBRSB_CONFIG=.*$/export LIBRSB_CONFIG=no LIBRSB_CONFIG_VERSION=1.3.0;/g' configure*;"
+
+# workaround to a configure defect manifesting during --rebuild: ensure librsbpp/configure sees and uses root dir rsb.h
+local_cppflags = 'CPPFLAGS="${CPPFLAGS} -I%(builddir)s"'
+prebuildopts = local_cppflags
+preinstallopts = local_cppflags
```
lexming, line 28: I found a very simple solution to those rebuild issues! All the trouble is caused by `$LIBRSB_CONFIG --I_opts` in `librsbpp/configure.ac`. That command will inject paths to the installation directory, which is emptied upon rebuild. Easy solution is to not use that command at all, delete those lines, regenerate the configure script and build as usual. I tested it and had no errors.

```suggestion
builddependencies = [
    ('Autotools', '20200321'),
]

dependencies = [
    ('zlib', '1.2.11')
]

# do not rely on `librsb-config --I_opts`, it will inject paths in install dir
preconfigopts = "sed -i '21,22d' librsbpp/configure.ac && "
preconfigopts += "autoreconf -fi &&"
```

lexming (changes requested):
@michelemartone thanks a lot for the update. I'm sorry to bother you again, but while I was re-testing this PR I found a very simple and clean solution to those issues. Please see my suggestion below

lexming:

lexming:
Test report by @lexming
**SUCCESS**
node316.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/83f90b75805e649deec4237a9969d57b for a full test report.

lexming:

lexming (approved):
@michelemartone thanks for keeping up with this PR.

LGTM

lexming:
Going in, thanks @michelemartone !

lexming:

lexming:
Closing and re-opening to unstuck tests

easybuild/easyconfigs/l/librsb/librsb-1.3.0.0-GCC-10.2.0.eb
lexming: @boegel my understanding was that this will be properly fixed upstream by @michelemartone , so it should not be needed in future versions.

PR #14881, opened 2022-01-31
trim test configurations for easyconfigs test suite: only test with Python 2.7 + 3.6 and Lmod 7.x + 8.x

boegel:
The changes in #14857 were a bit ambitious, testing with a wide range of Python 3.x versions quickly leads to a large backlog for easyconfig PRs...

Only testing with Python 2.7 + 3.6 (most popular Python 3.x version), and no longer testing with Lmod 6.x (deprecated for a while, see https://github.com/easybuilders/easybuild-framework/pull/3077) should be enough...

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -5,24 +5,12 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9, '3.10']
-        modules_tool: [Lmod-6.6.3, Lmod-7.8.22, Lmod-8.1.14]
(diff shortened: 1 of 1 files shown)


lexming (approved):
LGTM

lexming:
The PR is expecting the two excluded tests (Lmod-7 + Tcl). Those tests are not running, so they should not be expected. I'll close an reopen this PR to see if that refreshes the list of expected tests.

lexming:
Ok, the required checks have to updated in this branch protection rules. @boegel I cannot change that :slightly_smiling_face:

boegel:
@lexming fixed!

lexming:
Going in, thanks @boegel !

PR #14800, opened 2022-01-20
{bio}[foss/2021b] methylpy v1.2.9 w/ Python 3.9.6

ItIsI-Orient:
(created using `eb --new-pr`)

PR #14891, opened 2022-02-01
Remove incompatible version of Bowtie

ItIsI-Orient:
Using tbb dependency with too high version. See https://github.com/easybuilders/easybuild-easyconfigs/pull/14800


boegel:
The required changes to this Bowtie easyconfig are a part of #14800 now, so there's no need at all to remove this easyconfig... I'll close this PR

@ItIsI-Orient If something isn't clear here, let me know!

PR #14922, opened 2022-02-03
{chem}[gompi/2021b] PCMSolver v1.3.0 w/ Python 3.9.6

sassy-crick:
(created using `eb --new-pr`)

I tried to address this [issue](https://github.com/PCMSolver/pcmsolver/issues/159) a bit better and I could get it to build on an Intel Gold 5118 CPU without any issues. Unfortunately, I don't have access to Zen2 or newer Intel CPUs right now. 
My idea was to have the best possible build but avoiding `avx` and `avx2` which seem to cause the problem with one of the test jobs. It is a bit disappointing it has not been addressed by them right now. 

@boegel As you raised the issue, would you mind to have a look at that as well?


lexming:

lexming:
Test report by @lexming
**SUCCESS**
node302.hydra.os - Linux centos linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 2.7.5
See https://gist.github.com/0514fd6ce7a12b735ebd7fec6b288056 for a full test report.

lexming:

lexming (approved):
Tests cover Intel Haswell, Intel Skylake and AMD Zen2.

LGTM

sassy-crick:
> Tests cover Intel Haswell, Intel Skylake and AMD Zen2.

Many thanks for doing this, much appreciated! So that could be a way forward then. 
However, I was wondering if it does make more sense to apply that only for the `SphericalDiffuse.cpp` part of the code as the rest of the code does not seem to be broken. I don't know if that is possible with `CMake` but I know that GAMESS-US does a similar thing, to have a less aggressive compilation for some parts of the code so the results are correct but leave other parts of it as it is. Just an idea and right now I don't have any hardware to test anything available. 😕

sassy-crick:
@boegel What are we going to do with this rather old but somehow interesting PR? From what I can tell that is the latest release of the software package.

sassy-crick:
@boegel  Ad the broken green-spherical test: seen this [PR](https://github.com/PCMSolver/pcmsolver/pull/200) which looks like an upcoming release.

sassy-crick:
I am going to close this PR simply as it is too old.

PR #14972, opened 2022-02-10
{bio}[GCCcore/10.3.0] vamb v3.0.3 w/ Python 3.9.5

Ghepardo:
(created using `eb --new-pr`)


easybuild/easyconfigs/v/vamb/vamb-3.0.3-GCCcore-10.3.0.eb
```diff
@@ -0,0 +1,31 @@
...
+source_urls = [GITHUB_SOURCE]
+sources = ['v%(version)s.tar.gz']
+checksums = ['b96e075a1477b855a45e47ceac423b8ae3c56db105e5810c8bc34949462099b0']
+
+dependencies = [
+    ('binutils', '2.36.1'),
+    ('Pysam', '0.16.0.1', '', ('GCC', '10.3.0')),
+    ('Python', '3.9.5'),
+    ('PyTorch', '1.10.0', '', ('foss', '2021a')),
+    ('SciPy-bundle', '2021.05', '', ('foss', '2021a')),
```
branfosj, line 24: Toolchain inheritance is `foss` -> `gompi` -> `GCC` -> `GCCcore`. If one of the dependencies is at the highest level of these then this easyconfig should be at the same level. I.e. it should be `vamb-3.0.3-foss-2021a.eb` with `toolchain = {'name': 'foss', 'version': '2021a'}`

branfosj (changes requested):

akesandgren:
This easyconfig is not quite correct in form and we have a newer version in PR #17457

Closing this PR. In case you really need this older version please look at the above PR for how it should look.

PR #15039, opened 2022-02-23
Fix typo in OSU-Micro-Benchmarks-5.7.1-gompi-2021a-CUDA-11.3.1.eb

ocaisa:


branfosj:
Looking at the OSU easyconfigs, we have two separate setups:
1. Using `configopts = 'CC="$MPICC" CXX="$MPICC"'`
2. Using `toolchainopts = {'usempi': True}`

If we, instead, switch them all to use method 2 then this automatically sets `CC` and `CXX` as required?

ocaisa:
I'm on board, I also noticed that when creating #15018 
I'll update this and there as well.

ocaisa:

smoors (approved):
lgtm

smoors:
the failures are because our cluster does not currently support the iimpic toolchain.

smoors:
Going in, thanks @ocaisa!

PR #15063, opened 2022-02-28
Xtb python

steenlysgaard:
This adds an easyconfig file for xtb-python and an update for xtb in order to use the same intel toolchain.

The post install commands look a bit messy, the reason is that xtb-python needs to be compiled with Meson-Ninja but is imported like a Python package. There could be better ways of doing this.


boegel:
Closing this since `intel/2021b` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @steenlysgaard 

PR #15084, opened 2022-03-05
{vis}[GCCcore/10.2.0] matplotlib-inline v0.1.3

zdemat:
Adding an easyconfig recipe for a new sw called matplotlib-inline providing a Matplotlib Inline Back-end for IPython and Jupyter.

### DIFF
--- a/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
@@ -1,0 +1,27 @@
+easyblock = 'PythonPackage'
+
+name = 'matplotlib-inline'
+version = '0.1.3'
+
+homepage = 'https://github.com/ipython/matplotlib-inline'
+description = """Matplotlib Inline Back-end for IPython and Jupyter."""
+
+toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
+
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['a04bfba22e0d1395479f866853ec1ee28eea1485c1d69a6faf00dc3e24ff34ee']
+
+dependencies = [
+    ('Python', '3.8.6'),
+    ('IPython', '7.18.1'), # has traitlets
+]
+
+use_pip = True
+download_dep_fail = True
+sanity_pip_check = True
+
+# use matplotlib-inline backend as default
+# see https://matplotlib.org/tutorials/introductory/usage.html#what-is-a-backend
+modextravars = {'MPLBACKEND': 'module://matplotlib_inline.backend_inline'}
+
+moduleclass = 'vis'
--- a/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
@@ -11,9 +11,11 @@ toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
 sources = [SOURCELOWER_TAR_GZ]
 checksums = ['a04bfba22e0d1395479f866853ec1ee28eea1485c1d69a6faf00dc3e24ff34ee']
 
+builddependencies = [('binutils', '2.35')]
+
 dependencies = [
     ('Python', '3.8.6'),
-    ('IPython', '7.18.1'),
+    ('IPython', '7.18.1'), # has traitlets
 ]
 
 use_pip = True
--- a/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
@@ -15,7 +15,7 @@ builddependencies = [('binutils', '2.35')]
 
 dependencies = [
     ('Python', '3.8.6'),
-    ('IPython', '7.18.1'), # has traitlets
+    ('IPython', '7.18.1'),  # has traitlets
 ]
 
 use_pip = True
--- a/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
@@ -22,8 +22,4 @@ use_pip = True
 download_dep_fail = True
 sanity_pip_check = True
 
-# use matplotlib-inline backend as default
-# see https://matplotlib.org/tutorials/introductory/usage.html#what-is-a-backend
-modextravars = {'MPLBACKEND': 'module://matplotlib_inline.backend_inline'}
-
 moduleclass = 'vis'
--- a/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
+++ b/easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
@@ -13,7 +13,7 @@ checksums = ['a04bfba22e0d1395479f866853ec1ee28eea1485c1d69a6faf00dc3e24ff34ee']
 
 dependencies = [
     ('Python', '3.8.6'),
-    ('IPython', '7.18.1'), # has traitlets
+    ('IPython', '7.18.1'),
 ]
 
 use_pip = True


jfgrimm:

jfgrimm:
Test report by @jfgrimm
**SUCCESS**
node060.pri.viking.alces.network - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6138 CPU @ 2.00GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/897c7da7ab89c88ed4b26155081ba175 for a full test report.

easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,29 @@
...
+toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
+
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['a04bfba22e0d1395479f866853ec1ee28eea1485c1d69a6faf00dc3e24ff34ee']
+
+builddependencies = [('binutils', '2.35')]
+
+dependencies = [
+    ('Python', '3.8.6'),
+    ('IPython', '7.18.1'),  # has traitlets
```
casparvl, line 18: Shouldn't this depend on matplotlib as well?

jfgrimm: > A particularly interesting backend, provided by IPython, is the `inline` backend. This is available only for the Jupyter Notebook and the Jupyter QtConsole.

zdemat: It is a very minimalistic python module that builds without matplotlib. I guess it is rather ipython & jupyter extension and depends more on them than on matplotlib. Moreover matplotlib has many other dependencies and needs to be in MPI toolchains whereas this jupyter extension can stay in GCCcore.

The module is mostly useful for creating jupyter kernels depending on env modules.

zdemat: In particular the original purpose of this module was to help us create env modules based jupyter kernels that behave similarly to shared jupyter installations that does not have desktop backends. Notebooks executed in such environments behave similarly without explicit `matplotlib inline` statements and raising matplotlib desktop windows.

casparvl: Yeah, I had a look at the source code too, and I'm a bit confused: how can `matplotlib` _not_ be a dependency if the first line is `import matplotlib`? And even if that works somehow, is there a use case where you _don't_ combine the use of `matplotlib-inline` with `matplotlib` at runtime?

I see that their installation instructions (on the github) are `pip install notebook matplotlib`. That seems to suggest that `matplotlib` actually includes `matplotlib-inline`. If that's true, _and_ if there is a dependency of `matplotlib-inline` on `matplotlib`, then I don't really see the use of a seperate EasyConfig - unless I'm missing something...? Sure, you have this environment variable, but that seems more like a configuration thing. As far as I understand you want to set `MPLBACKEND` so that `matplotlib inline` is the default, and you don't have to explicitely execute the `%matplotlib inline` magic, right? If `matplotlib-inline` indeed comes with `matplotlib`, then I'd advise to (locally) modify the `matplotlib` EasyConfig and add the setting of the `MPLBACKEND` variable there (note: if you want to avoid holding/maintaining your own copies of EasyConfigs, you could also do this with a hook that you pass to the `--hooks` argument).

casparvl: But, please let me know if I'm overlooking something of course. I might just not properly understand what you're trying to do here :)

zdemat: I actually agree with most of the arguments, however I still think it is not clear if matplotlib_inline is part of matplotlib or IPython. A sentence at [matplotob_inline on GitHub](https://github.com/ipython/matplotlib-inline) says: *"This package is included in IPython and can be used in a Jupyter Notebook: ..."* However this is not important.

The issue I have not discussed yet is that setting `MPLBACKEND=module://matplotlib_inline.backend_inline` will result in `ModuleNotFoundError: No module named 'matplotlib_inline'` if matplotlib_inline module is not explicitly  installed. I saw my colleague is using exactly this method on a different production system. However I checked on Google Colab and they have `MPLBACKEND=module://ipykernel.pylab.backend_inline` which likely does not require any additional special module and has a similar effect. Maybe it was the meaning of a short message from @jfgrimm.

With the alternative settings, this module is really not needed and I should find a different configuration way how to *hook* the right MPLBACKEND variable into the jupyter kernel.

Thanks for carefully checking the easyconfig and a productive discussion.

casparvl: I checked on our system, and I don't get `ModuleNotFoundError: No module named 'matplotlib_inline'`. For me, it just picks it up form the `IPython` installation. I had the module `IPython/7.25.0-GCCcore-10.3.0` loaded, but no module for `matplotlib`:

```
%matplotlib inline
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
/tmp/ipykernel_10637/352233182.py in <module>
----> 1 get_ipython().run_line_magic('matplotlib', 'inline')

/sw/arch/Debian10/EB_production/2021/software/IPython/7.25.0-GCCcore-10.3.0/lib/python3.9/site-packages/IPython/core/interactiveshell.py in run_line_magic(self, magic_name, line, _stack_depth)
   2346                 kwargs['local_ns'] = self.get_local_scope(stack_depth)
   2347             with self.builtin_trap:
-> 2348                 result = fn(*args, **kwargs)
   2349             return result
   2350 

/sw/arch/Debian10/EB_production/2021/software/Python/3.9.5-GCCcore-10.3.0/lib/python3.9/site-packages/decorator.py in fun(*args, **kw)
    230             if not kwsyntax:
    231                 args, kw = fix(args, kw, sig)
--> 232             return caller(func, *(extras + args), **kw)
    233     fun.__name__ = func.__name__
    234     fun.__doc__ = func.__doc__

/sw/arch/Debian10/EB_production/2021/software/IPython/7.25.0-GCCcore-10.3.0/lib/python3.9/site-packages/IPython/core/magic.py in <lambda>(f, *a, **k)
    185     # but it's overkill for just that one bit of state.
    186     def magic_deco(arg):
--> 187         call = lambda f, *a, **k: f(*a, **k)
    188 
    189         if callable(arg):

/sw/arch/Debian10/EB_production/2021/software/IPython/7.25.0-GCCcore-10.3.0/lib/python3.9/site-packages/IPython/core/magics/pylab.py in matplotlib(self, line)
     97             print("Available matplotlib backends: %s" % backends_list)
     98         else:
---> 99             gui, backend = self.shell.enable_matplotlib(args.gui.lower() if isinstance(args.gui, str) else args.gui)
    100             self._show_matplotlib_backend(args.gui, backend)
    101 

/sw/arch/Debian10/EB_production/2021/software/IPython/7.25.0-GCCcore-10.3.0/lib/python3.9/site-packages/IPython/core/interactiveshell.py in enable_matplotlib(self, gui)
   3515         """
   3516         from IPython.core import pylabtools as pt
-> 3517         from matplotlib_inline.backend_inline import configure_inline_support
   3518         gui, backend = pt.find_gui_and_backend(gui, self.pylab_gui_select)
   3519 

/sw/arch/Debian10/EB_production/2021/software/IPython/7.25.0-GCCcore-10.3.0/lib/python3.9/site-packages/matplotlib_inline/backend_inline.py in <module>
      4 # Distributed under the terms of the BSD 3-Clause License.
      5 
----> 6 import matplotlib
      7 from matplotlib.backends.backend_agg import (  # noqa
      8     new_figure_manager,

ModuleNotFoundError: No module named 'matplotlib'
```

This proves two things to me:
1. `matplotlib-inline` is installed as part of `IPython` (as you can see from the directory structure `/sw/arch/Debian10/EB_production/2021/software/IPython/7.25.0-GCCcore-10.3.0/lib/python3.9/site-packages/matplotlib_inline/backend_inline.py` in the stacktrace), at least for the version of IPython that we have installed.
2. `matplotlib` is a runtime (but not install time) dependency for `matplotlib-inline` (since I get `ModuleNotFoundError: No module named 'matplotlib'` when running the magic)

As an additional test, I loaded the `matplotlib` module, set your environment variable, and then created a plot:

```
import matplotlib.pyplot as plt
plt.plot([1, 2, 3, 4])
plt.ylabel('some numbers')
plt.show()
<Figure size 640x480 with 1 Axes>
```
I only get a plot of the figure after I still run execute the matplotlib magic... I think it might be related to this discussion: https://github.com/ipython/ipython/issues/12190

In any case, if setting that env-var somehow does somehow work for you, I'd recommend doing it on the `matplotlib` module, by using a hook like:

```
#my_matplotlib_hook.py
def parse_hook(ec, *args, **kwargs):
    if ec.name == 'matplotlib':
        ec['modextravars']['MPLBACKEND'] = 'module://matplotlib_inline.backend_inline'
```

and then you'd install with e.g. `eb matplotlib-3.4.2-foss-2021a.eb --hooks=my_matplotlib_hook.py`. That should give you a matplotlib module that sets this environment variable. This _should_ the default `MPLBACKEND` that is already set by the EasyConfig for matplotlib ([here](https://github.com/easybuilders/easybuild-easyconfigs/blob/cd502a79f67c2b4d3e910d6c21d6fd341fb80977/easybuild/easyconfigs/m/matplotlib/matplotlib-3.4.2-foss-2021a.eb#L62)).

casparvl: @jfgrimm to have one more independent datapoint: can you check that `lib/python3.9/site-packages/matplotlib_inline/backend_inline.py` is also present in your installation of IPython? If so, I'd say that `matplotlib-inline` as a seperate EasyConfig (i.e. to install as a seperate package) might not make much sense, right?

jfgrimm: @casparvl I see it for the `GCCcore-10.3.0` generation version, but not for the `GCCcore-10.2.0` generation easyconfig

casparvl: Oh, I can see why: the `IPython` EasyConfig is a bundle, and it includes `matplotlib-inline` for `IPython-7.25.0-GCCcore-10.3.0.eb`, see [here](https://github.com/easybuilders/easybuild-easyconfigs/blob/cd502a79f67c2b4d3e910d6c21d6fd341fb80977/easybuild/easyconfigs/i/IPython/IPython-7.25.0-GCCcore-10.3.0.eb#L136). But, indeed, the EasyConfig for `IPython-7.18.1-GCCcore-10.2.0` [does not include that](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/i/IPython/IPython-7.18.1-GCCcore-10.2.0.eb).

So... essentially: accepting this EasyConfig and loading `matplotlib-inline/0.1.3-GCCcore-10.2.0` + `IPython/7.18.1-GCCcore-10.2.0` would give exactly the same set of python apcagkes as `IPython/7.25.0-GCCcore-10.3.0`. Since the latter also includes `matplotlib-inline` _without_ having a dependency on `matplotlib` itself, I say we should be consistent and accept it here as well. It remains a bit debateable, but if someone objects in the future then it should be changed not only for this PR, but also for `IPython-7.25.0-GCCcore-10.3.0.eb`.

easybuild/easyconfigs/m/matplotlib-inline/matplotlib-inline-0.1.3-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,29 @@
...
+]
+
+use_pip = True
+download_dep_fail = True
+sanity_pip_check = True
+
+# use matplotlib-inline backend as default
+# see https://matplotlib.org/tutorials/introductory/usage.html#what-is-a-backend
+modextravars = {'MPLBACKEND': 'module://matplotlib_inline.backend_inline'}
```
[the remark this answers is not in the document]
zdemat: I am ok with removing the `MPLBACKEND` variable. If the inline backend is already an integral part of IPython or matplotlib, moreover the newer toolchains do not need this module as matplotlib-inline is already included in other easyconfigs there, then I will have to find other way of setting the variable for jupyter kernels in envkernel. So this `matplotlib-inline` easyconfig is finally only supplying a small additional module missing in `IPython-GCCcore-10.2.0`.

casparvl:
Just to summarize the conclusion of our discussion in the review:

`IPython-7.25.0-GCCcore-10.3.0.eb` includes the `matplotlib-inline` extension, but `IPython-7.18.1-GCCcore-10.2.0.eb` does not. This PR adds the missing `matplotlib-inline` to that toolchain generation.

casparvl (approved):
LGTM!

casparvl:
Going in, thanks @zdemat!

PR #15153, opened 2022-03-22
Ensure Bazel 3.7.2 respects $TMPDIR

jfgrimm:
Haven't checked older versions; I assume that most versions will probably be affected

(created using `eb --new-pr`)


jfgrimm:

branfosj:
> Haven't checked older versions; I assume that most versions will probably be affected

0.20.0 and 2.0.0 both have the same line, but, annoyingly, the patch does not work for either. I am thinking that it would be better to fix this in the easyblock.

jfgrimm:
> > Haven't checked older versions; I assume that most versions will probably be affected
> 
> 0.20.0 and 2.0.0 both have the same line, but, annoyingly, the patch does not work for either. I am thinking that it would be better to fix this in the easyblock.

Makes sense, I'll have a look at the easyblock tomorrow

jfgrimm:
@branfosj I've opened an easyblocks PR https://github.com/easybuilders/easybuild-easyblocks/pull/2703

PR #15168, opened 2022-03-23
{bio[foss/2020b]} Mothur v1.46.1

sdx23:
Updated to current version and newer toolchain

### DIFF
--- a/easybuild/easyconfigs/m/Mothur/Mothur-1.46.1-foss-2020b.eb
+++ b/easybuild/easyconfigs/m/Mothur/Mothur-1.46.1-foss-2020b.eb
@@ -18,7 +18,7 @@ toolchainopts = {'usempi': True, 'cstd': 'c++11'}
 
 source_urls = ['https://github.com/mothur/mothur/archive/']
 sources = ['v%(version)s.tar.gz']
-checksums = ['29b500b3c92d726cde34922f697f2e47f0b7127d76d9a6fb167cc2b8ba3d00fd']    # mothur-1.46.1.tar.gz
+checksums = ['29b500b3c92d726cde34922f697f2e47f0b7127d76d9a6fb167cc2b8ba3d00fd']    # v1.46.1.tar.gz
 
 dependencies = [
     ('Python', '3.8.6'),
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/m/Mothur/Mothur-1.46.1-foss-2020b.eb
```diff
@@ -0,0 +1,34 @@
...
+homepage = 'https://www.mothur.org/'
+description = """Mothur is a single piece of open-source, expandable software
+ to fill the bioinformatics needs of the microbial ecology community."""
+
+toolchain = {'name': 'foss', 'version': '2020b'}
+toolchainopts = {'usempi': True, 'cstd': 'c++11'}
+
+source_urls = ['https://github.com/mothur/mothur/archive/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['29b500b3c92d726cde34922f697f2e47f0b7127d76d9a6fb167cc2b8ba3d00fd']    # mothur-1.46.1.tar.gz
```
smoors, line 21: @sdx23 the comment is not correct.

please re-add the checksum with `eb --inject-checksums --force <easyconfig>` to automatically get the correct comment

smoors (changes requested):

easybuild/easyconfigs/m/Mothur/Mothur-1.46.1-foss-2020b.eb
sdx23: Can you please elaborate? I saw such comments in other easyconfigs and found them quite useful, so I included it.

When I perform ``eb --inject-checksums --force Mothur-1.46.1-foss-2020b.eb``, it deletes the comment and following lines including the dependencies (This is EasyBuild 4.5.0 (framework: 4.5.0, easyblocks: 4.5.0)).

smoors: you're right, using `--inject-checksums` removes the comment, but that is because there is only one source, so there can be no confusion. if there is more than one source and/or patch, the comments are automatically added.

in your case though, the comment is wrong: the source file is `v1.46.1.tar.gz`, not `mothur-1.46.1.tar.gz`

boegel:

PR #15183, opened 2022-03-24
{phys}[foss/2022a] TRIQS v3.1.1 and v3.1.0 of TRIQS-cthyb, TRIQS-dft_tools and TRIQS-tprf

Wentzell:


Wentzell:

branfosj:
@boegelbot please test @ generoso / jsc-zen2

easybuild/easyconfigs/t/TRIQS/TRIQS-3.1.0-foss-2021b.eb
```diff
@@ -0,0 +1,70 @@
...
+    ('CMake', '3.22.1')
+]
+
+separate_build_dir = True
+
+# Remove installation directory before building. This fixes problems with
+# failing builds in the presence of preexisting installation.
+preconfigopts = "rm -rf %(installdir)s && "
+
+configopts = ""
```
smoors, line 44: @Wentzell you removed the configopts here because `-DBuild_Deps=Always` is now the default.

looking at the dependencies that are built:
```
Cpp2Py
GTest
itertools
mpi
h5
nda
```
I'm wondering if this can be changed to `-DBuild_Deps=IfNotFound`?
HDF5 and MPI are already included as deps, and googletest (GTest) can be added easily too.

Wentzell: Dear @smoors,

Yes we have switched to building the deps listed in [triqs/deps/CMakeLists.txt](https://github.com/TRIQS/triqs/blob/e24f9c8d38b4fea199a17cd240213803be81d629/deps/CMakeLists.txt#L63-L97) by default.
Note however that h5 and mpi refer to [TRIQS/h5](https://github.com/TRIQS/h5) and [TRIQS/mpi](https://github.com/TRIQS/mpi) which are only C++ Interfaces to the HDF5 and MPI C Libraries that are still searched for in the build environment. We don't won't to package those interfaces at the moment.

For GTest, which is only a build-time and not a runtime dependency, we would like to stick with the recommendation  to build it together with the project being tested.

Did the full build-test succeed? It seems the bot didn't report any test results?

smoors: thanks for the reply.
can you remove the `configopts` as it's an empty string?

not sure why the bot did not report back. I'll rerun after you've made the change.

Wentzell: Sure, I just changed this!

smoors:

easybuild/easyconfigs/p/python-igraph/python-igraph-0.9.8-foss-2021b.eb
```diff
@@ -17,7 +17,7 @@ dependencies = [
     ('Python', '3.9.6'),
     ('igraph', '0.9.5'),
     ('PyCairo', '1.20.1'),
-    ('Clang', '12.0.1'),
+    ('Clang', '13.0.1'),
```
boegel, line 20: @Wentzell Can you clarify why this change was made?

This will cause trouble when this easyconfig is combined with others from the same generation that depend on Clang/12.0.1, like `pocl-1.8-GCC-11.2.0.eb`

Wentzell: This change was made because two variants of Clang dependencies were found in the dependency resolution (see [log](https://github.com/easybuilders/easybuild-easyconfigs/runs/5684084504?check_suite_focus=true#step:8:146))

```
found 2 variants of 'Clang' dependency in easyconfigs using '2021b' toolchain generation
* version: 12.0.1; versionsuffix:  as dep for set(['leidenalg-0.8.8-foss-2021b.eb', 'python-igraph-0.9.8-foss-2021b.eb', 'scanpy-1.8.2-foss-2021b.eb'])
* version: 13.0.1; versionsuffix:  as dep for set(['TRIQS-tprf-3.1.0-foss-2021b.eb', 'TRIQS-cthyb-3.1.0-foss-2021b.eb', 'TRIQS-3.1.0-foss-2021b.eb', 'TRIQS-dft_tools-3.1.0-foss-2021b.eb'])
```

TRIQS depends on Clang-13 through the [Clang-Python-bindings-13.0.1](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/Clang-Python-bindings/Clang-Python-bindings-13.0.1-GCCcore-11.2.0.eb) dependency. The reason we chose 13 here is that TRIQS itself requires Clang v13+ to compile.

I checked at the time that I created the PR and `python-igraph` was the only `GCC-11.2.0` and `foss-2021b` easyconfig to explicitly list Clang as a dependency, so I updated this.

Would it be an option to update also the `pocl` or would this lead to other potential problems?

Wentzell: @boegel Please let me know how to proceed! It would be great to get this merged for the next easybuild release.

Wentzell:
@smoors Could you trigger a full build for this again?

robert-mijakovic:

Wentzell:
Thank you for triggering the build! Looks like everything is passing.
It would be great if this could either be merged or you could provide additional feedback.

smoors:
@Wentzell updating easyconfigs that are already released (as part of an EB release) should be avoided as much as possible, unless for bug fixes or to fix a broken installation.

I see 3 possible options going forward:

- move everything to the 2022a toolchain generation. this is the most straightforward solution, but it may require a lot of work from your side.
- make Clang-13 a build dependency instead of a dependency. this would then require users to separately load a Clang module to get the full functionality of the C++2Py Tool. looking at the packaged installation instructions for ubuntu, clang is installed with a simple `sudo apt-get`, which in ubuntu 20.04 is Clang-10, so this still gives the option of using an older Clang at runtime if needed for compatibility.
- add an exception for using Clang-13 in the TRIQS modules, and add a `-Clang-13.0.0` versionsuffix. this is of course also something we'd like to avoid if possible, so I'm not sure if this will be accepted.

what do you think?

would be good to get a second opinion on this, @boegel or @Micket ?

easybuild/easyconfigs/t/TRIQS-cthyb/TRIQS-cthyb-3.1.0-foss-2022a.eb
```diff
@@ -0,0 +1,69 @@
...
+
+toolchain = {'name': 'foss', 'version': '2022a'}
+toolchainopts = {'pic': True, 'usempi': True}
+
+source_urls = ['https://github.com/TRIQS/cthyb/releases/download/%(version)s/']
+sources = ['cthyb-%(version)s.tar.gz']
+checksums = ['8d6d2c4d5b3928d062b72fad4ea9df9aae198e39dd9c1fd3cc5dc34a5019acc0']
+
+dependencies = [
+    ('Python', '3.9.6'),
```
smoors, line 31:
```suggestion
    ('Python', '3.10.4'),
```

easybuild/easyconfigs/t/TRIQS-cthyb/TRIQS-cthyb-3.1.0-foss-2022a.eb
```diff
@@ -0,0 +1,69 @@
...
+    ('SciPy-bundle', '2022.05'),
+    ('Boost', '1.79.0'),
+    ('Clang', '13.0.1'),
+    ('Clang-Python-bindings', '13.0.1'),
+    ('GMP', '6.2.1'),
+    ('HDF5', '1.13.1'),
```
smoors, line 37: odd versions of HDF5 are development versions, stick to the stable even versions:
```suggestion
    ('HDF5', '1.12.2'),
```

Wentzell:
@smoors I updated everything to 2022a now. Please have a look and let me know if things look good.
Also, please go ahead and trigger a full build.

boegel:
@Wentzell There are still in-place updates in existing easyconfigs in this PR, see `pocl-1.8-GCC-11.2.0.eb` for example.

Are those really needed?

smoors:
@Wentzell as @boegel mentioned, I think pocl and python-igraph can now be removed from this PR?

Wentzell:
@smoors You are correct, I had missed to revert these changes. This is now done!

smoors:

Wentzell:
@smoors Fixed the NFFT 3.5.3 sha256 which was the reason for the first build failure

boegel:

boegel (approved):
lgtm

boegel:
Going in, thanks @Wentzell!

PR #790, opened 2014-03-29
IsoLasso-2.6.1-goolf-1.4.10.eb + patch + deps

pescobar:
It's working in my side but I had to remove the Qt dependency in CGAL-4.0-goolf-1.4.10-Python-2.7.5.eb because I was getting an error about "QGLWidget: No such file or directory" and anyway I don't need Qt support in CGAL

PR #15206, opened 2022-03-30
fix installation with setuptools>=61 by explicitly declaring there are no Python packages

branfosj:
fixes [#790](https://github.com/easybuilders/easybuild/issues/790)

Tested with:
```
$ python3 -m venv
$ source venv/bin/activate
$ cd easybuild-framework/
$ pip install .
$ cd ../easybuild-easyblocks/
$ pip install .
$ cd ../easybuild-easyconfigs/
$ pip install .
```

And `$ eb --search pytorch` shows expected output.

Two tests with:
* updating setuptools to latest (61.2.0)
* using EB installed Python 3.9.5, which has 56.0.0
* using OS Python 3.6.8 which has 39.2.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,6 +101,7 @@ versions, etc.).""",
         "Programming Language :: Python :: 3.9",
         "Topic :: Software Development :: Build Tools",
     ],
+    packages=[],
     platforms="Linux",
     requires=[
         "easybuild_framework(>=%s.0)" % MAJ_VER,


branfosj:
We might want to add a test install github action - for several python versions:
* update setuptools and pip
* install framework develop
* install blocks develop
* install easyconfigs PR

The downside of such testing is that it will cause a failure of PRs that is unrelated to the contents of the PR.

boegel:
@branfosj We could test this in a separate workflow, that runs daily (rather than running it on PRs)

boegel (approved):
I compared the source tarball that's created with `python3 setup.py sdist` with and without this change (using setuptools 57.4.0), they're identical (other than the `setup.py` itself), so good to go...

Thanks @branfosj!

akesandgren:
Does this solve https://github.com/easybuilders/easybuild-framework/issues/3984 too?

PR #15316, opened 2022-04-21
Remove pkg-config use from SeqLib configure patch

bartoldeman:
There was a syntax error in the autoconf macro, which causes an
error if pkg-config is available, and if not available simply
causes configure to issue a warning and ignore it.

Since the result is not used anyway, we can remove it.


jfgrimm:

jfgrimm:

jfgrimm (approved):
LGTM

PR #15375, opened 2022-04-29
{bio}[GCC/11.2.0] GMAP-GSNAP-2021-21-17

wdlingit:
added GMAP-GSNAP-2021-21-17-GCC-11.2.0.eb

PR #15376, opened 2022-04-29
{bio}[foss/2021a] rnaQUAST v2.2.2 w/ Python 3.9.6

wdlingit:
added rnaQUAST-2.2.2-foss-2021a-Python-3.9.6.eb

requires:

- [x] #15375

### DIFF
--- a/easybuild/easyconfigs/r/rnaQUAST/rnaQUAST-2.2.2-foss-2021b.eb
+++ b/easybuild/easyconfigs/r/rnaQUAST/rnaQUAST-2.2.2-foss-2021b.eb
@@ -2,7 +2,6 @@ easyblock = 'Tarball'
 
 name = 'rnaQUAST'
 version = '2.2.2'
-versionsuffix = '-Python-%(pyver)s'
 
 homepage = 'https://cab.spbu.ru/software/rnaquast/'
 description = """rnaQUAST is a tool for evaluating RNA-Seq assemblies using reference genome and gene database.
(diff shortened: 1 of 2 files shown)


wdlingit:
this one is depending on  #15375

migueldiascosta:
@wdlingit not sure if you meant to use `foss/2021b` with those versions of dependencies (except `GMAP-GSNAP`, which should then be the one mentioned in the comment above) or `foss/2021a`, in which case the versions of `Python`, `matplotlib` and `BLAST+` should be different

also note that in either case, the Python `versionsuffix` shouldn't be necessary anymore

wdlingit:
@migueldiascosta Thank you for pointing out this. I am confusing about the `versionsuffix`. Sorry to all that this should be foss/2021b. I mistakenly built the easyconfig of wrong version while I was updating this one.

migueldiascosta:
@wdlingit no worries, it can be confusing. the Python `versionsuffix` is no longer necessary when using the default version of Python for that toolchain.

also just noticed that upstream seems to have tagged `v.2.2.2` (note the extra dot after `v`, so this needs to be changed in `sources`) and the checksum is different?

wdlingit:
@migueldiascosta I have updated the file name and its contents ("rnaQUAST with corrected foss ersion" (typo again)). This one should be the one I used to build my programs.

easybuild/easyconfigs/r/rnaQUAST/rnaQUAST-2.2.2-foss-2021b.eb
```diff
@@ -0,0 +1,68 @@
+easyblock = 'Tarball'
+
+name = 'rnaQUAST'
+version = '2.2.2'
+versionsuffix = '-Python-%(pyver)s'
```
migueldiascosta, line 5:
```suggestion

```

wdlingit: @migueldiascosta Done and thank you!

migueldiascosta:

migueldiascosta (approved):
lgtm

migueldiascosta:
Going in, thanks @wdlingit!

PR #15416, opened 2022-05-04
{devel}[GCC/11.2.0] Autotools v20200321

wdlingit:
In some rare cases, autotools downgrade seems a solution. For example: https://karobben.github.io/2020/07/28/Bioinfor/Trinity/#Errors


boegel:
Closing this since `GCC/11.2.0` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @wdlingit 

PR #15482, opened 2022-05-10
{data}[foss/2020b] TomoPy v1.11.0, DXchange v0.1.8

jhein32:
New packages TomoPy and DXchange


jhein32:

SebastianAchilles:

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
```diff
@@ -0,0 +1,53 @@
...
+    ('scikit-build', '0.11.1'),
+    ('CMake', '3.18.4'),
+    ('Ninja', '1.10.1'),
+]
+
+use_pip = True
+download_dep_fail = True
+sanity_pip_check = True
+
+sources = [{
```
Micket, line 40: I would prefer the more conventional order sources -> builddeps -> deps

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
```diff
@@ -0,0 +1,53 @@
...
+    'filename': '%(name)s-%(version)s.tar.gz',
+    'git_config': {
+        'url': 'https://github.com/tomopy',
+        'tag': '%(version)s',
+        'recursive': True,
```
Micket, line 45: I would prefer to see an external PTL, with the benefit that it would actually be useful elsewhere, it would have a tagged release, it would remove the need to do a git checkout here so we can actually have checksums.

Seems quote easily doable with TOMOPY_EXTERNAL_LIBRARIES and TOMOPY_EXTERNAL_INCLUDE_DIRS.

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
```diff
@@ -0,0 +1,53 @@
...
+    ('scikit-image', '0.18.1'),
+    ('OpenCV', '4.5.1', '-contrib'),
+    ('imkl', '2020.4.304'),
+    ('numexpr', '2.8.1'),
+    ('DXchange', '0.1.8'),
+    ('PyWavelets', '1.1.1'),
+]
+
+builddependencies = [
+    ('scikit-build', '0.11.1'),
```
Micket, line 31: I suspect this isn't actually used at all? https://github.com/tomopy/tomopy/blob/master/CMakeLists.txt#L12 since no such parameter is specified? Isn't it just a normal cmake project?

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+toolchain = {'name': 'foss', 'version': '2020b'}
+
+dependencies = [
+    ('Python', '3.8.6'),
+    ('SciPy-bundle', '2020.11'),
+    ('scikit-image', '0.18.1'),
+    ('OpenCV', '4.5.1', '-contrib'),
+    ('imkl', '2020.4.304'),
```
Micket, line 24: Hm, this might a a non issue, but it worries me to mix in another BLAS here.

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
```diff
@@ -0,0 +1,53 @@
...
+    ('numexpr', '2.8.1'),
+    ('DXchange', '0.1.8'),
+    ('PyWavelets', '1.1.1'),
```
Micket, line 27: I can't actually find that these are actually dependencies of tomopy looking through setup.py and CMakeLists

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
```diff
@@ -0,0 +1,53 @@
...
+builddependencies = [
+    ('scikit-build', '0.11.1'),
+    ('CMake', '3.18.4'),
+    ('Ninja', '1.10.1'),
+]
+
+use_pip = True
+download_dep_fail = True
+sanity_pip_check = True
+
```
Micket, line 39: Looks like there are tests. Can we run them? Also, is there a sanity check command we can try?

Micket (changes requested):

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
zdemat: Hi @Micket, thanks for the comment. When we started thinking about this software we were aware of the fact that Conda is the most common distribution for this software. There is some sw engineering on [TomoPy](https://github.com/tomopy/tomopy) so I guess you are right and TomoPy can be build and installed without some dependencies. On the other side TomoPy is that type of software that it installs without some dependencies but some functionality is missing. We have tried to create a similar comfortable installation as with Conda so when defining dependencies we were looking also in its conda env recipes as e.g. [linux-38.yml](https://github.com/tomopy/tomopy/blob/master/envs/linux-38.yml). In particular `DXchange` is practically an I/O module for TomoPy. And it is very comfortable for the user to have it as the user can run immediately e.g. the [tooth example](https://github.com/tomopy/tomopy/blob/master/doc/source/ipynb/tomopy.ipynb).

So I suggest to keep these dependencies. An alternative is that we provide something like TomoPy-suite but that makes things just complicated.

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
zdemat: Hi @Micket , I think, based on a personal discussion with the scientific developer long time ago, it is opposite and they are heavily relying on `scikit-build` and are actually managing the build process with it. It is again based on TomoPy [azure-pipelines.yml](https://github.com/tomopy/tomopy/blob/1.11.0/azure-pipelines.yml).

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
zdemat: Hi @Micket , thanks for the comment. Again we were following the [azure-pipelines](https://github.com/tomopy/tomopy/blob/1.11.0/azure-pipelines.yml) and I think they have tests and even a special performance benchmark included but it seems to be based on [pyctest](https://github.com/jrmadsen/pyctest) and partially maintained by [Argonne-National-Laboratory](https://github.com/Argonne-National-Laboratory). I was struggling with `pyctest` for days and I did not mange to do anything with that so we dropped it as a dependency. Contrary simple python import tests helped me to identify bugs even in binary components.

So I would suggest to stay simple and do not add complex test dependencies.

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
zdemat: @Micket and @jhein32 

from what I know, [TomoPy](https://github.com/tomopy/tomopy) is heavily relaying on FT. They were using `fftw` long time ago but later they switched to `mkl_fft` as you can see also here in their conda env definition, e.g. [linux-38.yml](https://github.com/tomopy/tomopy/blob/master/envs/linux-38.yml). Note there is also `nomkl` version relying on conda `nomkl` (?).

Note also that from a discussion with leading scientific developers, it is known TomoPy has mkl optimizations included. It can even profit from the intel compiler. However we were preferring to submit the foss EasyConfigs first as having it in the foss toolchains allows to combine TomoPy with more sw.

easybuild/easyconfigs/a/astropy/astropy-4.2.1-foss-2020b.eb
```diff
@@ -0,0 +1,40 @@
...
+    }),
+    (name, version, {
+        'checksums': ['ed483e472241153daec45f4b0c318c2c63d9f47305b78e6e63d32fc388c18427'],
+    }),
+]
+
+sanity_check_paths = {
+    'files': [],
+    'dirs': ['lib/python%(pyshortver)s/site-packages/astropy'],
+}
```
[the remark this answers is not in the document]
zdemat: Hi @jhein32 

I did several comments to the changes suggest by @Micket. I agree many comments from @Micket are very relevant. Fixes for many of them are quite straightforward. I would go ahead with them but I guess it is easier if you push the changes.

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
jhein32: I am not sure I fully understand this.  I discussed in the slack that TomoPy will not build from a github source.   One needs a full fledged clone.   I you try from a github source you get an error message that suggest you to build from a clone.

@Micket Let me know if I didn't understand your request.

Micket: I'm talking about the order of things in this file.

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
Micket: The only reason i see for using the git checkout is because it uses a git submodule for the PTL dependency, which, like many other software projects do, they incorrectly "bundle" in their sources. 
Just like we do with lots of other software, we configure/patch the softwares to use this dependency externally instead. I.e. 1. make another easyconfig for PTL
2. switch to sources without PTL "bundled" (thus gaining the benefits of using a proper release than git checkout, meaning we can now also use checksums) 
3. configure/patch TomoPy to use it's PTL dependency externally instead.

zdemat: @jhein32 I guess the PTL dependency is in [tomopy/source](https://github.com/tomopy/tomopy/tree/1.11.0/source) directory and it is this project: [jrmadsen/PTL](https://github.com/jrmadsen/PTL). I guess this we can try to fix. That is really a serious bug.

easybuild/easyconfigs/t/TomoPy/TomoPy-1.11.0-foss-2020b.eb
jhein32: To test, I removed the scikit-build from the config and rebuild TomoPy.  So the dependency section looks like:

```
builddependencies = [
    ('CMake', '3.18.4'),
    ('Ninja', '1.10.1'),
]


```

and I get the error as below:

```
    ERROR: Command errored out with exit status 1:
     command: /sw/easybuild/software/Python/3.8.6-GCCcore-10.2.0/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/eb-fpX2V6/pip-req-build-x4mbpfab/\
setup.py'"'"'; __file__='"'"'/tmp/eb-fpX2V6/pip-req-build-x4mbpfab/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', \
'"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/eb-fpX2V6/pip-pip-egg-info-dxvpkp5o
         cwd: /tmp/eb-fpX2V6/pip-req-build-x4mbpfab/
    Complete output (5 lines):
      File "/tmp/eb-fpX2V6/pip-req-build-x4mbpfab/setup.py", line 9, in <module>
        from skbuild import setup
    ModuleNotFoundError: No module named 'skbuild'
```

So it is needed in our current set up.

jfgrimm: https://tomopy.readthedocs.io/en/latest/devguide.html#building-tomopy

boegel:

PR #15496, opened 2022-05-12
{vis}[GCCcore/11.2.0] PGPLOT v5.2.2 

sassy-crick:
(created using `eb --new-pr`)

PR #15497, opened 2022-05-12
{geo}[GCC/11.2.0] WCSLIB v7.11

sassy-crick:
(created using `eb --new-pr`)

Requires:
- [x] PR #15496


sassy-crick:
I tried to use a sanity check alike `sanity_check_commands = ['wcsware --help &> /dev/null']` but the return value was always *1*. The same happened when I used `--version` instead. 
Also, this program is the reason why only static libs are build in #15496 as with the dynamic ones, the binary does not find it at execution time. Open for suggestions though.

easybuild/easyconfigs/w/wcslib/wcslib-7.11-GCC-11.2.0.eb
```diff
@@ -0,0 +1,41 @@
...
+]
+
+dependencies = [
+    ('CFITSIO', '3.49'),
+    ('pgplot', '5.2.2'),
+    ('X11', '20210802'),
+]
+
+configure_cmd = './configure --with-cfitsiolib=$EBROOTCFITSIO/lib --with-cfitsioinc=$EBROOTCFITSIO/include '
+configure_cmd += '--with-pgplotlib=$EBROOTPGPLOT/lib --with-pgplotinc=$EBROOTPGPLOT/include '
```
jfgrimm, line 34: configure options should go in `configopts`, no need to change the `configure_cmd` itself
```suggestion
configopts = '--with-cfitsiolib=$EBROOTCFITSIO/lib --with-cfitsioinc=$EBROOTCFITSIO/include '
configopts += '--with-pgplotlib=$EBROOTPGPLOT/lib --with-pgplotinc=$EBROOTPGPLOT/include '
```

jfgrimm:
> Also, this program is the reason why only static libs are build in https://github.com/easybuilders/easybuild-easyconfigs/pull/15496 as with the dynamic ones, the binary does not find it at execution time. Open for suggestions though.

it looks like the problem is that it is actually looking for `libcpgplot.so.`/`libpgplot.so.` (note the trailing dot) when linking dynamically.

jfgrimm:
I see now, the patch file for pgplot adds lines like `SHARED_LD_PGPLOT_OPTS="-o $SHARED_LIB -Wl,-soname,libpgplot.so.\$(VM)"` so if `VM` isn't defined when building pgplot, you end up with an unversioned shared object that has a trailing dot

easybuild/easyconfigs/w/wcslib/wcslib-7.11-GCC-11.2.0.eb
```diff
@@ -0,0 +1,41 @@
...
+homepage = 'https://www.atnf.csiro.au/people/mcalabre/WCS/'
+description = """The FITS "World Coordinate System" (WCS) standard defines keywords 
+and usage that provide for the description of astronomical coordinate systems in a 
+FITS image header."""
+
+toolchain = {'name': 'GCC', 'version': '11.2.0'}
+toolchainopts = {'pic': True}
+
+source_urls = ['ftp://ftp.atnf.csiro.au/pub/software/wcslib/']
+sources = ['%(name)s-%(version)s.tar.bz2']
```
jfgrimm, line 18:
```suggestion
sources = [SOURCELOWER_TAR_BZ2]
```

jfgrimm (changes requested):

easybuild/easyconfigs/w/wcslib/wcslib-7.11-GCC-11.2.0.eb
sassy-crick: Thanks. I knew there was a better way of doing this but at the time I could not see the wood of the forest.

easybuild/easyconfigs/w/wcslib/wcslib-7.11-GCC-11.2.0.eb
```diff
@@ -0,0 +1,41 @@
+# Contribution from Imperial College London (UK)
+# Uploaded by J. Sassmannshausen
+
+easyblock = 'ConfigureMake'
+
+name = 'wcslib'
```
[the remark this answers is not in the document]
sassy-crick: Changes, also the name for the sources. Thanks.

sassy-crick:
I tried to use a sanity check alike `sanity_check_commands = ['wcsware --help &> /dev/null']` but the return value was always *1*. The same happened when I used `--version` instead. 
Also, this program is the reason why only static libs are build in #15496 as with the dynamic ones, the binary does not find it at execution time. Open for suggestions though. 

> I see now, the patch file for pgplot adds lines like `SHARED_LD_PGPLOT_OPTS="-o $SHARED_LIB -Wl,-soname,libpgplot.so.\$(VM)"` so if `VM` isn't defined when building pgplot, you end up with an unversioned shared object that has a trailing dot

It is not the first time I did not see the 'dot, so thanks for pointing that out. I have resolved that issue and now it is working as I expect it. Thanks for helping me here!

boegel:

boegel:

boegel (approved):
lgtm

boegel:

PR #15508, opened 2022-05-17
{tools}[system/system] vaspkit v1.3.3

anselmicz:
(created using `eb --new-pr`)


easybuild/easyconfigs/v/vaspkit/vaspkit-1.3.3.eb
```diff
@@ -0,0 +1,33 @@
...
+sources = ['%(namelower)s.%(version)s.linux.x64.tar.gz']
+patches = ['vaspkit-1.3.3-fix-setup-script.patch']
+checksums = [
+    'e428d5e3f3dde5f11fc17770f7c0864a',  # vaspkit.1.3.3.linux.x64.tar.gz
+    'f4f523c04d27c9f2165b140a0c0ccde0',  # vaspkit-1.3.3-fix-setup-script.patch
+]
+extract_sources = True
+
+dependencies = [
+    ('matplotlib', '3.3.3', '', ('foss', '2020b')),
```
boegel, line 26: @anselmicz Although there's no actual compilation going on here, only unpacking a tarball (and applying the patch), it's better to use `foss` as `toolchain` above, rather than having a "hidden" dependency on `foss` through `matplotlib`.

So, I suggest to make 3 changes:

* rename the file to `vaspkit-1.3.3-foss-2020b.eb`
* use `toolchain = {'name': 'foss', 'version': '2020b'}` above
* change this line to `('matplotlib', '3.3.3')`

easybuild/easyconfigs/v/vaspkit/vaspkit-1.3.3.eb
```diff
@@ -0,0 +1,33 @@
...
+]
+
+skipsteps = ['sanitycheck']
+
+modluafooter = 'execute{cmd="bash $EBROOTVASPKIT/setup.sh", modeA={"load"}}'
```
boegel, line 31: Hmm, are you sure this actually works? What does the `setup.sh` script do? Does it change the environment?
Because I don't think that's going to work at all...

Also, what if EasyBuild is configured to generate module files in Tcl syntax instead?

boegel (changes requested):

easybuild/easyconfigs/v/vaspkit/vaspkit-1.3.3.eb
anselmicz: > @anselmicz Although there's no actual compilation going on here, only unpacking a tarball (and applying the patch), it's better to use `foss` as `toolchain` above, rather than having a "hidden" dependency on `foss` through `matplotlib`.
> 
> So, I suggest to make 3 changes:
> 
>     * rename the file to `vaspkit-1.3.3-foss-2020b.eb`
> 
>     * use `toolchain = {'name': 'foss', 'version': '2020b'}` above
> 
>     * change this line to `('matplotlib', '3.3.3')`

sure, no problem

easybuild/easyconfigs/v/vaspkit/vaspkit-1.3.3.eb
anselmicz: > Hmm, are you sure this actually works? What does the `setup.sh` script do? Does it change the environment? Because I don't think that's going to work at all...

This is the vaspkit way of doing stuff, though it might be probably better to export most of the script into the module file itself. However, there's also .vaspkit file created in the /home/$USER directory, which includes the rest of the configuration (which the user will have to change around according to their needs). I'll try to write a patch that would circumvent most of this script.

> Also, what if EasyBuild is configured to generate module files in Tcl syntax instead?

I have no experience with the Tcl, so I cannot attest to this.

anselmicz:
My last update should fix almost all of the addressed issues:

* toolchain has been added and I have fixed the respective matplotlib dependency
* files have been renamed to reflect this
* I delved into the documentation and fixed my sanity check
* sha256sums have been added
* I have made changes to how_to_set_environment_variables and setup.sh files; now there is no need to adjust bashrc

Setup.sh script now only checks if user has .vaspkit file in their home, if not then copies for him a fresh copy. This is the main reason for the `modluafooter`, sadly I have no experience with Tcl and so can't add a Tcl analogy

I have verified (at least some, I'm no vaspkit expert) functionality using examples provided within the tarball.

PR #15515, opened 2022-05-18
{math}[gompi/2021a] Mmg v5.6.0

sassy-crick:
(created using `eb --new-pr`)

PR #15514, opened 2022-05-18
{math}[GCCcore/10.3.0] muParser v2.3.3

sassy-crick:
(created using `eb --new-pr`)

PR #15519, opened 2022-05-18
{numlib}[foss/2021a] PETSc v3.17.0 w/ Python 3.9.5

sassy-crick:
(created using `eb --new-pr`)

Requires:
- [x] PR #15515
- [x] PR #15514


easybuild/easyconfigs/p/PETSc/PETSc-3.17.0-foss-2021a.eb
```diff
@@ -0,0 +1,84 @@
...
+configopts = '--configModules=PETSc.Configure --optionsModule=config.compilerOptions '
+configopts += '--LIBS="-lm -lpthread -lrt" --with-mpi4py=0 '
+configopts += '--CFLAGS="-O2 -ftree-vectorize -march=native -fno-math-errno -fopenmp -fPIC" '
+configopts += '--CXXFLAGS="-O2 -ftree-vectorize -march=native -fno-math-errno -fopenmp -fPIC '
+configopts += '-DOMPI_SKIP_MPICXX -DMPICH_SKIP_MPICXX" --with-mpi=1 --with-build-step-np=4 '
+configopts += '--FFLAGS="-O2 -ftree-vectorize -march=native -fno-math-errno -fopenmp -fPIC" '
+configopts += '--with-shared-libraries=1 --with-debugging=0 --with-pic=1 --with-x=0 '
+configopts += '--with-windows-graphics=0 --with-numpy=1 --with-fftw=1 --with-scalapack=1 '
+configopts += '--with-blaslapack-lib="[/home/sassy/apps/software/FlexiBLAS/3.0.4-GCC-10.3.0/lib/'
+configopts += 'libflexiblas.a,libgfortran.a]" '
```
jfgrimm, line 73: this could do with cleaning up, probably something along the lines of:
```suggestion
configopts += '--LIBS="$LIBS -lrt" --with-mpi4py=0 '
configopts += '--CFLAGS="$CFLAGS" '
configopts += '--CXXFLAGS="$CXXFLAGS '
configopts += '-DOMPI_SKIP_MPICXX -DMPICH_SKIP_MPICXX" --with-mpi=1 --with-build-step-np=4 '
configopts += '--FFLAGS="$FFLAGS" '
configopts += '--with-shared-libraries=1 --with-debugging=0 --with-pic=1 --with-x=0 '
configopts += '--with-windows-graphics=0 --with-numpy=1 --with-fftw=1 --with-scalapack=1 '
configopts += '--with-blaslapack-lib="$BLAS_LAPACK_LIB_DIR"
'
```

easybuild/easyconfigs/p/PETSc/PETSc-3.17.0-foss-2021a.eb
```diff
@@ -0,0 +1,84 @@
...
+    ('METIS', '5.1.0'),
+    ('SCOTCH', '6.1.0'),
+    ('MUMPS', '5.4.0', '-metis'),
+    ('SuiteSparse', '5.10.1', '-METIS-5.1.0'),
+    ('Hypre', '2.21.0'),
+    ('zlib', '1.2.11'),
+    ('BLIS', '0.8.1'),
+    ('Szip', '2.1.1'),
+    ('HDF5', '1.12.1'),
+    ('Eigen', '3.3.9'),
```
jfgrimm, line 46: Eigen should only need to be a build dependency (since it's header-only). I imagine googletest isn't a runtime requirement either

easybuild/easyconfigs/p/PETSc/PETSc-3.17.0-foss-2021a.eb
sassy-crick: Thanks for the suggestion. much appreciated. The "$BLAS_LAPACK_LIB_DIR" is actually not needed, what is needed are the two libs `libflexiblas.a,libgfortran.a` which I replaced with `-lflexiblas -lgfortran` to stick to dynamic libs. 
Please have a look at the log file to make sure I am not accidentally build it with the wrong flags. From what I can see, it is all working as intended but that does not mean it is correct.

easybuild/easyconfigs/p/PETSc/PETSc-3.17.0-foss-2021a.eb
sassy-crick: True. I forgot about `Eigen` being only header. I was quite happy that it all worked after 2 days of playing with it to be honest. :-)

boegel:
@sassy-crick Is there a strict need for a newer version of PETSc, or could you also use the existing `PETSc-3.15.1-foss-2021a.eb`?

sassy-crick:
> @sassy-crick Is there a strict need for a newer version of PETSc, or could you also use the existing `PETSc-3.15.1-foss-2021a.eb`?

Well, the reason I put myself into that was the installation of [FreeFEM](https://github.com/easybuilders/easybuild-easyconfigs/pull/15537) where, when I looked at the 3rd party files, I found that:
```
download('PETSc','https://www.mcs.anl.gov/petsc/mirror/release-snapshots/petsc-3.17.0.tar.gz',
```
So I thought I better stick to what they are using. I agree with you that is not an ideal situation to have two different versions of the same programs in one toolchain. 
To make things more interesting: there is actually PETSc version 3.17.1 out, which in turn requires SLEPC-3.17.1
Of course, all of that also requires `SuperLU` next to the other dependencies where I have created PRs recently. 

I guess one way out of that is to push it up in the toolchain and see if the 3.17.1 versions are ok for that FreeFEM build. How far are we with `foss-2022a` ?

boegel:
> I guess one way out of that is to push it up in the toolchain and see if the 3.17.1 versions are ok for that FreeFEM build. How far are we with `foss-2022a` ?

Since 3.17.1 is a bugfix release, that should be fine for FreeFEM. `foss/2022a` isn't fully settled yet, but how about going with `foss/2021b` instead?

sassy-crick:
> Since 3.17.1 is a bugfix release, that should be fine for FreeFEM. `foss/2022a` isn't fully settled yet, but how about going with `foss/2021b` instead?

There are two things which we need to consider:
- that PR is only part of a larger sets of PRs which are needed for [PR 15537](https://github.com/easybuilders/easybuild-easyconfigs/pull/15537)
- also, I have asked the user to test FreeFEM to see if that is all working. I am waiting for the results to see if that is all working numerically as well. 
 
That said, I am happy to push that up to version 3.17.1 do the same for [PR 15550](https://github.com/easybuilders/easybuild-easyconfigs/pull/15550) (SLEPc) as they are mutually depending on each other in terms of version, and push the rest of the requirements up to `foss2021b` as well. I did not want to go for a too new compiler initially, just to make sure I am not opening yet another can of worms in that quite complex installation. 
Happy with that?

boegel:

sassy-crick:
Closed this PR as it is using a rather old toolchain.

PR #15344, opened 2022-04-26
{perf}[gompi/2021b] OSU-Micro-Benchmarks v5.9

branfosj:
(created using `eb --new-pr`)

* [ ] #14919

PR #15343, opened 2022-04-26
{perf}[gompi/2021a] OSU-Micro-Benchmarks v5.9

branfosj:
(created using `eb --new-pr`)

The NCCL benchmarks were new in OSU MB 5.8.

PR #14919, opened 2022-02-03
PoC: Add groundwork for separating actual cuda parts in OpenMPI. WIP WIP WIP

akesandgren:
(created using `eb --new-pr`)

NOTE: PoC only (even if it actually works)

This shows how we can separate the CUDA parts in OpenMPI in such a way that we can handle OpenMPI the same way we do UCX and UCX-CUDA.

To build the cuda split base version (i.e. OpenMPI) configure with --with-cuda=enable with no CUDA deps
To build the fully cuda enabled OpenMPI-CUDA, configure with --with-cuda=$EBROOTCUDA (as usual) and only install
```
libmca_common_cuda*
mca_btl_smcuda*
mca_pml_ob1*
mca_pml_ucx*
mca_btl_openib*
mca_btl_tcp*
mca_rcache_gpusm*
mca_rcache_grdma*
mca_rcache_rgpusm*
```
into OpenMPI-CUDA.../lib and lib/openmpi
And finally make sure OMPI_MCA_mca_component_path first lists OpenMPI-CUDA's lib and lib/openmpi and the the base OpenMPI's lib and lib/openmpi

The OpenMPI-4.1.1_fix_missing_OPAL_CUDA_GDR_SUPPORT_protection.patch doesn't really have anything to do with this, it's just a fix for an actual bug in OpenMPI that got uncovered due to this.

And CUDA_GDR support is still not handled since that will also need manual intervention during configure.

Depends on: https://github.com/easybuilders/easybuild-easyblocks/pull/2710

PR #15528, opened 2022-05-22
OpenMPI 4.1.1: patch and build --with-cuda=internal

bartoldeman:
Allow building Open MPI with --with-cuda=internal, by providing an
internal minimal cuda.h header file. This eliminate the CUDA
(build)dependency; as long as the runtime CUDA version is 8.0+,
libcuda.so will be dlopen'ed and used successfully.


bartoldeman:
See the discussion in #14919 

Note that @micket suggests to let `--with-cuda=internal` be a default, but that'd need an easyblock change

bartoldeman:
I can also add a speedup patch, on Tuesday at the latest.

Micket (approved):
LGTM. Should we wait for the performance patch or do that in a separate PR?

easybuild/easyconfigs/o/OpenMPI/OpenMPI-4.1.1-GCC-10.3.0.eb
```diff
@@ -39,6 +44,10 @@ dependencies = [
     ('PMIx', '3.2.3'),
 ]
 
+preconfigopts = './autogen.pl --force && '
+
+configopts = '--with-cuda=internal '
```
Micket, line 49: On second thought..

> Note that @Micket suggests to let `--with-cuda=internal` be a default, but that'd need an easyblock change

Given that it relies on a patch, to me is seems it actually seems like fine to me to have it in the easyconfig rather than trying to put some parts in the easyblock.
Just add a comment explaining this new option, and how you could opt-out of it (by using --without-cuda (and optionally remove the then unnecessary patch(es)))

Micket (changes requested):

easybuild/easyconfigs/o/OpenMPI/OpenMPI-4.1.1-GCC-10.3.0.eb
Micket: @bartoldeman Could we write a small comment along the lines of
```suggestion
# Patch and custom configure option can be removed if CUDA support isn't wanted.
configopts = '--with-cuda=internal '
```
to indicate that it is optional for those without CUDA hardware. I suppose all we need is to just remove the configopts and you'll have no CUDA support (or overhead), so, very simple change and i don't think we need any easyblock specialities

Micket:
Test report by @Micket
**SUCCESS**
alvis-s1 - Linux Rocky Linux 8.5, x86_64, Intel(R) Xeon(R) Silver 4314 CPU @ 2.40GHz, Python 3.6.8
See https://gist.github.com/a02966aa45ac98b2bdadec5f006df089 for a full test report.

branfosj:
Test report by @branfosj
**SUCCESS**
bask-pg0309u12a.cluster.baskerville.ac.uk - Linux RHEL 8.6, x86_64, Intel(R) Xeon(R) Platinum 8360Y CPU @ 2.40GHz (icelake), 1 x NVIDIA NVIDIA A100-SXM4-40GB, 470.57.02, Python 3.6.8
See https://gist.github.com/252c18c3e9aec873e856eddf9bd17bf3 for a full test report.

SebastianAchilles:
CORE_CNT=16
EB_ARGS="--buildpath=/dev/shm/$USER --installpath=/tmp/$USER/pr15528"

Micket (approved):
lgtm

I ran through all expected OSU tests;
```bash
#!/usr/bin/env bash
#SBATCH -n 2 -N 2
#SBATCH --gpus-per-node=A40:4
#SBATCH -t 1:00:00

ml OSU-Micro-Benchmarks/5.7.1-gompi-2021a-CUDA-11.3.1

for t in osu_bibw osu_bw osu_latency osu_mbw_mr osu_multi_lat osu_allgather osu_allgatherv osu_allreduce osu_alltoall osu_alltoallv osu_bcast osu_gather osu_gatherv osu_reduce osu_reduce_scatter osu_scatter osu_scatterv osu_iallgather osu_ialltoall osu_ibcast osu_igather osu_iscatter osu_alltoall osu_allreduce osu_reduce osu_alltoall
do
    echo "Running ${t}"
    mpirun ${t} -d cuda D D
    mpirun ${t} -d cuda H D
    mpirun ${t} -d cuda D H
    mpirun ${t} -d cuda H H
done
```
and it all worked without errors.

Anyone else wants to have a second check? I'm good with this getting merged.

branfosj:
> Anyone else wants to have a second check? I'm good with this getting merged.

I tested with the OSU 5.9 (#15343 and #15344). In addition to the above, this also allows running the NCCL based tests:
```bash
for t in osu_nccl_bibw osu_nccl_bw osu_nccl_latency osu_nccl_allgather osu_nccl_allreduce osu_nccl_bcast osu_nccl_reduce osu_nccl_reduce_scatter osu_nccl_reduce osu_nccl_allreduce
do
  echo ${t}
  mpirun -np 2 ${t} -d cuda D D
done
```

These were all fine. So LGTM :)

branfosj:
Going in, thanks @bartoldeman!

migueldiascosta:
I was running a benchmark on a CPU node with `foss/2022a`, getting bad performance, and noticed that OpenMPI was using `smcuda` as shared memory `btl`, forcing it to use `vader` significantly improved performance, by at least 30% (this was on a 2x96 core Genoa node, so right now it may be an extreme case of shared memory communication, but that's the trend...)

I think I'll just set `OMPI_MCA_btl=^smcuda` on non-GPU nodes, but just wanted to mention here that our builds may be trading off too much performance for generality?

Micket:
This CPU node had the cuda runtime installed? Otherwise i don't think the smcuda would have even been enabled at all.

migueldiascosta:
Not that I can tell - this was on benchmarking system provided by a vendor, running Ubuntu, but the only thing nvidia related I can find is `/usr/bin/nvidia-detector` from `ubuntu-drivers-common` 

Just tried it on our own (centos) nodes and indeed `smcuda` is not used, so not sure what is triggering it in the benchmarking system...

boegel:
If this should be discussed further, let's open an issue on this, discussing stuff in a merged PR is difficult to keep track of...

migueldiascosta:
done, https://github.com/easybuilders/easybuild-easyconfigs/issues/17854

migueldiascosta:
> This CPU node had the cuda runtime installed? Otherwise i don't think the smcuda would have even been enabled at all.

@Micket if I understand https://github.com/open-mpi/ompi/blob/9216ad4c49a16b3134d0dc47d8aa623f569d45ae/opal/mca/btl/smcuda/README.md correctly, `smcuda` was implemented as a copy of the old `sm`; without the cuda runtime, it obviously doesn't find any device or do anything cuda related, but it will use the old `sm` code instead of `vader`. The comment about `smcuda` having a higher priority than `sm` probably also applies to `vader`, which is why it's being selected (?)

PR #15690, opened 2022-06-14
{math,phys}[foss/2021a] deal.II v9.3.3, p4est v2.8

boegel:
(created using `eb --new-pr`)

PR #15692, opened 2022-06-15
correct configopts in deal.II v9.1.1 easyconfig

branfosj:
(created using `eb --new-pr`)

see https://github.com/easybuilders/easybuild-easyconfigs/pull/15690#discussion_r897127603


branfosj:

branfosj:
And correcting this leace deal.II failing to find some part of netCDF:
```
Include /tmp/boegelbot/dealII/9.1.1/intel-2019a/dealii-9.1.1/cmake/configure/configure_netcdf.cmake
-- NETCDF_INCLUDE_DIR not found! Call:
--     FIND_PATH(NETCDF_INCLUDE_DIR netcdfcpp.h HINTS /project/boegelbot/Rocky8/haswell/software/netCDF/4.6.2-iimpi-2019a PATH_SUFFIXES netcdf include)
-- NETCDF_CPLUSPLUS_LIBRARY not found! Call:
--     FIND_LIBRARY(NETCDF_CPLUSPLUS_LIBRARY NAMES netcdf_c++ netcdf_cpp HINTS /project/boegelbot/Rocky8/haswell/software/netCDF/4.6.2-iimpi-2019a PATH_SUFFIXES lib lib64 lib)
-- Found NETCDF_C_LIBRARY
--   NETCDF_LIBRARIES: *** Required variable "NETCDF_CPLUSPLUS_LIBRARY" set to NOTFOUND ***
--   NETCDF_INCLUDE_DIRS: *** Required variable "NETCDF_INCLUDE_DIR" set to NOTFOUND ***
-- Could NOT find NETCDF
-- DEAL_II_WITH_NETCDF has unmet external dependencies.
CMake Error at cmake/macros/macro_configure_feature.cmake:112 (MESSAGE):
  

  Could not find the netcdf library!

  Please ensure that a suitable netcdf library is installed on your computer.

  If the library is not at a default location, either provide some hints for
  autodetection,

      $ NETCDF_DIR="..." cmake <...>
      $ cmake -DNETCDF_DIR="..." <...>

  or set the relevant variables by hand in ccmake.
```

boegel:
@branfosj Yes, I ran into that as well...
Works fine if you just omit the configure options for netCDF, which I guess boils down to building deal.II without netCDF support (but we were doing that already, so it's not a regression).

It seems to want to old netCDF-C++ bindings?

branfosj:

branfosj:
> @branfosj Yes, I ran into that as well... Works fine if you just omit the configure options for netCDF, which I guess boils down to building deal.II without netCDF support (but we were doing that already, so it's not a regression).
> 
> It seems to want to old netCDF-C++ bindings?

Yes, but I think it wants both netCDF and netCDF-C++ and we'd have to patch it to know that both are there but in separate places. So, I've remove the netCDF line from the configopts in 9427350

boegel:
> > @branfosj Yes, I ran into that as well... Works fine if you just omit the configure options for netCDF, which I guess boils down to building deal.II without netCDF support (but we were doing that already, so it's not a regression).
> > It seems to want to old netCDF-C++ bindings?
> 
> Yes, but I think it wants both netCDF and netCDF-C++ and we'd have to patch it to know that both are there but in separate places. So, I've remove the netCDF line from the configopts in [9427350](https://github.com/easybuilders/easybuild-easyconfigs/commit/9427350d4c0bc8c44c0b3d6981c832a23fbf43e4)

Makes sense, because whatever netCDF support that was there was removed recently anyway, see https://github.com/dealii/dealii/pull/10387.
That explains why we're not seeing this problem with deal.II v9.3.3 in #15690, but it's weird that CMake doesn't trip over the `-DDEAL_II_WITH_NETCDF=ON` being used there then... 🤔

boegel (changes requested):
@branfosj `netCDF` dependency should also be removed since `-DDEAL_II_WITH_NETCDF=ON` was removed

branfosj:
> 

done in f65dd39

boegel:

boegel:
Going in, thanks @branfosj!

PR #15833, opened 2022-07-08
{bio}[GCC/11.2.0] Raven v1.8.1 w/ Python 3.9.6

ItIsI-Orient:
(created using `eb --new-pr`)


boegel (changes requested):
@ItIsI-Orient The tests aren't happy because you should set `download_dep_fail` and `use_pip` via `exts_default_options` (as opposed to in the extension itself in `exts_list`), like so:

```python
exts_default_options = {
    'download_dep_fail': True,
    'use_pip': True,
}
```

(that's just a style thing, functionally there's no difference)

easybuild/easyconfigs/r/Raven/Raven-1.8.1-GCC-11.2.0.eb
```diff
@@ -0,0 +1,53 @@
...
+]
+
+modextrapaths = {'PYTHONPATH': 'lib/python%(pyshortver)s/site-packages'}
+
+sanity_check_paths = {
+    'files': ['bin/raven'],
+    'dirs': []
+}
+
+sanity_check_commands = ["pip check"]
```
bedroge, line 51: Could you also add a simple `sanity_check_command` for the `raven` binary itself, e.g. `raven --help` or `raven --version`?

bedroge (changes requested):

ItIsI-Orient:
Test report by @ItIsI-Orient
**SUCCESS**
node3701.slaking.os - Linux RHEL 8.4, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 3.6.8
See https://gist.github.com/0e3296e4f9ee74dc5ba6142219dd4442 for a full test report.

bedroge:
The installation of the Python extension fails on my Ubuntu system:
```
  [25/26] Linking CXX static library lib/libraven.a
  [26/26] Linking CXX shared module lib/ravenpy.cpython-310-x86_64-linux-gnu.so
  error: [Errno 2] No such file or directory: '/data/eb/tmp/eb-aklw6_1o/pip-req-build-6c3ee0ie/build/lib.linux-x86_64-3.9/ravenpy_build/lib/ravenpy.cpython-39-x86_64-linux-gnu.so'
```
It looks like it's somehow mixing up the Python version from the OS (3.10) and the module (3.9)...

bedroge:
I've debugged this a bit more, and the issue seems to be caused by the way CMake is trying to find Python. This behavior has been changed in CMake 3.15: https://cmake.org/cmake/help/latest/policy/CMP0094.html#policy:CMP0094
With older versions/policies of CMake, it would just look for the newest Python it could find (and since my Ubuntu has Python 3.10, it prefers that one), with newer ones the location is preferred. We do use a newer CMake for this easyconfig, but because Raven's `CMakeLists.txt` has `cmake_minimum_required(VERSION 3.11)`, it still falls back to the old policy. We can change this by either patching that line in `CMakeLists.txt`, or by explicitly enabling the new policy in `setup.py`; I'd prefer the latter. On top of that, we should also make sure that `pybind` uses CMake's `FindPython` functionality by setting `PYBIND11_FINDPYTHON=ON`.

Finally, I see that the installation pulls in some dependencies, see:
https://github.com/lbcb-sci/raven/blob/1.8.1/Raven.deps.cmake
It would be cleaner if we would list these as dependencies in the easyconfig. For most(/all?) of them, easyconfigs already seem to exist, they may only require a version bump.

boegel:
> Will take a look into that.
> 
> Also I am not sure how to enable the new policy in `setup.py`, could you point me in the right direction about solving this one?

@ItIsI-Orient Via `-DPython3_FIND_STRATEGY` I guess, see https://cmake.org/cmake/help/latest/policy/CMP0094.html ?

boegel:
We fixed a very similar problem with the Python being used for VTK in #16741 via `-DPython3_EXECUTABLE`, which we should also use here (and actually everywhere where CMake is being used and Python is included as a dependency, so in the `CMakeMake` easyblock, but that's a big hammer...)

ItIsI-Orient:
Test report by @ItIsI-Orient
**SUCCESS**
node3703.slaking.os - Linux RHEL 8.6, x86_64, Intel(R) Xeon(R) CPU E5-2680 v3 @ 2.50GHz, Python 3.6.8
See https://gist.github.com/89f3570fd5c99e45fc384ccf907537b7 for a full test report.

bedroge:
Test report by @bedroge
**FAILED**
bob-Latitude-5300 - Linux Ubuntu 22.04, x86_64, Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz, Python 3.10.6
See https://gist.github.com/206dd0cc6458400f3d2af95b759d14b0 for a full test report.

bedroge:
This doesn't seem to solve the issue, as it's only passed to the build of Raven itself, while the issue pops up while RavenPy is being built. In order to pass it to that (CMake) build as well, I guess we have to modify the `setup.py` (e.g. using something like what I used in a previous suggestion here: https://github.com/easybuilders/easybuild-easyconfigs/pull/15833#pullrequestreview-1045043403), which also calls `cmake` with some predefined arguments.

bedroge:
Test report by @bedroge
**SUCCESS**
bob-Latitude-5300 - Linux Ubuntu 22.04, x86_64, Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz, Python 3.10.6
See https://gist.github.com/73b29b6588e22bf6d03ba1f6df8dfe64 for a full test report.

bedroge:
So, this does seem to work (but I had already tried that a while ago). I'm not sure though if this is what @boegel meant, as it only sets `-DPython_FIND_STRATEGY=LOCATION` now, while @boegel was referring to `-DPython3_EXECUTABLE` . I also tried setting only the latter for `ravenpy`, but that didn't work, so it looks like we do need  the first one as well (or maybe both to be sure?). This is a bit of a weird one anyway, as `setup.py` calls `cmake` itself, I'm not sure how usual this is.

@boegel any thoughts on what to do here?

boegel:
I'm surprised that only setting `-DPython3_EXECUTABLE` isn't sufficient...
What does using `-DPython_FIND_STRATEGY=LOCATION` actually imply?

@ItIsI-Orient Can you take a detailed look at that, and try and see if you can figure out why it seems that `-DPython_FIND_STRATEGY=LOCATION` is required (too)?

branfosj:
> I'm surprised that only setting `-DPython3_EXECUTABLE` isn't sufficient... What does using `-DPython_FIND_STRATEGY=LOCATION` actually imply?

From the CMake documentation
> LOCATION: Stops lookup as soon as a version satisfying version constraints is founded.

The alternative strategy is for it to find the latest version.

`-DPython3_EXECUTABLE` is being set for `Raven`, but not for `ravenpy`.

bedroge:
And, as mentioned in https://github.com/easybuilders/easybuild-easyconfigs/pull/15833#issuecomment-1190059574, I think the issue is that the `CMakeLists.txt` file sets the minimum required CMake version to a rather old one, which also leads to old policies being used. 

In this case it means that it will / seems to use `VERSION` instead of `LOCATION` for `Python_FIND_STRATEGY`; as Simon already mentioned, this means it will use the newest Python version it can find instead of looking for the first one (based on your `$PATH` I guess?) that satisfies the version constraints.

bedroge:
Test report by @bedroge
**SUCCESS**
bob-Latitude-5300 - Linux Ubuntu 22.04, x86_64, Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz, Python 3.10.6
See https://gist.github.com/7ec344fd25ef7bb885a1885d729ec361 for a full test report.

bedroge:
This version works for me. @boegel are you also okay with the current approach/version?

bedroge:

boegel:

boegel:
> ```
> sanity check command pip check exited with code 1 (output: poetry 1.1.7 has requirement keyring<22.0.0,>=21.2.0; python_version >= "3.6" and python_version < "4.0", but you have keyring 23.9.3.
> ```
> 
> I don't understand where this version 23.9.3 comes from, the Python installation has 21.2.0...

There was some stuff installed in the home directory of the bot that explains this:

```
[boegelbot@login1 ~]$ ml Python/3.9.6-GCCcore-11.2.0
[boegelbot@login1 ~]$ python3 -m pip list | grep keyring
keyring                           23.9.3
```

```
[boegelbot@login1 ~]$ python3
Python 3.9.6 (default, Jul 14 2022, 08:55:00)
[GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import keyring
>>> keyring.__file__
'/home/boegelbot/.local/lib/python3.9/site-packages/keyring/__init__.py'
```

I cleaned that up, should be fine now...

easybuild/easyconfigs/r/Raven/Raven-1.8.1-GCC-11.2.0.eb
```diff
@@ -0,0 +1,61 @@
...
+
+modextrapaths = {'PYTHONPATH': 'lib/python%(pyshortver)s/site-packages'}
+
+sanity_check_paths = {
+    'files': ['bin/raven'],
+    'dirs': []
+}
+
+sanity_check_commands = [
+    "pip check",
```
boegel, line 57: Better check with:

```
export PYTHONNOUSERSITE=1 && python%(pyshortver)s -m pip check

boegel (changes requested):

easybuild/easyconfigs/r/Raven/Raven-1.8.1-GCC-11.2.0.eb
boegel: fixed in 438320f

boegel:

boegel:
Test report by @boegel
**SUCCESS**
node3125.skitty.os - Linux RHEL 8.6, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/7c1b0a19f7e28bf2de4510f47adb7997 for a full test report.

boegel:

boegel (dismissed):

boegel:
Going in, thanks @ItIsI-Orient!

bedroge:
Test report by @bedroge
**FAILED**
bob-Latitude-5300 - Linux Ubuntu 22.04, x86_64, Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz, Python 3.10.6
See https://gist.github.com/d0c9cb1ee0f20f8d7c3a953d93174ca6 for a full test report.

boegel:

boegel:
Test report by @boegel
**FAILED**
node3133.skitty.os - Linux RHEL 8.6, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/boegel/9f8394523c1f9a613b6ff59ae2ec9af3 for a full test report.

boegel:

boegel:
Test report by @boegel
**SUCCESS**
node3133.skitty.os - Linux RHEL 8.6, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz (skylake_avx512), Python 3.6.8
See https://gist.github.com/boegel/acdb3455fc6d79d09e96b1e65a79b926 for a full test report.

boegel (approved):
lgtm

boegel:
Going in, thanks @ItIsI-Orient!

PR #15851, opened 2022-07-13
{math}[foss/2021b] molmod v1.4.8 w/ Python 3.9.6

sassy-crick:
(created using `eb --new-pr`)

PR #15854, opened 2022-07-13
{chem}[foss/2021b] yaff v1.6.0 w/ Python 3.9.6

sassy-crick:
(created using `eb --new-pr`)

Requires:
- [x] PR #15851


bedroge:

easybuild/easyconfigs/y/yaff/yaff-1.6.0-foss-2021b.eb
```diff
@@ -0,0 +1,59 @@
...
+description = """Yaff stands for 'Yet another force field'. It is a pythonic force-field code."""
+
+toolchain = {'name': 'foss', 'version': '2021b'}
+
+builddependencies = [('pkgconfig', '1.5.5', '-python')]
+
+dependencies = [
+    ('Python', '3.9.6'),
+    ('SciPy-bundle', '2021.10'),
+    ('HDF5', '1.10.8'),  # versions 1.12.x does not seem to work
```
bedroge, line 30: This requires an exception in the test suite, as the default HDF5 version for `foss/2021b` is 1.12.1:
https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/test/easyconfigs/easyconfigs.py#L474

sassy-crick: Yes, that makes sense. 
The reason for that YAFF build was as I wanted to use it for LAMMPS, as originally suggested in PR #14653 which I then used in my LAMMPS PR #15869 (update 3). As a newer PR for LAMMPS have been merged, I am not quite sure what to do with it. On one hand I think that test would be good to have, on the other hand there is the problem with the `HDF5` version for the `YAFF` build. 
@boegel Maybe something to touch on at the next EB user meeting?

sassy-crick: Sorry, that was on my to-do list which is too long. 
I have attached the log-file. The main problem I have is that I don't know how to fix that with the usual way of doing it:
https://portal.hdfgroup.org/display/HDF5/Migrating+from+HDF5+1.10+to+HDF5+1.12
I tried that but I failed miserably. I guess it shows my lack of deeper understanding here. 
This is the log file where I replaced the `1.10.x` version with `1.12.y` but left everything else the same as it is in this PR:

[easybuild-yaff-1.6.0-20220906.101056.LUfUV.log](https://github.com/easybuilders/easybuild-easyconfigs/files/9495216/easybuild-yaff-1.6.0-20220906.101056.LUfUV.log)

boegel:
closing this, since the PR it was required for (#15869) is also closed...

PR #15867, opened 2022-07-14
FlexiBLAS: Workaround miscompilation on PPC

Flamefire:
(created using `eb --new-pr`)


Flamefire:
Test report by @Flamefire
**SUCCESS**
taurusml4 - Linux RHEL 7.6, POWER, 8335-GTX (power9le), 6 x NVIDIA Tesla V100-SXM2-32GB, 440.64.00, Python 2.7.5
See https://gist.github.com/098cefbaf8edcd473f811bc0cb4ce2b0 for a full test report.

easybuild/easyconfigs/f/FlexiBLAS/FlexiBLAS-3.0.4-GCC-10.3.0.eb
```diff
@@ -11,6 +11,13 @@ toolchain = {'name': 'GCC', 'version': '10.3.0'}
 local_extra_flags = "-fstack-protector-strong -fstack-clash-protection"
 toolchainopts = {'pic': True, 'extra_cflags': local_extra_flags, 'extra_fflags': local_extra_flags}
 
+if ARCH == 'ppc64le':
```
ocaisa, line 14: I don't want to stall this, but in general I think it is better if we avoid logic in easyconfigs if possible (since they are not supposed to contain any). Instead you could create a `local_configopts` dict and just add `local_configopts[ARCH]` to the component.
I say this because this easyconfig will now fall over with some of the `try-*` options.

ocaisa: BTW ARCH is not defined is it? Shouldn't it be `%(arch)s`?

Flamefire:
The fix needs to go into OpenBLAS anyway. Closing this.

PR #15910, opened 2022-07-27
{cae}[GCCcore/11.2.0,system/system] MATSim v14.0

lexming:
(created using `eb --new-pr`)

PR #15932, opened 2022-07-29
exclude Binary, PackedBinary and JAR easyblocks from binutils build requirements

lexming:
Fixes failed tests in https://github.com/easybuilders/easybuild-easyconfigs/pull/15910


Micket (approved):
lgtm

PR #15968, opened 2022-08-04
Numpy: Fix broken test on PPC

Flamefire:
(created using `eb --new-pr`)

- [x] Requires https://github.com/easybuilders/easybuild-easyconfigs/pull/15973

PR #15973, opened 2022-08-05
refactor checksum test for extensions to use collect_exts_file_info

Flamefire:
Extension patches may use templates so use `collect_exts_file_info` to gather the information about extensions and their sources and patches.

See https://github.com/easybuilders/easybuild-easyconfigs/pull/15968


boegel (approved):
lgtm

PR #16003, opened 2022-08-09
{math}[GCCcore-11.2.0] CPLEX 22.10 w/ Python 3.9.6

tardigradus:


jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 16003` and update the PR to use `SYSTEM` instead.

easybuild/easyconfigs/c/CPLEX/CPLEX-22.10-GCCcore-11.2.0-Python-3.9.6.eb
```diff
@@ -0,0 +1,26 @@
+name = 'CPLEX'
+version = '22.10'
+versionsuffix = '-Python-%(pyver)s'
```
jfgrimm, line 3: drop the versionsuffix, we made python 3 the default from the 2020a toolchain generation onwards. The filename should be updated to reflect this as well
```suggestion
```

easybuild/easyconfigs/c/CPLEX/CPLEX-22.10-GCCcore-11.2.0-Python-3.9.6.eb
```diff
@@ -0,0 +1,26 @@
...
+# https://www.ibm.com/developerworks/community/blogs/jfp/entry/cplex_studio_in_ibm_academic_initiative?lang=en
+# a restricted "Community edition" version can be found on
+# https://www-01.ibm.com/software/websphere/products/optimization/cplex-studio-community-edition/
+sources = ['cplex_studio%s.linux_x86_64.bin' % ''.join(version.split('.'))]
+checksums = ['c8f8e6a2eb55433fddb5986c78ffe23cb4c6dfd09d6c24cebcf16d928a988a1d']
+
+builddependencies = [('binutils', '2.37')]
+
+dependencies = [
+    ('Java', '11.0.2', '', True),
```
jfgrimm, line 22: we should depend on the `Java` wrapper, rather than a specific version directly
```suggestion
    ('Java', '11', '', SYSTEM),
```

easybuild/easyconfigs/c/CPLEX/CPLEX-22.10-GCCcore-11.2.0-Python-3.9.6.eb
```diff
@@ -0,0 +1,26 @@
...
+description = """IBM ILOG CPLEX Optimizer's mathematical programming technology enables
+ analytical decision support for improving efficiency,
+ reducing costs, and increasing profitability."""
+
+toolchain = {'name': 'GCCcore', 'version': '11.2.0'}
+
+# the Academic Initiative version (as used in this file) can be downloaded as described on
+# https://www.ibm.com/developerworks/community/blogs/jfp/entry/cplex_studio_in_ibm_academic_initiative?lang=en
+# a restricted "Community edition" version can be found on
+# https://www-01.ibm.com/software/websphere/products/optimization/cplex-studio-community-edition/
```
jfgrimm, line 15: since we added support for `download_instructions`, it would be good to make use of that here. it accepts a string, and makes it more obvious why an easyconfig failed to build if the sources are not found (since they're not auto-downloaded)

boegel:
superseded by #17574

PR #16004, opened 2022-08-09
adding easyconfigs: RheoTool-5.0-foss-2019b.eb

deniskristak:
https://github.com/hpcugent/eb_inuits/issues/263

after loading the module, this must be run:
```source $FOAM_BASH```

shouldn't we add it as an automatic cmd during module loading?


deniskristak:
still only using workaround with the $HOME problem

boegel:
closing this one in favor of #16077

PR #15415, opened 2022-05-03
{compiler}[system/system] NVHPC v22.3 w/ CUDA 11.4.1

robert-mijakovic:
(created using `eb --new-pr`)

PR #15155, opened 2022-03-22
{compiler}[system/system] NVHPC v22.2 w/ CUDA 11.4.1

robert-mijakovic:
(created using `eb --new-pr`)

PR #13028, opened 2021-06-03
{compiler}[system/system] NVHPC v21.5

robert-mijakovic:
(created using `eb --new-pr`)

edit: requires ~https://github.com/easybuilders/easybuild-easyblocks/pull/2453~

PR #16006, opened 2022-08-09
{compiler}[system/system] NVHPC v22.7

hattom:
adding easyconfigs: NVHPC-22.7-CUDA-11.4.1.eb

(half-created with eb --new-pr, half not; due to token issues.)


hattom:
@robert-mijakovic -- I based this off your PRs, any comments welcome.

easybuild/easyconfigs/n/NVHPC/NVHPC-22.7-CUDA-11.4.1.eb
```diff
@@ -0,0 +1,69 @@
...
+            '6dd4fd382c22769e4fa9508714119abd7d1df3dc58c69414a14b0b0dbc34564f',
+        local_tarball_tmpl % 'x86_64':
+            '3ce1c346f8bc7e50defb41c545c8907fdc012ff60b27eb8985cf3213f19d863a',
+    }
+]
+
+local_gccver = '11.2.0'
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.37', '', ('GCCcore', local_gccver)),
```
boegel, line 33: Same as in #15155 and #15415: GCC should be used as toolchain to ensure that this dependency is reflected in the module name?

boegel (changes requested):

hattom:
Since the Nvidia HPC-SDK (NVHPC) 22.7 ships with CUDA {10.2,11.0,11.7}, should I be choosing one of those CUDA versions rather than 11.4.1 (based on the previous NVHPC module)?

hattom:
Similarly, not sure if GCC/11.2.0 is the right choice for GCC?

easybuild/easyconfigs/n/NVHPC/NVHPC-22.7-CUDA-11.4.1.eb
SebastianAchilles: For a the last few release of `NVHPC` we decided to put it into the `SYSTEM` toolchain and only add `GCCcore` as dependency. The reasoning was the same as with the `intel-compiler`: we wanted to have a shorter module name.

easybuild/easyconfigs/n/NVHPC/NVHPC-22.7-CUDA-11.4.1.eb
```diff
@@ -0,0 +1,69 @@
...
+    # This is necessary to avoid cases where just libnuma.so.1 is present in the system and -lnuma fails
+    ('numactl', '2.0.14', '', ('GCCcore', local_gccver)),
+    ('CUDA', '11.4.1', '', True),
+]
```
SebastianAchilles, line 37:
```suggestion
local_gccver = '11.3.0'
dependencies = [
    ('GCCcore', local_gccver),
    ('binutils', '2.38', '', ('GCCcore', local_gccver)),
    # This is necessary to avoid cases where just libnuma.so.1 is present in the system and -lnuma fails
    ('numactl', '2.0.14', '', ('GCCcore', local_gccver)),
    ('CUDA', '11.7.0', '', True),
]
```

SebastianAchilles (changes requested):
I would suggest to go with the newer GCC v11.3.0 and CUDA v11.7.0.

easybuild/easyconfigs/n/NVHPC/NVHPC-22.7-CUDA-11.4.1.eb
hattom: Looks like my message got lost.
I wanted to write yesterday:
"Can change it, but how is this different from `intel`?"

I'd favour the solution that @SebastianAchilles mentions, but I'll change it to whatever is required.

SebastianAchilles:

SebastianAchilles:

SebastianAchilles:
**SUCCESS**
bdw-opensuse-154 - Linux openSUSE Leap 15.4, x86_64, Intel(R) Core(TM) i7-6900K CPU @ 3.20GHz (broadwell), 2 x NVIDIA NVIDIA GeForce GTX 1060 6GB, 510.85.02, Python 3.6.15
See https://gist.github.com/d9b6d1399ad914857a7ec80e90436a90 for a full test report.

jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 16006` and update the PR to use `SYSTEM` instead.

hattom:
@Micket -- Not quite clear to me, shall I just fix this PR as-is, or does this become superfluous once the NV toolchain is done?

boegel:
Closing since several newer versions of NVHPC have been merged already, like:
- #13028

Issue #14320, opened 2021-11-10
realizing added Rust dependency as pre-build wheel


cmeesters:
Hi,

As half-way discussed in the slack channel, I would like to bring to attention the following thoughts:

As of 2021b when installing Python a build-dependency for Rust in introduced to satisfy the `cryptography` module. This is a considerable overhead as building the Rust module (a one-time need for those only installing Python with it) takes about 1.5 h (on our machine). Also, it is only possible on head nodes and not in a job context for those sites where internet access on compute nodes is disabled.

Perhaps this is a non-issue, when we envision more and more software written in Rust to be rolled out in the future. Bio-Rust for instance might gain impact on the field.

**IF** the build overhead of Rust to support Python is a considerable overhead, one could consider deviating from eb policies and replacing the extension with a pre-build wheel.

Cheers,
Christian

branfosj:
> I think using a pre-built wheel for `cryptography` to avoid `Rust` as a build dependency for `Python` is worth consdering...

In out 2021b generation Python we have `cryptography` 3.4.7. Looking at https://pypi.org/project/cryptography/3.4.7/#files I am only seeing wheels for Python 3.6 (`cp36`). Are they more generally usable?

Otherwise we hit the lack of POWER support, but there are issues with >= 2021a anyway that may make that an irrelevance.

Micket:
One possible alternative can also be to break out some extensions from Python, in particular to reduce the number of dependencies from Python. This could also save us from having to make a -bare python variant.
This would probably also mean that building a toolchain wouldn't pull in `Rust`, but many things would still depend indirectly on `Rust` since `cryptography` is such a common dep.

There is probably also some stuff we can do to improve the Rust build, perhaps an external LLVM, or make it build it's LLVM in parallel, or patch it in some way so that it doesn't do downloads during the build (based on comments from other maintainers, I haven't actually looked into this myself).
So, I'm inclined to say; lets keep it? But we could always prepare the python easyconfigs to that it is easy to opt-out of Rust for those who can and want to use the pre built cryptography.

migueldiascosta:
Regarding the time it takes to build `Rust`'s `LLVM`, one issue is that we disabled `Ninja` because it would introduce a circular dependence on `Python`, but that's not actually the case if we change the `Ninja` easyconfigs to use `Python` `-bare` as a builddep, same as we already do in `Rust`. Then we can add `Ninja` as a builddep in `Rust` and remove

```python
[...]

PR #16037, opened 2022-08-12
WIP: add amd toolchains

zemu-unile:
depends on 
- https://github.com/easybuilders/easybuild-framework/pull/4058
- https://github.com/easybuilders/easybuild-easyblocks/pull/2774

Adds easyconfigs for toolchains in https://github.com/easybuilders/easybuild-framework/pull/4058 with AOCC 3.2.0 and AOCL 3.2. GCC and OpenMPI versions are the same as in 2022a toolchains.

TODO:
- should `AOCC` have a `versionsuffix` for `GCCcore`?
- decide on which versioning to use. FFTW is using the upstream versioning 3.10. Everything else got the AOCL version (3.2)
- should we add the non-foss AOCL libraries to the `amd` toolchain (see comments in `amd-2022a.eb`)
- is there a way to get rid of the Rust dependency introduced by Python?

### DIFF
--- a/easybuild/easyconfigs/a/AOCC/AOCC-3.2.0.eb
+++ b/easybuild/easyconfigs/a/AOCC/AOCC-3.2.0.eb
@@ -1,0 +1,26 @@
+name = 'AOCC'
+version = '3.2.0'
+
+homepage = 'https://developer.amd.com/amd-aocc/'
+description = "AMD Optimized C/C++ & Fortran compilers (AOCC) based on LLVM 13.0"
+
+# Clang also depends on libstdc++ during runtime, but this dependency is
+# already specified as the toolchain.
+toolchain = SYSTEM
+
+source_urls = ['http://developer.amd.com/wordpress/media/files/']
+sources = ['aocc-compiler-%(version)s.tar']
+checksums = ['8493525b3df77f48ee16f3395a68ad4c42e18233a44b4d9282b25dbb95b113ec']
+
+clangversion = '13.0.0'
+
+local_gccver = '11.3.0'
+dependencies = [
+    ('GCCcore', local_gccver),
+    ('binutils', '2.38', '', ('GCCcore', local_gccver)),
+    ('ncurses', '6.3', '', ('GCCcore', local_gccver)),
+    ('zlib', '1.2.12', '', ('GCCcore', local_gccver)),
+    ('libxml2', '2.9.13', '', ('GCCcore', local_gccver)),
+]
+
+moduleclass = 'compiler'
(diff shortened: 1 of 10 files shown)


jfgrimm:
@zemu-unile thanks for kicking off an AMD toolchain PR!

> should AOCC have a versionsuffix for GCCcore?

in keeping with e.g. intel-compilers, no, a versionsuffix is not necessary

> should we add the non-foss AOCL libraries to the amd toolchain (see comments in amd-2022a.eb)

this should be opt-in IMO

> is there a way to get rid of the Rust dependency introduced by Python?

nope, it's required because Python's `cryptography>3.4` package is written in Rust

Micket:
Rust/cryptography discussion here: #14320  (I suspect we are moving in a direction where more and more python packages will be using Rust though, so i suspect this might be a bit inevitable)

boegel:
This hasn't seen any activity in quite some time now...

@zemu-unile Do you think it's still worth the effort to try and get this merged?

zemu-csi:
Sadly, I am not working on building software at our site anymore so I don't have time to finish this.

I only had a single user asking for AOCL in 2022 which caused me to open this PR but we don't even have it installed anymore.

The main TODOs here are:
- get the framework PR merged that adds the toolchain: https://github.com/easybuilders/easybuild-framework/pull/4058
- update ECs from this PR to new versions
- optionally add ECs for more AMD libraries to AOCL, most of them are proprietary and not included in this PR, see https://docs.amd.com/r/en-US/57404-AOCL-user-guide/Introduction for a list
- Testing/Benchmarking

Unless someone else wants to take over we should probably close this PR and the accompanying PR in easybuild-framework.

AMD is usually contributing their patches upstream so they will end up in the foss toolchain anyways.

Thyre:
I think having AOCC toolchain support generally helps aligning with e.g. Cray systems to better support their toolchains. LUMI for example has PrgEnv-AOCC, and supporting these toolchains at the moment requires patching EasyBlocks manually for their repository. This is not great.

Since AOCC is basically also just LLVM (with the old flang currently), we might be able to re-use the existing work to get the LLVM toolchain up and running. Then, I would probably go the route of using FlexiBLAS instead of just relying on the AMD libraries...

I think doing a separate attempt at an AOCC toolchain once the LLVM one is established is the best idea. Not to actually have many modules available, but to have the toolchain established in general...

boegel:
I'll go ahead and close this, since the `2022a` generation of easyconfigs is deprecated now (see https://docs.easybuild.io/policies/toolchains/#current-situation).

PR #16070, opened 2022-08-22
update CI workflows to use Ubuntu 20.04 (since Ubuntu 18.04 is deprecated)

boegel:
see also https://github.com/actions/runner-images/issues/6002


Micket (approved):
lgtm

PR #16118, opened 2022-08-24
{math}[GCCcore/10.2.0,GCCcore/11.2.0] MATLAB-Engine v2021a w/ Python 3.8.6 + v2021b w/ Python 3.9.6

Micket:
(created using `eb --new-pr`)
currently don't have MATLAB 2021b to test against, but it should be a compatible version of Python (<=3.9 for 2021b).


Micket:
Test report by @Micket
**SUCCESS**
alvis-c1 - Linux Rocky Linux 8.5, x86_64, Intel Xeon Processor (Skylake), Python 3.6.8
See https://gist.github.com/3a8e6e2d2a631c6edf173dc363fa695d for a full test report.

jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 16118` and update the PR to use `SYSTEM` instead.

lexming:
Test report by @lexming
**FAILED**
node300.hydra.os - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 3.6.8
See https://gist.github.com/0023c6577d887cf72d02699d9ebf1b08 for a full test report.

easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-GCCcore-11.2.0.eb
```diff
@@ -0,0 +1,36 @@
+easyblock = 'PythonPackage'
+
+name = 'MATLAB-Engine'
+version = '2021a'
```
Micket, line 4: i actually messed up here, pretty sure i meant for this to be 2021b, which should be compatible with python 3.9 out of the box if i recall correctly.

lexming: Actually I'm interested in making this work with our Python in 2021a and I think I got a workable fix. However, I'm hitting another error in both Python 3.9 and 3.8 where the engine libs are not found. Not sure what is happening yet

Micket: Is there any reason to think that would work?

Also, i think it would be more appropriate for the for this to just be the 2021b version of matlab as the default, since 2021b version of matlab happens to match the GCCcore 11.2.0 toolchain release.

and it this needs to be backported to an older matlab for a newer toolchain then add a suffix for that special case

Micket:
Test report by @Micket
**SUCCESS**
alvis-c1 - Linux Rocky Linux 8.5, x86_64, Intel Xeon Processor (Skylake), Python 3.6.8
See https://gist.github.com/9b1153254bc80b08302a6071a4b9a0f3 for a full test report.

easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-GCCcore-11.2.0.eb
lexming: yes, having version 2021b in GCCcore/11.2.0 makes most sense.

I'm dealing with this issue now:
https://www.mathworks.com/matlabcentral/answers/597865-matlab-linking-error-centos-7-r2020b#answer_1007455

Apparently in Centos7 the engines cannot be loaded out-of-box

easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,34 @@
...
+
+skipsteps = ['install']
+
+# Test that connection with MATLAB can be established successfully
+sanity_check_commands = ["python -c 'import matlab.engine; eng = matlab.engine.start_matlab(); eng.quit()'"]
+
+sanity_pip_check = True
+
+options = {'modulename': 'matlab.engine'}
+
```
lexming, line 33: Loading the `matlabengine*.so` fails in systems with old glibc. We have to load the same glibc used by MATLAB:
```suggestion

# load glibc used by MATLAB binaries
modluafooter = """
    setenv("LD_PRELOAD", pathJoin(os.getenv("EBROOTMATLAB"), "bin", "glnxa64", "glibc-2.17_shim.so"))
"""

```

easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,34 @@
...
+    ('Python', '3.8.6'),
+]
+
+use_pip = False
+download_dep_fail = True
+
+# Since this setup.py does not provide a separate --build-base for the install
+# step, both build and install must be performed in a single command.
+prebuildopts = "cd $EBROOTMATLAB/extern/engines/python && "
+buildopts = "--build-base=%(builddir)s install --prefix=%(installdir)s"
```
lexming, line 23: It would be better to copy the sources to `builddir` to be sure that nothing is modified in the existing installation of MATLAB. The following does the trick, what do you think?
```suggestion
# Copy sources from MATLAB root to build dir
prebuildopts = "cp -r $EBROOTMATLAB/extern/engines/python/* %(builddir)s/ && "
prebuildopts += 'sed -i "s|_cwd,.*_parent|\'$EBROOTMATLAB\'|" setup.py && '
# Since this setup.py does not provide a separate --build-base for the install
# step, both build and install must be performed in a single command.
buildopts = "--build-base=%(builddir)s install --prefix=%(installdir)s"
```

lexming (changes requested):

easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-GCCcore-10.2.0.eb
lexming: this is also needed for version 2021b

Micket: This will break things for using this in conjunction with other stuff from the module tree for everyone else though, since this is older than what others might have. It breaks matlab.engine itself for me on Rocky 8.5

I think this needs to be handled as a hook.

easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-GCCcore-10.2.0.eb
lexming: wait, we've been doing this wrong... MATLAB-Engine is actually available from [PyPI](https://pypi.org/project/matlabengine) and Mathworks has a [github repo](https://github.com/mathworks/matlab-engine-for-python) with the source code.

Those packages do even receive updates and are compatible with `pip`. So there is no reason to use the sources in the installation of MATLAB.

I tested it and it works very well. I made a PR to this PR with the updated easyconfigs. Please check https://github.com/Micket/easybuild-easyconfigs/pull/13

easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-GCCcore-10.2.0.eb
lexming: you're right, since matlab works fine in our centos7 system I thought that it was always preloading that shim library, but it is not the case. It checks for the version of glibc and only preloads the shim library on systems with that specific version.
```
# Preload glibc_shim in case of RHEL7 variants
  test -e /usr/bin/ldd &&  ldd --version |  grep -q "(GNU libc) 2\.17"  \
            && export LD_PRELOAD="${MCRROOT}/bin/glnxa64/glibc-2.17_shim.so"
```

So I changed this suggestion to just a comment in https://github.com/Micket/easybuild-easyconfigs/pull/13

lexming:
For the record, version 2021a seems to work with unsupported Python 3.9. If anybody is interested, easyconfig is in [lexming/easybuild-easyconfigs/MATLAB-Engine-2021a-9.10.1-GCCcore-10.3.0.eb](https://github.com/lexming/easybuild-easyconfigs/blob/vub-hpc/easybuild/easyconfigs/m/MATLAB-Engine/MATLAB-Engine-2021a-9.10.1-GCCcore-10.3.0.eb)

lexming:
Test report by @lexming
**SUCCESS**
node300.hydra.os - Linux CentOS Linux 7.9.2009, x86_64, Intel(R) Xeon(R) Gold 6148 CPU @ 2.40GHz, Python 3.6.8
See https://gist.github.com/97f31cf7527a013e727f0b3163e1ffdd for a full test report.

lexming (approved):
LGTM

lexming:
Going in, thanks @Micket !

PR #16126, opened 2022-08-25
make check for toolchain value in dependency spec in easyconfigs test suite aware that dumped easyconfig uses SYSTEM constant

boegel:
requires:
* https://github.com/easybuilders/easybuild-framework/pull/4069

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -42,6 +42,7 @@ from easybuild.base import fancylogger
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.easyblocks.generic.pythonpackage import PythonPackage
 from easybuild.framework.easyblock import EasyBlock
+from easybuild.framework.easyconfig.constants import EASYCONFIG_CONSTANTS
 from easybuild.framework.easyconfig.default import DEFAULT_CONFIG
 from easybuild.framework.easyconfig.format.format import DEPENDENCY_PARAMETERS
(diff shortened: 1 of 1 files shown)


boegel:
The tests for this PR will fail until https://github.com/easybuilders/easybuild-framework/pull/4069 is merged, but this PR will need to get merged ASAP after the merging of https://github.com/easybuilders/easybuild-framework/pull/4069 to avoid broken tests in PRs being opened...

Micket (approved):
lgtm

test/easyconfigs/easyconfigs.py
```diff
@@ -1429,7 +1430,13 @@ def template_easyconfig_test(self, spec):
                 # 4th value is toolchain spec
                 if len(dumped_dep) >= 4:
                     if len(orig_dep) >= 4:
-                        self.assertEqual(dumped_dep[3], orig_dep[3])
+                        # if True was used to indicate that dependency should use system toolchain,
+                        # then we need to compare the value for the dumped easyconfig more carefully;
+                        # see also https://github.com/easybuilders/easybuild-framework/pull/4069
+                        if orig_dep[3] == True:
```
Micket, line 1436:
```suggestion
                        if orig_dep[3] is True:
```

Micket (changes requested):

test/easyconfigs/easyconfigs.py
Micket: i guess this is the nicest in this setting, don't want implicit true conversions i think

jfgrimm (approved):

PR #16140, opened 2022-08-26
{chem}[foss/2020b] OpenBabel v3.1.1 w/ Python 3.8.6

DirectorSloan:
(created using `eb --new-pr`)


DirectorSloan:
don't worked with Eigen 3.3.8 but in order to don't break foss/2020b chain on upstream I will keep this easyconfig private only until 2022b will be released because we leave out 2021a and 2021b on our System

PR #16185, opened 2022-09-05
{lib}[GCC-11.2.0-CUDA-11.4.1] pocl-1.8

zdemat:
Adding `pocl/1.8` with support for `NVPTX` target.


jfgrimm (commented):
As decided in issue #16330, we have deprecated the use of `True` to signify a system-toolchain dependency (#16384), in favour of the more intuitive `SYSTEM` template constant. Due to the change in the test suite, please run `eb --sync-pr-with-develop 16185` and update the PR to use `SYSTEM` instead.

boegel:
Closing this since `GCC/11.2.0` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains, and newer versions of `pocl` have been merged, like:
- #16398 

PR #16200, opened 2022-09-06
{lib}[GCC/10.2.0,GCC/11.2.0] SLICOT v5.0

sassy-crick:
(created using `eb --new-pr`)

I am using the tarball from the Debian archive, rather than the one from the original website. You need to register there but it is basically 'free' software. The [GitHub](https://github.com/SLICOT/SLICOT-Reference) site only provides the source code for Windows and thus does not compile, specially as `make.inc` is missing as well.


easybuild/easyconfigs/s/SLICOT/slicot-makefile-flexiblas.patch
```diff
@@ -0,0 +1,29 @@
...
+@@ -16,10 +16,10 @@
+ #  selected.  Define LOADER and LOADOPTS to refer to the loader and 
+ #  desired load options for your machine.
+ #
+-FORTRAN  = f77
+-OPTS     = -O4 -native -u
+-NOOPT    = -u
+-LOADER   = f77
++FORTRAN  = gfortran
++OPTS     = -O2 -mtune=native -fPIC -fdefault-integer-8
```
boegel, line 16: @sassy-crick Rather than hardcoding things here, why not use `${FC}` and `${FCFLAGS}` which are set by EasyBuild in the build environment?

boegel (changes requested):

easybuild/easyconfigs/s/SLICOT/slicot-makefile-flexiblas.patch
sassy-crick: Good idea. I changed that but I left the -fdefault-integer-8 as that seems to be important. I hope that is not biting what `OpenBLAS` and `FlexiBLAS` are currently having set. Maybe keeping this PR open until we got `-fdefault-integer-8` sorted for them as well?

sassy-crick:

zao:
Test report by @zao
**SUCCESS**
eb-mix.zao.se - Linux Rocky Linux 9.2 (Blue Onyx), x86_64, AMD Ryzen 9 3900X 12-Core Processor (zen2), Python 3.9.16
See https://gist.github.com/zao/350a01f22b2d4fe0065f08e3b2583127 for a full test report.

sassy-crick:

sassy-crick:
This build is done with `-fdefault-integer-8`, which is required for the library to be compatible with `MATLAB`. It looks like that `Octave` does not need that:
https://git.dynare.org/Dynare/dynare/-/blob/master/README.md#fedora-centos-or-rhel

I am not sure if we should use a prefix for that in the name of the EasyConfig file.

sassy-crick:

sassy-crick:
Closed as it is is the foss-2021b toolchain version

PR #16247, opened 2022-09-12
{bio}[GCC/11.2.0] PsiCLASS v1.0.3, GeneMark-ET v4.69

deniskristak:
adding easyconfigs: PsiCLASS-1.0.3-foss-2021b.eb, SAMtools-0.1.19-GCC-11.2.0.eb and patches: PsiCLASS-samtool_as_dep.patch

will used as a dependency for https://github.com/vscentrum/vsc-software-stack/issues/21


easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-1.0.3-foss-2021b.eb
```diff
@@ -0,0 +1,32 @@
+easyblock = "MakeCp"
+
+name = "PsiCLASS"
+version = "1.0.3"
+
+homepage = "https://github.com/splicebox/PsiCLASS"
+description = """PsiCLASS is a reference-based transcriptome assembler for single or multiple RNA-seq samples."""
+
+toolchain = {"name": "foss", "version": "2021b"}
```
jfgrimm, line 9: GCC should be sufficient, it doesn't look like it needs MPI/BLAS
```suggestion
toolchain = {"name": "GCC", "version": "11.2.0"}
```

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-1.0.3-foss-2021b.eb
```diff
@@ -0,0 +1,32 @@
...
+
+source_urls = ["https://github.com/splicebox/PsiCLASS/archive/"]
+sources = ["v%(version)s.tar.gz"]
+patches = ["PsiCLASS-samtool_as_dep.patch"]
+
+dependencies = [
+    ("SAMtools", "0.1.19"),
+]
+
+files_to_copy = ["*"]
```
jfgrimm, line 19: this will copy the bundled `samtools` directory to the install dir as well, I think we would want to avoid that

jfgrimm (changes requested):

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-1.0.3-foss-2021b.eb
deniskristak: done, thank you

easybuild/easyconfigs/g/GeneMark-ET/GeneMark-ET-4.69-GCCcore-11.2.0.eb
```diff
@@ -0,0 +1,27 @@
...
+name = "GeneMark-ET"
+version = "4.69"
+
+homepage = "http://exon.gatech.edu/GeneMark"
+description = "Eukaryotic gene prediction suite with automatic training"
+
+toolchain = {"name": "GCCcore", "version": "11.2.0"}
+
+# download via http://exon.gatech.edu/GeneMark/license_download.cgi
+# rename gmes_linux_64.tar.gz to gmes_linux_64-4.69.tar.gz
```
jfgrimm, line 12:
```suggestion
download_instructions = """
1. complete the license form: http://exon.gatech.edu/GeneMark/license_download.cgi
2. rename the tarball: `mv gmes_linux_64.tar.gz gmes_linux_64-%(version)s.tar.gz`
"""
```

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-samtool_as_dep.patch
```diff
@@ -0,0 +1,52 @@
+diff -ruN PsiCLASS-1.0.3_orig/alignments.hpp PsiCLASS-1.0.3/alignments.hpp
```
jfgrimm, line 1: please add an attribution (author) and description for the patch

jfgrimm (changes requested):
got a few minor nit-picks (nothing that affects the build)

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-samtool_as_dep.patch
jfgrimm: (and don't forget to update the checksum for the patch afterwards)

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-1.0.3-GCC-11.2.0.eb
```diff
@@ -0,0 +1,39 @@
+easyblock = "MakeCp"
```
boegel, line 1: @deniskristak It's probably cleaner to use `Binary` as easyblock here, since you're just copying everything anyway...

Or, if that works, you should refine `files_to_copy` to only the binaries are copied, and copy them into a `bin` subdirectory?

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-1.0.3-GCC-11.2.0.eb
```diff
@@ -0,0 +1,39 @@
...
+source_urls = ["https://github.com/splicebox/PsiCLASS/archive/"]
+sources = ["v%(version)s.tar.gz"]
+patches = ["PsiCLASS-samtool_as_dep.patch"]
+checksums = [
+    'c60c0f33f9a076b97ccdba68b9140a5fb6936a5c5e309d70b3d0cf558828f73d',  # v1.0.3.tar.gz
+    '4cb29e2a1e9e14dd83b2d3b63a9ef508255ffb8c992e8f45565a7666d3f5e72a',  # PsiCLASS-samtool_as_dep.patch
+]
+
+dependencies = [
+    ("SAMtools", "0.1.19"),
```
boegel, line 20: As discussed, depending on an ancient version of SAMtools is going to be a PITA when `PsiCLASS` is needed as a dependency for a tool (like [`Finder`](https://github.com/sagnikbanerjee15/Finder)) which also depends on a whole bunch of other tools, some of which require SAMtools 1.x...

So, in this case, it may be better to just stick to the bundled `samtools-0.1.9`, and not provide it as a dependency installed via EasyBuild...

boegel (changes requested):

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-1.0.3-GCC-11.2.0.eb
deniskristak: done

easybuild/easyconfigs/p/PsiCLASS/PsiCLASS-1.0.3-GCC-11.2.0.eb
deniskristak: @boegel to be honest, whole directory is only 15 MB and I am a bit afraid that I might forget to copy something important which might show later...therefore, wouldn't it be better to just copy everything here?

boegel: OK, I guess it makes sense to stick to `MakeCp` with that in mind...

boegel (changes requested):
@deniskristak Since we've reverted back to using the bundled SAMtools with PsiCLASS, we should remove the patch file + SAMtools easyconfig from this PR?

Maybe we should also add a comment in the PsiCLASS easyconfig to make it clear that we're deliberately using the bundled SAMtools (and why, i.e. because it's ancient and including it as a proper dependency is a PITA w.r.t. conflicts with dependencies of other modules)

boegel:
@deniskristak Looks like this can be closed, PsiCLASS was merged via #16906, GeneMark-ET was merged via #16907

PR #16316, opened 2022-09-26
ncbi-vdb-3.0.0 requires multidep exception for HDF5

fizwit:
add ncbi-vdb 3.0.0 to test/easyconfigs/easyconfigs.py multidep execption for HDF5 1.10.x


casparvl (approved):
Looks good to me. Next time, it's probably best to put the multidep exception and the PR for the software to which it applies (`ncbi-cdb-3.0.0`) in the same PR. That way we can actually see that it works in the CI of said software.

For now, I've just checked the change real carefully and assume that it works, and we should be able to see this immediately in https://github.com/easybuilders/easybuild-easyconfigs/pull/16376 . But if it doesn't, we'd have to roll back this PR, which would be a bit annoying.

PR #16344, opened 2022-09-30
{lib}[foss/2021a] torchdata v0.3.0

VRehnberg:
Found no existing issue connected to this.


easybuild/easyconfigs/t/torchdata-0.3.0-PyTorch-1.11.0.eb
```diff
@@ -0,0 +1,30 @@
...
+    ('Python', local_python_version),
+    ('PyTorch', local_pytorch_version, local_pytorch_versionsuffix),
+]
+
+sanity_pip_check = True
+use_pip = True
+
+exts_list = [
+    (name, version, {
+        'source_tmpl': 'v%(version)s.tar.gz',
```
Micket, line 25: This could come bite us with conflicting filenames if used as part of a bigger bundle

Hm, do we not have a way of specifying download-name in extensions?

Micket (changes requested):
Filename needs to include version suffix -CUDA-11.3.1 as well

Going forward with PyTorch, we should definitely bundle this component with the normal PyTorch, as it would be much cleaner, less messy suffixes and such.

In fact, we can consider just updating the PyTorch easyconfig themselves to add this extension. 
It's one of the few in-place changes we can do to modules without concern since EB allows for a very simple `eb --skip --rebuild Foo.eb` that only builds any missing extensions.

easybuild/easyconfigs/t/torchdata-0.3.0-PyTorch-1.11.0.eb
Micket: ok we do, e.g.
```
'sources': [{'download_filename': 'v%(version)s.tar.gz', 'filename': '%(name)s-%(version)s.tar.gz'}],
```

easybuild/easyconfigs/t/torchdata-0.3.0-PyTorch-1.11.0.eb
```diff
@@ -0,0 +1,30 @@
...
+local_pytorch_versionsuffix = '-foss-2021a-CUDA-11.3.1'
+
+homepage = 'https://github.com/pytorch/data'
+description = "TorchData is a prototype library of common modular data loading primitives for easily constructing flexible and performant data pipelines."
+
+toolchain = {'name': 'foss', 'version': '2021a'}
+
+dependencies = [
+    ('Python', local_python_version),
+    ('PyTorch', local_pytorch_version, local_pytorch_versionsuffix),
```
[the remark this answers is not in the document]
VRehnberg:
```suggestion
    ('CUDA', '11.3.1', '', True)
    ('Python', '3.9.5'),
    ('PyTorch', local_pytorch_version, '-CUDA-%(cudaver)s'),
```
add CUDA dependency explicitly to be able to use `'%(cudaver)s'`

easybuild/easyconfigs/t/torchdata-0.3.0-PyTorch-1.11.0.eb
VRehnberg:
```suggestion
        'sources': [{'download_filename': 'v%(version)s.tar.gz', 'filename': '%(name)s-%(version)s.tar.gz'}],
```
changing to

Micket: Ignore this. revisit this idea if we decide to bundle this with PyTorch

Micket: This doesn't matter for this single thing here. But you can do the download filename if you want to. Would make it easier to just copy this over to a bundle if we decide to.

VRehnberg:
@boegel I've added the checksum. Should I do anything else to alert your bot?

Micket:
nope. bot responds when it sees finished CI (since it's your first PR, we have to manually start it)

Micket:
```
 /home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/easybuild/easyconfigs/t/torchdata/torchdata-0.3.0-foss-2021a-PyTorch-1.11.0-CUDA-11.3.1.eb:9:121: E501 line too long (154 > 120 characters)
 ```
 
 hint: eb --check-style Foo-1.2.3.eb

VRehnberg:
It still says requesting changes, but as far as I can tell all requested changes are resolved?

Micket (approved):
lgtm

Micket:

Micket:
Going in, thanks @VRehnberg!

Issue #16330, opened 2022-09-28
Shall we start asking new PRs to use `SYSTEM` instead of `True`


Micket:
Going forward, we agreed that it's a good idea to start using SYSTEM instead of True for system level dependencies. 

Discussed briefly on todays conf call, but I think it's best if we start by
1. Doing a big PR where we at least change most of the recent stuff to use this. Otherwise, it will be an never-ending stream of new contributions that copy old stuff and then get asked to change it. Very annoying and a time waste.

2. Have the CI check for this automatically.

jfgrimm:
>     1. Doing a big PR where we at least change most of the recent stuff to use this. Otherwise, it will be an never-ending stream of new contributions that copy old stuff and then get asked to change it. Very annoying and a time waste.

done in PR #16386, and a few that were originally missed (or merged in the meantime) in #16384

>     2. Have the CI check for this automatically.

addressed  in #16384 

A PR to update the easybuild documentation was also merged in https://github.com/easybuilders/easybuild/pull/827, and instances in the framework tests updated in https://github.com/easybuilders/easybuild-framework/pull/4094

Micket:
we do this now

PR #16384, opened 2022-10-11
deprecate use of `True` in favour of `SYSTEM` for system-toolchain dependencies

jfgrimm:
adds a CI check as suggested in https://github.com/easybuilders/easybuild-easyconfigs/issues/16330
part 2 (updating existing easyconfigs to use `SYSTEM`) coming soon


jfgrimm:
> @jfgrimm Wouldn't it be easier if this change is also part of #16386 ? _Together_ they should create a CI that passes successfully, right? That's not easy to verify when they are in two different PRs.

I was thinking merge the updates to easyconfigs first, and then the CI after. The switch from `True` to `SYSTEM` is already supported, so I'd consider that distinct from the CI enhancement. In any case, any oversights should be caught in this PR

jfgrimm:
@casparvl besides, the other PR is already so big that it's easier to see the actual CI change when it's separated out

casparvl:
> > @jfgrimm Wouldn't it be easier if this change is also part of #16386 ? _Together_ they should create a CI that passes successfully, right? That's not easy to verify when they are in two different PRs.
> 
> I was thinking merge the updates to easyconfigs first, and then the CI after. The switch from `True` to `SYSTEM` is already supported, so I'd consider that distinct from the CI enhancement. In any case, any oversights should be caught in this PR

Ah you're right, I didn't realize that that one should just pass already with the current develop. Then, indeed, this is the sane way to do it :)

casparvl (approved):
Looks good to me! The CI passes, which is the most important check here. Just as a sanity check, I've also confirmed that the regex indeed correctly matches e.g. 

```
gompi-2016.04.eb
 gompi-2016.06.eb
 gompi-2016.07.eb
 gompi-2016.09.eb
 gompi-2016a.eb
 gompi-2016b.eb
 gompi-2017a.eb
 gompi-2017b.eb
 gompi-2018.08.eb
 gompi-2018a.eb
 gompi-2018b.eb
 gompi-2019a.eb
 gompi-2019b.eb
 ``` 
 but not
 ```
 gompi-2020a.eb
 gompi-2020b.eb
 gompi-2021a.eb
 gompi-2021b.eb
 gompi-2022.05.eb
 gompi-2022a.eb
```
as intended.

casparvl:
Thanks @jfgrimm for going through this effort! :)

PR #16413, opened 2022-10-15
Update VASP-5.4.1-intel-2016.02-GCC-4.9.eb

biozit:


sassy-crick:
Closing this since `intel-2016` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @biozit 

PR #16414, opened 2022-10-15
Create CMake-3.24.2.eb

biozit:


Thyre:
By now, we have [CMake-3.18.4.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/CMake/CMake-3.18.4.eb) and [CMake-3.31.8.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/CMake/CMake-3.31.8.eb) on the SYSTEM toolchain level. Therefore, I think this EasyConfig isn't really necessary anymore...

boegel:
> By now, we have [CMake-3.18.4.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/CMake/CMake-3.18.4.eb) and [CMake-3.31.8.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/CMake/CMake-3.31.8.eb) on the SYSTEM toolchain level. Therefore, I think this EasyConfig isn't really necessary anymore...

Agreed, closing...

PR #16483, opened 2022-10-25
{bio}[GCCcore 9.3.0] parallel-fastq-dump w/ SRA-Toolkit-3.0.0

hajgato:


easybuild/easyconfigs/p/parallel-fastq-dump/parallel-fastq-dump-0.6.6-GCCcore-9.3.0-SRA-Toolkit-3.0.0-Python-3.8.2.eb
```diff
@@ -0,0 +1,45 @@
...
+github_account = 'rvalieris'
+source_urls = [GITHUB_SOURCE]
+sources = ['%(version)s.tar.gz']
+checksums = ['38a3db72e4f87b8ca2412147a5938372098df15e839c78cddc8a7bfa3dd35b64']
+
+builddependencies = [('binutils', '2.34')]
+
+dependencies = [
+    ('Python', '3.8.2'),
+    ('SRA-Toolkit', local_sra_version, '-centos_linux64', SYSTEM)
```
boegel, line 27: @hajgato Can we use a from-source installation of `SRA-Toolkit` instead? See for example [this easyconfig](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/s/SRA-Toolkit/SRA-Toolkit-2.10.9-gompi-2020b.eb)

hajgato: well, we already had the binary in the develop version, in the parallel-fastq-dump what I modified it had a binary version, and honestly I wanted to do here the minimum effort as I needed the module quickly. If you insist I will make the SRA-Toolkit compiled version.

Anyway, based on the researcher's report, it might be possible that all SRA-Tookit 2.x version rendered useless as "version 2.x SRA-Tools do not have access to the database anymore (as you need a minimum 3.0.x version to have the right credentials to query the database)"

hajgato: ok, from https://github.com/ncbi/sra-tools

```
ANNOUNCEMENT:

August 4, 2022 : Security Update

Due to updated security at NCBI, versions of the SRA Toolkit 2.9.6 and older will no longer be able to connect to the NCBI data location service. We advise impacted users to update to the [latest version](https://github.com/ncbi/sra-tools/wiki/01.-Downloading-SRA-Toolkit) of the SRA Toolkit.
```

so we have to use SRA tookit >= 2.10.0, older versions are not functional.

boegel:
closing this since `GCCcore/9.3.0` is deprecated, see https://docs.easybuild.io/policies/toolchains

PR #16153, opened 2022-08-29
replace HDF5 v1.13.1 with v1.12.1 as dependency, since we shouldn't use odd minor versions of HDF5 which are not stable releases

Micket:
(created using `eb --new-pr`)
Odd releases are just RCs apparently.

PR #16554, opened 2022-11-04
add hdf5-1.13.3

LukeLabrie:
including easyconfig for 1.13.3 since source url for hdf5-1.13.1 easyconfigs is broken


jfgrimm:
see https://github.com/easybuilders/easybuild-easyconfigs/pull/16153

the HDF5 1.13.x versions are actually experimental releases,

LukeLabrie:
ah that makes sense. thanks.

PR #16620, opened 2022-11-11
{lib}[foss/2022a] TensorFlow v2.9.1 w/ Python 3.10.4 + CUDA 11.7.0

VRehnberg:
(created using `eb --new-pr`)


VRehnberg:
Currently problem with nsync_cv.h not found from dependency during build. Might have to solve similarily to how this was solved for protobuf.

VRehnberg:
My current plan is to create a genrule that symlinks needed protobuf
includes from PROTOBUF_INCLUDE_PATH as dependency for those rules where
similar errors occurs.

I expect it to be slow going.

Den ons 16 nov. 2022 18:24Alexandre Strube ***@***.***> skrev:

surak:
Adding
```patch
+    "coded_stream": ("google/protobuf/io/coded_stream.h", []),
```
right after the port_def line fixes this

surak:
adding the aforementioned line goes further but then I get this:

```
ERROR: /easybuild/2020/build/TensorFlow/2.9.1/foss-2022a-CUDA-11.7.0/TensorFlow/tensorflow-2.9.1/tensorflow/core/BUILD:1640:16: Compiling tensorflow/core/util/work_sharder.cc failed: undeclared inclusion(s) in rule '//tensorflow/core:framework_internal_impl':
this rule is missing dependency declarations for the following files included by 'tensorflow/core/util/work_sharder.cc':
  'bazel-out/k8-opt/bin/external/com_google_protobuf/google/protobuf/io/coded_stream.h'
Target //tensorflow/tools/pip_package:build_pip_package failed to build
```

VRehnberg:
One possible reason for the latest error you write about is that the includes in coded_stream.h needs to be in the list of imports that is the second value in the include tuples.

I've seen the following include errors so far
 - <google/protobuf/port_def.inc>
 - "google/protobuf/io/coded_stream.h"
 - "nsync_cv.h"       // NOLINT
 - "double-conversion/double-conversion.h"
 - "google/protobuf/duration.pb.h"
adding duration.pb.h to the WELL_KNOWN_PROTOS dict gets a similar error as coded_stream.h, at least if the import list is empty. I haven't had time to test my hypothesis yet.

VRehnberg:
What I don't understand is about this PR https://github.com/tensorflow/tensorflow/pull/44144 that seem to have removed many of those include files that we now want to link to.

VRehnberg:
This issue https://github.com/tensorflow/tensorflow/issues/37861 describes our issue quite well. However, while the symptom is the same it doesn't seem to be the same cause. CPATH is specified correctly in the builds, but inlcude files are not found anyway.

The command that is run is
`external/local_config_cuda/crosstool/clang/bin/crosstool_wrapper_driver_is_not_gcc`
which is a
> Crosstool wrapper for compiling CUDA programs.
and uses(?) nvcc.

Perhaps, nvcc doesn't honor the CPATH? I've got no idea. Put it seems that going back to symlinking files is probably the wrong approach and should perhaps be avoided.

surak:
I couldn't care less for protobuf version, and I would be happy to have a second protobuf in the system only for tensorflow. 

They mention protobuf 3.6.1 in [their ci build](https://github.com/tensorflow/tensorflow/blob/v2.11.0/tensorflow/tools/ci_build/install/install_proto3.sh)
 - I tried that.

As it also depends on protobuf-python, I created both from 3.19.4, just removing the checksums and changing versions. protobuf-python 3.6.1 needs a little patch which I made here: [python3 patch for protobuf-python 3.6.0](https://gist.github.com/surak/cddbba4d1c0c8d364b8d299b4d33dc6a)

But then I get this: 
```
tensorflow/core/grappler/mutable_graph_view.cc:345:18: error: const class google::protobuf::MessageLite has no member named ShortDebugString
```
, which is correct - protobuf 3.6.1 doesn't have this `ShortDebugString` [protobuf 3.6 source](https://github.com/protocolbuffers/protobuf/blob/3.6.x/src/google/protobuf/message_lite.h) (first version which has it is 3.8.0)

So, I tried now with 3.8.0 to see if I see any improvement. Which made me arrive to... 

The same place. 

```c++
./tensorflow/core/platform/protobuf.h:28:10: fatal error: google/protobuf/io/coded_stream.h: No such file or directory
   28 | #include "google/protobuf/io/coded_stream.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

I don't know what to do further.

VRehnberg:
> ```c++
> ./tensorflow/core/platform/protobuf.h:28:10: fatal error: google/protobuf/io/coded_stream.h: No such file or directory
>    28 | #include "google/protobuf/io/coded_stream.h"
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> ```

google/protobuf/io/coded_stream.h exists at $PROTOBUF_INCLUDE_PATH. As such either CPATH=$PROTOBUF_INCLUDE_PATH:$CPATH should be specified or used by the build tool. From logs (building with `eb --debug --trace`) I can see that the correct CPATH is used for some SUBCOMMANDs in some actions  

With Bazel I learnt that --action_env command line variable should be used to specify this. And at least on my system it is and CPATH is used for some actions. However, not for all. Right now I'm trying to find the first commit that introduced the error and then I should know which rule to patch at least.

This is how it has been done in the past https://github.com/tensorflow/tensorflow/issues/43019

Other avenues are:
 - Figure out what goes wrong by understanding Bazel and TensorFlow such that you can tell how to fix it by looking at the log (I gave up on this in favor of brute force)
 - Create an issue with TensorFlow and hope that they know how to fix it (should probably create an issue sooner or later anyway)

Flamefire:
> With Bazel I learnt that --action_env command line variable should be used to specify this. And at least on my system it is and CPATH is used for some actions. However, not for all. Right now I'm trying to find the first commit that introduced the error and then I should know which rule to patch at least.
> 
> This is how it has been done in the past [tensorflow/tensorflow#43019](https://github.com/tensorflow/tensorflow/issues/43019)

I started getting back into TensorFlow ECs now and got 2.7.1 working on all platforms. Then I went to the next version we have which is 2.8.4 and it doesn't have a CUDA EC, so I added it and am testing it now:

I run into the same issue you see here: `CPATH` action_env is not passed to some commands resulting in e.g.

> ```
> In file included from bazel-out/k8-opt-exec-50AE0418/bin/tensorflow/compiler/mlir/tools/kernel_gen/compile_cache_item.pb.cc:4:
> bazel-out/k8-opt-exec-50AE0418/bin/tensorflow/compiler/mlir/tools/kernel_gen/compile_cache_item.pb.h:10:10: fatal error: google/protobuf/port_def.inc: No such file or directory
>    10 | #include <google/protobuf/port_def.inc>
> ```

So the issue appeared somewhere between 2.7.1 and 2.8.4 and only for the CUDA version, the non-CUDA version compiles fine for me.

As for `crosstool_wrapper_driver_is_not_gcc`: This is a wrapper on the compiler which can be either GCC or NVCC depending on what the build process (i.e. Bazel) passes to it. Not much going on there.

Previously I was able to "solve" the issue by replacing "`exec_tools=`" by "`tools=`" in the build files but that doesn't seem to work now. I'm now trying to find the commit introducing the issue in TF

Flamefire:
Happy new year! Just as an update on this: On the last (work) day of the last year I was able to figure out why the includes were missing. I summarized that in an [issue against Bazel](https://github.com/bazelbuild/bazel/issues/17062) pointing out the commit I found via `git bisect`: https://github.com/tensorflow/tensorflow/commit/07cbc7bb0bf899aac2bee5e21e1ba4eb40038682

I'm also currently working on a fix which will solve the issue here and in other TF ECs by changing the EasyBlock. So I'd ask for a bit of patience while I'm testing this.

VRehnberg:
> I'm also currently working on a fix which will solve the issue here and in other TF ECs by changing the EasyBlock. So I'd ask for a bit of patience while I'm testing this.

Happy new year and thanks for your effort! I started bisecting myself but I kept running into other bugs for commits between releases and feedback was so slow that I got tired of it and haven't touched it in a few weeks.

I wouldn't mind taking a look at something untested and could run tests on my end as well.

VRehnberg:
Closing in favor of https://github.com/easybuilders/easybuild-easyconfigs/pull/17092

PR #16842, opened 2022-12-08
{geo}[foss/2022a,gompi/2022a,system/system] CliMetLab v0.12.6, Magics v4.13.0, ecBuild v3.7.0 w/ Python 3.10.4

smoors:
(created using `eb --new-pr`)


easybuild/easyconfigs/e/ecBuild/ecBuild-3.7.0.eb
```diff
@@ -0,0 +1,35 @@
...
+        'source_urls': [GITHUB_SOURCE],
+        'filename': '%(version)s.tar.gz',
+        'extract_cmd': 'tar -xzf %s --strip-components=1',
+    },
+]
+checksums = ['44ba49b6b628e88bc0145f0cd1fdbdff37fbff2327c581f22a47cd10263a21b6']
+
+buildininstalldir = True
+
+skipsteps = ['install']
```
lexming, line 26: I'm curious what is this doing. `Tarball` will already copy the sources in the build directroy to install directory, isn't it?

easybuild/easyconfigs/c/CliMetLab/CliMetLab-0.12.6-foss-2022a.eb
```diff
@@ -0,0 +1,84 @@
...
+    ('PyYAML', '6.0'),
+    ('Magics', '4.13.0'),
+    ('cdsapi', '0.5.1'),
+]
+
+use_pip = True
+sanity_pip_check = True
+
+# stick to termcolor 1.x, to avoid hatchling required dependency
+# remove ecmwflibs requirement (provides ecCodes and Magics)
```
lexming, line 34: This was confusing at first, I suggest to expand this comment:
```suggestion
# remove ecmwflibs requirement as it just bundles ecCodes and Magics and there are no direct imports to it
```

easybuild/easyconfigs/e/ecBuild/ecBuild-3.7.0.eb
smoors: `buildininstalldir = True` will unpack the sources directly into the installdir, so there is no need for the extra install step.

lexming:

lexming (approved):
LGTM

lexming:
Going in, thanks @smoors !

PR #16846, opened 2022-12-09
{lib}[GCCcore/11.3.0] CFITSIO v4.2.0

surak:
(created using `eb --new-pr`)


smoors:
@surak 
Please add a short description at the top of the patch to describe what it does, why it's needed, and/or which errors it's fixing exactly.

Please also mention author or add a reference to upstream fix.

Don't forget that you will also need to change the SHA256 checksums for this patch file.

boegel:
No activity since Dec'22, and software version is no longer relevant either, so closing...

PR #17052, opened 2023-01-05
{data}[GCCcore/10.2.0,GCCcore/11.3.0] UDA v2.6.0

SimonPinches:
(created using `eb --new-pr`)


easybuild/easyconfigs/u/UDA/UDA-2.6.0-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,65 @@
...
+
+dependencies = [
+    ('libxml2', '2.9.10'),
+    ('Boost', '1.74.0'),
+    # OS dependency should be preferred if the os version is more recent then this version
+    # it's nice to have an up to date openssl for security reasons
+    # ('OpenSSL', '1.1.0e'),
+]
+
+osdependencies = [('openssl-devel', 'libssl-dev', 'libopenssl-devel')]
```
jfgrimm, line 31: we can use the `OpenSSL` wrapper module for this:

```suggestion
    ('OpenSSL', '1.1', '', SYSTEM),
]
```

easybuild/easyconfigs/u/UDA/UDA-2.6.0-GCCcore-11.3.0.eb
```diff
@@ -0,0 +1,65 @@
...
+
+dependencies = [
+    ('libxml2', '2.9.13'),
+    ('Boost', '1.79.0', '-no_mpi'),
+    # OS dependency should be preferred if the os version is more recent then this version
+    # it's nice to have an up to date openssl for security reasons
+    # ('OpenSSL', '1.1.0e'),
+]
+
+# osdependencies = [('openssl-devel', 'libssl-dev', 'libopenssl-devel')]
```
jfgrimm, line 31: we can use the `OpenSSL` wrapper module for this:

```suggestion
    ('OpenSSL', '1.1', '', SYSTEM),
]
```

easybuild/easyconfigs/u/UDA/UDA-2.6.0-GCCcore-11.3.0.eb
```diff
@@ -0,0 +1,65 @@
...
+builddependencies = [
+    ('binutils', '2.38'),
+    ('CMake', '3.23.1'),
+    ('libMemcached', '1.0.18'),
+    ('libtirpc', '1.3.2'),
+]
+
+dependencies = [
+    ('libxml2', '2.9.13'),
+    ('Boost', '1.79.0', '-no_mpi'),
```
jfgrimm, line 25: we already build `Boost` with the tagged layout, no need to introduce a `-no_mpi` variant
```suggestion
    ('Boost', '1.79.0'),
```

easybuild/easyconfigs/u/UDA/UDA-2.6.0-GCCcore-11.3.0.eb
```diff
@@ -0,0 +1,65 @@
+easyblock = 'CMakeMake'
+
+name = 'UDA'
+version = '2.6.0'
+
+homepage = 'https://github.com/ukaea/UDA'
+description = """Universal Data Access"""
+
+toolchain = {'name': 'GCCcore', 'version': '11.3.0'}
```
jfgrimm, line 9: we already have `Boost` at `GCC` level, so we should really use that (and bump this easyconfig to `GCC`)
```suggestion
toolchain = {'name': 'GCC', 'version': '11.3.0'}
```

easybuild/easyconfigs/u/UDA/UDA-2.6.0-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,65 @@
+easyblock = 'CMakeMake'
+
+name = 'UDA'
+version = '2.6.0'
+
+homepage = 'https://github.com/ukaea/UDA'
+description = """Universal Data Access"""
+
+toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
```
jfgrimm, line 9: we already have `Boost` at `GCC` level, so we should really use that (and bump this easyconfig to `GCC`)
```suggestion
toolchain = {'name': 'GCC', 'version': '10.2.0'}
```

jfgrimm (changes requested):
I would suggest removing the `GCCcore`-level `Boost` easyconfigs you're adding, and instead using the `GCC`-level ones we already have in the repository. This will require bumping `UDA` to `GCC` (note that the filenames should be changed to match)

boegel:

@SimonPinches If `UDA` is still relevant, please consider opening a new pull with a more recent toolchain

PR #17095, opened 2023-01-12
This is needed for JUWELS booster

surak:
This test gives a slightly higher number while installing on Juwels booster.


boegel:

boegel:
Patch seems to need work, doesn't apply for `jax-0.3.14-foss-2022a-CUDA-11.7.0.eb` for example.

Full test report coming up...

branfosj:
> Patch seems to need work, doesn't apply for `jax-0.3.14-foss-2022a-CUDA-11.7.0.eb` for example.
> 
> Full test report coming up...

The new part of the patch only applies in the latest Jax easyconfigs. So, `jax-0.3.9_relax-test-tolerance.patch` should be reverted to the original content and a new patch added to relax the tolerance in the failing test for 0.3.23.

boegel:
Hasn't been touched in quite a while, unlikely to still be relevant, so closing as is...

Issue #5151, opened 2017-09-21
checksums for source tarballs downloaded from github.com/.../.../archive can change over time


boegel:
First reported by @schiotz at https://github.com/easybuilders/easybuild-easyconfigs/pull/4871#issuecomment-331072093, several other projects have been hit by this as well:

* https://github.com/google/protobuf/issues/3619
* https://github.com/bazelbuild/bazel/issues/3722
* https://github.com/tensorflow/tensorflow/issues/12979

Some more details in https://github.com/libgit2/libgit2/issues/4343#issuecomment-328631745

Long story short: we should try to avoid downloading from `github.com/.../.../archive` (and try to use `github.com/.../.../releases` (or another located where 'packaged' tarballs are available) instead, if at all possible.

If not, the alternatives I see are:

* doing a `git clone` on the tagged version and creating the tarball ourselves (directly using `tar`), which should always give the same tarball?
* not including any checksums if we need to download from `github.com/.../.../archive`

boegel:
related discussion in Homebrew: https://github.com/Homebrew/homebrew-core/issues/18044

boegel:
Interesting quote by someone at GitHub (from https://github.com/Homebrew/homebrew-core/issues/18044#issuecomment-329313468)

```
Not all hashes will change. Most of the changes are due to the bugfix in git/git@22f0dcd, so only tarballs with filenames greater than 100 characters are affected.
```

and (from https://github.com/Homebrew/homebrew-core/issues/18044#issuecomment-329301763)

```
Historically, they don't change that often ...
```

So short-term, we can figure out which checksums we need to fix, and then figure out a better way of dealing with this.

boegel:
Idea by @wpoely86: we should also support `--ignore-checksums` and `--force-download`

tgamblin:
> doing a git clone on the tagged version and creating the tarball ourselves (directly using tar), which should always give the same tarball?

Note that this doesn't work.  [`tar` is known to be nondeterministic](https://lists.gnu.org/archive/html/help-tar/2015-05/msg00005.html) across systems/users/etc., so you can't really rely on this for checksumming.  You might be able to steal some of Debian's boilerplate from that issue, I guess.

But really why do this?  If you're already going to use `git clone` as your fetch mechanism, just clone at a particular commit and you're done.  The commit hash is already verified by `git` during the clone.

schiotz:
[...]

PR #17218, opened 2023-01-26
adding easyconfigs: Bowtie2-2.5.1-GCC-12.2.0.eb

benteb:


easybuild/easyconfigs/b/Bowtie2/Bowtie2-2.5.1-GCC-12.2.0.eb
```diff
@@ -0,0 +1,55 @@
...
+# The Francis Crick Institute
+# Modified by: Kurt Lust, UAntwerp
+# Modified by: Sebastien Moretti for non-x86_64 systems
+# SIB Swiss Institute of Bioinformatics
+# Modified by: Bente Barge <bente.s.barge@uit.no>
+
+name = 'Bowtie2'
+version = '2.5.1'
+_simde_version = '20220504'
+_simde_commit = 'cbef1c1'
```
jfgrimm, line 17: since we're not using a release version, perhaps we should bump to the latest commit?
```suggestion
_simde_version = '20230122'
_simde_commit = '6e9ac24'
```

easybuild/easyconfigs/b/Bowtie2/Bowtie2-2.5.1-GCC-12.2.0.eb
```diff
@@ -0,0 +1,55 @@
...
+# SIMD Everywhere implementations only, for non-x86_64 systems
+if ARCH != 'x86_64':
+    source_urls += ['https://github.com/simd-everywhere/simde-no-tests/archive']
+    sources += [
+        {'download_filename': '%s.zip' % _simde_commit,
+         'filename': 'simde-%s.zip' % _simde_version,
+         'extract_cmd': 'unzip %%s && mv simde*-%s*/* %%(namelower)s-%%(version)s/third_party/simde/' % _simde_commit,
+         }
+    ]
+    checksums += ['d01f084ef5ff69b0a9b96370ae314fe1e55ef3339b25afcd3385958ac0e6ad68']
```
jfgrimm, line 42: updated checksum if using commit `6e9ac24`
```suggestion
    checksums += ['d961f5f2720c2148d88e687a5d5262d15b402f81675f62f217f60603466f8a4e']
```

jfgrimm (changes requested):

easybuild/easyconfigs/b/Bowtie2/Bowtie2-2.5.1-GCC-12.2.0.eb
benteb: Hi. First I committed this suggestion. Now having a second look, I think you meant to updated the actual checksum, not this "checksum +=" inside this IF-loop that is triggered for non-x86_64 systems? See my latest commit where I undid this change and updated the actual checksum.

jfgrimm: no, I did mean to update the `simde` checksum
github made some changes to how they calculate checksums recently -- see [this comment](https://github.com/easybuilders/easybuild-easyconfigs/issues/5151#issuecomment-1410107202). The changes have since been reverted), so if you got a checksum error that was probably why

jfgrimm: the `simde` checksum needs updating since the other suggestion bumped the version it downloads to a newer commit (which obviously has a different checksum)

boegel:
We already have this merged via:
* #18097

So closing...

PR #17231, opened 2023-01-29
{bio}[foss/2021b] svFSI v2022.09.26

aebruno:
Easybuild recipe for svFSI (https://github.com/SimVascular/svFSI).

### DIFF
--- a/easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
+++ b/easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
@@ -18,8 +18,6 @@ checksums = [
     'be5a4889167cf242d863fa424ade7d8556c1d3d9525a809a230611a14ab095f8',  # svFSI-2022.09.26_cmake-rpath-fix.patch
 ]
 
-separate_build_dir = True
-
 builddependencies = [('CMake', '3.22.1')]
 
 configure_cmd = 'cmake -DCMAKE_INSTALL_PREFIX=%(installdir)s -DCMAKE_BUILD_TYPE=RelWithDebInfo '
--- a/easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
+++ b/easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
@@ -15,13 +15,12 @@ patches = [
 ]
 checksums = [
     '4f4fbf4112402c7d037f35043f8aa77b141276a59ca957614cf8de20543bd6eb',  # 2022.09.26.tar.gz
-    'be5a4889167cf242d863fa424ade7d8556c1d3d9525a809a230611a14ab095f8',  # svFSI-2022.09.26_cmake-rpath-fix.patch
+    '1fb62ea33f2970c4674441d5ad2fc698fd1f526c158536cbccb35751374cd49f',  # svFSI-2022.09.26_cmake-rpath-fix.patch
 ]
 
-builddependencies = [('CMake', '3.22.1')]
+srcdir = "Code"
 
-configure_cmd = 'cmake -DCMAKE_INSTALL_PREFIX=%(installdir)s -DCMAKE_BUILD_TYPE=RelWithDebInfo '
-configure_cmd += '../svFSI-%(version)s/Code'
+builddependencies = [('CMake', '3.22.1')]
 
 sanity_check_paths = {
     'files': ['SV/bin/svFSI', 'SV/lib/lib_simvascular_thirdparty_zlib.a'],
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
```diff
@@ -0,0 +1,35 @@
...
+sources = ['%(version)s.tar.gz']
+patches = [
+    '%(name)s-%(version)s_cmake-rpath-fix.patch',
+]
+checksums = [
+    '4f4fbf4112402c7d037f35043f8aa77b141276a59ca957614cf8de20543bd6eb',  # 2022.09.26.tar.gz
+    'be5a4889167cf242d863fa424ade7d8556c1d3d9525a809a230611a14ab095f8',  # svFSI-2022.09.26_cmake-rpath-fix.patch
+]
+
+separate_build_dir = True
```
jfgrimm, line 21: doesn't need to be explicit, this is already default behaviour
```suggestion
```

easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
```diff
@@ -0,0 +1,35 @@
...
+
+builddependencies = [('CMake', '3.22.1')]
+
+configure_cmd = 'cmake -DCMAKE_INSTALL_PREFIX=%(installdir)s -DCMAKE_BUILD_TYPE=RelWithDebInfo '
+configure_cmd += '../svFSI-%(version)s/Code'
```
jfgrimm, line 26: no need to override the default `CMakeMake` `configure_cmd`, just pass extra options via `configopts`. 
EasyBuild should already pass the install prefix, and correct source directory. For `BUILD_TYPE` we actually have another parameter to control this, e.g.: 
```suggestion
build_type = 'RelWithDebInfo'
```
if you want to stick with `RelWithDebInfo`; by default, EasyBuild targets a `Release` build

easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26_cmake-rpath-fix.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/Code/CMakeLists.txt b/Code/CMakeLists.txt
```
jfgrimm, line 1: please add an author and patch description

easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
```diff
@@ -0,0 +1,35 @@
...
+    '%(name)s-%(version)s_cmake-rpath-fix.patch',
+]
+checksums = [
+    '4f4fbf4112402c7d037f35043f8aa77b141276a59ca957614cf8de20543bd6eb',  # 2022.09.26.tar.gz
+    'be5a4889167cf242d863fa424ade7d8556c1d3d9525a809a230611a14ab095f8',  # svFSI-2022.09.26_cmake-rpath-fix.patch
+]
+
+separate_build_dir = True
+
+builddependencies = [('CMake', '3.22.1')]
```
jfgrimm, line 23: we should perhaps also add the optional dependencies that `svFSI` supports?
`Boost`, `HDF5`, `Hypre`, `MUMPS`, `Trilinos` should all have existing easyconfigs

jfgrimm (changes requested):
Thanks for your contribution @aebruno!
I've left a few suggestions, let me know if you have any questions about any of them

easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
aebruno: All fixed.

easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
aebruno: Thanks for the pointers! This should be all cleaned up. Adding the `srcdir` option solved the issue nicely.

easybuild/easyconfigs/s/svFSI/svFSI-2022.09.26-foss-2021b.eb
aebruno: @jfgrimm Unless I'm missing something, it looks like all those deps are related to Trilinos. I didn't see an easybuild recipe for Trilinos for foss-2021b which is why I was inclined to leave it off this recipe. Let me know what you think?

boegel:
Closing this since `foss/2021b` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @aebruno 

PR #17272, opened 2023-02-07
{ai}[foss/2021b] PyTorch v1.12.1 w/ Python 3.9.6 w/ CUDA 11.5.2 & dependencies

Flamefire:
(created using `eb --new-pr`)

Update of #17154 using CUDA 11.5 due to an incompatibility.

As this works while the other does not this   
Closes #17154

- [x] Requires #17289

PR #17289, opened 2023-02-09
Allow multiple ECs with same name if they differ by CUDA version

Flamefire:
Add `CUDA` to `versionsuffix_deps` next to `CUDAcore` as newer toolchains don't use `CUDAcore`.

Tested locally on top of #17272 : Before `test_dep_versions_per_toolchain_generation` fails, after this PR it succeeds.


boegel (approved):
lgtm

boegel:
@Flamefire Thanks, can you sync #17272 now that this is merged?

Issue #11824, opened 2020-12-04
Duplication between EasyConfigs that have the same name and different capitalization


mboisson:
I just found
https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/g/grace
and
https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/g/Grace

which are the same, with different capitalization

@boegel

mboisson:
For completeness, I also extracted the list of EasyConfigs that have the same name, just with different capitalization : 
```
      2 a/arb
      2 a/arrow
      2 a/assimp
      2 b/bamm
      2 b/blasr
      2 c/casper
      2 c/charmm
      2 c/check
      2 g/grace
      2 g/graphviz
      2 h/hyperopt
      2 i/ioapi
      2 l/libpsml
      2 m/magma
      2 n/nanofilt
      2 n/ncl
      2 o/orca
      2 x/xmlf90
```

SebastianAchilles:
`Grace` was fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/14493

Micket:
Discussed brielfy on todays meeting; i wanted to give an update. Current offenders are:
- a/arb - Arb vs ARB, but ARB is just 2 patches and should just be archived. Trivial.
- b/bamm - BAMM vs BamM, neither have been updated in some time and will be archived.
- b/blasr - BLASR vs blasr; but blasrc just has a leftover patch should be moved or archived
- c/charmm - CHARMM vs charmm same software
- c/check - check vs Check same software
- g/graphviz - graphviz vs Graphviz - graphviz is the python package, it should just get a suffix. It actually depends on Graphviz!
- h/hyperopt - hyperopt vs Hyperopt same software
- l/libpsml - libpsml vs libPSML same software
- m/magma -  magma vs MAGMA MAGMA should have a suffix to indicate it's the gene analysis tool
- n/nanofilt - NanoFilt vs nanofilt same software
- n/ncl - ncl vs NCL: ncl should probably be renamed "NEXUS" or something

lexming:
`Hyperopt` was fixed in #16697

Micket:
What would everyone prefer for MAGMA? I suggest
1. MAGMA-gene-analysis
2. MAGMA-CTGlab
3. MAGMA-CNCR

lexming:
[...]

PR #17303, opened 2023-02-11
Add case-insensitive name clash test

branfosj:
See #11824

```
FAIL: test_easyconfig_name_clashes (test.easyconfigs.easyconfigs.EasyConfigTest)
Make sure there is not a name clash when all names are lowercase
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 953, in test_easyconfig_name_clashes
    self.assertTrue(False, "EasyConfigs with case-insensitive name clash: %s" % duplicates)
AssertionError: False is not true : EasyConfigs with case-insensitive name clash: {'arb': ['ARB', 'Arb'], 'nanofilt': ['NanoFilt', 'nanofilt'], 'ncl': ['NCL', 'ncl'], 'check': ['Check', 'check'], 'charmm': ['charmm', 'CHARMM'], 'graphviz': ['Graphviz', 'graphviz'], 'magma': ['MAGMA', 'magma'], 'libpsml': ['libpsml', 'libPSML'], 'bamm': ['BAMM', 'BamM'], 'blasr': ['BLASR', 'blasr']}
```


test/easyconfigs/easyconfigs.py
```diff
@@ -926,6 +927,31 @@ def test_easyconfig_locations(self):
...
+        topdir = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+        names = defaultdict(list)
+        for (dirpath, _, _) in os.walk(topdir):
+            # ignore git/svn dirs & archived easyconfigs
+            if '/.git/' in dirpath or '/.svn/' in dirpath or '__archive__' in dirpath:
+                continue
+            dirpath_split = dirpath.replace(topdir, '').split(os.sep)
+            if len(dirpath_split) == 5:
+                name = dirpath_split[4]
+                names[name.lower()].append(name)
```
lexming, line 941: Can we use `robot_path` instead of jumping 3 levels up?  It would be even better if we could use a case insensitive version of `robot_find_easyconfig` from framework, then we can benefit from the indexes.

branfosj: The jump three levels up was copied from the test above this one. I am unsure what might happen is we use `robot_path` when running the tests locally if we have that set.

For `python3` (RHEL 8.6 default version) the speed is not an issue:
```
real    0m0.321s
user    0m0.108s
sys     0m0.184s
```

Sadly, for `python2` it is a lot slower:
```
real    0m49.083s
user    0m0.588s
sys     0m1.439s
```

test/easyconfigs/easyconfigs.py
```diff
@@ -926,6 +927,27 @@ def test_easyconfig_locations(self):
...
```
lexming, line 941: maybe we can make the list of ignored dirs easier to maintain and also get rid of the `continue`  for the same price:
```suggestion
        # ignore git/svn dirs & archived easyconfigs
        ignore_dirs = ['.git', '.svn', '__archive__']
        for (dirpath, _, _) in os.walk(topdir):
            if not any('/%s' % d in dirpath for d in ignore_dirs):
                dirpath_split = dirpath.replace(topdir, '').split(os.sep)
                if len(dirpath_split) == 5:
                    name = dirpath_split[4]
                    names[name.lower()].append(name)
```

lexming (approved):
LGTM

lexming:
Going in, thanks @branfosj !

PR #17346, opened 2023-02-15
Archive ARB and blasr patch files

branfosj:
for #11824 

for `ARB` and `blasr` there are only `.patch` files in the directory, so we'll move these into `__archive__` to remove the case-insensitive directory name collision.


Micket (commented):
I *think* blasr patch actually belongs to `__archive__/b/BLASR/`  but this patch is so old it's hard to even find the software. In fact, BLASR itself has basically been removed from the internet as far as i can tell, so i could be wrong here.

branfosj:
> I _think_ blasr patch actually belongs to `__archive__/b/BLASR/` but this patch is so old it's hard to even find the software. In fact, BLASR itself has basically been removed from the internet as far as i can tell, so i could be wrong here.

It is used in https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/__archive__/b/blasr/blasr-smrtanalysis-2.1-goolf-1.4.10.eb, so I moved it into that same directory. That should have been named `BLASR`, but I am not sure we should bother fixing the name of an archived easyconfig.

Micket (approved):
lgtm

PR #17348, opened 2023-02-15
Archive BAMM, BamM, and GroopM to avoid BAMM/BamM name collision

branfosj:
for #11824

All of the `BAMM` and `BamM` easyconfigs are old, so we'll archive them to avoid the case-insensitive name collision. `GroopM` depends on `BamM`, so we'll archive that as well.


Micket (approved):
lgtm

PR #17352, opened 2023-02-15
rename python graphviz to graphviz-python to avoid case-insensitiive name clash

branfosj:

I'm unable to build easyconfigs from these toolchains.


Micket (approved):
lgtm. to old to build, so i just very thoroughly looked throw the changes and that will have to do.

PR #17353, opened 2023-02-15
rename libPSML for consistency with libpsml

branfosj:


Micket (approved):
to bad this is basically picking the lesser name since it's not the capitalization used upstream, but i guess we have many such sins in other software with no non-painful way of changing them now.

branfosj:
> to bad this is basically picking the lesser name since it's not the capitalization used upstream, but i guess we have many such sins in other software with no non-painful way of changing them now.

I've put the alternative in #17359

branfosj:
closing in favour of `libPSML` in #17359

PR #17364, opened 2023-02-17
archive charmm and CHARMM easyconfigs to avoid case-insensitive name clash

branfosj:

`charmm` and `CHARMM` are different variants of the same software. However, all the easyconfigs are old. So, rather than sort out the naming now, we should archive them all and decide on the name if a new easyconfig is submitted.


Micket (approved):
lgtm

For reference, i think we could just have "CHARMM" and label the free version with a versionsuffix (presumably, you can only ever use one of them at a time anyway, so they should conflict)

PR #17217, opened 2023-01-26
{bio,data}[foss/2022a] Arrow v8.0.0, ont-remora v2.0.0 w/ Python 3.10.4

branfosj:
(created using `eb --new-pr`)

CUDA variant of #17183

PR #17183, opened 2023-01-23
{bio}[foss/2022a] ont-remora v2.0.0 w/ Python 3.10.4

branfosj:
(created using `eb --new-pr`)

* [x] #17179
* [x] #17181

PR #17422, opened 2023-02-24
{bio}[foss/2022a] Cellpose v2.2 w/ Python 3.10.4

branfosj:
(created using `eb --new-pr`)


easybuild/easyconfigs/c/Cellpose/Cellpose-2.2-foss-2022a-CUDA-11.7.0.eb
```diff
@@ -0,0 +1,56 @@
...
+version = '2.2'
+versionsuffix = '-CUDA-%(cudaver)s'
+
+homepage = "https://www.cellpose.org/"
+description = """a generalist algorithm for cellular segmentation"""
+
+toolchain = {'name': 'foss', 'version': '2022a'}
+
+dependencies = [
+    ('CUDA', '11.7.0', '', True),
```
jfgrimm, line 14:
```suggestion
    ('CUDA', '11.7.0', '', SYSTEM),
```

branfosj: Yeah. I'd not bothered fixing this as there is a wider PyTorch `1.12.0` vs `1.12.1` issue to decide on. See also #17183, #17217, and the easyconfigs for `cell2location` and `scvi-tools` that I've not yet PR-ed.

PR #17527, opened 2023-03-14
Added easyconfig for Apache Avro C++

ll4strw:
Added easyconfig for Apache Avro C++


easybuild/easyconfigs/a/avrocpp/avrocpp-1.11.1-GCC-11.2.0.eb
```diff
@@ -0,0 +1,38 @@
+easyblock = 'Binary'
```
casparvl, line 1: If I look at the install_cmd, this seems to be a CMakeMake installation. Wouldn't it be possible to use the `CMakeMake` easyblock? Binary is really for binary installations, i.e. binaries that are not compiled, but only installed. That means (among other things) that e.g. rpath-checking is skipped.

Note that the CMakeMake easyblock supports specifying a generator (see [here](https://github.com/easybuilders/easybuild-easyblocks/blob/fce1775ad011a9ccc3ad6963cbd6f350af6c3301/easybuild/easyblocks/generic/cmakemake.py#L198)).

I think the only thing custom to this install_cmd is the change in directory. I guess you can do this using `preconfigopts=cd lang/c++ &&` or something similar. I'm not 100% sure what the default startddir for EasyBuild will be when building, but you may need to specify a similar `prebuildopts` and `preinstallopts`.

casparvl (changes requested):

easybuild/easyconfigs/a/avrocpp/avrocpp-1.11.1-GCC-11.2.0.eb
```diff
@@ -0,0 +1,38 @@
...
+
+toolchain = {'name': 'GCC', 'version': '11.2.0'}
+
+source_urls = ['https://github.com/apache/avro/archive/refs/tags/']
+sources = ['release-%(version)s.tar.gz']
+checksums = ['599f96bb405f72a35154b2477caa6254d723bb4e3f6a0e54e9ae540664321752']
+
+builddependencies = [
+    ('binutils', '2.37', '', ('GCCcore', '11.2.0')),
+    ('CMake', '3.22.1', '', ('GCCcore', '11.2.0')),
```
boegel, line 17: No need to hardcode subtoolchain for dependencies. `binutils` is available automatically when using `GCC` as toolchain (only required as build dependency when using `GCCcore` toolchain):

```suggestion
    ('CMake', '3.22.1'),

File: easybuild/easyconfigs/a/avrocpp/avrocpp-1.11.1-GCC-11.2.0.eb [hunk 3]
```diff
@@ -0,0 +1,38 @@
...
+
+generator = "Unix Makefiles"
+separate_build_dir = True
+
+configopts = ' -DCMAKE_INSTALL_PREFIX=%(installdir)s/'
+configopts += ' -DCMAKE_BUILD_TYPE=Release '
+
+sanity_check_paths = {
+    'files': ["%(installdir)s/bin/avrogencpp"],
+    'dirs': ["%(installdir)s/lib"],
```
boegel, line 33: Paths in `sanity_check_paths` should be relative to installation prefix:

```suggestion
    'files': ['bin/avrogencpp'],
    'dirs': ['lib'],

File: easybuild/easyconfigs/a/avrocpp/avrocpp-1.11.1-GCC-11.2.0.eb [hunk 4]
```diff
@@ -0,0 +1,38 @@
...
+}
+
+sanity_check_commands = ["%(installdir)s/bin/avrogencpp -h"]
```
boegel, line 36: Don't use full path, just command name (should work since EasyBuild will automatically add `%(installdir)s/bin` to `$PATH` in module file):

```suggestion
sanity_check_commands = ["avrogencpp -h"]

File: easybuild/easyconfigs/a/avrocpp/avrocpp-1.11.1-GCC-11.2.0.eb [hunk 5]
```diff
@@ -0,0 +1,38 @@
...
+
+builddependencies = [
+    ('binutils', '2.37', '', ('GCCcore', '11.2.0')),
+    ('CMake', '3.22.1', '', ('GCCcore', '11.2.0')),
+]
+dependencies = [
+    ('Boost', '1.77.0'),
+]
+
+srcdir = "%(builddir)s/avro-release-%(version)s/lang/c++"
```
boegel, line 23: this path can be relative to unpacked sources:

```suggestion
srcdir = 'lang/c++'

**Review by boegel — DISMISSED** [review]
Thanks a lot for your contribution @ll4strw , and welcome to the wonderful world of EasyBuild!

Some additional suggestions for changes, do let us know if you have any questions...

File: easybuild/easyconfigs/a/avrocpp/avrocpp-1.11.1-GCC-11.2.0.eb [hunk 6]
```diff
@@ -0,0 +1,38 @@
+easyblock = 'CMakeMake'
+
+name = 'avrocpp'
```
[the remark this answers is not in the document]
casparvl: Maybe `avro-cpp`? We also have things like `yaml-cpp` (a YAML parser and emitter in C++) and `netcdf4-python` (python interface to netCDF). Imho, `avro-cpp` would be the most consistent with that. @boegel what do you think?

casparvl (changes requested):

casparvl:

casparvl:

casparvl:
Test report by @casparvl
**SUCCESS**
tcn216.local.snellius.surf.nl - Linux Rocky Linux 8.7, x86_64, AMD EPYC 7H12 64-Core Processor, Python 3.6.8
See https://gist.github.com/casparvl/73c3cdf80df636745e20bdaa02ec71a0 for a full test report.

casparvl (approved):
Looks good to me. Let's wait for the test on Generoso to finish, and if succesful, this PR can be merged imho

casparvl:
Going in, thanks @ll4strw!

boegel:
EB_ARGS="--sanity-check-only"

PR #17542, opened 2023-03-16
Adding easyconfig of XZ/5.4.1 for GCC/12.2.0 

absrocks:


casparvl:
Any particular reason for adding another `XZ` version in `GCCcore-12.2.0`? I'm afraid it might cause confusion: as a policy, whenever we need to use a package (A) as a dependency for another package (B), we always use the one version (of A) per toolchain. The only exception to this is if B simply _does not work_ with A, or if there is a substantial performance benefit.

As `XZ` is quite low level, I'm afraid that adding another one might trigger people to use this as a dependency in `GCCcore-12.2.0`, even though `XZ-5.2.7` is already used in that toolchain.

absrocks:
Hi, I installed amk on the HPC cluster which had a dependency on XZ latest version. But I can reinstall it based on the XZ old version. I will close the PR then.

PR #16385, opened 2022-10-11
{devel}[foss/2021b] PyTorch v1.11.0 w/ Python 3.9.6 + CUDA 11.4.1

sassy-crick:
(created using `eb --new-pr`)

PR #17580, opened 2023-03-20
Horovod 0.26.1 for PyTorch 1.11.0

absrocks:


easybuild/easyconfigs/h/Horovod/Horovod-0.26.1-foss-2021b-CUDA-11.4.1-PyTorch-1.11.0.eb
```diff
@@ -0,0 +1,48 @@
...
+toolchain = {'name': 'foss', 'version': '2021b'}
+
+builddependencies = [
+    ('CMake', '3.22.1'),
+    ('flatbuffers', '2.0.0'),
+]
+dependencies = [
+    ('Python', '3.9.6'),
+    ('PyYAML', '5.4.1'),
+    ('CUDA', '11.4.1', '', True),
```
casparvl, line 21: Please use `SYSTEM` instead of `True` (that's what is tripping up the CI right now)

easybuild/easyconfigs/p/PyTorch/PyTorch-1.11.0-foss-2021b-CUDA-11.4.1.eb
```diff
@@ -0,0 +1,126 @@
...
+
+osdependencies = [OS_PKG_IBVERBS_DEV]
+
+builddependencies = [
+    ('CMake', '3.22.1'),
+    ('hypothesis', '6.14.6'),
+]
+
+dependencies = [
+    ('CUDA', '11.4.1', '', True),
```
casparvl, line 66: Please use `SYSTEM` instead of `True` (that's what is tripping up the CI right now)

easybuild/easyconfigs/p/PyTorch/PyTorch-1.11.0-foss-2021b-CUDA-11.4.1.eb
```diff
@@ -0,0 +1,126 @@
...
+    ('protobuf-python', '3.17.3'),
+    ('pybind11', '2.7.1'),
+    ('SciPy-bundle', '2021.10'),
+    ('typing-extensions', '4.3.0'),
+    ('PyYAML', '5.4.1'),
+    ('MPFR', '4.1.0'),
+    ('GMP', '6.2.1'),
+    ('numactl', '2.0.14'),
+    ('FFmpeg', '4.3.2'),
+    ('Pillow', '8.2.0'), # 8.3.2
```
casparvl, line 79: Note that Pillow 8.3.2 is used in this toolchain as dependency. I see you put that in a comment, was there a particular reason not to use it here?

easybuild/easyconfigs/p/PyTorch/PyTorch-1.11.0-foss-2021b-CUDA-11.4.1.eb
```diff
@@ -0,0 +1,126 @@
...
+
+dependencies = [
+    ('CUDA', '11.4.1', '', True),
+    ('Ninja', '1.10.2'),  # Required for JIT compilation of C++ extensions
+    ('Python', '3.9.6'),
+    ('protobuf', '3.17.3'),
+    ('protobuf-python', '3.17.3'),
+    ('pybind11', '2.7.1'),
+    ('SciPy-bundle', '2021.10'),
+    ('typing-extensions', '4.3.0'),
```
casparvl, line 73: There is no EasyConfig for this in the `easybuild-easyconfigs` repository. I'm assuming you created one locally. Can you include it in this PR?

casparvl (changes requested):

easybuild/easyconfigs/p/PyTorch/PyTorch-1.11.0-foss-2021b-CUDA-11.4.1.eb
absrocks: I initially tried with Pillow 8.3.2, but it has a compatibility error. The configuration requires Pillow lower version. Hence I used Pillow 8.2.0 as the same version was used in PyTorch-1.11.0-foss-2021a-CUDA-11.3.1.eb

casparvl: Ok, in that case two things will need to be done:

1. A clear comment should be put in the EasyConfig as to why the version is different. Typically a one-sentence mention of the error you'd get if you use newer pillow, plus a reference link to an issue describing the error more extensively (if needed). Feel free to just open an issue for that on `easybuild-easyconfigs`, describe the error with a full trace of the output/error you get, and close your own issue with a remark that the solution is to downgrade to Pillow 8.2.0. 

2. An exception needs to be made for the CI to accept that a different version of Pillow will be used here than in other parts of the toolchain. You can do that by adding `Pillow` to the dictionary [here](https://github.com/easybuilders/easybuild-easyconfigs/blob/fdcaa750ca492ff2bb5cc122b51806767b9a4128/test/easyconfigs/easyconfigs.py#L488). The key should be `Pillow` and the value should be a list of tuples, where the first element is the version of `pillow` that should be accepted (i.e. `'8.2.0'` in this case) and the second should be a list of the packages that are allowed to use this dependency version of `Pillow`, i.e. in your case `r'PyTorch-1\.11\.0-`. These are regular expressions, so that's why the `.` should be escaped there (there are plenty of examples in that file).

By the way, I see the CI is also tripping over the `typing-extensions` version, for the same reason (another version of `typing-extensions` is already in use in this toolchain). The same two points as above will apply to that. I.e. justify & document why we deviate from the version already use in the toolchain (1) and add it as an exception to CI (2).

casparvl:
Wait a second... this is the same PyTorch as in https://github.com/easybuilders/easybuild-easyconfigs/pull/16385/files, right?

Because there, it _does_ use a newer Pillow. 

Then, there is also https://github.com/easybuilders/easybuild-easyconfigs/pull/17272

As PyTorch is rather difficult to get merged (lots of issues pop up, with varying systems), I'd like to focus on getting either of those two merged, and simply build this Horovod on top of that...

PR #17600, opened 2023-03-24
Added easyconfig for Catch2, GCC toolchain

ll4strw:
Existing Catch2 easyconfig is based on SYSTEM toolchain, but lists GCC in build dependencies. This one is on the top of GCC toolchain.


smoors:
@ll4strw i don't see why you need this?
as mentioned in the description, this is a header-only library, so you don't need GCC at all as a runtime dependency.
modules in the SYSTEM toolchain can always be loaded together with GCC toolchains, so there is no conflict.

smoors:
@ll4strw `Catch2-2.13.4.eb` is already in the easyconfigs repo, so shouldn't you just close this PR?

smoors:
> That is right, but it looks like some of the build dependencies in original eb aren't needed after all.

it may not be needed for your system, but it may be needed for other clusters.

- binutils should always be added when using GCCcore, needed for old systems
- GCCcore its should according to the comment be added for hierarchical module naming scheme

smoors:
> I see. Then this PR could be closed unless you think it is a good idea to upgrade the existing easyconfig with a new one based on `GCCcore v11.2.0`. Thanks.

that's fine, but then i would use the latest version `GCCcore-12.2.0` instead.
`GCCcore-10.2.0` is not that old though, so it's not urgent.

boegel:
no longer relevant, see:

* #19270

PR #17603, opened 2023-03-27
fix sanity check for libxml2 shared library

DirectorSloan:
- builtin sanity check searches for libxml2.a
- on our CentOS7 system no *.a  libraries were created


branfosj:
Which version of EB are you running? This should not be needed with the latest easyblock for libxml2 - https://github.com/easybuilders/easybuild-easyblocks/pull/2825

smoors:
i can confirm that the following libs are created with the dev version of EB on centos7:
libxml2.a  libxml2.la  libxml2.so  libxml2.so.2  libxml2.so.2.10.3

SebastianAchilles:
Enabling the build of static libraries for libxml2 >= 2.10 was added in https://github.com/easybuilders/easybuild-easyblocks/pull/2825 which is part of the EasyBuild v4.7.0 release.

This test report shows that the build of `libxml2-2.10.3-GCCcore-12.2.0.eb` works with EB v4.7.2 on CentOS 7.9: https://gist.github.com/SebastianAchilles/fea95a24400bce3ce110e1608d8e9924

Issue #17701, opened 2023-04-12
libcerf-Fortran


hattom:
As discussed today (slack & conf. call), the fortran bindings (included in a subfolder of the libcerf repository[1], but not built) of libcerf can be a new easyconfig -- which allows us to keep libcerf itself in GCCcore.
Due to the fortran .mod files, the fortran bindings need to go in iccifort / GCC.

There's only a single .F90 file that needs compiling (generates: 1 .o file, 1 .mod file) -- (and a second file which can be used to test).

@SimonPinches already has a non-upstreamed libcerf-Fortran easyconfig. I will ask him if I can fix his file, tidy, and then upstream.

[1] https://jugit.fz-juelich.de/mlz/libcerf/-/tree/main/ | https://jugit.fz-juelich.de/mlz/libcerf/-/tree/main/fortran/ccerflib_f95_interface

PR #17702, opened 2023-04-12
WIP: {math} libcerf-Fortran-2.3: GCC/12.2.0, intel-compilers/2022.2.1

hattom:
Pretty sure this is still a bit of a mess. Requires further testing.
Not sure what the best naming scheme is for the .a/.so files is (the documentation refers only to building `.mod` and `.o` files, but this seems a bit little).

Was contemplating adding a Makefile with a patch, not sure if that's better than just putting the $FC commands in like this.

See also: https://github.com/easybuilders/easybuild-easyconfigs/issues/17701

### DIFF
--- a/easybuild/easyconfigs/l/libcerf-Fortran/libcerf-Fortran-2.3-GCC-12.2.0.eb
+++ b/easybuild/easyconfigs/l/libcerf-Fortran/libcerf-Fortran-2.3-GCC-12.2.0.eb
@@ -1,0 +1,35 @@
+easyblock = 'ConfigureMake'
+
+name = 'libcerf-Fortran'
+version = '2.3'
+
+homepage = 'https://jugit.fz-juelich.de/mlz/libcerf'
+description = """
+ libcerf is a self-contained numeric library that provides an efficient and
+ accurate implementation of complex error functions, along with Dawson,
+ Faddeeva, and Voigt functions.
+"""
+
+toolchain = {'name': 'GCC', 'version': '12.2.0'}
+toolchainopts = {'pic': True}
+
+#patches = ["libcerf-Fortran-Makefile.patch"]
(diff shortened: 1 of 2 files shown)


sassy-crick:
@hattom You might want to sync your PR with development, something like this: 
```
```
Also, it appears we got `libcerf-2.3-GCCcore-12.2.0.eb` already merged in `EasyBuild-5.1.0`. Maybe remove that version from your PR?

hattom:
@sassy-crick -- not quite.
`libcerf-2.3` is the C library
`libcerf-Fortran-2.3` are the fortran interfaces

Btw -- The reason why it makes sense to have a separate package rather than to add the fortran bindings into the libcerf library is that the libcerf library is GCCcore, the fortran bindings would have to be GCC or intel-compilers. We discussed this quite some time ago either in Slack or at a conf call, probably around the time I made the PR.

sassy-crick:
> @sassy-crick -- not quite. `libcerf-2.3` is the C library `libcerf-Fortran-2.3` are the fortran interfaces
> 
> Btw -- The reason why it makes sense to have a separate package rather than to add the fortran bindings into the libcerf library is that the libcerf library is GCCcore, the fortran bindings would have to be GCC or intel-compilers. We discussed this quite some time ago either in Slack or at a conf call, probably around the time I made the PR.

@hattom My wrong, apologies, I did not read properly. 

What still needs to be done is the `eb --sync-pr 17702` to get the checks started. :-)

sassy-crick:
Your line seem to be too long:
```
libcerf-Fortran-2.4-GCC-13.2.0.eb:30:121: E501 line too long (140 > 120 characters)
```

hattom:
Thanks. `--try-toolchain --copy-ec` changed the formatting and created a super-line.

boegel:
@hattom Sorry for not picking up on this...

Seems like this should be closed in favor of:
- #22162

If you agree, please close this PR

PR #17855, opened 2023-05-04
{bio}[gompi/2021b] KofamScan 1.3.0

wdlingit:
Two eb files, one for program and the other for database. This is mimicking eb files for InterProScan.


easybuild/easyconfigs/k/KofamScan_data/KofamScan_data-20230401-gompi-2021b.eb
```diff
@@ -0,0 +1,50 @@
...
+    'ko_list.gz',
+    'profiles.tar.gz',
+]
+checksums = [
+    '099a62576b1754f31d3182e54d8ea310b61a2cacd3f6b3d5762be51ab0c711dd',  # ko_list.gz
+    'e017ead8fd52677d0a20558b9fd504876558e1723e77ec6be79aa72e6dee90a2',  # profiles.tar.gz
+]
+
+dependencies = [
+    ('KofamScan',  '1.3.0'),  # hard coded latest program version
```
Micket, line 32: I would have expected the dependency to be the other way around? Maybe not?

easybuild/easyconfigs/k/KofamScan_data/KofamScan_data-20230401-gompi-2021b.eb
```diff
@@ -0,0 +1,50 @@
...
+
+toolchain = {'name': 'gompi', 'version': '2021b'}
+
+source_urls = [
+    'https://www.genome.jp/ftp/db/kofam/',  # add archives/YYYY-MM-DD/ if moved
+]
+sources = [
+    'ko_list.gz',
+    'profiles.tar.gz',
+]
```
Micket, line 25: This will be problematic; the download source file doesn't contain any information on what version this is.
In fact, I can't tell if this is a 20230401 version from anything here, in fact, the time stamps on the ftp states 24-Apr-2023 and 26-Apr-2023, so it's not clear what this is supposed to be.

We'll have to do a few things, use `filename` and `download_filename` to indicate that these differ.
Something along the lines of
```suggestion
source_urls = [
    'https://www.genome.jp/ftp/db/kofam/%(version)s',  # prefer the archive url present after new releases
    'https://www.genome.jp/ftp/db/kofam/',
]
sources = [
    {'download_filename': 'ko_list.gz', 'filename': 'ko_list_%(version)s.gz'}
    {'download_filename': 'profiles.tar.gz', 'filename': 'profiles_%(version)s.tar.gz'}
]
```
might be sufficient to protect reproducibility here, but I had to guess what the version string will be in the future. 

Maybe this means we would prefer the 2023-04-26 version as the version string. 

(there would of course be several modifications that follow from this renaming)

easybuild/easyconfigs/k/KofamScan/KofamScan-1.3.0-gompi-2021b.eb
```diff
@@ -0,0 +1,59 @@
...
+    ('Perl',     '5.34.0'),
+    ('parallel', '20210722'),
+    ('HMMER',    '3.3.2'),
+]
+
+extract_sources = False
+
+# NOTE The data can be added manually afterwards with for example a symlink called *data*
+# in %(installdir)s/
+# But simpler to use the InterProScan_data-*.eb file. The regular InterProScan data package
```
Micket, line 40: confusing to mention interproscan here

easybuild/easyconfigs/k/KofamScan_data/KofamScan_data-20230401-gompi-2021b.eb
```diff
@@ -0,0 +1,50 @@
...
+
+extract_sources = False
+
+install_cmd = 'mkdir -p %(installdir)s/db/ && '
+install_cmd += 'tar xfz profiles.tar.gz --directory %(installdir)s/db/ && '
+install_cmd += 'gzip -dc ko_list.gz > %(installdir)s/db/ko_list && '
+install_cmd += 'rm -f profiles.tar.gz && '
+install_cmd += 'rm -f ko_list.gz && '
+install_cmd += 'rm -rf $EBROOTKOFAMSCAN/db && '
+install_cmd += 'ln -s %(installdir)s/db $EBROOTKOFAMSCAN/db'
```
Micket, line 43: oh, hm, so this modifies another installation? 

This seems backwards to me. Rather than symlinking stuff in here, of KofamScan depended on KofamScan_data, it could just write it's own configuration file to look for the files in this module instead, removing this need for a software to modify another installation (which will break in several setups, EB can write protect installation directories to ensure one doesn't do things like this)

Micket (changes requested):

easybuild/easyconfigs/k/KofamScan/KofamScan-1.3.0-gompi-2021b.eb
wdlingit: Thank you. Will fix this in the next commit.

easybuild/easyconfigs/k/KofamScan_data/KofamScan_data-20230401-gompi-2021b.eb
wdlingit: It seems to me that KofamScan is updating the database files monthly and rather not associate database versions with program versions. There are tens of versions of database files but only four program versions. I am imaging that people may download KofamScan_data more frequently than install KofamScan module. So it is better to specify the most updated KofamScan when doing KofamScan_data installation.

easybuild/easyconfigs/k/KofamScan_data/KofamScan_data-20230401-gompi-2021b.eb
wdlingit: Thank you. I did know that we can do this filename versioning locally. Will take the codes in the next commit. I will also take "2023-04-26" as the version string, but not sure how will KofamScan name the current database version when they move files into the archive.

easybuild/easyconfigs/k/KofamScan_data/KofamScan_data-20230401-gompi-2021b.eb
wdlingit: This is mimicking InterProScan_data, and I think doing this (symlink to program install dir) for KofamScan is rather reasonable than InterProScan. It seems to me that database files and program files are bundled together for InterProScan so we used to see one set of database files and program files under a directory of one version. For KofamScan, database files are monthly updated so it is possible to install a new version of database for the same program.

I am not very sure what to do to fit easybuild's standard. Do you think let KofamScan (the program module) depending on KofamScan_data and symlink a db dir (under KofamScan and by KofamScan.eb) to the db dir of KofamScan_data is acceptable?

easybuild/easyconfigs/k/KofamScan/KofamScan-1.3.0-gompi-2021b.eb
```diff
@@ -0,0 +1,57 @@
...
+high score assignments are highlighted with asterisks '*' in the output.  
+The K number assignments facilitate the interpretation of the annotation 
+results by linking the user's sequence data to the KEGG pathways and EC 
+numbers.  [code only: libraries and external binaries but no data]."""
+# software_license = 'MIT'
+
+toolchain = {'name': 'gompi', 'version': '2021b'}
+
+source_urls = [
+    'https://www.genome.jp/ftp/tools/kofam_scan/',  # add archives/kofam_scan-%(version)s/ if moved
```
Micket, line 20:
```suggestion
    'https://www.genome.jp/ftp/tools/kofam_scan/',
    'https://www.genome.jp/ftp/tools/kofam_scan/archives/kofam_scan-%(version)s/',
```

should be possible to preempt this, EB will try to next url if it fails.

Micket (changes requested):

easybuild/easyconfigs/k/KofamScan/KofamScan-1.3.0-gompi-2021b.eb
wdlingit: Thank you!

easybuild/easyconfigs/k/KofamScan_data/KofamScan_data-20230401-gompi-2021b.eb
wdlingit: Thank you for the comment! So how about let's just delete KofamScan_data and add some comment and modloadmsg to ask people who installed KofamScan.eb to do some necessary modification?

Micket: I would remove
```
install_cmd += 'echo "profile: %(installdir)s/db/profiles" > %(installdir)s/config.yml && '
install_cmd += 'echo "ko_list: %(installdir)s/db/ko_list" >> %(installdir)s/config.yml && '
```
from the software installation, forcing users to specify this themselves explicitly. 

and either do one of:
1. Don't provide the database at all (users just get the version they want for their research into their project area like many other things)
2. Provide Kofam-Data at SYSTEM level, add a modloadmsg to tell users how to use it; something along the lines of `"use specify kofam_scan --profile $EBROOTKOFAM_DATA/profile_stuff --ko-list $EBROOTKOFAM_DATA/ko-stuff ..."`

Since the database seems to be completely independent, and I suspect we don't want to keep up with PRs for this dataset every month, I'm leaning towards option 1, and leave that to the users themselves. 

I don't think we need any modloadmsg in this central repo, since there hasn't been any release yet with KofamScan included, we aren't breaking anything backwards compatibility. If you already provide this at your center and like to move it closer to upstream behavior, you could consider adding it locally.

wdlingit: I did corresponding changes, removed KofamScan_data from this PR, and added some comments in KofamScan EB file. I kept modloadmsg but with some minor modification to remind users the command to enter and the example config file.

boegel:
Closing this since `gompi/2021b` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @wdlingit 

PR #17857, opened 2023-05-04
{math}[foss/2022a] rustworkx v0.12.1 w/ Python 3.10.4

Micket:
(created using `eb --new-pr`)

- [x] new cargo easyblock https://github.com/easybuilders/easybuild-easyblocks/pull/2902
- [x] https://github.com/easybuilders/easybuild-easyblocks/pull/2934

PR #17858, opened 2023-05-04
{lib}[foss/2022a] SnapATAC2 v2.3.0 w/ Python 3.10.4

Micket:
(created using `eb --new-pr`)
- [x] depends on #17857

Didn't compile, fails due to broken code and the rust compiler refusing since it's unsafe. But, shows how the new easyblock might be possible to use for messier projects.


Micket:
This should be revisited in a newer toolchain with a more up to date Rust compiler

PR #17867, opened 2023-05-05
FLINT needs BLAS so move it up from GCC level

akesandgren:
(created using `eb --new-pr`)
FLINT depends on BLAS so it must be higher than GCC. It also depends on NTL. Due to moving FLINT up we also need to move Arb and polymake in the same way.


Micket (approved):
lgtm 
no choice but to make this change to make things right i guess

akesandgren:
This one is a bit hard to test while FLINT@GCC is still in develop since that's what it picks up as deps for the Arb/polymake ec's
I'll split them into two separate ones instead.

akesandgren:
This one is a bit hard to test while FLINT@GCC is still in develop since that's what it picks up as deps for the Arb/polymake ec's
I'll split them into two separate ones instead.

PR #17958, opened 2023-05-22
Two bazel versions

DirectorSloan:
HTH


boegel:

easybuild/easyconfigs/b/Bazel/Bazel-5.3.0-GCCcore-12.2.0.eb
```diff
@@ -0,0 +1,26 @@
...
+version = '5.3.0'
+
+homepage = 'https://bazel.io/'
+description = """Bazel is a build tool that builds code quickly and reliably.
+It is used to build the majority of Google's software."""
+
+toolchain = {'name': 'GCCcore', 'version': '12.2.0'}
+
+source_urls = ['https://github.com/bazelbuild/%(namelower)s/releases/download/%(version)s']
+sources = ['%(namelower)s-%(version)s-dist.zip']
```
jfgrimm, line 11: the 5.1.1 easyconfig still includes some patches:
```
patches = [
    'Bazel-3.7.1_fix-protobuf-env.patch',
    'Bazel-3.7.2_fix-GCC-11.patch',
]
```

can you confirm whether those are no longer required?

jfgrimm: patches apply cleanly to 5.3.0, so should be kept there at least. For 6.2.0, `Bazel-3.7.1_fix-protobuf-env.patch` can't determine patch level, but `Bazel-3.7.2_fix-GCC-11.patch` succeeds and should be kept.

jfgrimm (changes requested):
see comments r.e. patches

easybuild/easyconfigs/b/Bazel/Bazel-5.3.0-GCCcore-12.2.0.eb
DirectorSloan: > the 5.1.1 easyconfig still includes some patches:
> 
> ```
> patches = [
>     'Bazel-3.7.1_fix-protobuf-env.patch',
>     'Bazel-3.7.2_fix-GCC-11.patch',
> ]
> ```
> 
> can you confirm whether those are no longer required?

Yes, the patches aren't applyable to both versions. Hunk errors etc.

Issue #17455, opened 2023-03-02
Julia: user depot not accesible with Pkg.add


lexming:
The Julia easyconfigs and JuliaPackages follow the usual approach of prepending installation paths to the environment. In this case, new modules prepend their installation path to `JULIA_DEPOT_PATH`.

However, this breaks the capability of end-users to install new packages with `Pkg.add()` in Julia. This tool only uses the top path in `DEPOT_PATH` as destination of new installations. This seems to be by design as described in the [Pkg documentation](https://pkgdocs.julialang.org/v1/glossary/)
> The depot path is controlled by the Julia DEPOT_PATH global variable which is populated at startup based on the value of the JULIA_DEPOT_PATH environment variable. The first entry is the “user depot” and should be writable by and owned by the current user. The user depot is where: registries are cloned, new package versions are installed, named environments are created and updated, package repos are cloned, newly compiled package image files are saved, log files are written, development packages are checked out by default, and global configuration data is saved. Later entries in the depot path are treated as read-only and are appropriate for registries, packages, etc. installed and managed by system administrators.

So we would have to append paths to `JULIA_DEPOT_PATH` to not break `Pkg`. The issue is that this approach allows user installs to overwrite what is loaded with modules from EasyBuild (not desirable).

An alternative approach might be to use [JULIA_LOAD_PATH](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_LOAD_PATH) in our easyconfigs to control locations with available installations and only set `JULIA_DEPOT_PATH` to the user's depot.

lexming:
So I tested several configurations for this issue and there is no setup that allows users to easily install their own Julia packages on top of our installations.

With https://github.com/easybuilders/easybuild-easyblocks/pull/2935 we get back the user depot `~/.julia` as the top entry in `JULIA_DEPOT_PATH`. This is an improvement over the current situation, but users will not be able to add their own packages yet.

Regardless of the contents of `JULIA_DEPOT_PATH`, the active environment is defined by its associated `Project.toml`. In the case of our JuliaPackage modules, they have their own environment defined in `$EBROOTXXX/environments/v1.x/Project.toml`. For instance this is the environment after loading `IJulia`:

```bash
$ julia -E 'Base.DEPOT_PATH'
["~/.julia",
"~/easybuild/install/skylake/software/Julia/1.8.5-linux-x86_64/share/julia",
[...]

PR #17976, opened 2023-05-25
set JULIA_DEPOT_PATH in Julia easyconfigs with default value

### LINKED ISSUE #17961: Remove user directory from JULIA_DEPOT_PATH during installation
The easyconfig of Julia adds `~/.julia` to `JULIA_DEPOT_PATH` to allow users to add their own stuff on top of the installation. This is picked during installation though, which might lead to issues if the user has custom environments. See https://github.com/easybuilders/easybuild-easyconfigs/pull/16720#issuecomment-1558838707

Solution: strip down  `JULIA_DEPOT_PATH` during installation.

Note: this still applies after the switch to `JULIA_LOAD_PATH` proposed in https://github.com/easybuilders/easybuild-easyconfigs/issues/17455

lexming:
(created using `eb --new-pr`)

As reported in #17455, user depot should be the top entry in `JULIA_DEPOT_PATH`. This PR fixes that issue by simply setting `JULIA_DEPOT_PATH` to `:`, which is the default value with user's depot as top entry.

Having `JULIA_DEPOT_PATH` defined in the base Julia easyconfigs is needed for the changes in https://github.com/easybuilders/easybuild-easyblocks/pull/2935 to fix #17961


lexming:

lexming:

exaexa:
Hi all,
I just created https://github.com/easybuilders/easybuild-easyconfigs/pull/17995 which might be similar or straight out duplicated of this (didn't check the list properly, sorry).

One point to discuss: do we _really_ want to point to the "shared" julia depots? Since 1.9.0 Julia will actually recompile much of the base packages anyway (depending on dependency updates etc), so I'd say that any attempts to push a package directory to it might be completely vain. The other concern is that the `..../share/julia` directory that EB now points the depots to **is not a proper julia depot** and can't really be used for anything (at least on all Julia versions I tried).

And one note: I'm not sure if ~ in the configs here works as intended (especially for the history dir -- see the other PR).

Thanks!

lexming:
@exaexa I checked your PR #17995 and as I commented there, entirely unsetting `JULIA_DEPOT_PATH` is not an option because then any other `JuliaPackage` loaded on top of Julia will reset the `DEPOT_PATH` and disable the user's depot.

I agree with you that using `~` is dangerous, so I simplified my approach by just setting `JULIA_DEPOT_PATH` to `:`. This is equivalent to use the default `DEPOT_PATH` and is compatible with loading `JuliaPackages` distributed with EasyBuild.

Let me know if this approach works for you.

lexming:

lexming:

lexming:

branfosj (approved):

branfosj:
Test report by @branfosj
**SUCCESS**
Build succeeded for 12 out of 12 (12 easyconfigs in total)
bear-pg0105u03a.bear.cluster - Linux RHEL 8.6, x86_64, Intel(R) Xeon(R) Platinum 8360Y CPU @ 2.40GHz (icelake), Python 3.6.8
See https://gist.github.com/branfosj/5ce7226cec7a9240b112197fbc33dd99 for a full test report.

branfosj:
Going in, thanks @lexming!

PR #17804, opened 2023-04-26
{lang}[GCCcore/12.2.0] Cython v0.29.34 w/ Python 3.10.8

SimonPinches:
(created using `eb --new-pr`)

PR #17987, opened 2023-05-28
cython

karoshi-0815:


Micket:
Thank you for your contribution, however, lack of Cython iasn't an oversight, but intentional, as it's part of the bundle already;

https://github.com/easybuilders/easybuild-easyconfigs/pull/17804#issuecomment-1523755949

> As is tradition when I see a cython PR:
> We already have 0.29.32 as part of Python bundle, is this updated version required by something as a build dep? I suspect this is a case where this was accidentally added since we used to have some older cython easyconfigs.

PR #17995, opened 2023-05-31
[Julia-1.8.5] Prevent setting environment that breaks Julia

exaexa:
Julia breaks if the depot path is set to an unwritable location. At the same time, setting ~ into environment variables typically forces a creation of actual directory named `~`, creating mild risk of fun `rm -fr` stories and impacting data of novice usrers.

Removing all environment fixes the problem and is generally safe (Julia will properly use a well-defaulted depot path).

cc @laurentheirendt @ekieffer (Thanks @ULHPC team for applying and quicktesting this fix).

### DIFF
--- a/easybuild/easyconfigs/j/Julia/Julia-1.8.5-linux-x86_64.eb
+++ b/easybuild/easyconfigs/j/Julia/Julia-1.8.5-linux-x86_64.eb
@@ -35,12 +35,4 @@ _depot_paths = ':'.join([
     '~/.julia',
 ])
 
-modextravars = {
-    'JULIA_DEPOT_PATH': _depot_paths,
-    # 'set JULIA_HISTORY to user's DEPOT_PATH (~/.julia)
-    # by default it will point to the DEPOT_PATH (install dir) of the last module loaded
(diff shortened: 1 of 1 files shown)


exaexa:
Possibly fixes/duplicates also:
- https://github.com/easybuilders/easybuild-easyconfigs/issues/17455
- https://github.com/easybuilders/easybuild-easyconfigs/pull/17976

lexming (commented):
This change breaks Julia packages distributed through EasyBuild like [IJulia](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/i/IJulia). Loading a Julia package through a module file on top of this easyconfig of Julia will disable the user's depot.

I agree with you in the dangers of using `~`. I addressed that issue in the recent changes of #17976. I suggest to close this PR and maybe we can continue on that one, since that covers many more Julia easyconfigs.

exaexa:
@lexming Hi, thanks, I'm OK with closing this (actually let's do that right away). If I get it right, the `:` depot path from the other PR serves as a default and is still sufficiently safe in a way the IJulia recipe can append its depot to it?

lexming:
@exaexa that's exactly it

PR #18009, opened 2023-06-02
also run easyconfigs test suite with Python 3.11

boegel:
WIP because `sdist` tests fails with Python 3.11, because the `CVS` easyconfigs are not being included (cfr. #10326)

requires:
* https://github.com/easybuilders/easybuild-easyconfigs/pull/18985


Micket:
Looks like disabling pruning is no longer allowed, CVS files are excluded despite 
```
[sdist]
# Don't remove CVS and RCS directories
prune = 0
```
sheesh python..

branfosj:
This is because, for `setuptools >= 60`, we are getting `setuptools.setup` instead of `distutils.core.setup`. This is part of the deprecation of `distutils` and moving people over to `setuptools`. From what I can tell, `setuptools.setup` does not have the `prune` option.

Starting point:
```
$ pip list
Package    Version
---------- -------
pip        23.0.1
setuptools 66.1.1
```

No CVS in dist:
* `python3 setup.py sdist`
* `pip install setuptools==60.0.0 && python3 setup.py sdist`

CVS in dist:
* `SETUPTOOLS_USE_DISTUTILS=stdlib python3 setup.py sdist`
* `pip install setuptools==59.1.0 && python3 setup.py sdist`
* `pip uninstall setuptools && python3 setup.py sdist`

branfosj:
Also, from the setuptools [release notes](https://setuptools.pypa.io/en/latest/history.html):
* `v59.7.0`: Require Python 3.7
* `v60.0.0`: Setuptools once again makes its local copy of distutils the default. To override, set SETUPTOOLS_USE_DISTUTILS=stdlib.

branfosj:
From https://packaging.python.org/en/latest/guides/using-manifest-in/
> After processing the `MANIFEST.in` file, setuptools removes the `build/` directory as well as any directories named `RCS`, `CVS`, or `.svn` from the sdist, and it adds a `PKG-INFO` file and an `*.egg-info` directory. This behavior cannot be changed with `MANIFEST.in`.

I also could not find any indication in that documentation on how to stop it happening.

boegel:
@Micket Let's merge?

Micket (approved):
lgtm

PR #17737, opened 2023-04-15
make excluded tests for PyTorch 1.12.1 consistent + also exclude flaky test_optim rather than adding patch to try and fix it

branfosj:
Based on the failures in  #17732 and #17733 I think we should revert this. It looks to be a case of fixing one flaky test and making other tests less reliable. :(

We'll also add `test_optim` to excluded tests, which it has been for 1.9.0 to 1.12.0.

I've also made sure all 1.12.1 easyconfigs skip the tests metnioned in #17615.

PR #18062, opened 2023-06-08
Remove unused PyTorch patch

Flamefire:
The usage of this patch was removed in #17737 so this patch is no longer required to be kept.


casparvl (approved):
Makes sense, it indeed no longer seems to be used. Lgtm!

PR #18098, opened 2023-06-15
allow multiply dependency variants of ncbi-vdb for SRA-Toolkit per easyconfigs generation, as long as versions match

smoretti:
Test fix to allow several (sync) ncbi-vdb/SRA-Toolkit in the same toolchain


boegel (approved):
lgtm

PR #18145, opened 2023-06-19
remove old UCC (1.1.0) from GCCcore 12.3.0

Micket:
(created using `eb --new-pr`)
I probably should have git mv'ed UCC-1.1.0-GCCcore-12.3.0.eb into UCC-1.2.0-GCCcore-12.3.0.eb, but i missed that. 

This just cleans up the GCCcore a bit as to not cause any confusion.


boegel (approved):
lgtm

PR #18258, opened 2023-07-04
Create Serf-1.3.9_sslbuckets.patch

levlafayette:
Without this patch, Subversion will generate an error (libserf-1.so: undefined reference to `ERR_GET_FUNC') from the Serf dependency.


easybuild/easyconfigs/s/Serf/Serf-1.3.9_sslbuckets.patch
```diff
@@ -0,0 +1,13 @@
+--- serf-1.3.9/buckets/ssl_buckets.c.errgetfunc
```
jfgrimm, line 1: please add an author and description to the patch

boegel:
This was intended for:
* #18259

Since that PR was closed, I'm also closing this one...

PR #18267, opened 2023-07-05
{math}[system/system] MATLAB v2023a

sassy-crick:
(created using `eb --new-pr`)

[Test report](https://gist.github.com/sassy-crick/5630af99972b68a862e3f7e1f3eb4335)


easybuild/easyconfigs/m/MATLAB/MATLAB-2023a-r3.eb
```diff
@@ -0,0 +1,28 @@
...
+homepage = 'https://www.mathworks.com/products/matlab'
+description = """MATLAB is a high-level language and interactive environment
+ that enables you to perform computationally intensive tasks faster than with
+ traditional programming languages such as C, C++, and Fortran."""
+
+toolchain = SYSTEM
+
+sources = ['R%s_Update_%s_Linux.iso' % (version, _update)]
+checksums = ['e80d78f5f0e08a0da24234569b6472fc0dacac70d77b53e79d831ba6a0b1aa33']
+
```
Micket, line 15:
```suggestion
download_instructions = 'Download %s from mathworks.com' % sources[0]

```

Micket (dismissed):

easybuild/easyconfigs/m/MATLAB/MATLAB-2023a-r3.eb
```diff
@@ -0,0 +1,28 @@
...
+java_options = '-Xmx2048m'
+
+osdependencies = [('p7zip-plugins', 'p7zip-full')]  # for extracting iso-files
+
+# Use EB_MATLAB_KEY environment variable or uncomment and modify license key
+# key =
```
Micket, line 21:
```suggestion
# key = '00000-00000-00000-00000-00000-00000-00000-00000-00000-00000-00000-00000'
```
lets not change this, just more fore the review pr to needlessly mention

arkdavy:
Hi All, 

One of our systems shows a failure on missing read permissions for 
```bash
%(installdir)s/toolbox/local/settings
```
which we could not debug for a long time.

I am not sure at which conditions this folder is created, but its timestamp is at a later moment than the last installation log report. It looks like it inherits the permissions pattern for my easybuild user account and is, most likely, not related to my launches of the Matlab executable. The easybuild fix like that 

```python
postinstallcmds = [
    'mkdir -p %(installdir)s/toolbox/local/settings && '
    'chmod o+rx %(installdir)s/toolbox/local/settings'
]
```
should do the job, but it was not excessively tested.

I am wondering whether this can be included as a comment to easyconfig at some point. 

The launch failure is actually diverse across the versions and the operating systems we use (centOS7 and Rocky9), for example, we had the following reported failures

```bash
The 2021b_Update_2 : "std::exception: The parent object contains a child with the same name already.". 
```

```bash
The 2022b version : "Fatal Startup Error: 
Dynamic exception type: std::runtime_error
std::exception::what: 22".
```

and failure with no message at all

sassy-crick:
> Hi All,
> 
> One of our systems shows a failure on missing read permissions for
> 
> ```shell
> %(installdir)s/toolbox/local/settings
> ```
> 
> which we could not debug for a long time.
> 
> I am not sure at which conditions this folder is created, but its timestamp is at a later moment than the last installation log report. It looks like it inherits the permissions pattern for my easybuild user account and is, most likely, not related to my launches of the Matlab executable. The easybuild fix like that
> 
> ```python
> postinstallcmds = [
>     'mkdir -p %(installdir)s/toolbox/local/settings && '
>     'chmod o+rx %(installdir)s/toolbox/local/settings'
> ]
> ```
> 
> should do the job, but it was not excessively tested.
> 
> I am wondering whether this can be included as a comment to easyconfig at some point.
> 
> The launch failure is actually diverse across the versions and the operating systems we use (centOS7 and Rocky9), for example, we had the following reported failures
> 
> ```shell
> The 2021b_Update_2 : "std::exception: The parent object contains a child with the same name already.". 
> ```
> 
> ```shell
> The 2022b version : "Fatal Startup Error: 
> Dynamic exception type: std::runtime_error
> std::exception::what: 22".
> ```
> 
> and failure with no message at all

I have not observed these issues before on our cluster with any of the `MATLAB` installations we have and they are using more or less the same EasyConfig file with the main difference being the version. As you know `centOS7` is EOL so I will not look into that. We are using `RHEL8` on our cluster. That said, I did see similar things with packed binaries before where some folders got the wrong permissions, simply due to copying them from the `BUILDDIR` to the `INSTALLDIR`. I put that down as specific file system settings as we are using `/dev/shm` for building. 
So for now, I would suggest to leave it as it is but keep an eye on it. 
Feel free to raise an issue if that is persisting with `Rocky9` and sorry for the delay in getting back on this.

akesandgren:
Closing this since since we already have newer MATLAB versions merged.

PR #18357, opened 2023-07-20
adding easyconfigs: TINKER-8.10.5-foss-2022b.eb

michaellass:
The build system used by the TINKER easyblock (running compile.make, library.make and link.make) is currently broken upstream and needs a small patch. It was submitted upstream at https://github.com/TinkerTools/tinker/pull/138.

The tests performed after building TINKER for me took about 15 hours on a 128 core machine. That's maybe not optimal but would have to be dealt with in the TINKER easyblock. There are already two tests disabled due to long runtime: https://github.com/easybuilders/easybuild-easyblocks/blob/fa9468436edba63d681179dd50bed48372a3bc60/easybuild/easyblocks/t/tinker.py#L129


michaellass:
I'll close this as it has been open for nearly a year without progress.

PR #17483, opened 2023-03-08
{lang}[GCCcore/12.2.0] gnuplot v5.4.6, libcerf v2.3, libgd v2.3.3, ...

furstj:
(created using `eb --new-pr`)

PR #17105, opened 2023-01-13
{devel}[GCCcore/12.2.0] GLPK v5.0, GSL v2.7, libgit2 v1.5.0, ...

jfgrimm:
(created using `eb --new-pr`)

PR #18498, opened 2023-08-08
remove GSL/2.7-GCC-12.2.0 as we have GSL/2.7-GCCcore-12.2.0

branfosj:
(created using `eb --new-pr`)

#17105 added `GSL-2.7-GCC-12.2.0.eb` and #17483 added `GSL-2.7-GCCcore-12.2.0.eb`. As GSL is being used as a dependency in `GStreamer-1.22.1-GCCcore-12.2.0.eb` the less complicated change is to remove the `GCC` version.


boegel:
Although the changes in #18501 are a bit more painful, I think that's the right approach to fix this, rather than this one used here...

PR #18596, opened 2023-08-18
Add new maturin for GCCcore 12.3.0 and change existing one to use CargoPythonPackage, since it's build with Rust

casparvl:
(created using `eb --new-pr`)


easybuild/easyconfigs/m/maturin/maturin-1.1.0-GCCcore-12.2.0.eb
```diff
@@ -21,6 +21,8 @@ dependencies = [
 sources = [SOURCE_TAR_GZ]
 checksums = ['4650aeaa8debd004b55aae7afb75248cbd4d61cd7da2dcf4ead8b22b58cecae0']
 
+offline = False
```
Micket, line 24: Specifying the crates should work fine here

casparvl: Ah, I just stole from https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/f/fastparquet/fastparquet-2023.4.0-gfbf-2022b.eb#L27 and assumed that was ok. I don't know right away how to find the names of the crates. I'll look for some example EasyConfigs that do this, and I'll have a look in the build logs - hoping I can find the crate names there?

I think the docs at https://docs.easybuild.io/version-specific/generic-easyblocks/#extra-easyconfig-parameters-specific-to-cargopythonpackage-easyblock could use a more extensive explaination of what 'offline' means, and maybe also the `crates` keyword. From your comment, I _guess_ it is something like this: for most builds, you specify the crates, and if the crates are specified you can (and thus should) build offline (`offline = True`). It sounds like there are exceptions, where the build can really not be done offline (maybe we should list an example scenario?), and in that case you don't specify the crates and put `offline = false`.

casparvl: @Micket explained on chat: can be found by running `python3 -m easybuild.easyblocks.Cargo path/to/unpacked/tarball/` which scans the `cargo.lock` file.

Micket: As an extension, you probably could put the entire (huge) list of `crates` inside the extension block, but that's very ugly and cumbersome. I created CargoPythonBundle https://github.com/easybuilders/easybuild-easyblocks/pull/2964  but it's currently blocked because there is no good way for us to handle optional kwargs since we still have to support python2. 

I added offline option since it's still better than not using the easyblock at all, and who knows what some package might actually require. `cargo --offline `  really just blocks internet access.

casparvl: Aaah, right, so the reason I can do it here is essentially because it's CargoPythonPackage. In which case it's fine: I can define that _long_ and ugly crate list in the main easyconfig, instead of in an extension (which indeed would be enormous, especially if you have multiple extensions based on Cargo I guess).

Right, so in _those_ cases (or any other where a package might _really_ require internet connection, because it pulls something in on the fly), we have the fallback of using `offline = False`. In any case, these should really be the exceptions then.

I'll try my best to find the time to do a PR to make the docs a bit more extensive on this :)

casparvl: Ok, added the crate list and took out the `offline = True`

casparvl:
Test report by @casparvl
**SUCCESS**
tcn1.local.snellius.surf.nl - Linux RHEL 8.6, x86_64, AMD EPYC 7H12 64-Core Processor, Python 3.6.8
See https://gist.github.com/casparvl/2eca40f02c73987acbf2214547099ab5 for a full test report.

casparvl:

casparvl:

Micket (approved):
lgtm

Micket:
Test report by @Micket
**SUCCESS**
vera-c1 - Linux Rocky Linux 8.7, x86_64, Intel Xeon Processor (Skylake), Python 3.6.8
See https://gist.github.com/Micket/64b6d1578b9f56be9bdc84a5b6a5d2e8 for a full test report.

PR #18633, opened 2023-08-23
(draft/WIP) LAMMPS for Hopper GPUs

rarensu:
Howdy Y'all, 

I was able to get LAMMPS to build for Hopper GPUs. My Easyconfig is not perfect, but it at least mostly works. That's progress. If you wanted to help me finish this out, that would be great. Of course, since the H100 are so new, hardly anyone has them... likely this isn't a high priority. 

Known bug: python module is not installed. The new LAMMPS build procedure doesn't install the python module by default. I was too lazy to put the optional install command into the easyconfig. 
Known bug: only sm90 libraries are built. It's common practice to build for multiple targets and give the executables different names. E.g., lmp_mpi. I have not done that, instead I just force-picked the sm90. 
Known bug: lammps complains that OpenMP builds are not the best for GPU-only runs. Probably we should consider turning that off at build time. 

Richard

### DIFF
--- a/easybuild/easyconfigs/l/LAMMPS/LAMMPS-15Jun2023-foss-2022a-kokkos-sm90.eb
+++ b/easybuild/easyconfigs/l/LAMMPS/LAMMPS-15Jun2023-foss-2022a-kokkos-sm90.eb
@@ -1,0 +1,173 @@
+name = 'LAMMPS'
+version = '15Jun2023'
+versionsuffix = '-kokkos'
+_cuda_suffix = '-CUDA-%(cudaver)s'
+versionsuffix += _cuda_suffix
+cuda_compute_capabilities = ['9.0']
+versionsuffix += '-sm90'
+
+homepage = 'https://www.lammps.org'
+description = """LAMMPS is a classical molecular dynamics code, and an acronym
+for Large-scale Atomic/Molecular Massively Parallel Simulator. LAMMPS has
+potentials for solid-state materials (metals, semiconductors) and soft matter
+(biomolecules, polymers) and coarse-grained or mesoscopic systems. It can be
+used to model atoms or, more generically, as a parallel particle simulator at
+the atomic, meso, or continuum scale. LAMMPS runs on single processors or in
+parallel using message-passing techniques and a spatial-decomposition of the
+simulation domain. The code is designed to be easy to modify or extend with new
+functionality.
+"""
+
+toolchain = {'name': 'foss', 'version': '2022a'}
+toolchainopts = {'openmp': True, 'cstd': 'c++14', 'usempi': True}
+
+# 'https://github.com/lammps/lammps/archive/'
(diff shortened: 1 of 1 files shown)


sassy-crick:
@rarensu You might want to sync your PR with development, something like this: 
```
```

laraPPr:
@rarensu would you still like some help getting this pr over the finish line? I see that the LAMMPS version is not that recent anymore. And we might also want to update the toolchain.

easybuild/easyconfigs/l/LAMMPS/LAMMPS-15Jun2023-foss-2022a-kokkos-sm90.eb
```diff
@@ -0,0 +1,173 @@
+name = 'LAMMPS'
+version = '15Jun2023'
+versionsuffix = '-kokkos'
+_cuda_suffix = '-CUDA-%(cudaver)s'
+versionsuffix += _cuda_suffix
+cuda_compute_capabilities = ['9.0']
+versionsuffix += '-sm90'
```
laraPPr, line 7: This is to specific. We want easyconfigs that can work on any architecture not just for one architecture but for all of them. You can set the `cuda_compute_capabilities` when building it on your system by building with the flag `--cuda_compute_capabilities=9.0` or setting the cuda_compute_capabilities through eb_hooks.
```suggestion
```

easybuild/easyconfigs/l/LAMMPS/LAMMPS-15Jun2023-foss-2022a-kokkos-sm90.eb
```diff
@@ -0,0 +1,173 @@
...
+    ('kim-api', '2.3.0'),
+    ('Eigen', '3.4.0'),
+    ('PLUMED', '2.8.1'),
+    ('ScaFaCoS', '1.0.4'),
+    ('SciPy-bundle', '2022.05'),
+    # VTK package is auto-disabled if this dep is not available
+    ('VTK', '9.2.2'),
+]
+
+install_cmd = 'make install install-python'
```
laraPPr, line 61: This command should not be required when you use the following [patch](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/l/LAMMPS/LAMMPS-2Aug2023_install_lammps_python_package_in_eb_software_module.patch) and the updates made to the [easyblock](https://github.com/easybuilders/easybuild-easyblocks/blob/ca4372790f8ef81028aebc789a3939a4802f742c/easybuild/easyblocks/l/lammps.py#L510-L516). 
If you would like to still have the patch_15JUN2023 than we'll need some minor changes but I can help with that if you want.
```suggestion
```

laraPPr (changes requested):

boegel:

PR #11148, opened 2020-08-20
Remove whitespace at EOL from ECs

Flamefire:
First step of enabling flake8 for ECs.

Cmds run:
- `flake8 --filename '*.eb' --extend-ignore F821 -j4 --count` to find and count issues excluding unknown variables
- `flake8 --filename '*.eb' --select W291 -j5` to find only EOL whitespace
- `autopep8 -i --select W291 --max-line-length 120` to fix only this issue, in particular: `autopep8 -i --select W291 --max-line-length 120 $(flake8 --filename '*.eb' --select W291 -j5 --format='%(path)s' | sort -u)`

Note that `git diff develop --ignore-space-at-eol` turns up empty except for the CI disabling

PR #18646, opened 2023-08-25
remove executable flag from easyconfigs

Flamefire:
I found a bunch of ECs that were marked executable which they clearly should not. Might even be bad for security

I fixed that and added a test, that (changed) ECs are readable, owner-writable and non-executable, i.e. at least 0644 without any executable flag.

An example failure I forced to test the test:
`AssertionError: CFITSIO-3.34-ictce-5.5.0.eb must be readable (owner, group, other), not executable, at least owner writable, is: 0o100511`

As some of the ECs (mostly/only? in `__archive__`) didn't parse correctly anymore making the test fail I did the minimal changes to fix them.


Flamefire:
Lot's of issues with existing ECs in test_pr_sanity_check_paths, test_pr_sha256_checksums & test_pr_python_packages

How do you want to proceed?

Flamefire:
@smoors Rebased after merge of #18647 
The files tab shows that only permissions got changed (except for the commit attempting to fix CI by addressing the parse failures)

Would be great to get this merged as there were 4 new files with wrong permissions (likely C&P) but getting all the tests to pass is more work than it's worth

boegel:
this should be a significant help here:

* #18753

Flamefire:
Rebased to include #18753, let's see if that fixes CI

Flamefire:
@boegel How to proceed here?

boegel:
@Flamefire It's probably better to re-do this exercises in the `5.0.x` branch (which will soon-ish become `develop` once the last EasyBuild 4.x release has been published), since a lot of easyconfigs have been archived there already

Flamefire:
Agree to do that in addition but I'd expect users to still use the (next) 4.x release to copy ECs from and submit PRs so fleshing out those issues is IMO cheap enough to do here once in bulk

boegel:
> Agree to do that in addition but I'd expect users to still use the (next) 4.x release to copy ECs from and submit PRs so fleshing out those issues is IMO cheap enough to do here once in bulk

OK, but the failing CI is making that a really painful merge...

Do we have a test in place to verify that easyconfigs touched in a PR don't have exec permissions?

Flamefire:
> OK, but the failing CI is making that a really painful merge...

Can't you ignore the failing CI and still merge that? Other PRs should succeed then (There already was one where the exec permissions led to failures unrelated to the changes in that PR)
Also the CI running on the branch (i.e. after merge) doesn't take those changed ECs into account and hence wouldn't fail either, would it?

Hence I think it is less painful to do that straight-forward merge here once and ignore CI than to have contributors have to deal with the possible many times in the future.

> Do we have a test in place to verify that easyconfigs touched in a PR don't have exec permissions?

Yes, since https://github.com/easybuilders/easybuild-easyconfigs/pull/18647

Flamefire:
Maybe we can merge this and #11148 when collapsing develop and 5.0x? I guess we'll need to ignore CI for that step anyway and a few failures (like "no `use_pip`) will already go away with the new defaults and checks

Flamefire:
Closing after https://github.com/easybuilders/easybuild-easyconfigs/pull/19059 for 5.x

PR #18647, opened 2023-08-25
add test for easyconfig file permissions

Flamefire:
Extracted the test only from #18646 to check the file permissions of EC files on PR.

Test: ECs are readable, owner-writable and non-executable, i.e. at least 0644 without any executable flag.

An example failure I forced to test the test:
`AssertionError: CFITSIO-3.34-ictce-5.5.0.eb must be readable (owner, group, other), not executable, at least owner writable, is: 0o100511`

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -31,6 +31,7 @@ import glob
 import os
 import re
 import shutil
+import stat
 import sys
 import tempfile
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


Micket (commented):
Hm, I'd like to see what this looks like when it actually fails once, but i'm not sure how to trigger such a test locally

smoors:
did a quick test, seems to work fine (although running a single test did not work for me, i had to run all of them).

smoors (approved):
lgtm

PR #18683, opened 2023-08-29
{devel}[GCCcore/12.3.0] Qt6 v6.5.2

appolloford:
(created using `eb --new-pr`)

PR #18698, opened 2023-08-31
remove html5lib-1.1-GCCcore-12.3.0.eb as html5lib is in Python-bundle-PyPI/2023.06-GCCcore-12.3.0

branfosj:
(created using `eb --new-pr`)

[Python-bundle-PyPi contains html5lb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/p/Python-bundle-PyPI/Python-bundle-PyPI-2023.06-GCCcore-12.3.0.eb#L398), so we should use that.

It was added for #18683, but we should just do what I say in https://github.com/easybuilders/easybuild-easyconfigs/pull/18683#pullrequestreview-1605242431


smoors (approved):
lgtm

PR #18700, opened 2023-09-01
Update postinstallcmds to fix bin paths set for overlap.sh fixes #18699

### LINKED ISSUE #18699: MaSuRCA postinstallcmds miss some paths used to generate `overlap.sh`
The current `postinstallcmds` in the MaSuRCA easyconfig miss a couple of entries where the bin path is set and used to create a script `overlap.sh`.  In our case, this breaks MaSuRCA depending on how is run as `overlap.sh` will have `/scicore/soft/apps/MaSuRCA/Linux-amd64/bin/overlapInCore` instead of `/scicore/soft/apps/MaSuRCA/4.0.9-foss-2021a-Perl-5.32.1/bin/overlapInCore`

The code that requires some extra patching exists in:
- `PBcR`
- `runCA`
- `runCA-dedupe`
- `runCA-overlapStoreBuild`

The code to be replaced is:
- ` my $path = "$installDir/$syst-$arch/bin";`
- `$string .= "bin=\"$installDir/\$syst-\$arch/bin\"\n";`

i-mtz:
I have updated the postinstallcmds to match the missing strings that manipulate the bin path in several executables from `MaSuRCA`:
- `PBcR`
- `runCA`
- `runCA-dedupe`
- `runCA-overlapStoreBuild`

The PR only fixes the issue for `MaSuRCA-4.0.9-foss-2021a-Perl-5.32.1` but I have looked at the installed by `MaSuRCA-4.1.0-GCC-11.3.0` and the fix also works there.


boegel:
Closing this since `foss/2021a` is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @i-mtz 

PR #18713, opened 2023-09-04
{bio}[foss/2022a] PyTom 1.1, CuPy 12.1.0 /w CUDA 11.7.0

ThomasHoffmann77:
(created using `eb --new-pr`)
requires #19167


boegel:

@ThomasHoffmann77 Sorry for never picking up on this :(

PR #16388, opened 2022-10-11
{tools}[GCCcore/11.3.0] Apptainer v1.1.2

bartoldeman:
(created using `eb --new-pr`)

PR #18793, opened 2023-09-15
{tools}[system/system] Apptainer 1.2.3

rsdmse:


rsdmse:
I'm aware of #16388 but instead of a `Bundle` here I use `ConfigureMake`. The `osdependencies` are taken directly from https://apptainer.org/docs/admin/1.0/installation.html#id1 and it seems to work without installing `squashfuse` separately.

rsdmse:
Just wanted to check if this is on hold? Or can anyone review it?

andre-geldenhuis:
I don't have write access, but I ran your pull request to install Apptainer.  It failed due to the checksum being different.  The Apptainer 1.2.3 [sha256sums](https://github.com/apptainer/apptainer/releases/download/v1.2.3/sha256sums) is different than the one in your pull request.

Updating that made it build and work successfully.  Can't say if easybuild will merge this, but I found it handy, thanks!

PR #18795, opened 2023-09-16
prepend `$EBROOTFREESURFER/lib/qt/lib` to `$LD_LIBRARY_PATH` in order to override the qt libraries from `$EBROOTMCR`

### LINKED ISSUE #18831: FreeSurfer/7.4.1-centos8_x86_64 GUI not starting
With the latest EasyConfig of FreeSurfer, `FreeSurfer-7.4.1-centos8_x86_64.eb`, attempts to start the GUI on RHEL-8.5 using the `freeview` command fails, with the following error message:
```
Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.
```
Tracing it down with the help of @migueldiascosta we noticed that the X libraries from `MCR/R2019b.8` are being used, not the ones from `FreeView`. A simple change of the order of the `LD_LIBRARY_PATH` to `LD_LIBRARY_PATH=$EBROOTFREESURFER/lib/qt/lib:$LD_LIBRARY_PATH` , i.e. make sure the X libraries from `FreeView` are loaded first, seem to have sorted that issue. 

We propose to add
```
modextrapaths = {'LD_LIBRARY_PATH': ['lib/qt/lib']}
```
to the `FreeView` module. 

Maybe somebody could test that on other installations as well to make sure it is not a RHEL specific issue?

Thanks

migueldiascosta:
(created using `eb --new-pr`)

@sassy-crick reported on Slack that `freeview` failed to run when the `MCR` module was loaded, and we tracked it down to it loading the `libQt5*` libraries from `$EBROOTMCR` instead of from `$EBROOTFREESURFER/lib/qt/lib`, this fixes PT that

I don't *think* it breaks FreeSurfer's use of `MCR`, but I don't know for sure...

edit: fixes https://github.com/easybuilders/easybuild-easyconfigs/issues/18831


boegel:
@migueldiascosta Is this still relevant?

migueldiascosta:
That's a good question, did it work for you @sassy-crick ?

Even if it did, looking at https://github.com/easybuilders/easybuild-easyconfigs/issues/18831#issuecomment-3451100014, looks like it's not an universal fix...

sassy-crick:
I am wondering if we not simply should updated to a newer version of `MCR`, ie. `MCR-R2024b.eb` for example, or is `FreeSurfer` really bound to that particular `MCR` version?

PR #18797, opened 2023-09-18
Ignore EasyConfig cache file

Flamefire:
EasyBuild creates the file `.eb-path-index` in `easybuild/easyconfigs` to cache searches which shows up in git checkouts when used directly.

As we never want to add this file to git ignore it explicitly


SebastianAchilles (approved):
lgtm

SebastianAchilles:
Going in, thanks @Flamefire!

PR #18967, opened 2023-10-10
Archive pre 2016a compilers

Micket:
We had a bunch of pre-2016a GCC's that i also included. It's a bit messy back then so I might have missed something.


Micket:
bah, wrong branch, i don't see how to edit that

PR #18973, opened 2023-10-10
{vis}[GCCcore/12.3.0] setuptools v68.2.2

garadar:
(created using `eb --new-pr`)


garadar:
@boegel is this PR is OK for you ?

boegel:
@garadar Do you have a particular use case for this?

Python 3.11.3 already includes `setuptools` 67.7.2, is that not recent enough for a particular use case?

boegel:
no reply for over 2 years, still unclear why this would be useful, so closing...

PR #18958, opened 2023-10-09
Archive 2016a

Micket:
So many oddball toolchains back then. Good riddance!

PR #18982, opened 2023-10-11
remove old archived easyconfigs (EasyBuild 4.x archive)

boegel:
We should clean up the easyconfigs that were archived in EasyBuild 4.x, to avoid things piling up in the archive.

It's quite unlikely that there's anything useful in there by now.

The archive in EasyBuild 5.0 will only contain easyconfigs from the `2016a` era onwards (see #18958, #18968, #18976, #18978, ...)


Micket (approved):
lgtm

PR #18996, opened 2023-10-11
WIP: {tools}[GCCcore/13.2.0] Jupyter-bundle-2023{nn}{nn} + deps

jose-d:
(created using `eb --new-pr`)

Missing dependencies to have Jupyter-bundle ready for Python at GCCcore/13.2.0 level

- [x] [maturin](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/m/maturin) - provided in https://github.com/easybuilders/easybuild-easyconfigs/pull/19099
- [x] [Python-bundle-PyPI](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/p/Python-bundle-PyPI) - provided in https://github.com/easybuilders/easybuild-easyconfigs/pull/19143
- [x] [PyYAML](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/p/PyYAML)
- [ ] [PyZMQ](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/p/PyZMQ)
- [ ] [aiohttp](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/a/aiohttp)
- [x] [poetry](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/p/poetry) - https://github.com/easybuilders/easybuild-easyconfigs/pull/19099

splitted into different PR:

- [x] IPython-8.17.2-GCCcore-13.2.0.eb - #19277
- [x] jedi-0.19.1-GCCcore-13.2.0.eb - #19277

Updated / implemented at 13.2.0 level in this PR:

- [x] anyio-4.0.0-GCCcore-13.2.0.eb
- [x] hatch-jupyter-builder-0.8.3-GCCcore-13.2.0.eb
- [x] ~hatchling-1.18.0-GCCcore-13.2.0.eb~ - inbetween implemented in different PR
- [x] hatch-nodejs-version-0.3.2-GCCcore-13.2.0.eb
- [x] prometheus-client-0.17.1-GCCcore-13.2.0.eb
- [x] requests-2.31.0-GCCcore-13.2.0.eb
- [x] tornado-6.3.3-GCCcore-13.2.0.eb

Remains to update/implement:

- ~jsonschema (depends on `maturin`)~ - part of Python-bundle-PyPI-2023.10-GCCcore-13.2.0.eb
- JupyterLab (needs indirectly `Python-bundle-PyPI` )
- JupyterNotebook (needs indirectly `Python-bundle-PyPI` )
- jupyter-resource-usage (needs indirectly `Python-bundle-PyPI` )
- jupyter-server (needs `PyYAML`,`PyZMQ` + others )
- jupyter-server-proxy (needs indirectly `Python-bundle-PyPI`)
- python-lsp-server (needs `poetry` + others )
- aiohttp (needs `poetry` + others )
______________

There is new major version of `jupyterlab-lsp` - 5.0.0 - finally compatible with JupyterLab-4+, so it allows us to combine `jupyterlab-lsp` and `python-lsp-server` bringing style and syntax checks into our notebooks.

Few refactorings were done, including pulling some dependencies from PythonBundles into new standalone easyconfigs, which are part of this PR:

* jedi
* hatch-jupyter-builder
* hatch-nodejs-version
* jsonschema
* anyio
* prometheus-client
* requests

I think this refactoring could avoid duplicate library installs, and make regular updates easier.


jose-d:
I am switching to draft, still considering if not pull even more libraries out of bundles into standalone easyconfigs.

casparvl:
Is there any particular reason to update all of those dependency-versions (i.e. `IPython`, `JupyterLab`, `JupyterNotebook`, `jupyter-server` and `jupyters-server-proxy`)? In principle, we always use the same versions for dependencies within a toolchain generation (i.e. `GCCcore/12.3.0`/`foss-2023a`), unless there is a good reason not to. That's also why our CI checks it, and complains now for your PR that you are introducing multiple versions for the same dependency in the same toolchain.

Good reasons are for example

- (substantial) Performance benefits
- Your software simply doesn't work with the existing version of that dependency.

Unless I'm mistaking, these don't really apply here (unless `jupyter-lsp` has no way of working with the existing versions of those dependencies - but the version differences are so minor that I cannot imagine that to be the case?).

I would personally just use the versions of  `IPython`, `JupyterLab`, `JupyterNotebook`, `jupyter-server` and `jupyters-server-proxy` that are already there. In terms of the `jupyter-bundle`, we can discuss (and I'd like to get some other maintainer opinions as well) what would be the most sensible approach: update the existing one, or create a new one (with `jupyter-lsp`). I'd personally just update the existing one, as I think we've done similar things with other bundles in the past... But again, would be good to get some more input on that.

jose-d:
hello @casparvl !

> Is there any particular reason to update all of those dependency-versions

The upstream-required dependency is between `jupyterlab-lsp` (first (!) working with JupyterLab4+)  [needing](https://github.com/jupyter-lsp/jupyterlab-lsp/releases/tag/v5.0.0) at least `JupyterLab` 4.0.6.

The rest .. it's more that I saw no reason to not use the recent software versions. I wrote some _request-for-comments_ into EB Slack [here](https://easybuild.slack.com/archives/C34UA1HT7/p1697101605544689) - but it can be I wrote it in too boring/lenghty way, or it is simply quite marginal topic, as possibly most easybuild sites use Conda to deliver Jupyter to users.

I/my site/ plan to stick to 2023a for at least a year or two (I mentioned reasons into the Slack post mentioned above), so I have quite good motivation to keep Jupyter-* stack up to date..  (people doing `pip install --user` on top of this)
..if it's unthinkable to merge this PR, we can keep it open as a draft, and once 2023b and related Python will be clear/known, I can try to rewrite it to use GCCcore from 2023b toolchain..?

>  these don't really apply here (unless jupyter-lsp has no way of working

Please note the difference between [`jupyter-lsp` ](https://pypi.org/project/jupyter-lsp/)(already supported by Easybuild) and [`jupyterlab-lsp`](https://pypi.org/project/jupyterlab-lsp) (Easybuild support introduced in this PR).

jose-d:
@casparvl oki, so after very informative explanation from @Micket at Slack, I suggest to keep this PR in draft mode, and once 2023b will be out, upgrade easyconfigs and possibly merge?

casparvl:
@jose-d I think that's actually a clever suggestion from @Micket . Indeed, it is quite uncommon for us to use multiple versions of the same dependency. In this case, since it is quite particular to the Jupyter stack (and you'd never mix those anyway), I would have personally been ok with putting in a second version in `2023a` - but putting it in `2023b` is more elegant. And since that toolchain is already [coming together quite nicely](https://github.com/easybuilders/easybuild-easyconfigs/pulls?q=is%3Apr+label%3A2023b+is%3Aclosed), it shouldn't be too long before that's possible

Micket:
Huh, I don't think I made any clever suggestions? (I just provided the usual explanation as to why multiple deps within the toolchain generation is painful)

casparvl:
In that case I guess @jose-d drew this clever conclusion on his own :D in any case: I'm in favour, I think it's a good approach :D

jhgoebbert:
You might want to add `Python-bundle-PyPI` as dependency for `python-lsp-server` as this would shrink your ebconfig file for quite some packages.

jhgoebbert:
Thank you for this great draft!

I tried to stick close to it and updated/extended the easybuild files and dependencies for our installation we use at https://jupyter.jsc.fz-juelich.de which gives access to all HPC systems and the cloud at Jülich Supercomputing Centre.

You can find our easyconfigs for the JupyterLab modules here: 
https://github.com/easybuilders/JSC/blob/2024/Golden_Repo/j/Jupyter-bundle/Jupyter-bundle-20240520-GCCcore-12.3.0.eb

I will try to adopt this to upstream as PRs as soon as possible.

jose-d:
hi @jhgoebbert, 

Good to hear that this draft was useful for you. I'm about to close this PR, as we have decided to manage our JupyterLab differently.

Our problem was that creating and updating a whole tree of easyconfigs for _Jupyter and friends_ is quite demanding. Also, PRs into easybuild-easyconfigs that are not aligned with usual toolchain releases (202Xa-b, etc.) are usually not accepted, and I understand the reasons.

Discussion in the EB Slack revealed that a more efficient way — at least for us — is to have JupyterLab installed in a Python virtual environment (venv). This venv is kept up-to-date by creating a new one regularly using standard Python tools.

To keep the runtime environment reproducible, the JupyterLab from the venv starts (using kernelspec) IPython built by EasyBuild. For purposes for scientific app runtime, we're fine staying with the standard toolchain release cycle—2023a, b, etc.

This separation of Jupyter and (i)Python exposed to users ensures an up-to-date Jupyter and no need to create a new JupyterLab easyconfig for every Jupyter release.

cheers

josef

jhgoebbert:
Thanks @jose-d 
We will stay with our current approach of JupyterLab as modules for the next software stage (which will be default until Nov. 2025) but I will keep an eye on your venv-solution. I fully understand the motivation - even though it rises some open questions.

jose-d:
@jhgoebbert

> even though it rises some open questions.

Any questions are welcomed. I can not say mentioned solution is the final and best one, and any input can possibly lead to improvement...  
Cheers

jhgoebbert:
Hi @jose-d 

at our side JupyterLab has become an alternative to ssh for login to our HPC systems (login- and compute-nodes) without anything else than a browser ([slides](https://www.fz-juelich.de/de/ias/jsc/aktuelles/termine/kurse/2023/supercomputing-2/08-jupyterlab)). It is extremely easy to have this comparable (ssh vs JupyterLab) access as we provide exactly the same environment based on EasyBuild by default.

If JupyterLab comes in its own venv the most critical is probably to explicitly ensure that users will use the performance optimized Python modules, like numpy etc., from the software stack when using the default Python kernel. 🤔 
But I am sure this is solvable as kernels are running in their own environment anyway.

For now a pure venv adds for me - as the one who installs/maintains the JupyterLab software at JSC - a change with less benefit. I am quite happy that all Python packages in a software stage are compatible so that the JupyterLab is by design a solid and reliable UI. Users are less interested in new features than in robustness when it comes to the interface they use to interact with the HPC systems - if we update the version of JupyterLab once a year it is fine.

But I can image that a smart use of venv comes with other benefits. It might for example allow users in future to install new extensions in JupyterLab 4 through the Plugin Manager. EasyBuild just have to ensure that this JupyterLab works fine by default together with all the other modules of the software stage or we can directly switch to containers - like Jupyter Docker Stacks & Co.

jose-d:
closing this one as we changed approach to running Jupyter at our site.

PR #13385, opened 2021-07-08
Remove duplicated PYTHONPATH module entries

Flamefire:
Those ECs set the PYTHONPATH again which is already done by the used easyblock leading to a slower module/python

Running into https://github.com/easybuilders/easybuild-easyconfigs/issues/13432 while testing this, hence fixed here

Fixes #13432

PR #19061, opened 2023-10-23
Add test for duplicate `PYTHONPATH` in `modextrapaths`

Flamefire:
And similar test for `sanity_check_paths`.

Intended to avoid issues fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/13385

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -1112,6 +1112,24 @@ class EasyConfigTest(TestCase):
                 if not sanity_pip_check and not any(re.match(regex, ec_fn) for regex in whitelist_pip_check):
                     failing_checks.append("sanity_pip_check should be enabled in %s" % ec_fn)
 
+            # Avoid duplicated PYTHONPATH entries and checks (set by the easyblock)
+            if easyblock == 'PythonBundle' or easyblock.endswith('PythonPackage'):
+                extra_python_path = ec.get('modextrapaths', dict()).get('PYTHONPATH')
+                if extra_python_path == 'lib/python%(pyshortver)s/site-packages':
+                    failing_checks.append("modextrapaths should not contain %s "
+                                          "(automatically added by easyblock)" % extra_python_path)
+                sanity_check_dirs = ec.get('sanity_check_paths', dict()).get('dirs') or []
+                default_dirs = (
+                    'lib/python%(pyshortver)s/site-packages',
+                    'lib/python%(pyshortver)s/site-packages/%(name)s',
+                    'lib64/python%(pyshortver)s/site-packages',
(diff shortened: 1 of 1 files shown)


Flamefire:
@boegel Anything missing here? This is a CI test change only so shouldn't cause any regressions for users (but actually helps avoiding them)

Flamefire:
Retargeteed 5.0x

boegel:
@Flamefire Sync with `5.0.x` now that #22061 is merged?

Flamefire:
@boegel Rebased this again and extended to checking the resolved values as well as one instance used `/lib/python%s...' % local_pyshortver` which will now be caught too

Can this be merged before new ECs with those paths get introduced?

boegel:
@Flamefire Needs more love (again), sorry...

Flamefire:
Rebased. Trivial merge conflict after a cleanup of a condition

test/easyconfigs/easyconfigs.py
```diff
@@ -1330,30 +1330,54 @@ def test_pr_python_packages(self):
...
+                'lib/python%(pyshortver)s/site-packages',
+            ]
+            with ec.allow_unresolved_templates():
+                default_dirs += [ec.resolve_template(d) for d in default_dirs]
+            extra_python_path = ec.get('modextrapaths', dict()).get('PYTHONPATH')
+            failing_checks_ec.extend(f"modextrapaths should not contain {d} (automatically added by EasyBlock)"
+                                     for d in default_dirs if d in extra_python_path)
+            if easyblock == 'PythonBundle' or easyblock.endswith('PythonPackage'):
+                sanity_check_dirs = ec.get('sanity_check_paths', dict()).get('dirs') or []
+                default_dirs = (
```
akesandgren: Should this  use [ and not ( ??

Flamefire: The general guideline I follow is: If the list doesn't need to be modified it will be a tuple

Not sure if this is also faster (due to less capabilities to cater for) but at least it avoids accidental changes where none are intended

akesandgren: But you are modifying it in the with ec.allow... for loop, line 1376

Flamefire: Good catch, not sure how I missed that as it would always fail

akesandgren (dismissed):
LGTM

akesandgren:
@boegel Please check this out. It looks good and clean to me...

akesandgren:
@Flamefire Hmm... I just caused this to have a conflict when merging 21362

Please adjust it

Flamefire:
Rebased and fixed

PR #19106, opened 2023-10-30
{bio}[GCCcore/10.3.0] mdust v20150102

wdlingit:
The mdust program for low-complexity removal. From a github historical archive.


easybuild/easyconfigs/m/mDust/mDust-20231029-GCC-10.3.0.eb
```diff
@@ -0,0 +1,37 @@
+easyblock = 'MakeCp'
+
+name = 'mDust'
```
bedroge, line 3: On the github page and in the source code they use `mdust` everywhere, so perhaps we should stick to that? This is also used in the description below.

easybuild/easyconfigs/m/mDust/mDust-20231029-GCC-10.3.0.eb
```diff
@@ -0,0 +1,37 @@
...
+version = '20231029'
```
bedroge, line 4: It's better and clearer to stick to the date of the used commit here.
```suggestion
version = '20150102'
```

easybuild/easyconfigs/m/mDust/mDust-20231029-GCC-10.3.0.eb
```diff
@@ -0,0 +1,37 @@
...
+
+toolchain = {'name': 'GCC', 'version': '10.3.0'}
+
+sources = [{
+    'filename': 'mdust.tar.gz',
+    'git_config': {
+        'url': 'https://github.com/lh3',
+        'repo_name': 'mdust',
+        'commit': '3e3fed8',
+        'clone_into': 'mdustSrc',
```
bedroge, line 17: The default is to use the name of the repo (`mdust`), which is fine I guess, so I'd leave this out:
```suggestion
```

easybuild/easyconfigs/m/mDust/mDust-20231029-GCC-10.3.0.eb
```diff
@@ -0,0 +1,37 @@
...
+    },
+}]
+# git clone, might have to adjust this checksum
+checksums = ['6fe21806e6bcd3810babed8d83dfb9ff072fcb9d3a31cd43da9cb9670c0ec6f0']
```
bedroge, line 21: Since this can change, we usually set the checksum to `None`:
```suggestion
checksums = [None]
```

easybuild/easyconfigs/m/mDust/mDust-20231029-GCC-10.3.0.eb
```diff
@@ -0,0 +1,37 @@
...
+
+configure_without_installdir = True
```
bedroge, line 23: `MakeCp` doesn't do a `configure`, so this can be removed.
```suggestion
```

easybuild/easyconfigs/m/mDust/mDust-20231029-GCC-10.3.0.eb
```diff
@@ -0,0 +1,37 @@
...
+
+files_to_copy = []
+
+postinstallcmds = ['mkdir %(installdir)s/bin',
+                   'cp mdust %(installdir)s/bin']
```
bedroge, line 28: Instead of doing a manual copy, you can use the `files_to_copy` (the `bin` directory is created automatically):
```suggestion
files_to_copy = [(['mdust'], 'bin')]
```

bedroge (changes requested):
Thanks for the contribution. I've made a few comments, also note that the filename would have to be changed to reflect the requested changes in the name and version.

bedroge:

bedroge:

bedroge (approved):
Looks good to me.

bedroge:
Going in, thanks @wdlingit!

bedroge:
Going in, thanks @wdlingit!

PR #19171, opened 2023-11-07
{tools}[GCCcore/10.2.0,GCCcore/12.2.0] f90nml v1.4.4

SimonPinches:
(created using `eb --new-pr`)


easybuild/easyconfigs/f/f90nml/f90nml-1.4.4-GCCcore-10.2.0.eb
```diff
@@ -0,0 +1,27 @@
+easyblock = 'PythonPackage'
+
+name = 'f90nml'
+version = '1.4.4'
+
+homepage = 'https://github.com/marshallward/f90nml'
+description = """A Python module and command line tool for parsing
+    Fortran namelist files"""
+
+toolchain = {'name': 'GCCcore', 'version': '10.2.0'}
```
akesandgren, line 10: This belongs to the foss/2020b toolchain which has been deprecated (https://docs.easybuild.io/policies/toolchains/) 

I suggest dropping it from the PR.

akesandgren (changes requested):

easybuild/easyconfigs/f/f90nml/f90nml-1.4.4-GCCcore-10.2.0.eb
SimonPinches: Ah, it wasn't deprecated when I submitted this :-)
I hope it's okay if I move it to the __archive__ folder.  We still actively use 2020b so it would be nice to keep this EB file somewhere.
I've also added the GCCcore/13.2.0 version.

akesandgren (approved):
LGTM

akesandgren:

PR #19179, opened 2023-11-08
{numlib,toolchain}[iompi/2022a,system/system] iomkl v2022a, iompi v2022a, imkl v2022.1.0, ...

branfosj:
(created using `eb --new-pr`)


branfosj:
We switched to using Intel MPI for 2022a, which means this is untested. So I'll close it.

PR #19265, opened 2023-11-20
clean up easyconfigs that explicitly set `use_pip`, `sanity_pip_check` and `download_dep_fail`

SebastianAchilles:
The cleanup was done with
```
sed -i "/use_pip = True\|sanity_pip_check = True\|download_dep_fail = True\|'use_pip': True\|'download_dep_fail': True\|'sanity_pip_check': True/d" easybuild/easyconfigs/*/*/*.eb
```

In addition it is necessary to take care of too many blank lines. This was done with:
```
git status --porcelain | sed s/^...// | grep ".eb" | xargs -n1 perl -0777pi -e 's/\n{3,}/\n\n/g'
```

Depends on:
- [x] https://github.com/easybuilders/easybuild-easyblocks/pull/3022


test/easyconfigs/easyconfigs.py
```diff
@@ -1107,16 +1106,7 @@ def test_pr_python_packages(self):
...
-                    fail = "'*download_dep_fail' should not be set in %s since PythonBundle easyblock is used" % ec_fn
-                    failing_checks.append(fail)
-                if pure_ec.get('exts_default_options', {}).get('source_urls') == python_default_urls:
-                    failing_checks.append("'source_urls' should not be defined in exts_default_options when using "
-                                          "the default value in %s" % ec_fn)
-
-            elif exts_defaultclass == 'PythonPackage':
+            if exts_defaultclass == 'PythonPackage':
                 # bundle of Python packages should use PythonBundle
                 if easyblock == 'Bundle':
```
branfosj, line 1111: The failures in https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/7818343578/job/21368323894?pr=19265 are all triggered in the `else` side of this `if` (I cannot comment there directly as it is too far from code being changed).

Do we also need to remove that?

boegel: fixed in 0fd98988ae

boegel (approved):

boegel:
Going in, thanks @SebastianAchilles!

PR #19374, opened 2023-12-07
Replace `parallel` by `maxparallel`

Flamefire:
The latter is the new, better option which makes it clear how it works with `--parallel` passed in the config or commandline.

See https://github.com/easybuilders/easybuild-framework/pull/3842

Note: CI fails for unrelated reasons and fixing all those ECs is unfeasible, so I'd suggest to bite the bullet and merge this even though CI fails.


Flamefire:
@boegel Resolved conflicts

Flamefire:
@boegel What about this (and #19375)? Unfortunately CI won't pass (E.g. "FAIL: test_pr_CMAKE_BUILD_TYPE (test.easyconfigs.easyconfigs.EasyConfigTest)") and as mentioned in the last conf call I don't see much worth in it given the actual change. So maybe just bite the bullet, do the replacement once (for develop and 5.0x branches) and merge faster than conflicts (re)appear?

Flamefire:
Closing after the merge to 5.0x in https://github.com/easybuilders/easybuild-easyconfigs/pull/19375

PR #19375, opened 2023-12-07
Replace `parallel` by `maxparallel`

Flamefire:
The latter is the new, better option which makes it clear how it works with `--parallel` passed in the config or commandline.

5.x target of #19374  required by https://github.com/easybuilders/easybuild-framework/pull/4398

Note: CI fails for unrelated reasons and fixing all those ECs is unfeasible, so I'd suggest to bite the bullet and merge this even though CI fails.


Flamefire:
@boegel Resolved conflicts

Flamefire:
Rebased and updated. Let's see what CI says ;-)

boegel:
@Flamefire We should add a small check in the easyconfig test suite to make sure that no easyconfigs are still setting `parallel`?

Flamefire:
> @Flamefire We should add a small check in the easyconfig test suite to make sure that no easyconfigs are still setting `parallel`?

I added a CI check for all deprecated EC parameters.

The setting I mentioned in the confcall that would detect any deprecated use is here: https://github.com/easybuilders/easybuild-easyconfigs/blob/9b6e0a1b400569b82c234eed4f07ccebc0a2ce4b/test/easyconfigs/suite.py#L49
It might be better to enable this instead to check for **any** deprecated behavior.

However I disabled that in https://github.com/easybuilders/easybuild-easyconfigs/pull/13257 with the comment
> Remove unused setting of `EASYBUILD_DEPRECATED` which, if in effect, would cause many failures due to deprecated ECs (using old versions) or even using LMod 6 (on CI) is already deprecated

I'm not sure why I determined this as unused back then, so maybe that has changed. If I do enable it the test fails earlier:
```
  File "/git/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 251, in _get_changed_easyconfigs
    ec = process_easyconfig(ec_file)
  File "/git/easybuild-framework/easybuild/framework/easyconfig/easyconfig.py", line 2147, in process_easyconfig
    raise EasyBuildError("Failed to process easyconfig %s: %s", spec, err.msg, exit_code=exit_code)
easybuild.tools.build_log.EasyBuildError: "Failed to process easyconfig /git/easybuild-easyconfigs/easybuild/easyconfigs/b/Bracken/Bracken-2.7-GCCcore-11.2.0.eb: DEPRECATED (since v5.1) functionality used: Easyconfig parameter 'parallel' is deprecated, use 'maxparallel' instead.; see https://docs.easybuild.io/deprecated-functionality/ for more information"
```

Micket (approved):
So, I tried to follow the discussion across multiple PRs, maybe I'm missing something, but this should just be good to go right? I browsed through the changes and I think they look fine. The extra check looks OK to me. Regardless of the changes in framework or easyblocks, this is just  a necessary and uncontroversial merge right?

Micket:
Sounds good to me. I see no objections here so I will push forward and we can follow up if there is something.

boegel:
Setting `$EASYBUILD_DEPRECATED` was done in #21885

PR #19382, opened 2023-12-08
Molpro 2022

nikishe:
Molpro eb

and also th SciPy-bundle-2022.05-foss-2022a.eb


smoors:
@nikishe can you clarify why you are adding `SciPy-bundle-2022.05-foss-2022b.eb`?
we already have `SciPy-bundle-2023.02-gfbf-2022b.eb`, is there a specific reason why that version is not sufficient?

smoors:
> > ady have `SciPy-bundle-2023.02-gfbf-2022b.eb`, is there a specific reason why that version is not sufficient?
> 
> I needed a Foss toolchain. It makes it easier in getting HPC newbies to pay attention to the toolchain use

`gfbf` is a subtoolchain of `foss`, isn't that good enough?

smoors:
> True , but doesnt that mean no mpi though?

indeed, `mpi4py` has been removed from `SciPy-bundle` exactly for that reason, so that MPI is no longer needed and can be lowered to `gfbf` subtoolchain

nikishe:
Thanks Smoors , Ill cancel the request

> > True , but doesnt that mean no mpi though?
> 
> indeed, `mpi4py` has been removed from `SciPy-bundle` exactly for that reason, so that MPI is no longer needed and can be lowered to `gfbf` subtoolchain

smoors:
> Thanks Smoors , Ill cancel the request

but the Molpro easyconfig is still ok, right?
if you reopen the PR, remove SciPy-bundle, and send a test report for Molpro, we can merge it without problem.

PR #19411, opened 2023-12-13
Create Rcorrector-1.0.6-GCC-11.3.0.eb

levlafayette:
New version, new toolchain.

### DIFF
--- a/easybuild/easyconfigs/r/Rcorrector/Rcorrector-1.0.6-GCC-11.3.0.eb
+++ b/easybuild/easyconfigs/r/Rcorrector/Rcorrector-1.0.6-GCC-11.3.0.eb
@@ -1,0 +1,31 @@
+name = 'Rcorrector'
+version = '1.0.6'
+
+homepage = 'https://github.com/mourisl/Rcorrector'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/r/Rcorrector/Rcorrector-1.0.6-GCC-11.3.0.eb
```diff
@@ -0,0 +1,31 @@
+name = 'Rcorrector'
```
jfgrimm, line 1: @levlafayette you're not specifying a generic easyblock here, and we don't have an existing software-specific easyblock for `Rcorrector`

akesandgren: The old version uses
```
easyblock = 'MakeCp'
```

akesandgren (changes requested):

sassy-crick:
@levlafayette You might want to sync your PR with development, something like this: 
```
```

boegel:

PR #19472, opened 2024-01-02
{data}[GCCcore/13.2.0] MDSplus v7.132.0

SimonPinches:
(created using `eb --new-pr`)


easybuild/easyconfigs/m/MDSplus/mdsplus-7.132.0_paths.patch
```diff
@@ -0,0 +1,239 @@
+diff -Nru mdsplus-java-7.131.6-orig/mdsplus/java/jdispatcher/jDispatcherIp.template mdsplus-java-7.131.6/mdsplus/java/jdispatcher/jDispatcherIp.template
```
akesandgren, line 1: Please add description and author/date info to the patch (and update checksums)

akesandgren (changes requested):

easybuild/easyconfigs/m/MDSplus/mdsplus-7.132.0_paths.patch
SimonPinches: Thanks @akesandgren, I've added a very brief description to the patch and updated as you suggested.

akesandgren:
@boegelbot Please test @ generoso

easybuild/easyconfigs/m/MDSplus/MDSplus-7.132.0-GCCcore-13.2.0.eb
```diff
@@ -0,0 +1,77 @@
...
+    'git_config': {
+        'url': 'https://github.com/' + name,
+        'repo_name': '%(namelower)s',
+        'commit': 'stable_release-' + version.replace('.', '-'),
+        'recursive': True,
+    },
+}]
+patches = ['%(namelower)s-%(version)s_paths.patch']
+checksums = [
+    {'mdsplus-7.132.0.tar.gz': '130b584c426e09167e042f2da829c03f59c57e9b352844bb4ef0880ccf93c8a7'},
```
akesandgren, line 24: Since this is pulled down using git_config it will not have a stable checksum.
I.e. you need to change it to
```
{'mdsplus-7.132.0.tar.gz': None},
```
But I'm not sure if we've manged to fix the bugs we had in framework regarding this.

It would be better to pull down the https://github.com/MDSplus/mdsplus/archive/refs/tags/stable_release-7-132-0.tar.gz file and the corresponding version of https://github.com/MDSplus/3rd-party-apis.git unfortunately it doesn't have releases, so the problem will remain...

akesandgren (changes requested):

easybuild/easyconfigs/m/MDSplus/MDSplus-7.132.0-GCCcore-13.2.0.eb
SimonPinches: Thanks @akesandgren, this is good to know!  I've updated the EB file to reflect this.  I had to change the syntax slightly since `--check-style` wasn't happy with what you proposed.

akesandgren:
This unfortunately trips on https://github.com/easybuilders/easybuild-framework/issues/4142 and the problems around that.
We'll probably need to wait until that is fixed.

SimonPinches:
This PR now relies upon #19620.

Louwrensth:
@boegelbot Please test @ generoso

akesandgren:
We already have 7.153.3 for GCCcore/13.2.0 merged so closing this one.

PR #19505, opened 2024-01-05
Julia 1.10.0

carstenbauer:
Tested locally.

### DIFF
--- a/easybuild/easyconfigs/j/Julia/Julia-1.10.0-linux-x86_64.eb
+++ b/easybuild/easyconfigs/j/Julia/Julia-1.10.0-linux-x86_64.eb
@@ -1,0 +1,23 @@
+easyblock = 'Tarball'
+
+name = 'Julia'
+version = '1.10.0'
+versionsuffix = '-linux-x86_64'
+
+homepage = 'https://julialang.org'
+description = "Julia is a high-level, high-performance dynamic programming language for numerical computing"
+
+toolchain = SYSTEM
+
+source_urls = ['https://julialang-s3.julialang.org/bin/linux/x64/%(version_major_minor)s/']
+sources = ['%(namelower)s-%(version)s%(versionsuffix)s.tar.gz']
+checksums = ['a7298207f72f2b27b2ab1ce392a6ea37afbd1fbee0f1f8d190b054dcaba878fe']
+
+sanity_check_paths = {
+    'files': ['bin/julia', 'include/julia/julia.h', 'lib/libjulia.%s' % SHLIB_EXT],
+    'dirs': ['bin', 'etc', 'include', 'lib', 'share']
+}
+
+sanity_check_commands = ['julia --help']
+
+moduleclass = 'lang'


smoors:
@carstenbauer why did you remove `JULIA_DEPOT_PATH`?
see Julia-1.9.3-linux-x86_64.eb
```python
modextravars = {
    # Use default DEPOT_PATH where first entry is user's depot
    # JULIA_DEPOT_PATH must be set to be able to load other JuliaPackage modules and have
    # those installations appended to DEPOT_PATH without disabling any of the default paths
    'JULIA_DEPOT_PATH': ':',
}
```

carstenbauer:
@smoors First of all, I didn't intentionally drop it but just added the config that I use on our cluster (which doesn't have this part). But looking at it, I don't think setting `JULIA_DEPOT_PATH` this way is a good idea for two reasons

1) Users should be able to set `JULIA_DEPOT_PATH` themselves. Loading a module with this extra bit would overwrite the user settings.
2) For Julia, it is generally not advisable to provide packages (like MPI.jl, or CUDA.jl) in a central fashion (as I think is the idea behind "JuliaPackage" modules). Instead it is recommended to provide light-weight configuration files for these packages such that, when a user decides to add them to a Julia project, they automatically behave in a certain way (e.g. use a system MPI or CUDA, for example).

That said, I don't care all too much. If you want to keep this part, I can add it (but won't use it 😉).

bartoldeman:
we're using
`append_path("JULIA_DEPOT_PATH",":")`
via modluafooter (or it could be done via a custom easyblock).
This allows users to set `JULIA_DEPOT_PATH` in `.bashrc` (or similar) but makes sure extensions can still append to it via the juliapackage easyblock.

smoors:
> @smoors First of all, I didn't intentionally drop it but just added the config that I use on our cluster (which doesn't have this part)

i didn't think you did this intentionally :)
it's good practice to always compare with existing easyconfigs for the same software, to make sure important customizations/fixes don't get lost in newer versions.
you can do this for example with `eb --preview-pr <easyconfig>`

smoors:
> we're using `append_path("JULIA_DEPOT_PATH",":")` via modluafooter (or it could be done via a custom easyblock). This allows users to set `JULIA_DEPOT_PATH` in `.bashrc` (or similar) but makes sure extensions can still append to it via the juliapackage easyblock.

i like this approach. i'm adding support for `modextrapaths_append` to make this a bit more user friendly
https://github.com/easybuilders/easybuild-framework/pull/4436

smoors:
> 2\. For Julia, it is generally not advisable to provide packages (like MPI.jl, or CUDA.jl) in a central fashion (as I think is the idea behind "JuliaPackage" modules). Instead it is recommended to provide light-weight configuration files for these packages such that, when a user decides to add them to a Julia project, they automatically behave in a certain way (e.g. use a system MPI or CUDA, for example).

it's still useful to have some julia packages centrally installed to make it easier for users who don't want to have to install it themselves, or as part of a bigger software project.

can you give some more info about those light-weight configuration files, maybe an example?

bartoldeman:
@carstenbauer there's also the problem of using lots of inodes in home directories with a lot of duplication, which isn't ideal on parallel file systems.
It's the same issue as with Anaconda.

This has been going on for a while, e.g.
https://discourse.julialang.org/t/how-does-one-set-up-a-centralized-julia-installation/13922/22?page=2

Ideally we'd have a more robust approach for having centrally installed packages, but of course most of Julia is used and developed on individual laptops and workstations where putting them in $HOME (on fast SSDs) isn't that much of an issue so I understand it's not a priority.

smoors:
@carstenbauer i've added back `JULIA_DEPOT_PATH` but appending as suggested by @bartoldeman

smoors:

smoors:

smoors (approved):
lgtm

smoors:
Going in, thanks @carstenbauer!

smoors:
this is not yet working as intended, fix in
- https://github.com/easybuilders/easybuild-easyconfigs/pull/19634

PR #19568, opened 2024-01-12
update Gaussian-16.A.03-AVX2.eb and  Gaussian-16.C.02-AVX2.eb

geoffreyweal:
The changes are as follows:

* easybuild/easyconfigs/g/Gaussian/Gaussian-16.A.03-AVX2.eb: Changes Gaussian 09 to Gaussian 16 in the comments. Only comments are changed in this file.
* easybuild/easyconfigs/g/Gaussian/Gaussian-16.C.02-AVX2.eb: This is a new file for installing Gaussian 16 version C.02 for AVX2-enabled CPUs. 

**Note** that the Gaussian tarball is a ```tbJ``` file (using the ```xz``` compression algorithm). Gaussian looks like it has moved from using the ```bzip2``` compressed```tbz``` format to ```xz``` compressed ```tbJ``` format (probably because Gaussian is distributed *via* DVD, so the tarball needs to be below 4.7 GB. The ```bzip2``` compressed ```tbz``` file is 7.2 Gb, while the ```xz``` compressed ```tbJ``` file is 4.4G).

To uncompress this tarball, easybuild will need to use the command ```tar -xJf 16.C.02-AVX2-E6B-432N.tbJ```. If the Gaussian-16.C.02-AVX2.eb file is to be accepted into the main branch, easybuild will need to be setup to allow for ```xz``` decompression of ```tbJ``` files using ```tar -xvJf 16.C.02-AVX2-E6B-432N.tbJ```

I have given the checksum (SHA256) for the 16.C.02-AVX2-E6B-432N.tbJ file (```a4f793e84819b8bf979f2e9d603945b54ee2874e9c496d0cefbe5331c7ede757```) in the Gaussian-16.C.02-AVX2.eb file.

I have tested that this script works for an ordinary ```bzip2``` compressed ```tbz``` file and it runs successfully and Gaussian works. The only thing that is different between a eb file for running Gaussian with a ```bzip2``` compressed ```tbz``` tarball and a ```xz``` compressed ```tbJ``` tarball is the checksum (SHA256) given, so hopefully this works if easybuild allows for decompression of ```xz``` compressed ```tbJ``` tarballs

**Summary of other changes that need to be made to easybuild**: Allow decompression of ```tbJ``` file using ```xz``` compression using command ```tar -xJf tarballname.tbJ``` 

Thanks


easybuild/easyconfigs/g/Gaussian/Gaussian-16.C.02-AVX2.eb
```diff
@@ -0,0 +1,65 @@
...
+"""
+
+toolchain = SYSTEM
+
+# Gaussian is licensed software.
+# You need to pick the tar file of your choice from the DVD,
+# put it in your sources tree with a proper name.
+# (should be a tree that ordinary users can't read due to license restrictions)
+# and change the name in "sources"
+sources = ['%(version)s%(versionsuffix)s-'+local_developmentid+'.tbJ']
```
jfgrimm, line 32: we can list the sources in a slightly different way so that we can specify the `extract_cmd` used to extract the sources:
```suggestion
sources = [{
    'filename': '%%(version)s%%(versionsuffix)s-%s.tbJ' % local_developmentid,
    'extract_cmd': 'tar xvfJ %s',
}]
```

easybuild/easyconfigs/g/Gaussian/Gaussian-16.C.02-AVX2.eb
```diff
@@ -0,0 +1,65 @@
...
+This is the official gaussian AVX2 build.
+"""
+
+toolchain = SYSTEM
+
+# Gaussian is licensed software.
+# You need to pick the tar file of your choice from the DVD,
+# put it in your sources tree with a proper name.
+# (should be a tree that ordinary users can't read due to license restrictions)
+# and change the name in "sources"
```
jfgrimm, line 31: we should use `download_instructions` for this, so that a nicer error is given when the sources are not found

jfgrimm (changes requested):
thanks for the PR, I've made a couple of suggestions

easybuild/easyconfigs/g/Gaussian/Gaussian-16.C.02-AVX2.eb
geoffreyweal: Excellent. Thanks for that edit!

easybuild/easyconfigs/g/Gaussian/Gaussian-16.C.02-AVX2.eb
geoffreyweal: Also, when I was installing G16 just now, Gaussian had a problem that when the g16 folder (which contains all the Gaussian files) was placed in the software folder, it had 777 permissions. This needed to be changed to 700 before Gaussian was happy to run. Is there a way in the eb file to get easybuild to change the permissions on ```.../easybuild/software/Gaussian/16.C.02-AVX2``` to prevent this issue arising when you first use Gaussian?

geoffreyweal: Also, thats cool I didn't realise there was a ```download_instructions``` entry. As I have just experienced all the things you have to do to install Gaussian recently, I am happy to write something for this.

geoffreyweal:
Also, when I was installing G16 just now, Gaussian had a problem that when the g16 folder (which contains all the Gaussian files) was placed in the software folder, it had 777 permissions. This needed to be changed to 700 before Gaussian was happy to run. Is there a way in the eb file to get easybuild to change the permissions on ```.../easybuild/software/Gaussian/16.C.02-AVX2``` to prevent this issue arising when you first use Gaussian? 

How would you fix this in the eb file?

easybuild/easyconfigs/g/Gaussian/Gaussian-16.C.02-AVX2.eb
geoffreyweal: I have just commited an update that includes a ```download_instructions```. Its a bit messy, so make changes as needed.

geoffreyweal (commented):
commited changes to ```sources``` and added a ```download_instructions``` section. The ```download_instructions``` section is a bit messy, so suggestions to make the python code for the ```download_instructions``` string nicer would be welcomed

easybuild/easyconfigs/g/Gaussian/Gaussian-16.C.02-AVX2.eb
```diff
@@ -0,0 +1,76 @@
...
+    'LD_LIBRARY_PATH': 'g%(version_major)s',
+}
+
+download_instructions = """
+    Gaussian is licensed software. You need to obtain a copy of %s %s%s. 
+    Once you have the DVD from Gaussian that contains the Gaussian installation files on it, 
+    copy the %s.tbJ file from the Gaussian DVD (GaussianDVD/tar/%s.tbJ) 
+    into the Gaussian source folder in EasyBuild (PATH_TO_EASYBUILD_FOLDER/sources/g/%s).
+    Make sure to change the name of the tbJ file in the sources folder to %s%s-%s.tbJ
+""" % (name, version, versionsuffix, local_developmentid, local_developmentid, name, version, versionsuffix, local_developmentid)
```
jfgrimm, line 71: I think using `.format()` is nicer here (really we want f-strings, but EasyBuild still supports Python 2 at the moment)
```suggestion
download_instructions = """
    Gaussian is licensed software. You need to obtain a copy of Gaussian {version}{versionsuffix}. 
    Once you have the DVD from Gaussian that contains the Gaussian installation files on it, 
    copy the {local_developmentid}.tbJ file from the Gaussian DVD (GaussianDVD/tar/{local_developmentid}.tbJ) 
    into the Gaussian source folder in EasyBuild (PATH_TO_EASYBUILD_FOLDER/sources/g/Gaussian).
    Make sure to change the name of the tbJ file in the sources folder to {version}{versionsuffix}-{local_developmentid}.tbJ
""".format(version=version, versionsuffix=versionsuffix, local_developmentid=local_developmentid)
```

geoffreyweal: Awesome. Thanks for that. Yes this looks much nicer!

jfgrimm:
> Also, when I was installing G16 just now, Gaussian had a problem that when the g16 folder (which contains all the Gaussian files) was placed in the software folder, it had 777 permissions. This needed to be changed to 700 before Gaussian was happy to run. Is there a way in the eb file to get easybuild to change the permissions on `.../easybuild/software/Gaussian/16.C.02-AVX2` to prevent this issue arising when you first use Gaussian?
> 
> How would you fix this in the eb file?

This isn't something that we currently do in the easyconfig. You should either add `--umask=007` to the command line when building Gaussian (as mentioned in a comment in the easyconfig), or add a `group` parameter to the easyconfig which will lock the software to a particular unix group.

geoffreyweal:
Oh right I was wondering what the ```--umask=007``` meant. Ok cool, I might play around with this later as I am very very very new to easybuild and dont want to break anything before properly testing.

PR #19591, opened 2024-01-16
{tools}[GCCcore/11.3.0] ROCm v5.6.0

akesandgren:
(created using `eb --new-pr`)

PR #19604, opened 2024-01-16
{bio}[foss/2022a] GROMACS v2023.3 w/ AdaptiveCpp 23.10.0 ROCm 5.6.0

MaximeVdB:
This adds easyconfigs for GROMACS v2023.3 with support for AMD GPUs via the SYCL backend (in the implementation provided by AdaptiveCpp). I went for a quite recent UCX version because the standard one for 2022a (v1.12.1) resulted in segfaults in GROMACS runs in "libMPI" mode using multiple GPUs.

Note: I just called it `GROMACS-2023.3-foss-2022a-ROCm-5.6.0.eb` because using AdaptiveCpp's SYCL implementation would be the regular way of building GROMACS for AMD GPUs (AMD provies a HIP port but that version is not officially supported). But please tell if you judge that e.g.  `GROMACS-2023.3-foss-2022a-ROCm-5.6.0-AdaptiveCpp-23.10.0.eb` would be more appropriate.

I've tested this on the arcturus partition on Vaughan (with MI100 GPUs) and got quite decent single-GPU performance on the `stmv` benchmark (~80% of the speed of AMD's HIP port of GROMACS). Using multiple GPUs for that benchmark brings less benefit than expected, but this is presumably related to the [incomplete support for direct GPU-GPU communication](https://gitlab.com/gromacs/gromacs/-/issues/4372) in the SYCL backend of GROMACS.

Depends on:

- https://github.com/easybuilders/easybuild-easyblocks/pull/3077
- https://github.com/easybuilders/easybuild-easyconfigs/pull/19591 (and all its dependencies, of course :)

### DIFF
--- a/easybuild/easyconfigs/a/AdaptiveCpp/AdaptiveCpp-23.10.0-GCC-11.3.0-ROCm-5.6.0.eb
+++ b/easybuild/easyconfigs/a/AdaptiveCpp/AdaptiveCpp-23.10.0-GCC-11.3.0-ROCm-5.6.0.eb
@@ -1,0 +1,88 @@
+easyblock = 'CMakeMake'
+
+name = 'AdaptiveCpp'
+version = '23.10.0'
+local_rocmver = '5.6.0'
+versionsuffix = '-ROCm-%s' % local_rocmver
+
+homepage = 'https://adaptivecpp.github.io/'
+
+whatis = [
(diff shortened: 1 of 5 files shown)


sassy-crick:
@MaximeVdB You might want to sync your PR with development, something like this: 
```
```

boegel:

PR #19626, opened 2024-01-18
{tools}[system] Micromamba v1.5.6-0

dirkpetersen:
right now a patch is needed to work around these issues: 
- https://github.com/easybuilders/easybuild-easyblocks/pull/3082


boegel:
No follow up in required eaysblock PR (https://github.com/easybuilders/easybuild-easyblocks/pull/3082) since March 2024, so closing...

PR #19398, opened 2023-12-11
{tools}[foss/2023a] pyiron v0.5.1 w/ Python 3.11.3

lcniel:
(created using `eb --new-pr`)

ETA: I have limited time to work on this and did not expect it to be as involved as it turned out to be, it should ideally be split up and merged as separate PR:s. Currently my available time to work on splitting this up is very limited as software building is only one aspect of a position I currently work one day a week at.

PR #19705, opened 2024-01-24
TEST PR ONLY! Pyiron

Flamefire:


Flamefire:
See https://github.com/easybuilders/easybuild-easyconfigs/pull/19398 & https://github.com/lcniel/easybuild-easyconfigs/pull/2 for the real PR

PR #19736, opened 2024-01-27
remove `easybuild_version = '3.5.0'` from R easyconfigs

branfosj:
This removes
```
# specify that at least EasyBuild v3.5.0 is required,
# since we rely on the updated easyblock for R to configure correctly w.r.t. BLAS/LAPACK
easybuild_version = '3.5.0'
```
from all not-archived R easyconfigs in the `5.0.x` branch. I cannot see that having this set serves any useful purpose in EasyBuild 5.


boegel (approved):
lgtm

PR #19620, opened 2024-01-17
{lang}[system/system] Java v21, Java v21.0.1

SimonPinches:
(created using `eb --new-pr`)

PR #19880, opened 2024-02-16
{tools}[gfbf/2023b] JPype v1.5.0

SimonPinches:
(created using `eb --new-pr`)

This PR depends upon
- [x] #19620


SimonPinches:
Depends upon PR #19620.

Could also be renamed to `Jpype` rather that `Jpype1` (which is the name still used on PyPi) which would seem more natural.

sassy-crick:
@SimonPinches You might want to sync your PR with development, something like this: 
```
```

easybuild/easyconfigs/j/JPype/JPype-1.5.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,28 @@
...
+
+homepage = 'http://www.jpype.org'
+description = "A Python module to provide full access to Java from within Python"
+
+toolchain = {'name': 'gfbf', 'version': '2023b'}
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+    ('Java', '21', '', SYSTEM),
```
boegel, line 14: @SimonPinches Does this really need `Java/21`?

For the `2023b` generation, we usually stick with `Java/11`...

SimonPinches: @boegel Some of our applications need `Java/21` as remarked here in PR #20945.  It is also the recommended LTS version (see https://whichjdk.com/).

boegel: We can make an exception to this, but you're swimming upstream a bit for this generation of easyconfigs...

SimonPinches: Thanks. I've added an exception which shouldn't be needed when we move to using the new 2025b toolchains :-)

akesandgren:
@boegelbot Please test @ jsc-zen3

akesandgren (approved):
LGTM

akesandgren:

PR #19881, opened 2024-02-17
detect easyconfigs where missing information about downloading sources

branfosj:
closes #19790 

I'll be PR-ing adding `download_instructions` to easyconfigs that this detects.

* #19882
* #19889
* #19890
* #19935
* #19905
* #19904
* #19903
* #19885
* #19931
* #19907
* #19932
* #19902
* #19906
* #19908
* #19909
* #19934
* #19910
* #19926
* #19927
* #19928
* #19933
* #19911
* #19888
* #19936 and #19937
* #19924
* #19920
* #19923
* #19929
* #19919
* #19912
* #19918
*  #19900
* #19887
* #19901
* #19883
* #19899
* #19898
* #19884
* #19897
* #19896
* #19895
* #19894

PR #19890, opened 2024-02-17
add `download_instructions` to AGeNT easyconfigs

branfosj:
for #19881


Micket (approved):
lgtm 

edit: I'll just verify that the download message is what it should be, i can't build all of these, so i'll just skip the build reports.

PR #19899, opened 2024-02-18
improve and add `download_instructions` to SignalP easyconfigs

branfosj:
for https://github.com/easybuilders/easybuild-easyconfigs/pull/19881


Micket (approved):
lgtm

PR #19901, opened 2024-02-18
add `download_instructions` to PEAR easyconfigs

branfosj:
for https://github.com/easybuilders/easybuild-easyconfigs/pull/19881

This copies in the instructions from the latest easyconfig (`PEAR-0.9.11-GCC-11.3.0.eb`) which uses a more direct link to the download page.


Micket (approved):
lgtm

PR #19911, opened 2024-02-18
add `download_instructions` to Maple easyconfigs

branfosj:
for #19881


Micket (changes requested):
Also update homepage to https, page redirects to
https://www.maplesoft.com/products/maple/

Micket (approved):
lgtm

PR #9739, opened 2020-01-20
{bio}[system/system] netMHC-4.0a: fix postinstallcmds and clarify download instructions

lexming:
(created using `eb --new-pr`)

PR #19912, opened 2024-02-18
add `download_instructions` to netMHC and netMHCII easyconfigs

branfosj:
for #19881

### DIFF
--- a/easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
+++ b/easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
@@ -9,8 +9,6 @@ description = """NetMHC 4.0 predicts binding of peptides to a number of differen
 
 toolchain = SYSTEM
 
-# netMHC is proprietary software, but free for academics.
-# It can be requested at https://www.cbs.dtu.dk/cgi-bin/nph-sw_request?netMHC
 sources = [
     {
         'filename': '%(name)s-%(version)s.Linux.tar.gz',
@@ -25,6 +23,8 @@ checksums = [
     'd9938388c50bf49d748e5c1293435d7d9fae5d0a33303d795e767429508e368a',  # netMHC-4.0a.Linux.tar.gz
     'f3fb6d77ec11bc51de963e96fd75a0622f5098a2d1e37d750afa3709d5b876d1',  # netMHC-4.0a-data.tar.gz
 ]
+download_instructions = f"""{name} is proprietary software, but free for academics.
+It can be requested at https://www.cbs.dtu.dk/cgi-bin/nph-sw_request?netMHC"""
 
 keepsymlinks = True
 
--- a/easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
+++ b/easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
@@ -3,7 +3,7 @@ easyblock = 'Tarball'
 name = 'netMHC'
 version = '4.0a'
 
-homepage = 'https://www.cbs.dtu.dk/services/NetMHC/'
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
```diff
@@ -25,6 +23,8 @@ checksums = [
     'd9938388c50bf49d748e5c1293435d7d9fae5d0a33303d795e767429508e368a',  # netMHC-4.0a.Linux.tar.gz
     'f3fb6d77ec11bc51de963e96fd75a0622f5098a2d1e37d750afa3709d5b876d1',  # netMHC-4.0a-data.tar.gz
 ]
+download_instructions = f"""{name} is proprietary software, but free for academics.
+It can be requested at https://www.cbs.dtu.dk/cgi-bin/nph-sw_request?netMHC"""
```
Micket, line 27: This easyconfig confuses me. 
What's going on with download_filename "data.tar.gz" but the filename is different? I suppose this should be mentioned clearly here.

easybuild/easyconfigs/n/netMHCII/netMHCII-2.3.eb
```diff
@@ -27,6 +25,8 @@ checksums = [
     '8b184b2463b93bbcbb93a7351c6c0bae12d82c7827c451ebc5153b257dd0be86',  # netMHCII-2.3.Linux.tar.gz
     '6bf8af9e2e89d96b8fe9a2752164606417c4583872bdc2e81637f6668dee30a0',  # netMHCII-2.3-data.Linux.tar.gz
 ]
+download_instructions = f"""{name} is proprietary software, but free for academics.
+It can be requested at https://www.cbs.dtu.dk/cgi-bin/nph-sw_request?netMHCII"""
```
Micket, line 29: Same with data.Linux.tar.gz here.

easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
jfgrimm: the link doesn't work for me anymore, looks like it should be `https://services.healthtech.dtu.dk/cgi-bin/sw_request?software=netMHC&version=4.0&packageversion=4.0a&platform=Linux` instead

jfgrimm: regarding `data.tar.gz`, that looks to be downloadable from: `https://services.healthtech.dtu.dk/services/NetMHC-4.0/data.tar.gz`
I guess the confusion was caused by https://github.com/easybuilders/easybuild-easyconfigs/pull/9739 removing the `source_urls`, which means `data.tar.gz` is no longer auto-downloaded

easybuild/easyconfigs/n/netMHCII/netMHCII-2.3.eb
jfgrimm: same comments apply here as for `netMHC`

branfosj: f3fb169

easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
branfosj: I think f3fb169 fixes this. Once you have the netMHC/netMHCII source then the data file will download directly.

jfgrimm (approved):
lgtm

easybuild/easyconfigs/n/netMHC/netMHC-4.0a.eb
jfgrimm: yes, that works

PR #19937, opened 2024-02-19
archive oldest MATLAB easyconfigs

branfosj:
for #19881

These are both from the old style of unpacking DVDs and then creating a tarball from them.


Micket (approved):
lgtm

PR #19978, opened 2024-02-27
{tools}[foss/2023a] SentencePiece v0.2.0, gperftools v2.15, huggingface-hub v0.20.3, ... w/ PyTorch 2.1.2 CUDA 12.1.1

NPotvin:
(created using `eb --new-pr`)


easybuild/easyconfigs/t/torchaudio/torchaudio-2.2.0-foss-2023a-PyTorch-2.1.2-CUDA-12.1.1.eb
```diff
@@ -0,0 +1,49 @@
+easyblock = 'PythonPackage'
```
Micket, line 1: Rather than the individual torch* packages, i think we want to use a PyTorch-bundle PR. 

There were several conflicting PRs open right now with torch related stuff. I tried to unify the approach into this PR (just the PyTorch bundle part)
https://github.com/easybuilders/easybuild-easyconfigs/pull/19987

PR #19985, opened 2024-02-27
use Python version of Z3 as dependency for Clang 16.0.6

pavelToman:
Without python version it fails to load either the Clang and PyTorch - has same version of Z3 but with python version.


pavelToman:

lexming:
Following the discussion in Slack, we took the opposite approach. Alternative fix in https://github.com/easybuilders/easybuild-easyconfigs/pull/20050

PR #20001, opened 2024-02-29
Fix installation of packages in Python for GCCcore 8.2.0

Flamefire:
(created using `eb --new-pr`)

PR #20006, opened 2024-03-01
migrate archive of easyconfigs to easybuild-easyconfigs-archive

lexming:
see https://github.com/easybuilders/easybuild-easyconfigs-archive/pull/2


Micket:
Hm, how can there be conflicts here. There shouldn't have been anything touching the archive right?

lexming:
@Micket no idea what happened, I just rebased from current `5.0.x`

Micket:
@branfosj do you think we should decline #20001 as to avoid troublesome merge conflicts later?

boegel (commented):
lgtm

boegel (approved):

PR #20032, opened 2024-03-05
{vis}[GCCcore/13.2.0] Vulkan v1.3.279

Micket:
(created using `eb --new-pr`)

I was hopeful, but i'm still not getting anything useful from a vulkaninfo and vkcube just segfaults.


sassy-crick:
Native build on RHEL-8, within my userspace works but when I run `vulkaninfo` I get:
```
$ vulkaninfo
WARNING: [Loader Message] Code 0 : terminator_CreateInstance: Received return code -3 from call to vkCreateInstance in ICD /usr/lib64/libvulkan_lvp.so. Skipping this driver.
ERROR: [Loader Message] Code 0 : setup_loader_term_phys_devs:  Failed to detect any valid GPUs in the current config
ERROR at /dev/shm/jsassman/Vulkan/1.3.279/GCCcore-13.2.0/Vulkan-Tools-1.3.279/vulkaninfo/./vulkaninfo.h:245:vkEnumeratePhysicalDevices failed with ERROR_INITIALIZATION_FAILED

$ ls /usr/lib64/libvulkan_lvp.so 
/usr/lib64/libvulkan_lvp.so
```
For me, it looks like it is falling back to the OS provided library. Is that the correct behaviour?

Micket:
Probably depends on what you have installed on the OS. With e.g. EGL we still have it (intentionally) pick up nvidia drivers and use that. 

But the little i knew about graphics setup gets thrown out the window with Vulkan, as it's not only a new API for rendering, but also does away with all the old GL-context things and replaces it with their own window integration system (WSI), which i know even less about.

Micket:
With Mesa Vulkan support, software rendering works fine. Maybe some day we can figure out how/if hardware rendering can also work, but I have no success with that even with OS installed libraries yet over VNC, so, don't know what to fix as even just vulkan cube demo crashes. I proclaim this ready as is.

Micket:

Micket:

Micket:
I'm just not doing to spend time debugging this. The newer version I have gotten things to work with mesa 25, so i'm just going with that.

PR #20105, opened 2024-03-13
Added QuantumESPRESSO easyconfigs for 7.3 (both intel and foss)

Crivella:
Added easyconfig files for QE 7.3 working with new easyblock from [PR-3241](https://github.com/easybuilders/easybuild-easyblocks/pull/3241)

Also ran some performance testing on a workstations with an intel i9-13900k.
After taking into account the presence of both performance and efficiency cores, and running all tests with the same pinning for reproducibility, both the `-O3` and `-ffast-math` do not give any visible improvement at the scale this tests are being ran.

In order to properly compare the efficiency with respect to the EESSI version (compiled with older easyblock), a multi-node heavy run would be needed as what has been fixed is mostly related to Scalapack and ELPA libraries usage.

image

The run reported are generated using the ReFrame test from [PR-3134](https://github.com/reframe-hpc/reframe/pull/3134) with `ecut=100` and `nbnd=100`

PR #20138, opened 2024-03-18
{chem} [foss/2023a + intel/2023a] QuantumESPRESSO 7.3.1  with cmake

Crivella:
These are the easyconfigs for QE 7.3.1 
- [x] depending on the new [CMake easyblock from this PR](https://github.com/easybuilders/easybuild-easyblocks/pull/3257).
- [x] https://github.com/easybuilders/easybuild-easyblocks/pull/3338
While for deployment on EESSI it makes more sense to use the 7.3 with Configure/Make in order to have OpenMP enabled, i am also adding the CMake version for future reference (7.4) with several problem related to standard and plugin compilation being worked on [see comment on issue](https://gitlab.com/eessi/support/-/issues/55#note_1817215763).


Crivella:
One thing that should be tested on multiple architecture for the intel version is running with only 1 core.
In my case i was getting segmentation faults due to the implementation of OFI, which i was able to solve by setting `I_MPI_FABRICS=shm mpirun` from the default `tcp:ofi`

`shm` should run on intra-node jobs only and while i think that using `tcp` fabric with `mpirun -n 1` is not the best usecase, it is still possible that users will run an heavy multinode calculation followed by many single core postprocessing in the same job, so I believe this is a problem worth investigating further

casparvl:
@ocaisa is this one you can also review, since you're also looking into https://github.com/easybuilders/easybuild-easyconfigs/pull/20105 ?

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.3.1-intel-2023a.eb
```diff
@@ -0,0 +1,58 @@
...
+    'usempi': True,
+    'openmp': False # QE not linking FFTW and ELPA correctly with OpenMP + cmake
+    }
+
+sources = [
+    {
+        'filename': 'q-e-qe-%(version)s.tar.gz',
+        'extract_cmd': 'mkdir -p %(builddir)s/qe-%(version)s && tar xzvf %s --strip-components=1 -C $_',
+        'source_urls': ['https://gitlab.com/QEF/q-e/-/archive/qe-%(version)s']
+    },
```
Micket, line 24: Do we not want the gipaw or wannier90 stuff anymore? I don't know much about QE, just comparing with older easyconfigs.

Crivella: With the `cmake` build recipe, those are pulled by QE as git submodules. This was actually already happening also for other components since 7.0.

Unfortunately letting EB handle the download of the sources is not trivial.
The procedure available for fully offline build (https://gitlab.com/QEF/q-e/-/wikis/Support/Build-without-internet) delegates the download to the make script by pulling in the files as git submodules.

Another possible way would be to pull the sources and also create a `.git` folder to trick the [initialization script](https://gitlab.com/QEF/q-e/-/blob/develop/external/initialize_external_repos.sh?ref_type=heads) to think that the submodules were already pulled

I think this should be discussed and also decide if it is worth implementing

Crivella: Having an actual option in the cmake would definitely be the more viable and long term solution.

I don't think deleting the file would work. Since we are building from a tarball the `IS_GIT_PROJECT` variable will be false
https://gitlab.com/QEF/q-e/-/blob/develop/CMakeLists.txt?ref_type=heads#L399
which ends up leading us to this branch in the `qe_git_submodule_update` function
https://gitlab.com/QEF/q-e/-/blob/develop/cmake/qeHelpers.cmake?ref_type=heads#L135
checking for the `.git` folder and raising an error if the hash file does not exists
https://gitlab.com/QEF/q-e/-/blob/develop/cmake/qeHelpers.cmake?ref_type=heads#L173

I will test this possible implementation and get back at you

On a side note, (not sure it is worth fixing it there as well) this was already happening with the previous config/make since 7.0 where QE moved some of its functionalities to separate libs `mbd` `lapack` (only if internal) and `devxlib`

Crivella: Tested the files in the [last commit](https://github.com/easybuilders/easybuild-easyconfigs/pull/20138/commits/d6889a4eaa512d5d3e9887392f067fab97883afc) with the modifications in the latest [eb commit](https://github.com/Crivella/easybuild-easyblocks/commit/b3759f5458f20959afb0d6d48bc4995462c52589).

The output from `cmake` in the log shows that the downloads from QE are being skipped.

```
...
-- Parallel HDF5 enabled!
-- Installing Wannier90 via submodule
-- Previous clone found at /home/crivella/.local/easybuild/build/QuantumESPRESSO/7.3.1/foss-2023a/qe-7.3.1/external/wannier90.
-- Installing MBD via submodule
-- Previous clone found at /home/crivella/.local/easybuild/build/QuantumESPRESSO/7.3.1/foss-2023a/qe-7.3.1/external/mbd.
-- Installing DeviceXlib via submodule
-- Previous clone found at /home/crivella/.local/easybuild/build/QuantumESPRESSO/7.3.1/foss-2023a/qe-7.3.1/external/devxlib.
-- Installing D3Q via submodule
-- Previous clone found at /home/crivella/.local/easybuild/build/QuantumESPRESSO/7.3.1/foss-2023a/qe-7.3.1/external/d3q.
-- Installing pw2qmcpack via submodule
-- Previous clone found at /home/crivella/.local/easybuild/build/QuantumESPRESSO/7.3.1/foss-2023a/qe-7.3.1/external/pw2qmcpack.
...
```

Since plugins are not tested in the test-suite, i did run a few of the examples for d3q and pw2qmcpack manually and they work fine (gipaw still disabled as there is no fix to make it work with 7.3.1 yet).

Considering that the supported solution for offline installation (https://gitlab.com/QEF/q-e/-/wikis/Support/Build-without-internet) uses git to pull the submodules and also checks the existance of the `.git` folder for skipping (same as in the non git-clone version of QE), i think this should be a reliable check (most likely it won't be changed by the devs in the near future)

@Micket Let me know if this could work

Crivella:
On a side node also working on a fix to make OpenMP work, was able to have it link to `fftw3f_omp` and `fftw3f`  want to check if QE should also be able to link with `fftw3_mpi`

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3257"

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3257"

ocaisa:

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.3.1-foss-2023a.eb
```diff
@@ -0,0 +1,153 @@
...
+    # {'d3q-%s.tar.gz' % local_d3q_hash: '43e50753a56af05d181b859d3e29d842fb3fc4352f00cb7fe229a435a1f20c31'},
+    # {'fox-%s.tar.gz' % local_fox_hash: '99b6a899a3f947d7763aa318e86f9f08db684568bfdcd293f3318bee9d7f1948'},
+    # {'qe-gipaw-%s.tar.gz' % local_qe_gipaw_hash: '9ac8314363d29cc2f1ce85abd8f26c1a3ae311d54f6e6034d656442dd101c928'},
+    # {'pw2qmcpack-%s.tar.gz' % local_qmcpack_hash: 'a8136da8429fc49ab560ef7356cd6f0a2714dfbb137baff7961f46dfe32061eb'},
+    # {'wannier90-%s.tar.gz' % local_w90_hash: 'f989497790ec9777bdc159945bbf42156edb7268011f972874dec67dd4f58658'},
+]
+
+builddependencies = [
+    ("M4", "1.4.19"),
+    ("CMake", "3.26.3"),
```
ocaisa, line 122:
```suggestion
    ("CMake", "3.26.3"),
    ("pkgconf", "1.9.5"),
```

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.3.1-intel-2023a.eb
```diff
@@ -0,0 +1,148 @@
...
+    # {'d3q-%s.tar.gz' % local_d3q_hash: '43e50753a56af05d181b859d3e29d842fb3fc4352f00cb7fe229a435a1f20c31'},
+    # {'fox-%s.tar.gz' % local_fox_hash: '99b6a899a3f947d7763aa318e86f9f08db684568bfdcd293f3318bee9d7f1948'},
+    # {'qe-gipaw-%s.tar.gz' % local_qe_gipaw_hash: '9ac8314363d29cc2f1ce85abd8f26c1a3ae311d54f6e6034d656442dd101c928'},
+    # {'pw2qmcpack-%s.tar.gz' % local_qmcpack_hash: 'a8136da8429fc49ab560ef7356cd6f0a2714dfbb137baff7961f46dfe32061eb'},
+    # {'wannier90-%s.tar.gz' % local_w90_hash: 'f989497790ec9777bdc159945bbf42156edb7268011f972874dec67dd4f58658'},
+]
+
+builddependencies = [
+    ('M4', '1.4.19'),
+    ('CMake', '3.26.3'),
```
ocaisa, line 121:
```suggestion
    ('CMake', '3.26.3'),
    ("pkgconf", "1.9.5"),
```

ocaisa (changes requested):
I noticed in the logs that the configure is picking up on `pkgconf` from the host

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.3.1-intel-2023a.eb
ocaisa: This has been implemented, and is also ready for checksums when git repo tarballs have a consistent hash in EB 5.0

ocaisa:

ocaisa:

ocaisa (approved):

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.3.1-foss-2023a.eb
```diff
@@ -0,0 +1,154 @@
...
+            "commit": local_w90_hash,
+        },
+    },
+]
+checksums = [
+    {'q-e-qe-7.3.1.tar.gz': '2c58b8fadfe4177de5a8b69eba447db5e623420b070dea6fd26c1533b081d844'},
+    # Holding off checksum checks untill 5.0.x
+    # https://github.com/easybuilders/easybuild-framework/pull/4248
+
+    # {'lapack-%s.tar.gz' % local_lapack_hash: 'c05532ae0e5fe35f473206dda12970da5f2e2214620487d71837ddcf0ea6b21d'},
```
ocaisa, line 110: @Crivella CI is failing because we should just be retaining the entries, but using a `None` value for the checksum values (same is true for all the rest as well)

ocaisa (changes requested):

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.3.1-foss-2023a.eb
Crivella: @ocaisa Is there a standard way to mix `dict` checksums with None?

Tried with 
```
[
    { XXX: YYY },
    { ZZZ: None},
]
```

which fails the typechecking
and than
```
[
    { XXX: YYY },
   None,
]
```

which ends up converting the checksum string into a list of single characters due to https://github.com/easybuilders/easybuild-framework/blob/65591dd32a78850144ce39238aa577df6fab3f9f/easybuild/framework/easyconfig/types.py#L528
(Not sure if this is a known bug or if i should open an issue)

If i try (to get around the previous bug)
```
[
    { XXX: [YYY] },
   None,
]
```

then i hit again a typechecking failure

ocaisa:
@Crivella I'll tweak these to use https://github.com/easybuilders/easybuild-easyblocks/pull/3306 (but they will then be tied to that PR)

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3306"

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3306"

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.3.1-foss-2023a.eb
Crivella: Until we re-add the checksums I have convert them into a list without dicts [0781213](https://github.com/easybuilders/easybuild-easyconfigs/pull/20138/commits/07812133646be2c42d59d8f7b045258cd4a1e11e), this should work even without fixing the previously mentioned behavior

Crivella:
The latest commit requires a change to the EB that will be implemented once https://github.com/easybuilders/easybuild-easyblocks/pull/3306 is merged

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3338"

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3338"

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3338"

ocaisa:
EB_ARGS="--include-easyblocks-from-pr 3338"

ocaisa (approved):
LGTM

ocaisa:
Closing and reopening to trigger tests

PR #20146, opened 2024-03-19
{perf}[gompi/2023b] Score-P v8.4 w/ CUDA 12.4.0

Flamefire:
(created using `eb --new-pr`)

Requires:
- [x]  https://github.com/easybuilders/easybuild-easyconfigs/pull/19832
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/20107

If the build fails with

> configure: WARNING: nvcc -ccbin mpicxx compilation failed. Disabling CUDA support.

the cause is very likely a missing `include-fixed` folder for GCC 13, see https://github.com/easybuilders/easybuild-easyblocks/pull/3254. So either rename the include-fixed folder renamed by easybuild or regenerate it with `$EBROOTGCC/libexec/gcc/x86_64-pc-linux-gnu/13.2.0/install-tools/mkheaders`

PR #20158, opened 2024-03-20
Remove UCX-CUDA with incompatible CUDA-GCC-combination

Flamefire:
CUDA 12.3.0 is not compatible with GCC 13 hence having this EC only causes confusion.

See https://github.com/easybuilders/easybuild-easyconfigs/pull/20146#issuecomment-2007845694


Micket (approved):
lgtm

PR #20173, opened 2024-03-20
Fix duplicate dict entries

Flamefire:
This enables fatal error checking for ECs via `flake8` which includes unknown/undefined variables used and other serious issues.

Those found in current ECs are fixed here. Mostly duplicate dict keys which are obviously a mistake.

I kept the last entry each which is what python would use


casparvl (approved):
Lgtm. All changes are for EasyConfigs for which the toolchains are deprecated. That makes it difficult, but also less crucial to test. I'll force a merge on this one.

PR #19987, opened 2024-02-28
{lib}[GCCcore/12.3.0] PyTorch-bundle v2.1.2, SoX v14.4.2, libmad v0.15.1b, ... w/ CUDA 12.1.1

Micket:
(created using `eb --new-pr`)

I have combined what I found from several conflicting PRs here to make what I think we want to have as the PyTorch-bundle. 

I made some choices here:

1. Stole some from other PRs to combine into this one
2. Using SentencePiece with the existing gperftools instead of introducing a duplicate
3. Verified torchvision and torchtext was using supported versions for PyTorch 2.1
4. Added torchaudio
5. kept torchdata despite it being abandoned (since it was only just abandoned relatively recently)

(I don't really mind how this gets merged, if someone wants to update their PR instead, feel free to close this or just use it as a reference).

PR #20175, opened 2024-03-21
CUDA-enabled torchvision/v0.16.0

moravveji:
This PR adds a new easyconfig file for building `torchvision/0.16.0-foss-2023a-CUDA-12.1.1` and complements a similar easyconfig file for a CPU-only build.


easybuild/easyconfigs/p/PyTorch/PyTorch-2.1.2-foss-2023a-CUDA-12.1.1.eb
```diff
@@ -171,7 +171,7 @@ dependencies = [
     ('expecttest', '0.1.5'),
     ('networkx', '3.1'),
     ('sympy', '1.12'),
-    ('Z3', '4.12.2'),
+    ('Z3', '4.12.2', '-Python-%(pyver)s', ('GCCcore', '12.3.0')),
```
casparvl, line 174: This was taken out on purpose, see https://github.com/easybuilders/easybuild-easyconfigs/pull/20050
We now have _one_ `Z3` which has Python support. Ie. no more build with/without suffix. Please undo this change. You can then build your `torchvision` against PR 20050.

VRehnberg:
This will probably build without JPEG support and is probably not running the torchvision test-suite. Compare against https://github.com/easybuilders/easybuild-easyconfigs/pull/19987/

Micket (changes requested):
I strongly propose we shouldn't use these at all, but use the PyTorch-bundle's instead.

PR #20205, opened 2024-03-25
Improve test for validity of easyconfig files

Flamefire:
Aggregate failures and report all together.
This is especially useful when the same error is found multiple times, e.g. multiple patch files are missing.
This avoids multiple back-and-forth between CI and user updating the PR.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -1362,7 +1362,7 @@ def template_easyconfig_test(self, spec):
         # cache the parsed easyconfig, to avoid that it is parsed again
         EasyConfigTest._parsed_easyconfigs.append(ecs[0])
     else:
-        self.assertTrue(False, "easyconfig %s does not contain blocks, yields only one parsed easyconfig" % spec)
+        self.fail("easyconfig %s does not contain blocks, yields only one parsed easyconfig" % spec)
 
     # check easyconfig file name
     expected_fn = '%s-%s.eb' % (ec['name'], det_full_ec_version(ec))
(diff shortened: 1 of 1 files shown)


Flamefire:
> lgtm. I would like to test this though, and i'm not sure how. Could you open a test PR with this change + some easyconfigs with multiple errors?

#20221 
> AssertionError: Verification for FFTW-3.3.9-intel-2021a.eb failed:
Found use of '$root', not compatible with modules in Lua syntax, use '%(installdir)s' instead: ["runtest = 'check $root'"]
There should be no redefined easyconfig parameters, found 1: runtest (2)
Old URL 'hpcugent.github.io/easybuild' found
Patch file /home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/easybuild/easyconfigs/f/FFTW/this.patch is missing
Patch file /home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/easybuild/easyconfigs/f/FFTW/that.patch is missing
'source' step should not be skipped, since that implies not verifying checksums

Micket:
Yes I'm happy with this change I just wanted to check with other maintainers first; we used to have boegelbot write error summary as a message, which I suspect would be impacted by this. Currently this feature seems to have been broken for some time now (I haven't seen such comments in a while now), but I still wanted to check first.

Flamefire:
> Yes I'm happy with this change I just wanted to check with other maintainers first; we used to have boegelbot write error summary as a message, which I suspect would be impacted by this. Currently this feature seems to have been broken for some time now (I haven't seen such comments in a while now), but I still wanted to check first.

I would expect those summaries being just picked up as before and hence be even more helpful. But yes, haven't seen them for a while either.

PR #20221, opened 2024-03-26
Trigger error in FFTW-3.3.9-intel-2021a.eb

Flamefire:
Test for #20205


Flamefire:
@Micket This is the one you request in #20205

PR #20067, opened 2024-03-07
{bio}[foss/2023a] scCODA v0.1.9, rpy2 v3.5.15, scikit-bio v0.5.9, ...

pavelToman:
(created using `eb --new-pr`)

PR #20300, opened 2024-04-05
replace incorrect scikit-bio 0.5.9 with scikit-bio 0.6.0 as dependency for scCODA

boegel:
The easyconfig for `scikit-bio` 0.5.9 that got merged via #20067 was using a commit ID that doesn't correspond to the 0.5.9 release at all (commit `7565847` is from 2024-03-05, while `0.5.9` release is from 2023-08-03).

This was done because `scikit-bio` 0.5.9 required a `scipy` version <= 1.10.1, which is older than what we have in SciPy-bundle 2023.07 (scipy 1.11.1).

`scikit-bio` 0.6.0 was not released yet at the time of PR #20067, but it is now (2024-03-28), so we should rectify this...
Going with 0.6.0 makes most sense, since the commit that was used is quite close to that version already (was closer than it was to 0.5.9).

The easyconfig for `biom-format` is there because `scikit-bio` 0.6.0 now requires it (it didn't yet in v0.5.9)

### DIFF
--- a/easybuild/easyconfigs/b/biom-format/biom-format-2.1.15-foss-2023a.eb
+++ b/easybuild/easyconfigs/b/biom-format/biom-format-2.1.15-foss-2023a.eb
@@ -1,0 +1,48 @@
+##
+# This is a contribution from DeepThought HPC Service, Flinders University, Adelaide, Australia
+# Homepage:     https://staff.flinders.edu.au/research/deep-thought
+#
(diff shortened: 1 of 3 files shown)


boegel:

boegel:

Micket (approved):
lgtm

Micket:
Going in, thanks @boegel!

PR #20324, opened 2024-04-09
change arrow-R dependency of Bioconductor v3.18 to v14.0.1 (which depends on required matching Arrow v14.0.1)

bedroge:
Bioconductor had a dependency on arrow-R 14.0.0.2, which depends on Arrow 14.0.1. However, those versions are not compatible, as the build log shows:
```
**** Not using: C++ library version (14.0.1) does not match R package (14.0.0.2)
```
It will then try to build its own version of Arrow. The script used by the build to check this has the following comment:
```
    # Patch releases we do for CRAN feedback get an extra x.y.z.1 version.
    # These should work with the x.y.z C++ library (which never has .1 added)
```

There are no CRAN source tarballs for arrow-R 14.0.1.x though (see https://cran.r-project.org/src/contrib/Archive/arrow/), so I fixed it for this version by taking the full Arrow source tarball for version 14.0.1, which contains everything (C++/R/Python sources).

I also modified Bioconductor, so that it now uses this version of arrow-R. I couldn't find any other packages that depend on arrow-R.


bedroge:
I've also added the same comment about compatible Arrow versions to some older arrow-R easyconfigs to prevent this from happening again. I'm just not sure what to do with `arrow-R-14.0.0.2-foss-2023a-R-4.3.2.eb`, as it violates this and is somewhat broken (it will build its own Arrow).

verdurin:
Test report by @verdurin
**SUCCESS**
easybuild-c7.novalocal - Linux CentOS Linux 7.9.2009, x86_64, Intel Xeon Processor (Skylake, IBRS), Python 3.6.8
See https://gist.github.com/verdurin/b844665e7a17baec2b9a1c9b826aaca9 for a full test report.

verdurin:

verdurin:

easybuild/easyconfigs/a/arrow-R/arrow-R-14.0.0.2-foss-2023a-R-4.3.2.eb
```diff
@@ -20,7 +20,7 @@ checksums = ['7138a52d66f1b94ec31c25e8929d6f92b1640df852a10817600a82ab68ba8ab7']
 dependencies = [
     ('R', '4.3.2'),
     ('R-bundle-CRAN', '2023.12'),
-    ('Arrow', '14.0.1'),
+    ('Arrow', '14.0.1'),  # arrow-R x.y.z[.N] only works with Arrow x.y.z
```
boegel, line 23: @bedroge Doesn't this imply that this `arrow-R` easyconfig should be removed?

bedroge: There is definitely no good use for it anymore, but I wasn't sure what the policy is for removing easyconfigs. But since it sounds like it's fine to remove it, I've done that in https://github.com/easybuilders/easybuild-easyconfigs/pull/20324/commits/ed616365aa5c904b0807e33188cd994e1b4dffcf.

boegel: If there's a good reason to remove it (like there is here), we should, even if it comes with fallout.

It's pretty limited here, since `arrow-R` is not a common dependency at all.

boegel:

boegel:
Test report by @boegel
**SUCCESS**
node3113.skitty.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/boegel/d5da50c58811df5c5bc99869ce509113 for a full test report.

boegel (approved):
lgtm

PR #20365, opened 2024-04-16
ObsPy 1.4.0

grpiccoli:
SQLAlchemy installer is missing typing-extensions dependency

### DIFF
--- a/easybuild/easyconfigs/o/ObsPy/ObsPy-1.4.0-intel-2023a-Python-3.11.3.eb
+++ b/easybuild/easyconfigs/o/ObsPy/ObsPy-1.4.0-intel-2023a-Python-3.11.3.eb
@@ -1,0 +1,41 @@
+# This file is an EasyBuild reciPY as per https://github.com/hpcugent/easybuild
+# Author: Jordi Blasco <jordi.blasco@hpcnow.com>; Guillermo Piccoli <guillermo.piccoli@hpcnow.com>
+easyblock = "PythonBundle"
+
+name = 'ObsPy'
+version = '1.4.0'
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/o/ObsPy/ObsPy-1.4.0-intel-2023a-Python-3.11.3.eb
```diff
@@ -0,0 +1,41 @@
...
+dependencies = [
+    ('Python-bundle-PyPI', '2023.06'),
+    ('lxml', '4.9.2'),
+    ('matplotlib', '3.7.2'),
+    ('SQLAlchemy', '2.0.25'),
+]
+
+download_dep_fail = True
+sanity_pip_check = True
+use_pip = True
```
akesandgren, line 25: These are the default nowadays and should be removed

easybuild/easyconfigs/s/SQLAlchemy/SQLAlchemy-2.0.25-GCCcore-12.3.0.eb
```diff
@@ -20,6 +20,7 @@ dependencies = [
     ('Python', '3.11.3'),
     ('psycopg2', '2.9.9'),  # optional, postgresql extra
     ('Mako', '1.2.4'),  # needed by alembic
+    ('typing-extensions', '4.9.0'),
 ]
 
 use_pip = True
```
akesandgren, line 26: And while you're at it, this one should also be dropped, and the sanity_pip_check at the bottom

PR #20378, opened 2024-04-16
add patch to fix hardcoded `/bin/mv` path in Rhdf5lib extension included in R-bundle-Bioconductor v3.16 + v3.18

bedroge:
This is now using a patch instead of the nested `sed` commands used in #20316.


bedroge:

PR #20428, opened 2024-04-26
use regular `configure` instead of wrapper script for recent UCX easyconfigs

bedroge:
The UCX easyconfigs for version 1.9.0 and newer are currently using `configure_cmd = contrib/configure-release`; this is a wrapper script around `./configure` which passed some predefined flags (these haven't changed for many years), see:
https://github.com/openucx/ucx/blob/master/contrib/configure-release

While trying to install UCX on RISC-V, I ran into issues with the outdated `config.guess` included with UCX. The ConfigureMake easyblock should download a newer version and use that one, but it turns out that it only does that when the `configure_cmd` script includes a certain string (`Generated by GNU Autoconf`). Though the `configure` script does include it, this is not the case for the `contrib/configure-release` wrapper script, hence that step is skipped. By appending the options that this wrapper uses to `configopts`, we can just call `./configure` itself, and then EB does make sure that an updated `config.guess` is being used. Additionally, it's more explicit in the easyconfig which flags are being passed.

### DIFF
--- a/easybuild/easyconfigs/u/UCX/UCX-1.13.1-GCCcore-12.2.0.eb
+++ b/easybuild/easyconfigs/u/UCX/UCX-1.13.1-GCCcore-12.2.0.eb
@@ -40,10 +40,10 @@ dependencies = [
     ('numactl', '2.0.16'),
 ]
 
-configure_cmd = "contrib/configure-release"
-
 configopts = '--enable-optimizations --enable-cma --enable-mt --with-verbs '
 configopts += '--without-java --without-go --disable-doxygen-doc '
+# include the configure options from contrib/configure-release
+configopts += '--disable-logging --disable-debug --disable-assertions --disable-params-check '
 
 buildopts = 'V=1'
 
--- a/easybuild/easyconfigs/u/UCX/UCX-1.14.0-GCCcore-12.2.0.eb
+++ b/easybuild/easyconfigs/u/UCX/UCX-1.14.0-GCCcore-12.2.0.eb
@@ -37,10 +37,10 @@ dependencies = [
     ('numactl', '2.0.16'),
 ]
 
-configure_cmd = "contrib/configure-release"
-
 configopts = '--enable-optimizations --enable-cma --enable-mt --with-verbs '
 configopts += '--without-java --without-go --disable-doxygen-doc '
+# include the configure options from contrib/configure-release
+configopts += '--disable-logging --disable-debug --disable-assertions --disable-params-check '
 
 buildopts = 'V=1'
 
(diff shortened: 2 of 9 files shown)


bedroge:

bedroge:

branfosj:
We also use `contrib/configure-release` in [`ucx_plugins.py`](https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/u/ucx_plugins.py#L80). Do we want to look at changing that as well?

bedroge:
> We also use `contrib/configure-release` in [`ucx_plugins.py`](https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/u/ucx_plugins.py#L80). Do we want to look at changing that as well?

Ah, good catch, didn't know it was used in other places. I suppose it makes sense to use the same approach everywhere, so I can make a PR for that easyblock as well. 

However, @SebastianAchilles found out that this patch is actually not really required anymore to solve the RISC-V issue (since the `autoreconf` command added in #20429 apparently updates the included `config.guess`, and then it does recognize RISC-V), so in that sense we don't really need this PR anymore. On the other hand, this approach may still be a bit cleaner anyway, since we don't depend on that wrapper script anymore?

bedroge:
I've applied the same change to the UCX plugins easyblock in https://github.com/easybuilders/easybuild-easyblocks/pull/3315. I've also opened a PR for UCX 1.16.0 (#20431), which demonstrates that this change is also still useful for RISC-V builds.

SebastianAchilles (approved):
> this approach may still be a bit cleaner anyway

Yes, I share your opinion, using the regular `configure` seems a bit cleaner.

> I've applied the same change to the UCX plugins easyblock in [easybuilders/easybuild-easyblocks#3315](https://github.com/easybuilders/easybuild-easyblocks/pull/3315). I've also opened a PR for UCX 1.16.0 (#20431), which demonstrates that this change is also still useful for RISC-V builds.

Since this change will be useful for future UCX easyconfigs for RISC-V I think it makes sense to merge this PR.

@branfosj Do you agree?

boegel (approved):
make sense to circumvent the `contrib/configure-release` script, since it doesn't add much value, and causes trouble since updated `config.guess` is not picked up

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 9 out of 9 (9 easyconfigs in total)
node3147.skitty.os - Linux RHEL 8.8, x86_64, Intel(R) Xeon(R) Gold 6140 CPU @ 2.30GHz, Python 3.6.8
See https://gist.github.com/boegel/17a637b6b2616e41c6cafbcf3d01f8e9 for a full test report.

boegel:
Going in, thanks @bedroge!

PR #20435, opened 2024-04-29
add 'archive' extra to optionally install easybuild-easyconfigs-archive

lexming:
Depends on:

* [x] https://github.com/easybuilders/easybuild-easyconfigs-archive/pull/5

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -109,6 +109,9 @@ versions, etc.).""",
         "easybuild_framework(>=%s.0)" % MAJ_VER,
         "easybuild_easyblocks(>=%s)" % MAJMIN_VER,
     ],
+    extras_require={
+        "archive": ["easybuild-easyconfigs-archive==%s" % VERSION],
+    },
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -109,6 +109,9 @@ def get_data_files():
         "easybuild_framework(>=%s.0)" % MAJ_VER,
         "easybuild_easyblocks(>=%s)" % MAJMIN_VER,
     ],
+    extras_require={
+        "archive": ["easybuild-easyconfigs-archive==%s" % VERSION],
```
boegel, line 113: @lexming This will only work if we start publishing `easybuild-easyconfigs-archive` packages on PyPI with those exact versions, I'm not sure that makes much sense?

lexming: @boegel true, we can just require anything over the current major version for the archive

jfgrimm (approved):
lgtm

Issue #20417, opened 2024-04-25
Checksums have changes for all libxc versions


schiotz:
Hi,

It looks like checksums have changes for all .tar.gz files downloaded by all (recent?) versions of ``libxc`` easyconfigs.  The files are "release assets" on gitlab, I don't suppose those should change.  Does anyone know if this is anyhow expected - I am wary of just changing the checksums in case something fishy is going on.

Also, I notices that *two* checksums are given although only one file is downloaded, could that be because this has happened before and the new checksums have just been added to the list?

/Jakob

CC: @OleHolmNielsen

zao:
Someone ran into this on Slack today, I took a look and it seems like there's not much stability in the checksums generated for these tarballs on GitLab.

On the dual checksums, the first set is from when we used to take libxc sources from another site; GitLab as a source was added in late March this year in https://github.com/easybuilders/easybuild-easyconfigs/pull/20163 as the first place became defunct. This introduced a new checksum.

Unfortunately, it seems like GitLab is still GitLab and doesn't necessarily have stable checksums on "release assets", which probably are generated on the fly from tags and other repo state.

For example for `libxc-5.1.6.tar.gz`, I'm getting one checksum and a reporter on Slack is getting another:
```
bcdf597593a80e53c4e2f6cab486c435fb497d2afe1819debd8e225597015913  xc-new/libxc-5.1.6.tar.gz
0d8e62f619b72b3af1694f13cfdffad94ccbeefcbdc5e2765d2e9f930c6f5792 /home/redacted/ebsrcs/l/libxc/libxc-5.1.6.tar.gz
```

This might need some investigation into whether GitLab is being silly again with regard to stable archive generation.

boegel:
see also https://github.com/easybuilders/easybuild-easyconfigs/pull/20452

boegel:
I've opened an issue upstream for this, since I'm not seeing any differences for other projects hosted on GitLab: https://gitlab.com/libxc/libxc/-/issues/511

appolloford:
Sorry, I just read through all the discussion. I only noticed that a few files are different since the last time the source was downloaded from Gitlab. I don't know whether they can be modified from Gitlab autogenerated tarball, but if so that may mean the history has been rebased?
```
Only in new/libxc-4.3.4: .codecov.yml
Only in new/libxc-4.3.4: .gitignore
Only in new/libxc-4.3.4: Doxyfile
[...]

PR #20452, opened 2024-05-02
Fix libxc-4.3.4 checksum

appolloford:
(created using `eb --new-pr`)


boegel:
@appolloford Are you consistently getting the same checksum when downloading multiple times?

It seems like the checksums are changing over time, which suggests that downloads from GitLab are not "stable", at least two people have reported getting another checksum in the EasyBuild Slack ([see thread](https://easybuild.slack.com/archives/C34UA1HT7/p1714645247438349))

boegel:
see also discussion in https://github.com/easybuilders/easybuild-easyconfigs/issues/20417

Micket:
i think it's fine to add this, but this PR can't be synced anymore, a conflict. I doubt anyone wants to fight the conflict, as it's easier to just open a new PR with `2d5878dd69f0fb68c5e97f46426581eed2226d1d86e3080f9aa99af604c65647` added, so i'm closing this.

PR #20508, opened 2024-05-10
{math}[system/system,GCCcore/13.2.0] MATLAB v2023b r5

Louwrensth:
(created using eb `--new-pr` and manually creating the PR)


Louwrensth:
Why `GCCcore` toolchain?
- To have MATLAB link to the same version of libstdc++ as provided by toolchain
- Allows running mex files built with a more recent version of [GCC that MATLAB supports](https://mathworks.com/support/requirements/supported-compilers-linux.html) (which is typically lagging)

Why drop `versionsuffix`?
- `MATLAB-2023b-GCCcore-13.2.0-r5` looks odd, and `MATLAB-2023b-r5-GCCcore-13.2.0` is found less confusing.
- The `-r5` is akin to a [patch version for MATLAB](https://mathworks.com/support/faq/mathworks-update-notifications.html), so can EB also regard it as such, like it does for other software with [patch version as in semver](https://semver.org/spec/v2.0.0.html)?


BTW what matlab says to `version`: 
```
$ matlab -nodisplay -batch version

ans =

    '23.2.0.2459199 (R2023b) Update 5'
```

Micket:
> a more recent version of [GCC that MATLAB supports](https://mathworks.com/support/requirements/supported-compilers-linux.html) (which is typically lagging)

I guess you meant to say "*than* what MATLAB supports"? It's a bit confusing read otherwise.

And first, this just works? I could have assumed there are compatibilities with the matlab kernel itself when it comes to linking to these libraries, and just removing them isn't a wise idea.

Louwrensth:
> 
> I guess you meant to say "_than_ what MATLAB supports"? It's a bit confusing read otherwise.

Yes, I absolutely meant that, thanks for catching the slip up.

> 
> And first, this just works? I could have assumed there are compatibilities with the matlab kernel itself when it comes to linking to these libraries, and just removing them isn't a wise idea.

Yes, this is the adaptation of the [`LD_PRELOAD` suggestion by a MathWorks staff member](https://mathworks.com/matlabcentral/answers/1907290-how-to-manually-select-the-libstdc-library-to-use-to-resolve-a-version-glibcxx_-not-found). I forgot to link this solution.

Of course, I am not 100% sure this will work in future versions, but I believe it should: `libstdc++.so` contains backwards compatibility via those `GLIBCXX_*` version symbols, so that Matlab's kernel can find older version symbols in a newer `libstdc++` too. (The reverse, an older GCC with a newer Matlab is not going to happen -- at least never with this `.eb` file.)

easybuild/easyconfigs/m/MATLAB/MATLAB-2023b-r5-GCCcore-13.2.0.eb
```diff
@@ -0,0 +1,44 @@
...
+local_versiontuple = tuple(version.split('-'))
+
+homepage = 'https://www.mathworks.com/products/matlab'
+description = """MATLAB is a high-level language and interactive environment
+ that enables you to perform computationally intensive tasks faster than with
+ traditional programming languages such as C, C++, and Fortran."""
+
+toolchain = {'name': 'GCCcore', 'version': '13.2.0'}
+
+sources = ['R%s_Update_%s_Linux.iso' % local_versiontuple]
```
SimonPinches, line 12: I think this doesn't work since it looks for `R2023b_Update_r5_Linux.iso` when it should be `R2023b_Update_5_Linux.iso`.

Louwrensth: Thanks! I made a fixup

SimonPinches (approved):

sassy-crick:
@Louwrensth You might want to sync your PR with development, something like this: 
```
```
Also, as `MATLAB` is proprietary software could you upload a test-report please? Finally, I checked the `MATLAB` versions we are having thus far and they are usually in the `SYSTEM` toolchain, i.e. `MATLAB-2022b-r5.eb`.

PR #20456, opened 2024-05-02
{tools}[foss/2023a] aiida-core-2.5.1, plumpy-0.21.6

Crivella:
Added Python bundles for `aiida-core==2.5.1` and its dependencies

Actual usage of AiiDA will require also for RabbitMQ (needs #20441), or another independent RabbitMQ instance running + a PostgreSQL database (already imported from the `psycopg2` dependency).

EDIT: There are modes to use it without RabbitMQ and Postgres with limited functionalities

PR #20441, opened 2024-04-30
{lang,tools}[foss/2023a] RabbitMQ 3.13.1 with dependencies

Crivella:
This PR includes EC files for RabbitMQ and its dependencies

- RabbitMQ-3.13.1-GCC-12.3.0.eb

Deps:

- ErlangOTP-26.2.4-GCC-12.3.0.eb
- Elixir-1.16.2-GCC-12.3.0.eb
- xmlto-0.0.28-GCC-12.3.0.eb

PR #20516, opened 2024-05-13
{tools}[foss/2023a] AiiDA

Crivella:
Set of bootstrap/cleanup scripts for aiida

Depends on PRs:

- #20456 
- #20441 

The bootstrap script will

- PostgreSQL
  - Create a database in user space
  - Set the database to make use of UNIX sockets only to avoid opening ports
  - Launch the database
- RabbitMQ
  - Create a rabbitmq instance
  - Looks for 2 open ports above 25672
  - Launch the rabbitmq instance
  - Add advanced configuration for avoiding timeouts [SEE HERE](https://aiida.readthedocs.io/projects/aiida-core/en/latest/intro/troubleshooting.html#rabbitmq-incompatibility)
- AiiDA
  - Setup the profile with the information about the created database and rabbitmq instances
    - Updates the rabbitMQ ports in case the services where stopped and the ports changed after rerunning the bootstrap script.
  - launch the daemon

The user can specify the following:

- Location of the PostgreSQL datadir (to allow running on a fast and/or backed up partition)
- Name/Surname/Email/Affiliation for the AiiDA profile (defaults are generated based on the username and hostname)
- Location of the aiida repository

After the first run of the bootstrap script, configuration details are stored under `$HOME/.eb_aiida` and will be recalled in future runs:

- Location of the database data dir
- Database user password (with 600 permission)
- Ports for RabbitMQ
- RabbitMQ user password (with 600 permission) (this is also stored by AiiDA in its configuration file)
- Location of the aiida repository

### DIFF
--- a/easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
+++ b/easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
@@ -3,8 +3,8 @@ easyblock = 'Binary'
 name = 'AiiDA'
 version = '2.5.1'
 
-homepage = ''
-description = """"""
+homepage = 'https://www.aiida.net/'
+description = """Scripts for bootstrapping and cleaning up an AiiDA installation."""
 
 toolchain = {'name': 'gfbf', 'version': '2023a'}
 
--- a/easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
+++ b/easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
@@ -12,7 +12,10 @@ sources = [
     'aiida_bootstrap.sh',
     'aiida_cleanup.sh',
 ]
-checksums = []
+checksums = [
+    {'aiida_bootstrap.sh': 'c92bc1c0ff23b2d54be2a5367df88abc94a413555b34fab830c5dcea84507650'},
+    {'aiida_cleanup.sh': '5f6a46c1f9b4381fe012c94323c38e72d8a314e5f7b81a4f3517ebe6034f3cfd'},
+]
 
 builddependencies = [
     ('binutils', '2.40'),
--- a/easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
+++ b/easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
@@ -1,0 +1,47 @@
+import os
+
+easyblock = 'Binary'
+
+name = 'AiiDA'
+version = '2.5.1'
+
+homepage = ''
+description = """"""
+
+toolchain = {'name': 'gfbf', 'version': '2023a'}
+
+sources = [
+    'aiida_bootstrap.sh',
+    'aiida_cleanup.sh',
+]
+checksums = []
+
+builddependencies = [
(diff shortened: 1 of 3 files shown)


sassy-crick:
@Crivella You might want to sync your PR with development, something like this: 
```
```

easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
```diff
@@ -0,0 +1,49 @@
...
+modluafooter = """
+local home = os.getenv("HOME")
+local eb_config_dir = pathJoin(home, ".eb_aiida")
+
+setenv("EB_CONFIG_DIR", eb_config_dir)
+setenv("PGHOST", eb_config_dir)
+setenv("RMQ_BASEDIR", pathJoin(eb_config_dir, "rabbitmq"))
+setenv("RABBITMQ_NODENAME", "rabbit-" .. os.getenv("USER"))
+setenv("AIIDA_PATH", eb_config_dir)
+"""
```
ocaisa, line 49: This is problematic as it is lua-specific but I also don't think you need it any more. There is now support for environment variables in `modextravars` so you should be able to leverage that.

ocaisa: I would probably add a `modloadmsg` about the usage of $HOME as on a number of systems the space here can be very limited (the end user may want to create a symlink from `EB_CONFIG_DIR` to a less constrained location for example)

easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
```diff
@@ -0,0 +1,49 @@
+easyblock = 'Binary'
+
+name = 'AiiDA'
+version = '2.5.1'
+
+homepage = 'https://www.aiida.net/'
+description = """Scripts for bootstrapping and cleaning up an AiiDA installation."""
```
ocaisa, line 7: You probably want to add some usage information. I don't see the scripts really explain what they do either so it is probably good to add that here.

easybuild/easyconfigs/a/AiiDA/AiiDA-2.5.1-gfbf-2023a.eb
ocaisa: When you run the bootstrap script, what will happen when they close their session? Should we be recommending running it in a `screen` session?

Crivella: > I would probably add a modloadmsg about the usage of $HOME as on a number of systems the space here can be very limited (the end user may want to create a symlink from EB_CONFIG_DIR to a less constrained location for example)

Was thinking of adding a `--help` to the scripts and adding what scripts are being added by the module (and possibly what environment variables they use), pointing to the `--help` on how to run them


> When you run the bootstrap script, what will happen when they close their session? Should we be recommending running it in a screen session?

The services are all ran in detached mode, so they will keep running until something kills them.
Guess in that spirit we could also add a `..._stop.sh` script to stop the services without also removing all the data

Crivella:
In hindsight, probably out of scope for EB.

Would required either making a separate package with the scripts, but they are also not really needed anymore with the newer versions of AiiDA that can work (even if with less features) without RabbitMQ and with a local `SQLite` database

PR #20102, opened 2024-03-13
{bio,tools}[foss/2023b] GROMACS v2024.1, networkx v3.2.1

boegel:
(created using `eb --new-pr`)

PR #20522, opened 2024-05-14
use the included gmxapi for GROMACS 2024.1

bedroge:
Seems like https://github.com/easybuilders/easybuild-easyconfigs/pull/20102 forgot to bump the gmxapi version and source tarball, it was still using an older version from the 2023.3 tarball. Looking at https://gitlab.com/gromacs/gromacs/-/blob/v2024.1/python_packaging/gmxapi/src/gmxapi/version.py?ref_type=tags#L68, the 2024.1 release includes gmxapi 0.5.0.

### DIFF
--- a/easybuild/easyconfigs/g/GROMACS/GROMACS-2024.1-foss-2023b.eb
+++ b/easybuild/easyconfigs/g/GROMACS/GROMACS-2024.1-foss-2023b.eb
@@ -54,7 +54,7 @@ checksums = [
 
 builddependencies = [
     ('CMake', '3.27.6'),
-    ('scikit-build', '0.17.6'),
+    ('scikit-build-core', '0.9.3'),
 ]
 
 dependencies = [


bedroge:
CORE_CNT=16

bedroge:
CORE_CNT=16

ocaisa:
@bedroge Looks like it needs another build dependency:
```
ModuleNotFoundError: No module named 'scikit_build_core'
```

bedroge:
> @bedroge Looks like it needs another build dependency:
> 
> ```
> ModuleNotFoundError: No module named 'scikit_build_core'
> ```

Opened a PR for a new `scikit-build-core`, which seems to work (based on a manual/interactive test): https://github.com/easybuilders/easybuild-easyconfigs/pull/20526

bedroge:
> > @bedroge Looks like it needs another build dependency:
> > ```
> > ModuleNotFoundError: No module named 'scikit_build_core'
> > ```
> 
> Opened a PR for a new `scikit-build-core`, which seems to work (based on a manual/interactive test): #20526

That PR just got merged, and I've done another test build without scikit-build (as I couldn't find anywhere that this is required) but with scikit-build-core, and that worked fine.

bedroge:
CORE_CNT=16

bedroge:
CORE_CNT=16

SebastianAchilles (approved):
lgtm

SebastianAchilles:
Going in, thanks @bedroge!

ocaisa:
What about adding the SVE patch in https://gitlab.com/gromacs/gromacs/-/merge_requests/4299 ?

SebastianAchilles:
> What about adding the SVE patch in https://gitlab.com/gromacs/gromacs/-/merge_requests/4299 ?

I would suggest to open a follow-up PR for that.

bedroge:
> What about adding the SVE patch in https://gitlab.com/gromacs/gromacs/-/merge_requests/4299 ?

I haven't tested their patch yet, and since it's unrelated to the change in this PR, we can do it in a follow-up PR. But maybe we should also be a bit careful and wait for it to get merged upstream (or only apply it to Neoverse V1 or something)?

boegel:
> > What about adding the SVE patch in https://gitlab.com/gromacs/gromacs/-/merge_requests/4299 ?
> 
> I haven't tested their patch yet, and since it's unrelated to the change in this PR, we can do it in a follow-up PR. But maybe we should also be a bit careful and wait for it to get merged upstream (or only apply it to Neoverse V1 or something)?

I think they're waiting for feedback from us to confirm that the fix works on our side too, I'm looking into that (but it's difficult to find time for that during ISC'24)...

easybuild/easyconfigs/g/GROMACS/GROMACS-2024.1-foss-2023b.eb
```diff
@@ -74,12 +74,12 @@ exts_default_options = {
 }
 
 exts_list = [
-    ('gmxapi', '0.4.2', {
+    ('gmxapi', '0.5.0', {
         'preinstallopts': 'export CMAKE_ARGS="-Dgmxapi_ROOT=%(installdir)s ' +
                           '-C %(installdir)s/share/cmake/gromacs_mpi/gromacs-hints_mpi.cmake" && ',
-        'source_tmpl': 'gromacs-2023.3.tar.gz',
+        'source_tmpl': 'gromacs-2024.1.tar.gz',
```
boegel, line 80: @bedroge @ocaisa We should use `%(version)s` here to avoid running into this again in the future?

And a comment on how to determine the `gmxapi` version that is shipped with GROMACS

PR #20586, opened 2024-05-22
remove ipympl v0.9.3 from foss/2023a in favour of gfbf/2023a

lexming:
Delete `ipympl-0.9.3-foss-2023a.eb`. Correct toolchain for `ipympl` is `gfbf/2023a`, and the easyconfig in `gfbf` already uses the better `EB_ENV_JUPYTER_ROOT` environment variable to configure the jupyter environment.


bedroge (approved):
Lgtm

PR #20598, opened 2024-05-23
easyconfig for R/4.4.0-gfbf-2023a

moravveji:
- In response to the published security vulnerability [CVE-2024-27322](https://nvd.nist.gov/vuln/detail/CVE-2024-27322) with R versions earlier than 4.4 (read more [in this blog post](https://thehackernews.com/2024/04/new-r-programming-vulnerability-exposes.html)), I could simply top up R version to 4.4.0 starting from a recent easyconfig for the 2023a toolchain
- I have successfully installed this R on our Intel Skylake, Cascadelake, Icelake and Sapphire Rapids nodes with Rocky 8.9
- I have ditched the patch `R-4.x_fix-CVE-2024-27322.patch`, because this version of R is supposed to contain a fix for the vulnerability
- To avoid having two easyconfigs for 2023a toolchain, I have removed `R-4.3.2-gfbf-2023a.eb`


moravveji:
I see the following error three times:
```
easybuild.tools.build_log.EasyBuildError: 'Missing dependencies: R/4.3.2-foss-2023a, R/4.3.2-gfbf-2023a (no easyconfig file or existing module found)'
```
... and it surprises me that some version of `R` is expected as a dependency for building this specific version of `R` (or perhaps I am misinterpreting the error message). Can someone please throw a tip on how to resolve this?

tardigradus:
>     * To avoid having two easyconfigs for 2023a toolchain, I have removed `R-4.3.2-gfbf-2023a.eb`

What exactly did you do? I am no expert, but in general I would think there is no necessity to remove any existing EBs when you write a new one.

You might also consider raising the matter on the mailing list, as it may take a while before someone with more expertise than me comes across this PR.

moravveji:
> > ```
> > * To avoid having two easyconfigs for 2023a toolchain, I have removed `R-4.3.2-gfbf-2023a.eb`
> > ```
> 
> What exactly did you do? I am no expert, but in general I would think there is no necessity to remove any existing EBs when you write a new one.
> 
`git rm`, but that was not the most brilliant idea.

moravveji:
I rather close this PR.

PR #20687, opened 2024-05-30
Removed timm-0.9.7-foss-2022a-CUDA-11.7.0.eb crate duplicates

thielemans:
Removed all crate duplicates as they would trigger extraction errors. They could not be successfully extracted (a second time) on eb 4.9.1.


Micket (approved):
lgtm  (i wonder why safetensor is included as an extension here, we have that as it's own easyconfig as well)

Micket:

Micket:
Going in, thanks @thielemans!

PR #20692, opened 2024-05-30
{lang}[foss/2022b] numba v0.58.1

zdemat:
Adding numba/0.58.1 for foss/2022b and CUDA/12.0.0.


Micket:
We don't need to do this, numba-0.58.1-foss-2022b.eb can do CUDA stuff (if you just load CUDA runtime of choice)

So, it's better if this doesn't exist, as the with and without CUDA suffix will just cause incompatibility between loaded modules. It's best to just use same numba package everywhere.

PR #20781, opened 2024-06-10
archive outdated example easyconfigs for Fujitsu toolchain

migueldiascosta:
these only worked on Fugaku, and most likely even there don't work anymore, they depend on an external module version that doesn't exist there anymore:

https://github.com/easybuilders/easybuild-easyconfigs/blob/6c0fc3b6d6775418a0270d683dcea160b1efab4b/easybuild/easyconfigs/f/FCC/FCC-4.5.0.eb#L14

also, the Fujitsu toolchain framework support itself needs to be made more flexible - it currently still depends on a hardcoded module name and environment variable that are system dependent: at least Fugaku, Isambard and Deucalion each use different modules

https://github.com/easybuilders/easybuild-framework/blob/f9134b1bd39da288d24cbc4f406dfbd84f203bc5/easybuild/toolchains/compiler/fujitsu.py#L42-L43

Will follow up on the framework support at https://github.com/easybuilders/easybuild/issues/704


boegel (approved):

boegel:
squeezing this in for EasyBuild v4.9.2...

boegel:
Actually no, will need to wait until first release after EasyBuild v4.9.2, we're too far in release cycle already to squeeze this in, sorry...

migueldiascosta:
no problem

PR #20793, opened 2024-06-10
{chem}[intel/2023a] FHI-aims v240507

sassy-crick:
(created using `eb --new-pr`)


smoors:

sassy-crick:
Test report can be found [here](https://gist.github.com/sassy-crick/be9c857f9fbefaf8060766b3c283da13)

easybuild/easyconfigs/f/FHI-aims/FHI-aims-240507-intel-2023a.eb
```diff
@@ -0,0 +1,57 @@
...
+Details on available license options and how to register to access
+FHI-aims club may be found at:
+https://fhi-aims.org/get-the-code-menu/get-the-code """
+
+sources = ['%(namelower)s.%(version)s.tar.gz']
+checksums = ['dea6483834d80ef5d8826171bd710ebcb035c4b971c4c958c82fdb9f03af6a6b']
+
+builddependencies = [('CMake', '3.26.3')]
+
+configopts = ' -DCMAKE_Fortran_COMPILER="$MPIF90" -DCMAKE_CXX_COMPILER="icpc"  -DCMAKE_C_COMPILER="icc" '
```
Flamefire, line 40: To honor toolchain options `oneapi` and `usempi`:

```suggestion
configopts = ' -DCMAKE_Fortran_COMPILER="$FC" -DCMAKE_CXX_COMPILER="$CXX_SEQ"  -DCMAKE_C_COMPILER="$CC_SEQ" '
```

According to https://fhi-aims-club.gitlab.io/tutorials/cmake-tutorial/cmake_tutorial they need an MPI fortran compiler --> `'usempi': True` in toolchainopts

sassy-crick: Thanks for that! I done a bit more digging and with the help of the FHI-aims team I think I got it all sorted so we can use `oneAPI` as the compiler now. From what I understand, there are only a few places where `MPI` is being used and thus when you link it is important to use `MPI` but not when compiling. I got told that eliminates problems with wrapper-scripts. Nonetheless, I used `MPI` for all three languages. Let me know what you think of the improvements done.

easybuild/easyconfigs/f/FHI-aims/FHI-aims-240507-intel-2023a.eb
```diff
@@ -0,0 +1,63 @@
...
+checksums = ['dea6483834d80ef5d8826171bd710ebcb035c4b971c4c958c82fdb9f03af6a6b']
+
+builddependencies = [('CMake', '3.26.3')]
+
+configopts = ' -DCMAKE_Fortran_COMPILER="$MPIF90" -DCMAKE_CXX_COMPILER="$MPICXX"  -DCMAKE_C_COMPILER="$MPICC" '
+configopts += ' -DLIBS="mkl_scalapack_lp64 mkl_blacs_intelmpi_lp64 mkl_intel_lp64 mkl_sequential mkl_core" '
+configopts += ' -DLIB_PATHS="$EBROOTIMKL/mkl/latest/lib/intel64" '
+configopts += ' -DCMAKE_Fortran_FLAGS="$FFLAGS -fc=ifx -finline-functions" '
+configopts += ' -DCMAKE_CXX_FLAGS="$CXXFLAGS -cxx=icpx -finline-functions -std=c++11" '
+configopts += ' -DCMAKE_C_FLAGS="$CFLAGS -cc=icx -finline-functions -std=gnu99" '
```
Flamefire, line 46: Are you sure any of this is required?

- `-fc` etc. should be set by EasyBuild via the `I_MPI` variables
- finline-functions is set by `-O2` and above
- `-std` shouldn't be required as the defaults for those intel compilers are likely at least that

Removing might help updating the EC later.

sassy-crick: I took these flags from the template they provided. The `-finline-functions` comes from the dropping of `-ip` and that was the replacement I found. 
That said: you are right, it works without them as well. I will update the PR and will include one which makes use of an external `ELPA` installation.

sassy-crick:
I have included a second EasyConfig file which is using an external `ELPA` installation. The thing I am not sure how to handle best is: which `ELPA` kernel has being used? `ELPA` itself is setting the kernel but I don't know how to retrieve that information for the `FHI-aims` installation.

steenlysgaard:
> I have included a second EasyConfig file which is using an external `ELPA` installation. The thing I am not sure how to handle best is: which `ELPA` kernel has being used? `ELPA` itself is setting the kernel but I don't know how to retrieve that information for the `FHI-aims` installation.

I have achieved this by using `elpa2_print_kernels`. For example use that in `preconfigopts`, like this:
```python
# Check the AVX capabilities of the ELPA installation
preconfigopts = """
AVX_CAP=$(elpa2_print_kernels | grep -o 'AVX[0-9]*' | sort -r | head -n1)
if [ -z "$AVX_CAP" ]; then
    AVX_CAP="Generic"
fi
"""
```
And then in `configopts` set `-DELPA2_KERNEL="$AVX_CAP"`

sassy-crick:
> > I have included a second EasyConfig file which is using an external `ELPA` installation. The thing I am not sure how to handle best is: which `ELPA` kernel has being used? `ELPA` itself is setting the kernel but I don't know how to retrieve that information for the `FHI-aims` installation.
> 
> I have achieved this by using `elpa2_print_kernels`. For example use that in `preconfigopts`, like this:
> 
> ```python
> # Check the AVX capabilities of the ELPA installation
> preconfigopts = """
> AVX_CAP=$(elpa2_print_kernels | grep -o 'AVX[0-9]*' | sort -r | head -n1)
> if [ -z "$AVX_CAP" ]; then
>     AVX_CAP="Generic"
> fi
> """
> ```
> 
> And then in `configopts` set `-DELPA2_KERNEL="$AVX_CAP"`

This is an excellent idea! I have included that in the EasyConfig file, many thanks. 
I would suggest when the next version is out (not v240920_2 as I already done that) we create a `fhi-aims.py` maybe and include an automated CPU detection so we can set the compiler flags closer to what the `FHI-aims` developer suggest, instead of using `xHosts`.

sassy-crick:
Test report by @sassy-crick
**SUCCESS**
hp - Linux Debian GNU/Linux 12, x86_64, 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz, Python 3.11.2
See https://gist.github.com/sassy-crick/0cd806872b6f21ea38c08b4de9550446 for a full test report.

sassy-crick:
Test report by @sassy-crick
**SUCCESS**
hp - Linux Debian GNU/Linux 12, x86_64, 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz, Python 3.11.2
See https://gist.github.com/sassy-crick/a668f3e301e63f2b8a892f1099637de1 for a full test report.

PR #20867, opened 2024-06-20
Use external deps for safetensors and tokenizers in Transformers-4.30.2

Micket:
(created using `eb --new-pr`)


Micket:
bah, i messed up, redoing this

PR #20960, opened 2024-07-03
stop using `modextrapaths` to update `$PYTHONPATH` with standard path to installed Python packages (`lib/python%(pyshortver)s/site-packages`)

Micket:
All the trivial cases cleaned up here. Some others i could potentially partly clean up, which uses a mix of several PYTHONPATH (standard + nonstandard location, Spark is an example of such an easyconfig).

Should of course not be merged until the corresponding framework PR is merged.

- [ ] depends on https://github.com/easybuilders/easybuild-framework/pull/4539/


Flamefire:
That reminds me of an old PR of mine: https://github.com/easybuilders/easybuild-easyconfigs/pull/13385

However all but 1 of the ECs are now deprecated and that one is fixed here. However there was also an enhancement to CI which I extracted to #20963

And a regexp I used which you might try in your tree: `grep -E ": 'lib/python[^/]+/site-packages'" $(grep --files-with-matches -Er "\\beasyblock = '(ConfigureMakePythonPackage|PythonBundle)'")`

Flamefire:
The CI requires https://github.com/easybuilders/easybuild-easyconfigs/pull/21405 after https://github.com/easybuilders/easybuild-framework/pull/4634 which requires https://github.com/easybuilders/easybuild-framework/pull/4644

Flamefire:
CI requires #21405 (which requires https://github.com/easybuilders/easybuild-framework/pull/4644) after https://github.com/easybuilders/easybuild-framework/pull/4634

test/easyconfigs/easyconfigs.py
```diff
@@ -1764,6 +1764,21 @@ def template_easyconfig_test(self, spec):
             fail_msg += f"'{orig_val}' vs '{dumped_val}'"
             failing_checks.append(fail_msg)
 
+    # don't allow updating of $PYTHONPATH with standard lib/python*/site-packages path,
+    # since that's already taken care of by EasyBuild framework now
+    # (cfr. https://github.com/easybuilders/easybuild-framework/pull/4539)
+    modextrapaths = ec.get('modextrapaths', {})
+    regex = re.compile(r'^lib*/python[0-9]\.[0-9]+/site-packages$')
+    for key, value in modextrapaths.items():
+        value = resolve_template(value, ec.template_values)
```
Flamefire, line 1773: `ec.get` has a `resolve` parameter defaulting to `True`, so it seems the value is already resolved at this point, isn't it?

If it was indeed required I'd suggest putting it below the check to avoid resolving every value without using almost any

Flamefire (changes requested):

test/easyconfigs/easyconfigs.py
boegel: Ah, maybe, I didn't actually verify that tbh... Will check

test/easyconfigs/easyconfigs.py
```diff
@@ -1764,6 +1764,20 @@ def template_easyconfig_test(self, spec):
             fail_msg += f"'{orig_val}' vs '{dumped_val}'"
             failing_checks.append(fail_msg)
 
+    # don't allow updating of $PYTHONPATH with standard lib/python*/site-packages path,
+    # since that's already taken care of by EasyBuild framework now
+    # (cfr. https://github.com/easybuilders/easybuild-framework/pull/4539)
+    modextrapaths = ec.get('modextrapaths', {}, resolve=True)
```
Flamefire, line 1770: We usually denote explicitly when we do **not** want to resolve values so I'd not be so explicit (only) here and given `ec.get('key') == `ec['key']`

```suggestion
    modextrapaths = ec.get('modextrapaths', {})
```

boegel: I prefer making it very explicit here that we're dealing with a resolved value, otherwise the regular expression being used below doesn't make sense (since it doesn't take into account templates like `%(pyshortver)s`).

Flamefire: Ok. For me that wouldn't have been a question because we always deal with resolved values when working with an `EasyConfig` instance and usually have to do extra work when we don't want/can't do that. So this felt inconsistent to the other uses of `ec.get` we have.  
But as it did cause confusion, even for someone as familiar with the codebase, it might make sense to be so explicit here.

boegel:
Test report by @boegel
**FAILED**
Build succeeded for 207 out of 209 (201 easyconfigs in total)
node3573.doduo.os - Linux RHEL 8.8, x86_64, AMD EPYC 7552 48-Core Processor, Python 3.6.8
See https://gist.github.com/boegel/7aa02cde418f967a759fb7895d895a0d for a full test report.

boegel:
The two failures have nothing to do the changes being made here:

For ` BUSCO-5.4.5-foss-2022a.eb`:
```
024-10-06 23:52:05 ERROR:	A run with the name /arcanine/scratch/gent/vo/000/gvo00002/vsc40023/easybuild_tests/RHEL8-eb5dev-rpath/zen2-ib/software/BUSCO/5.4.5-foss-2022a/test_eukaryota.out already exists...
	If you are sure you wish to overwrite existing files, please use the -f (force) option
2024-10-06 23:52:05 ERROR:	BUSCO analysis failed!
```

For `FunGAP-1.1.1-foss-2022a.eb`:
```
Missing modules for dependencies (use --robot?): BUSCO/5.4.5-foss-2022a
```

So it won't let it block this PR...

boegel (approved):

boegel:
Going in, thanks @Micket!

Flamefire:
Updated the PR for adding the CI check: https://github.com/easybuilders/easybuild-easyconfigs/pull/19061

PR #20963, opened 2024-07-04
whitelist ConfigureMakePythonPackage for sanity_check_paths CI check

Flamefire:
Also extend the `sanity_pip_check` whitelist for all 2016 ECs although this won't really matter going forward. Just kept that commit as long as that check is there


Micket (approved):
lgtm, though this should be cleaned up further for 5.0.x where we no longer have 2016[ab]

Flamefire:
> lgtm, though this should be cleaned up further for 5.0.x where we no longer have 2016[ab]

Right. But as long as the configparser check for 2016 is there it is now "complete".

PR #20965, opened 2024-07-04
add easyconfig files for openmc software stack

LukeLabrie:
This pr introduces a set of easyconfig files for the [OpenMC](https://github.com/openmc-dev/openmc) software stack. OpenMC is a monte-carlo particle transport software. It can be configured with or without [DAGMC](https://github.com/svalinn/DAGMC), a geometry engine for running particle transport on CAD models, and thus includes an easyconfig and relevant dependencies for both cases.

### DIFF
--- a/easybuild/easyconfigs/d/DAGMC/DAGMC-3.2.2_find-MOAB.patch
+++ b/easybuild/easyconfigs/d/DAGMC/DAGMC-3.2.2_find-MOAB.patch
@@ -1,0 +1,21 @@
+--- /home/myuser/.local/easybuild/build/dagmc_dev/3.2.2/foss-2022a/DAGMC/cmake/FindMOAB.cmake.orig	2024-04-11 11:00:09.538052758 +0000
++++ /home/myuser/.local/easybuild/build/dagmc_dev/3.2.2/foss-2022a/DAGMC/cmake/FindMOAB.cmake	2024-04-11 11:06:11.078058472 +0000
+@@ -2,9 +2,15 @@
+ 
+ # Find MOAB cmake config file
+ # Only used to determine the location of the HDF5 with which MOAB was built
+-set(MOAB_SEARCH_DIRS)
+-file(GLOB MOAB_SEARCH_DIRS ${MOAB_SEARCH_DIRS} "${MOAB_DIR}/lib*/cmake/MOAB")
+-string(REPLACE "\n" ";" MOAB_SEARCH_DIRS ${MOAB_SEARCH_DIRS})
++
++# Directly set MOAB_SEARCH_DIRS
++set(MOAB_SEARCH_DIRS
++    "${MOAB_DIR}/lib/cmake/MOAB"
++    "${MOAB_DIR}/lib64/cmake/MOAB"
++    "${MOAB_DIR}/cmake/MOAB"
++    "${MOAB_DIR}/MOAB"
++    "${MOAB_DIR}"
++    )
+ find_path(MOAB_CMAKE_CONFIG
+   NAMES MOABConfig.cmake
+   PATHS ${MOAB_SEARCH_DIRS}
--- a/easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
+++ b/easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
@@ -1,0 +1,36 @@
+name = 'double-down'
+version = '1.1.0'
+
+homepage = 'https://github.com/pshriwise/double-down'
+description = "double-down is a double precision interface to Embree via the Mesh Oriented dAtaBase (MOAB)."
+software_license = 'LicenseGCC'
+software_license_urls = ['https://github.com/pshriwise/double-down/blob/develop/LICENSE']
+toolchain = {'name': 'foss', 'version': '2022a'}
+
+sources = [{
+    'filename': 'double-down.tar.gz',
+    'git_config': {
+        'url': 'https://github.com/pshriwise',
+        'repo_name': 'double-down',
+        'tag': f'v{version}'
+    },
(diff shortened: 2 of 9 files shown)


easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
```diff
@@ -0,0 +1,37 @@
...
+toolchain = {'name': 'foss', 'version': '2022a'}
+
+sources = [{
+    'filename': 'double-down.tar.gz',
+    'git_config': {
+        'url': 'https://github.com/pshriwise',
+        'repo_name': 'double-down',
+        'tag': f'v{version}'
+    },
+}]
```
Flamefire, line 18: With `git_config` you can't add checksums. So better download the file from the releases:
```suggestion
source_urls = ['https://github.com/pshriwise/double-down/archive/']
sources = ['v%(version)s.tar.gz']
```

easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
```diff
@@ -0,0 +1,37 @@
...
+
+builddependencies = [('CMake', '3.23.1'),('make','4.3')]
+
+dependencies = [
+    ('Doxygen','1.9.4'),
+    ('Embree','3.13.4','',{'name': 'system', 'version': 'system'}),
```
Flamefire, line 24: There is a constant:
```suggestion
    ('Embree','3.13.4' ,'', SYSTEM),
```

easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
```diff
@@ -0,0 +1,37 @@
...
+    ('MOAB','5.5.0'),
+    ('tbb','2021.5.0')
+]
+
+easyblock = 'CMakeMake'
+
+_moab_dir = '$EBROOTMOAB'
+configopts = f"-DMOAB_DIR={_moab_dir}"
```
Flamefire, line 32: We can't use fstrings on possible older versions and it's not required here anyway:
```suggestion
configopts = "-DMOAB_DIR=$EBROOTMOAB"
```

easybuild/easyconfigs/d/DAGMC/DAGMC-3.2.2_find-MOAB.patch
```diff
@@ -0,0 +1,25 @@
+This patch corrects the search CMake search paths for MOAB. for some reason, 
+when built with easybuild, MOAB_SEARCH_DIRS does not become a list. The patch 
+therefore defines the search paths for explicity.
```
Flamefire, line 3: Can you give an example failure? The original code looks correct to me

easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
LukeLabrie: Ok, so as a rule easyconfigs should not contain fstrings?

easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
LukeLabrie: Makes sense. I've done this for the other easyconfigs that don't require submodules. The ones that do still use `git_config`.

LukeLabrie: @Flamefire, shouldn't checksums still work if you're pulling from a specific tag?

Flamefire: > I've done this for the other easyconfigs that don't require submodules. The ones that do still use git_config.

You can still use the tarballs by providing the sources and the directory to extract to. Often you don't even need those as many vendor dependencies as submodules and we can just provide them ourselves.

For OpenMC I see (ticks show available ECs):
- [x] Catch2
- [x] fmt
- [ ] gsl-lite
- [x] pugixml
- [x] xtensor
- [ ] xtl

So for most of them you don't need the submodules. For the remaining 2 it might make sense to create ECs or (for now at least) download the submodules

> shouldn't checksums still work if you're pulling from a specific tag?

No because a plain clone (done by EB with `git_config`) isn't fully deterministic.

easybuild/easyconfigs/d/double-down/double-down-1.1.0-foss-2022a.eb
Flamefire: Yes

easybuild/easyconfigs/o/openmc/openmc-0.15.0_turn_off_submodule_check.patch
```diff
@@ -0,0 +1,48 @@
+diff --unified --recursive --text openmc-0.15.0.orig/CMakeLists.txt openmc-0.15.0.new/CMakeLists.txt
```
Flamefire: Is this patch from upstream? If so please include a link (and a description in general)

If not, wouldn't it be easier to just remove the check instead of this large patch (due to aligning existing options) and passing an extra configure flag? If the patch is just for EB the flexibility isn't necessary, is it?

LukeLabrie: I suppose you might want the flexibility so it can be left on when using git clone. But I've only run into this for EB so for now I've just simplified the patch and added a description.

sassy-crick:
@LukeLabrie You might want to sync your PR with development, something like this: 
```
```

boegel:

PR #20966, opened 2024-07-04
remove duplicate (explicit) versionsuffix from UCC-CUDA

Flamefire:
It is cleaner with less room for mistakes to use the existing variable instead of redefining the same string multiple times.

I suggest to do the same for all other ECs using a (CUDA) versionsuffix and dependencies using the same.


Thyre:
EB_ARGS="--installpath=/tmp/$USER/ecpr-20966"

boegel:
Going in, thanks @Flamefire!

PR #20999, opened 2024-07-11
add ToFu-1.7.9-gfbf-2023b.eb ToFu-1.7.9-iimkl-2023b.eb

Louwrensth:
and a tiny patch

### DIFF
--- a/easybuild/easyconfigs/t/ToFu/ToFu-1.7.9-gfbf-2023b.eb
+++ b/easybuild/easyconfigs/t/ToFu/ToFu-1.7.9-gfbf-2023b.eb
@@ -1,0 +1,50 @@
+easyblock = 'PythonBundle'
+
+name = 'ToFu'
+version = '1.7.9'
+
+homepage = 'https://github.com/ToFuProject/tofu'
+description = """Tomography for Fusion."""
+
+toolchain = {'name': 'gfbf', 'version': '2023b'}
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+    ('matplotlib', '3.8.2'),
+    ('astropy', '6.1.0'),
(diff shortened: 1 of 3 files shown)


sassy-crick:
@Louwrensth You might want to sync your PR with development, something like this: 
```
```

easybuild/easyconfigs/t/ToFu/ToFu-1.7.9-gfbf-2023b.eb
```diff
@@ -0,0 +1,48 @@
...
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+    ('matplotlib', '3.8.2'),
+    ('astropy', '6.1.0'),
+    ('PyYAML', '6.0.1'),
+]
+
+use_pip = True
```
sassy-crick, line 19: I don't think we need the `use_pip = True` any more, which is why the tests are failing

easybuild/easyconfigs/t/ToFu/ToFu-1.7.9-iimkl-2023b.eb
```diff
@@ -0,0 +1,48 @@
...
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+    ('matplotlib', '3.8.2'),
+    ('astropy', '6.1.0'),
+    ('PyYAML', '6.0.1'),
+]
+
+use_pip = True
```
sassy-crick, line 19: I don't think we need `use_pip = True` any more, which is why the tests are failing

sassy-crick:
It appears you are missing some dependencies:
`Missing dependencies: astropy/6.1.0-gfbf-2023b, astropy/6.1.0-iimkl-2023b `

Taken from the failing tests.

PR #21021, opened 2024-07-19
VEP-112 eb. Based on VEP-111.

sebastiaandvriend-bc:
Added eb for VEP-112 with GCC-12.2.0. This is copied from VEP-111-GCC-12.2.0.eb

### DIFF
--- a/easybuild/easyconfigs/v/VEP/VEP-112-GCC-12.2.0.eb
+++ b/easybuild/easyconfigs/v/VEP/VEP-112-GCC-12.2.0.eb
@@ -1,0 +1,42 @@
+name = 'VEP'
+version = '112'
+
+homepage = 'https://www.ensembl.org/info/docs/tools/vep'
+description = """Variant Effect Predictor (VEP) determines the effect of your
+ variants (SNPs, insertions, deletions, CNVs or structural variants) on genes,
+ transcripts, and protein sequence, as well as regulatory regions.
+ Includes EnsEMBL-XS, which provides pre-compiled replacements for frequently
+ used routines in VEP."""
+
+toolchain = {'name': 'GCC', 'version': '12.2.0'}
+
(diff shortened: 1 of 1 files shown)


sebastiaandvriend-bc:
Could a maintainer please have a look at this pull request? :-)

sassy-crick:
@sebastiaandvriend-bc 
There is already `VEP-111-GCC-12.2.0.eb` so the question is is that really required to have it in the same toolchain?
If so:
You might want to sync your PR with development, something like this: 
```
```

sebastiaandvriend-bc:
> @sebastiaandvriend-bc There is already `VEP-111-GCC-12.2.0.eb` so the question is is that really required to have it in the same toolchain? If so: You might want to sync your PR with development, something like this:
> 
> ```
> eb --sync-pr-with-develop NUMBER_OF_PR
> ```

@sassy-crick I've created this config last year for version 112, and based it on the config from version 111. I noticed that this specific version is not (yet) integrated in the main branch by someone else. Could this config still be useful?

sassy-crick:
> > @sebastiaandvriend-bc There is already `VEP-111-GCC-12.2.0.eb` so the question is is that really required to have it in the same toolchain? If so: You might want to sync your PR with development, something like this:
> > ```
> > eb --sync-pr-with-develop NUMBER_OF_PR
> > ```
> 
> @sassy-crick I've created this config last year for version 112, and based it on the config from version 111. I noticed that this specific version is not (yet) integrated in the main branch by someone else. Could this config still be useful?

@boegel  What do you think?

boegel:
Sorry for only picking up on this now...

This has been superseded by now, see for example:
- #24466

In addition, the `GCC/12.2.0` toolchain is now basically deprecated (will be reflected at https://docs.easybuild.io/policies/toolchains/#current-situation soon), so I don't see the point of putting more effort in this.

@sebastiaandvriend-bc I hope this won't discourage you from contributing...

PR #21030, opened 2024-07-22
update libarchive OpenSSL to 3, fix missing libxml dependency

CJCShadowsan:
Build fails without libxml dependency with an xmlParser error for -lxml2


sassy-crick:
@CJCShadowsan You might want to sync your PR with development, something like this: 
```
```

PR #21061, opened 2024-07-25
adding easyconfigs: HDF5-1.14.4.3-gompi-2023b.eb, HDF5-1.14.4.3-iimpi-2023b.eb

Louwrensth:
Because HDF5-1.14.3 is a broken release: https://github.com/HDFGroup/hdf5/issues/3831

### DIFF
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.14.4.3-gompi-2023b.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.14.4.3-gompi-2023b.eb
@@ -1,0 +1,31 @@
+name = 'HDF5'
+# Note: Odd minor releases are only RCs and should not be used.
+version = '1.14.4.3'
+
+homepage = 'https://portal.hdfgroup.org/display/support'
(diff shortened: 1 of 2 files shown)


boegel:
@Louwrensth As discussed during today's EasyBuild conf call, introducing another `HDF5` version in the `2023b` generation would be pretty painful at this point.

How about creating a patch file for the existing HDF5 1.14.3 easyconfigs based on the bug fix in https://github.com/HDFGroup/hdf5/pull/3837 instead?

Louwrensth:
> @Louwrensth As discussed during today's EasyBuild conf call, introducing another `HDF5` version in the `2023b` generation would be pretty painful at this point.
> 
> How about creating a patch file for the existing HDF5 1.14.3 easyconfigs based on the bug fix in [HDFGroup/hdf5#3837](https://github.com/HDFGroup/hdf5/pull/3837) instead?

Yep good suggestion. I'll create a new PR for it.

Louwrensth:
Closing this in favor of
- easybuilders/easybuild-easyconfigs/pull/21280

PR #21092, opened 2024-07-30
Salmon 1.10.1 GCC 12.2.0

w34sel:


smoors:
@w34sel Salmon-1.10.1-GCC-12.2.0.eb is already available in EasyBuild, and it looks like there are no important differences with your PR?

w34sel:
Sry didn‘t see that there was already one available

PR #21069, opened 2024-07-26
{ai,lang,vis}[foss/2022a] DeepDRR v1.1.3, PyCUDA v2024.1, PyVista v0.43.8 w/ CUDA 11.7.0

smoors:
(created using `eb --new-pr`)

PR #21115, opened 2024-08-03
{ai,vis}[foss/2022a] DiffDRR v0.3.12, TorchIO v0.19.6 w/ CUDA 11.7.0

smoors:
(created using `eb --new-pr`)

depends on:
- #21069


lexming:

lexming:
This cannot be merged due to a dependency conflict. Most easyconfigs in `foss/2022a` depend on PyTorch v1.12.0, but `DiffDRR` needs `PyTorch-bundle` and there is only v1.12.1 in this toolchain. IDK how we got into this situation, but it's too much work to fix at this point and we already have newer versions of DiffDRR in the repo.

PR #21136, opened 2024-08-06
Ensure that recent BioPerl easyconfigs use `Bundle` easyblock

ocaisa:
This was causing problems in https://github.com/EESSI/software-layer/pull/658 as `eb` was attempting to copy the same easyblock twice and that was failing if `EasyBuild` had a read-only installation.

### DIFF
--- a/easybuild/easyconfigs/b/BioPerl/BioPerl-1.7.8-GCCcore-12.2.0.eb
+++ b/easybuild/easyconfigs/b/BioPerl/BioPerl-1.7.8-GCCcore-12.2.0.eb
@@ -5,7 +5,7 @@
 #  Fred Hutchinson Cancer Research Center
 #  Thomas Eylenbosch - Gluo NV 
 
-easyblock = 'PerlModule'
+easyblock = 'Bundle'
 
 name = 'BioPerl'
 version = '1.7.8'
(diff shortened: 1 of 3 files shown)


ocaisa:

ocaisa:

ocaisa:

boegel (approved):
lgtm

boegel:
I guess the intention with using `PerlModule` rather than `Bundle` was to avoid that these easyconfigs need to set `PERL5LIB` via `modextrapaths`, but they're still doing it.

@ocaisa Can you clarify why using `PerlModule` rather than `Bundle` is actually causing trouble here?
What is `Bundle` doing differently that the "`Permission denied`" error that occurs when using `PerlModule` doesn't happen?

ocaisa:
The top-level easyblock is copied over to the `reprod` directory and then the extensions are installed. The extensions also copy their easyblocks to the `reprod` directory, but they are using the same easyblock. This is fine when the permissions on the original file allow it to be overwritten, but when you start from a read-only installation of EasyBuild you end up with a permissions issue.

The comment in https://github.com/easybuilders/easybuild-framework/issues/4602#issuecomment-2271082442 still stands, the framework could be a little more graceful here and recognise that the files have the same content (so the copy could be skipped).

boegel:
> The top-level easyblock is copied over to the `reprod` directory and then the extensions are installed. The extensions also copy their easyblocks to the `reprod` directory, but they are using the same easyblock. This is fine when the permissions on the original file allow it to be overwritten, but when you start from a read-only installation of EasyBuild you end up with a permissions issue. If we switch to `Bundle` as the top-level easyblock we sidestep the problem.
> 
> The comment in [easybuilders/easybuild-framework#4602 (comment)](https://github.com/easybuilders/easybuild-framework/issues/4602#issuecomment-2271082442) still stands, the framework could be a little more graceful here and recognise that the files have the same content (so the copy could be skipped).

OK makes sense now, thanks for clarifying!

boegel:
Going in, thanks @ocaisa!

PR #21143, opened 2024-08-07
Fix checksum checks for easyconfigs using a `Bundle`-like easyblock in easyconfigs test suite

Flamefire:
Some bundle easyconfigs have a checksums parameter. The current test code clears that unconditionally which makes the test fail when it should not.
As we know the easyconfig dict will be modified we can just make a copy here.


boegel (approved):
lgtm

PR #20898, opened 2024-06-25
{lang}[foss/2023b,gfbf/2023b] R v4.4.1, R-bundle-CRAN v2024.06

branfosj:
(created using `eb --new-pr`)

* [x] #20893
* [x] #20894
* [x] #20892

PR #20756, opened 2024-06-07
R-bundle-CRAN-2023b, Brunsli-0.1 patch, and easyconfigs.py 

moravveji:
In this PR:
- the latest version of R packages from CRAN are built with R/4.3.3-gfbf-2023b (see #20060)
- few additional dependency packages are added to the extension list
- the build succeeds on Intel Sapphire Rapids with Rocky 8.9 Linux. I will try it on Intel SkyLake, CascadeLake and IceLake soon

Remarks:
- As of `Brotli/1.1.0-GCCcore-13.2.0` (compared to v.1.0.9), the `libbrotli*-static.a` are no longer built. As a result of that, building `Brunsli/0.1-GCCcore-13.2.0` fails. To fix this, a new patch is introduced (`Brunsli-0.1-fix-Brotli-static-library-names.patch`)
- This PR introduces a new .eb for `libgeotiff, Brunsli, GEOS, GDAL`, and so does PR #20675. So, some care has to be practiced here to avoid duplication

PR #21181, opened 2024-08-14
Add some more packages to R-bundle-CRAN-2024.06-foss-2023b

Flamefire:
Extracted the remaining change from #20756 after that got missed when merging #20898


Flamefire:
Test report by @Flamefire
**SUCCESS**
i7003 - Linux Rocky Linux 8.9 (Green Obsidian), x86_64, AMD EPYC 7702 64-Core Processor (zen2), Python 3.8.17
See https://gist.github.com/Flamefire/3371ede393336858db2faef2f24abbb3 for a full test report.

PhoenixEmik:
I believe these packages exist in `R-4.4.1-gfbf-2023b`. I wonder why they need to be added to `R-bundle-CRAN-2024.06-foss-2023b` duplicately?

Flamefire:
You are right, not needed.

@Thyre Should the 2023b etc. label maybe moved to the bot to add it automatically?

Thyre:
> @Thyre Should the 2023b etc. label maybe moved to the bot to add it automatically?

I think that this would generally be a good idea, yeah. However, if I remember discussions correctly, we want to avoid parsing EasyConfigs in our labeling job due to permissions, so we would need to do this via file names only.

Flamefire:
> we want to avoid parsing EasyConfigs in our labeling job due to permissions, so we would need to do this via file names only.

Isn't the parsing already happening when searching for related easyconfigs?

Thyre:
> Isn't the parsing already happening when searching for related easyconfigs?

No, it is searching based on the parent folder of the EasyConfig, and uses `git.Repo` to compare the files.
See https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/.github/workflows/tagbot.py

PR #21189, opened 2024-08-15
{devel}[GCCcore/13.2.0] ReFrame v4.6.2

gkaf89:
(created using `eb --new-pr`)
ReFrame needs control of the user environment, so normally relies on the
system Python to avoid polluting the test and compilation environment
with preloaded modules. However, ReFrame now offers the `--purge-env`
option that clears the environment modules before launching any
compilation or test operations. Given that the toolchain Python provides
a much better control over the environment, this module installs ReFrame
using the toolchain Python; use with the `--purge-env` to avoid any
module compatibility issues.


casparvl:
I considered reviewing this, but did a deep-dive into how we install ReFrame and decided we should change quite a number of things (some of which come from this PR :)). Thus, I'm closing this in favour of my own https://github.com/easybuilders/easybuild-easyconfigs/pull/21269 - hope you don't mind @gkaf89 , it seemed faster than giving you feedback on all of those things I intended to see changed (plus, I had to experiment to see if things worked the way I think they did).

PR #21202, opened 2024-08-17
Provide SciPy-bundle-2023.12 and dependents for intel and foss alike

Louwrensth:
Unlike version `2023.11`, `SciPy-bundle-2023.12` works for both `gfbg/2023b` and `iimkl/2023b`

I found a couple of dependent softwares that didn't have their intel counterpart so I provided them.

Also, I figured that the foss versions should have the version of the `SciPy-bundle` dep bumped to match, but I wouldn't want to bother existing installations, so I provide alternatives using `versionsuffix = '-b'`

### DIFF
--- a/easybuild/easyconfigs/a/Armadillo/Armadillo-12.8.0-intel-2023b.eb
+++ b/easybuild/easyconfigs/a/Armadillo/Armadillo-12.8.0-intel-2023b.eb
@@ -1,0 +1,25 @@
+name = 'Armadillo'
+version = '12.8.0'
+
+homepage = 'https://arma.sourceforge.net/'
+description = """Armadillo is an open-source C++ linear algebra library (matrix maths) aiming towards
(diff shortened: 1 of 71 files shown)


sassy-crick:
@Louwrensth 
This is a rather large PR with too many dependencies in it. Try and keep the number below say 10 as that will make it easier for the maintainer to look at them. 
Furthermore, `SciPy-bundle-2023.11-gfbf-2023b.eb` has already been merged in `EasyBuild-5.1.0`. I would suggest to remove anything which has already been merged, change the versions of the software you need to build for the `intel` toolchain so they are in sync with what is in the `2023b` toolchains.
Then you might want to sync your PR with development, something like this: 
```
```
resolve any still existing conflicts and remove the `draft` label. You done quite a bit of work and right now there is no `intel` toolchain version of `SciPy-bundle`.

PR #21250, opened 2024-08-23
Create Mathematica-14.1.0.eb

jose-d:
new version of Mathematica.

part of this release was also partial rebranding.

needs https://github.com/easybuilders/easybuild-easyblocks/pull/3419

### DIFF
--- a/easybuild/easyconfigs/m/Mathematica/Mathematica-14.1.0.eb
+++ b/easybuild/easyconfigs/m/Mathematica/Mathematica-14.1.0.eb
@@ -1,0 +1,25 @@
+name = "Mathematica"
+version = "14.1.0"
+
+homepage = "https://www.wolfram.com/mathematica"
+docurls = "https://reference.wolfram.com/language/"
+description = """
+    Mathematica is a computational software program
+    used in many scientific, engineering, mathematical and computing fields.
+"""
+download_instructions = "Download from https://user.wolfram.com/portal/myProducts.html"
+postinstallmsgs = [
+    "This Mathematica/Wolfram installer includes Qt6. Please ensure that your compute nodes ",
+    "have the necessary dependencies available (see https://support.wolfram.com/59044), or alternatively, ",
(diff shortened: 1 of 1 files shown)


jose-d:
testreport copypaste from https://github.com/easybuilders/easybuild-easyblocks/pull/3419#issuecomment-2306653369 - done with our license server so I think I can not do `--from-pr`.

Test report by @jose-d
Overview of tested easyconfigs (in order)

    SUCCESS Mathematica-14.1.0.eb

cvmfs2.phoebe.lan - Linux Rocky Linux 8.10 (Green Obsidian), x86_64, AMD EPYC 73F3 16-Core Processor, Python 3.11.9
See https://gist.github.com/jose-d/413bba692824202a747211266300db2d for a full test report.

dithwick:
I don't think the original easyblock/easyconfig for Mathematica checks for OS packages. Not sure where is the right place to make those checks (maybe in the easyconfig) but the package list is at https://support.wolfram.com/59044

jose-d:
> I don't think the original easyblock/easyconfig for Mathematica checks for OS packages. Not sure where is the right place to make those checks (maybe in the easyconfig) but the package list is at https://support.wolfram.com/59044

well there are some GUI related deps.
anyway one can run mathematica without gui, and without need of listed deps.

In current version, the GUI binary is `wolframnb`, and the CLI only binary is `wolfram`.

sassy-crick:
The quickest was to get the X stuff seems to be just to load `Qt` and here I guess any version will do. I am wondering if adding something like that to the module file, so it is printed out when the module is loaded, might help?

jose-d:
@sassy-crick like this?  ↑

boegel:
@jose-d I changed to target branch in this PR from `5.0.x` to `develop`, you should synchronize your PR branch with current `develop` branch (which has received a massive update after the release of EasyBuild v5.0.0), for example using `eb --sync-pr-with-develop`.

jose-d:
closing as deprecated.

PR #21278, opened 2024-08-28
add `*.eb_bak_*`, `*.out`, `.venv/` to `.gitignore`

Louwrensth:
Such paths should never be committed:
 - `.eb_bak` files are remnants after `eb --inject-checksums`
 - `.out` files are log remnants after `eb --job`
 - `.venv/` are probably created for ad-hoc testing


boegel (approved):
lgtm

PR #21311, opened 2024-09-04
{tools}[GCCcore/13.3.0] versioningit v3.1.2, zlib-ng v2.2.1

Flamefire:
(created using `eb --new-pr`)

This contains the Python bindings for zlib-ng

However they have different versioning, so maybe we should rather have zlib-ng and zlib-ng-python like with protobuf?


smoors:
they are also in a different account on github, so it seems indeed better to not bundle the python bindings with zlib-ng.

the official name seems to be `python-zlib-ng` so i would take that

Flamefire:
Depends on where you look for the official name, on PyPI it is "zlib-ng": https://pypi.org/project/zlib-ng but with a "python-zlib-ng" header.

Confusing...

I've just seen that this has been merged as part of #21424 which is now in 4.9.4

I have prepared the split of the Python bindings to a new EC but it now requires modifying the now existing zlib-ng EC and pytest-workflow-2.1.0-GCCcore-13.3.0.eb which depends on the python bindings.

Do we still want to do that?

smoors:
> I have prepared the split of the Python bindings to a new EC but it now requires modifying the now existing zlib-ng EC and pytest-workflow-2.1.0-GCCcore-13.3.0.eb which depends on the python bindings.
> 
> Do we still want to do that?

the merged easyconfigs are not in a release yet, so it's still fine to modify them

Flamefire:
> the merged easyconfigs are not in a release yet, so it's still fine to modify them

They are in 4.9.4, aren't they? https://github.com/easybuilders/easybuild-easyconfigs/blob/easybuild-easyconfigs-v4.9.4/easybuild/easyconfigs/z/zlib-ng/zlib-ng-2.2.1-GCCcore-13.3.0.eb

Flamefire:
Closed by https://github.com/easybuilders/easybuild-easyconfigs/pull/21424 followup in #22379

PR #21325, opened 2024-09-05
{perf} OTF2: Use toolchain Python instead of system one

### LINKED ISSUE #21266: OTF2 is using system Python instead of Python version of toolchain
Since the release of Python 3.12, several people have noticed problems when trying to install OTF2, independently of the usage of EasyBuild or not. This is due to the removal of the `imp` module which was long deprecated. 

While working on https://github.com/easybuilders/easybuild-easyconfigs/pull/21244 and https://github.com/easybuilders/easybuild-easyconfigs/pull/21135, adding a patch to OTF2 for `2024a` worked fine, but the build on my private system failed even for `2023b`, where Python 3.11 is available.

Looking into it a bit more, I noticed that the OTF2 module uses the system Python.

https://github.com/easybuilders/easybuild-easyconfigs/pull/21244#discussion_r1731293358
```
Configure command:
  ./configure                   '--prefix=/data/EasyBuild-develop/software/OTF2/3.0.3-GCCcore-13.3.0' \
                                '--build=x86_64-pc-linux-gnu' \
                                '--host=x86_64-pc-linux-gnu' \
                                '--enable-shared' \
                                '--with-nocross-compiler-suite=gcc' \
                                '--with-libbfd-include=/data/EasyBuild-develop/software/binutils/2.42-GCCcore-13.3.0/include' \
                                '--with-libbfd-lib=/data/EasyBuild-develop/software/binutils/2.42-GCCcore-13.3.0/lib' \
                                '--with-sionlib=/data/EasyBuild-develop/software/SIONlib/1.7.7-GCCcore-13.3.0-tools/bin' \
                                'build_alias=x86_64-pc-linux-gnu' \
                                'host_alias=x86_64-pc-linux-gnu'

Loaded modules:
  module load                   GCCcore/13.3.0 \
                                zlib/1.3.1-GCCcore-13.3.0 \
                                binutils/2.42-GCCcore-13.3.0 \
                                SIONlib/1.7.7-GCCcore-13.3.0-tools

Configuration summary:
  OTF2 3.0.3:
    Platform:                   linux (auto detected)
    Cross compiling:            no (auto detected)

    OTF2 (backend):
      C99 compiler:             /data/EasyBuild-develop/software/GCCcore/13.3.0/bin/gcc (gnu 13.3.0)
      Link mode:                static=yes, shared=yes
      SIONlib support:          yes, using /data/EasyBuild-develop/software/SIONlib/1.7.7-GCCcore-13.3.0-tools/bin/sionconfig --be \
                                --gen --threadsafe --64
      GCC atomic builtins:      yes
      TLS support:              yes, using __thread and the initial-exec model
      Entropy sources:          clock_gettime getpid sysinfo gethostid
      Python bindings support:  yes, using /usr/bin/python3
      Running tests:            no
        Parallel tests:         yes 
```

Looking at an HPC system like JURECA-DC, OTF2 is also using the system Python:
```console
$ ls /p/software/jurecadc/stages/2024/software/OTF2/3.0.3-GCCcore-12.3.0/lib/
libotf2.a  libotf2.la  libotf2.so  libotf2.so.10  libotf2.so.10.0.0  pkgconfig  python3.6
$ module load Python OTF2/.3.0.3
$ which python3
/p/software/jurecadc/stages/2024/software/Python/3.11.3-GCCcore-12.3.0/bin/python3
$ python3 -c "import otf2"
ModuleNotFoundError: No module named 'otf2'
```

From my perspective, using the system Python is incorrect. Instead, the bindings should use the Python version of the toolchain instead.

Thyre:
Add Python as a dependency to OTF2 versions to ensure that the toolchain Python is used. To verify, a sanity check is added which ensures that the OTF2 Python module can be loaded. As `six` is required, newer toolchains also need Python-bundle-PyPI.

Resolves #21266

### DIFF
--- a/easybuild/easyconfigs/o/OTF2/OTF2-3.0.3-GCCcore-13.2.0.eb
+++ b/easybuild/easyconfigs/o/OTF2/OTF2-3.0.3-GCCcore-13.2.0.eb
@@ -3,6 +3,7 @@
 # Copyright:: Copyright 2013-2019 Juelich Supercomputing Centre, Germany
 # Authors::   Bernd Mohr <b.mohr@fz-juelich.de>
 # Markus Geimer <m.geimer@fz-juelich.de>
+# Jan Andre Reuter <j.reuter@fz-juelich.de>
 # License::   3-clause BSD
 #
 # This work is based on experiences from the UNITE project
@@ -34,17 +35,22 @@ builddependencies = [
 dependencies = [
     # SIONlib container support (optional):
     ('SIONlib', '1.7.7', '-tools'),
+    # OTF2 Python bindings
+    ('Python', '3.11.5'),
+    ('Python-bundle-PyPI', '2023.10'),
 ]
 
+local_pyshortver = '3.11'
 configopts = '--enable-shared'
 
+modextrapaths = {'PYTHONPATH': ['lib64/python%s/site-packages' % local_pyshortver]}
 
 sanity_check_paths = {
     'files': ['bin/otf2-config', 'include/otf2/otf2.h',
               'lib/libotf2.a', 'lib/libotf2.%s' % SHLIB_EXT],
     'dirs': [],
 }
-
-sanity_check_commands = ['%(namelower)s-config --help']
+sanity_check_commands = ['%(namelower)s-config --help',
+                         'python -c "import %(namelower)s"']
 
 moduleclass = 'perf'
--- a/easybuild/easyconfigs/o/OTF2/OTF2-3.0-GCCcore-11.3.0.eb
+++ b/easybuild/easyconfigs/o/OTF2/OTF2-3.0-GCCcore-11.3.0.eb
@@ -3,6 +3,7 @@
 # Copyright:: Copyright 2013-2019 Juelich Supercomputing Centre, Germany
 # Authors::   Bernd Mohr <b.mohr@fz-juelich.de>
 # Markus Geimer <m.geimer@fz-juelich.de>
+# Jan Andre Reuter <j.reuter@fz-juelich.de>
 # License::   3-clause BSD
 #
 # This work is based on experiences from the UNITE project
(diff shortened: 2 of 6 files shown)


SebastianAchilles:

SebastianAchilles:
CORE_CNT=16

SebastianAchilles (approved):
lgtm

SebastianAchilles:
Going in, thanks @Thyre!

easybuild/easyconfigs/o/OTF2/OTF2-3.0.3-GCCcore-13.2.0.eb
```diff
@@ -34,17 +35,22 @@ builddependencies = [
...
 
 sanity_check_paths = {
     'files': ['bin/otf2-config', 'include/otf2/otf2.h',
               'lib/libotf2.a', 'lib/libotf2.%s' % SHLIB_EXT],
     'dirs': [],
 }
-
-sanity_check_commands = ['%(namelower)s-config --help']
+sanity_check_commands = ['%(namelower)s-config --help',
+                         'python -c "import %(namelower)s"']
```
Flamefire, line 54: I would suggest to [use `-s`](https://docs.python.org/3/using/cmdline.html#cmdoption-s) here as otherwise the $HOME packages might be used causing failures/success:

```diff
-                         'python -c "import %(namelower)s"']
+                         'python -s -c "import %(namelower)s"']
``````

boegel: @Flamefire This is relevant for several easyblocks too where we use `python -c`?

Flamefire: Yes it is

Flamefire: -> https://github.com/easybuilders/easybuild-easyblocks/pull/3440

PR #21350, opened 2024-09-09
This is the easyconfig file for the new cp2k.py in PR(3433)

beeebiii:
Hi, this is the easyconfig of `CP2K-2024.2-foss-2023a-CUDA-12.2.0.eb` that i created using the easyblock pull-request(https://github.com/easybuilders/easybuild-easyblocks/pull/3433). This recipe has a successfull test of almost 4000 tests and only fails for 50 test.


beeebiii:
Now using https://github.com/easybuilders/easybuild-easyblocks/pull/3433, all the tests passed and i was able to build CP2K+GPU enabled installation. Although i had to do hardcode some stuff, because withouth those, it doesnot work!!

sassy-crick:
@beeebiii You might want to sync your PR with development, something like this: 
```
```

PR #21394, opened 2024-09-13
Fix test_easyconfig_locations when cache is present

Flamefire:
The test fails if the easyconfig cache file .eb-path-index is present. This is caused by checking all files instead of only easyconfigs. Fix that, directly ignore the TEMPLATE.eb and replace `assertTrue(False`


smoors (approved):
lgtm

PR #21405, opened 2024-09-13
Use pycodestyle in CI and fix style issues in EasyConfigs

Flamefire:
framework uses pycodestyle if available and pep8 only as a fallback and later likely not at all (pep8 was renamed to pycodestyle and the existing package is no longer updated)
So we need to check for either both or (more future proof) only for pycodestyle.

Install it in CI to actually test it and fix the found issues to make the test pass again.


boegel:
@Flamefire Seems like this needs a bit more love?

You seem to be touching easyconfigs that will no longer be included with EasyBuild v5.0.0, no point in putting effort in those imho...

boegel:
Doing this cleanup in `develop` branch is way too painful, so let's only do it for the `5.0.x` branch (which will very soon become `develop`), see #21502.

It's really only relevant for the `5.0.x` branch too, since https://github.com/easybuilders/easybuild-framework/pull/4634 was changed to target `5.0.x`

Flamefire:
Ok I see. However I did like the line breaks here more than the ones in #21502  resulting from the line wrapping. See https://github.com/easybuilders/easybuild-easyconfigs/pull/21502/files#r1782351340

PR #21416, opened 2024-09-16
take into list of crates for easyconfigs using Cargo-based easyblock when determining checksums for patches

boegel:
for for failing CI:
```
ERROR: test__parse_easyconfig_DeltaLake-0.15.1-gfbf-2023a.eb (test.easyconfigs.easyconfigs.EasyConfigTest)
Test for easyconfig DeltaLake-0.15.1-gfbf-2023a.eb
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/easybuild/tools/filetools.py", line 1274, in verify_checksum
    checksum = checksum[filename]
KeyError: 'DeltaLake-0.15.1_remove-obsolete-pyarrow-hotfix.patch'

During handling of the above exception, another exception occurred:

  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 1693, in innertest
    template_easyconfig_test(self, spec_path)
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 1533, in template_easyconfig_test
    if not verify_checksum(patch_full, checksum):
  File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/easybuild/tools/filetools.py", line 1276, in verify_checksum
    raise EasyBuildError("Missing checksum for %s in %s", filename, checksum)
easybuild.tools.build_log.EasyBuildError: "Missing checksum for DeltaLake-0.15.1_remove-obsolete-pyarrow-hotfix.patch in {'addr2line-0.21.0.tar.gz': '8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb'}"
```

This problem arose due to the change in https://github.com/easybuilders/easybuild-easyblocks/pull/3448, where `self.cfg = self.cfg.copy()` was added to the `Cargo` constructor before `self.cfg['sources']` is being changed...

The checksum for `DeltaLake-0.15.1_remove-obsolete-pyarrow-hotfix.patch` is already there in `DeltaLake-0.15.1-gfbf-2023a.eb`

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -40,6 +40,7 @@ from unittest import TestCase, TestLoader, main, skip
 import easybuild.main as eb_main
 import easybuild.tools.options as eboptions
 from easybuild.base import fancylogger
+from easybuild.easyblocks.generic.cargo import Cargo
 from easybuild.easyblocks.generic.configuremake import ConfigureMake
 from easybuild.easyblocks.generic.pythonpackage import PythonPackage
 from easybuild.framework.easyblock import EasyBlock
@@ -1507,6 +1508,12 @@ def template_easyconfig_test(self, spec):
                     failing_checks.append("OpenSSL should not be listed as OS dependency")
 
     src_cnt = len(ec['sources'])
+
+    # if easyblock derives from Cargo easyblock we need to take into account list of crates,
+    # which get added to list of sources, to correctly determine number of checksums for sources (incl. crates)
+    if isinstance(app, Cargo):
+        src_cnt += len(ec.get('crates', []))
+
     patch_checksums = ec['checksums'][src_cnt:]
 
     # make sure all patch files are available


test/easyconfigs/easyconfigs.py
```diff
@@ -1507,6 +1508,12 @@ def template_easyconfig_test(self, spec):
                     failing_checks.append("OpenSSL should not be listed as OS dependency")
 
     src_cnt = len(ec['sources'])
+
+    # if easyblock derives from Cargo easyblock we need to take into account list of crates,
+    # which get added to list of sources, to correctly determine number of checksums for sources (incl. crates)
+    if isinstance(app, Cargo):
+        src_cnt += len(ec.get('crates', []))
```
Flamefire, line 1515: Can't we use the actual list of sources as will be done by the checksums step?
I.e. `app.cfg['sources']` as discussed in the other PR/issue

Flamefire:
> where self.cfg = self.cfg.copy() was added to the Cargo constructor before self.cfg['sources'] is being changed...

Just for clarification: Actual issue is not the change of of `self.cfg` but the implicit change of the passed in `ec` instance we use here for the test. See the comment.

test/easyconfigs/easyconfigs.py
boegel: Not since the change in https://github.com/easybuilders/easybuild-easyblocks/pull/3448, because `ec['sources']` no longer contains also the crates since that change, that's exactly why this fix is required

Flamefire: Since that change `app.cfg['sources`]` should be populated with the crates while `ec['sources']` should not. Or did I miss anything? See the [comment](https://github.com/easybuilders/easybuild-easyblocks/pull/3448#issuecomment-2351656529)
> `x = app_class(EasyConfig(file_content))`
> `x.cfg <-- always do this`

Flamefire: I created an alternative PR: https://github.com/easybuilders/easybuild-easyconfigs/pull/21419

Taking the sources, patches and checksums from `app.cfg` there and in a single place.

PR #21419, opened 2024-09-16
account for crates for easyconfigs using Cargo-based easyblock when determining checksums for patches

Flamefire:
Fix for failing CI

> easybuild.tools.build_log.EasyBuildError: "Missing checksum for DeltaLake-0.15.1_remove-obsolete-pyarrow-hotfix.patch in {'addr2line-0.21.0.tar.gz': '8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb'}"

This problem arose due to the change in https://github.com/easybuilders/easybuild-easyblocks/pull/3448, where `self.cfg = self.cfg.copy()` was added to the Cargo constructor before `self.cfg['sources']` (was equal to the input `ec` instance) is being changed.

Closes #21416 which addresses the issue in a more specific way. This PR is more generic.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -1459,6 +1459,9 @@ def template_easyconfig_test(self, spec):
     ]
     failing_checks.extend("Old URL '%s' found" % old_url for old_url in old_urls if old_url in ec.rawtxt)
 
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

boegel:
Makes perfect sense, is indeed a better alternative to #21416, thanks a lot @Flamefire!

PR #21493, opened 2024-09-27
{bio}[SYSTEM] BayesTraits V4.1.3

asalois-msu:
Added a new version of Bayes Traits as it was released recently


sassy-crick:
@asalois-msu You might want to sync your PR with development, something like this: 
```
```

PR #21504, opened 2024-09-30
adding easyconfigs: Palabos-cavity3d-2.3.0-foss-2023a.eb

garadar:


sassy-crick:
@garadar You might want to sync your PR with development, something like this: 
```
```

PR #21568, opened 2024-10-07
Add easyconfig Archive-Zip-1.68-GCCcore-13.2.0.eb

julianmorillo:


bedroge:
Looks like this was recently added with PR https://github.com/easybuilders/easybuild-easyconfigs/pull/21952, so let's close this one.

PR #21540, opened 2024-10-04
Upgrade WCSLIB version in easyconfig casacore-3.5.0-foss-2023b.eb

julianmorillo:
tarball file of the previous version of WCSLIB is not available any more on the FTP server pointed by the easyconfigs

PR #21577, opened 2024-10-07
Upgrade casacore version in AOFlagger-3.4.0-foss-2023b.eb

julianmorillo:
Needs PR https://github.com/easybuilders/easybuild-easyconfigs/pull/21540 to work


sassy-crick:
@julianmorillo You might want to sync your PR with development, something like this: 
```
```
Also: is there a specific reason for that change, like a bug in the version which is shipped with `EasyBuild-5.1.0` ?

julianmorillo:
@sassy-crick , the thing is that WCSLIB 7.1 (dependency of casacore as it is shipped with `EasyBuild-5.1.0`) is not available for downloading anymore (see PR #21540)

sassy-crick:
I have opened an issue regarding the problem with the `wcslib-7.11.tar.bz2` download. Thanks for finding this.

Issue #20910, opened 2024-06-26
easyconfig file for installing Siesta 5.0


luciandf:
Hello all,

Has any of you installed Siesta 5.0 with easybuild? Can you share the eb file please?

Thank you

iortizdeluzu:
@luciandf Here I send you an easyconfic that works for our cluster. I hope it works also for you.

[Siesta-5.0.1-intel-2023a.txt](https://github.com/user-attachments/files/17071191/Siesta-5.0.1-intel-2023a.txt)

luciandf:
thank you so much! I will try it as soon as possible!

luciandf:
@iortizdeluzu so I tried your eb  file with the foss2023b toolchain. I have a more difficult problem of installing the intel2023a toolchain from your file.

I got quite far but it stopped with this error:

```
-- Searching for s-dftd3 - found
-- Adding support with SIESTA_WITH_FLOOK=TRUE
--   Searching for flook library
--     trying pkg-config
--     Checking for module 'flook>=0.8.4'
--       Package 'flook', required by 'virtual:world', not found
--     trying pkg-config - not found
--     Checking libraries (flookall) in FLOOK_ROOT=
--     Checking libraries (flookall) in FLOOK_ROOT= - not found
--     flook: will compile on-the-fly from sources: /home/builder/2024-software-install/f/flook/0.8.4.tar.gz
--     Searching for library [dl]
--     Searching for library [dl] - Found library [dl]: /usr/lib64/libdl.so
--     Searching for library [readline]
CMake Error at External/Lua-Engine/CMakeLists.txt:48 (find_library):
  Could not find _flook_lib_readline using the following names: readline
Call Stack (most recent call first):
  External/Lua-Engine/CMakeLists.txt:243 (flook_add_dependencies)


-- Configuring incomplete, errors occurred!
 (at easybuild/software/EasyBuild/4.9.4/lib/python3.6/site-packages/easybuild/tools/run.py:695 in parse_cmd_output)
== 2024-10-01 22:31:46,688 build_log.py:267 INFO ... (took 2 mins 7 secs)
== 2024-10-01 22:31:46,689 filetools.py:2025 INFO Removing lock /opt/ohpc/pub/thor/software/.locks/_opt_ohpc_pub_thor_software_Siesta_5.0.1-foss-2023b.lock...
[...]

PR #21605, opened 2024-10-09
{phys}[intel/2023a] Siesta 5.0.1

steenlysgaard:
Easyconfig file for Siesta 5.0.1 with the intel toolchain.

Based on the work here: https://github.com/easybuilders/easybuild-easyconfigs/issues/20910

It runs the tests but does not verify the results (as described here: https://docs.siesta-project.org/projects/siesta/en/latest/installation/build-manually.html#tests). If I enable the verification, some tests fail with tiny errors on some values. That's probably to be expected since, as far as I can see, the test references were obtained with GNU compilers.


Micket:
triggering the PR bot

Micket (approved):
lgtm

Micket:

PR #21640, opened 2024-10-14
Fix finding libraries when linking ImageMagick

Flamefire:
(created using `eb --new-pr`)

PR #21646, opened 2024-10-15
Add easyconfig ImageMagick-7.1.1-38-GCCcore-13.2.0.eb

julianmorillo:


Flamefire:
There is already an existing ImageMagick for 13.2.0: https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/i/ImageMagick/ImageMagick-7.1.1-34-GCCcore-13.2.0.eb

Have you missed that or do you need that specific version?

julianmorillo:
> There is already an existing ImageMagick for 13.2.0: https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/i/ImageMagick/ImageMagick-7.1.1-34-GCCcore-13.2.0.eb
> 
Yes, but when I tried this one, it gave me an error (probably related to the lack of patch: https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/i/ImageMagick/ImageMagick-7.1.1-38_fix-linking.patch). So, instead of replicating the patch for the 34 version I preferred to go for the 38 version.

Flamefire:
> Yes, but when I tried this one, it gave me an error (probably related to the lack of patch: https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/i/ImageMagick/ImageMagick-7.1.1-38_fix-linking.patch). So, instead of replicating the patch for the 34 version I preferred to go for the 38 version.

There is a policy to prefer one version of each software per toolchain generation (with some exceptions) because introducing another can lead to conflicts when different other modules depend on different versions of that software. There is also a CI check in place to detect that. This PR introduces EasyConfigs where we (likely) already have another version for 13.2.0. I guess you copied the EasyConfigs from 13.3.0?

In this case the patch applied cleanly to 7.1.1-34 and as I ran into this issue too while testing something else I added it to the other ImageMagick ECs in #21640 That likely didn't exist when you created your ECs. Can you check if that works for you? You can do `eb --from-pr 21640 ImageMagick-7.1.1-34-GCCcore-13.2.0.eb` to just install that single module.

Hope that helps :)

julianmorillo:
> > Yes, but when I tried this one, it gave me an error (probably related to the lack of patch: https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/i/ImageMagick/ImageMagick-7.1.1-38_fix-linking.patch). So, instead of replicating the patch for the 34 version I preferred to go for the 38 version.
> 
> There is a policy to prefer one version of each software per toolchain generation (with some exceptions) because introducing another can lead to conflicts when different other modules depend on different versions of that software. There is also a CI check in place to detect that. This PR introduces EasyConfigs where we (likely) already have another version for 13.2.0. I guess you copied the EasyConfigs from 13.3.0?
> 
I see, I did not know about such policy. And yes, I took the easyconfig for 13.3.0 and just adapted for 13.2.0.

> In this case the patch applied cleanly to 7.1.1-34 and as I ran into this issue too while testing something else I added it to the other ImageMagick ECs in #21640 That likely didn't exist when you created your ECs. Can you check if that works for you? You can do `eb --from-pr 21640 ImageMagick-7.1.1-34-GCCcore-13.2.0.eb` to just install that single module.
> 
> Hope that helps :)
I already used `--from-pr` option to install the 38 version using my own PR #21646 and it worked just fine. So, as I already have ImageMagick correctly installed and this PR does not look useful to the community, I'll close it.

Thank you for letting me know about the version policy :)

Flamefire:
> Thank you for letting me know about the version policy :)

I hope I could explain the reasoning behind that well enough, i.e. the possibly conflicting versions when loading multiple modules.

> And yes, I took the easyconfig for 13.3.0 and just adapted for 13.2.0.

A suggestion for that: If you start with an EC and change the toolchain version then before creating ECs for its dependencies check for existing ECs in that toolchain first and just adapt the version in the `dependencies` list. Not only avoids this the issue with possibly multiple versions of the same software in a toolchain it also is less work, as you don't have to create more ECs than necessary.

There is also `eb --try-toolchain-version <xxx> --try-update-deps` that partially automates that, i.e. it searches for existing ECs while updating the toolchain. Together with `--copy-ec <destination>` this provides a good starting point.

PR #21656, opened 2024-10-15
Update easyconfigs.py - add 'alt_location' to exts patches

pavelToman:
For now when you add 'alt_location' to the patch in exts_list it fails.
See https://github.com/easybuilders/easybuild-easyconfigs/pull/21141/files# and https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/10453456099/job/28943965086?pr=21141

PR #21700, opened 2024-10-18
fix CI check for extension patches using `alt_location`

Flamefire:
While `alt_location` can be used for patches in the main EasyConfig part and in extensions the verification code handled them differently and missed to take the `alt_location` into account for extension patches.

Factor our the checking code to a function that can be used for both cases.
This also removes a redundant condition in one of the code instances before verifying the checksums.

More general solution of #21656


boegel (approved):
lgtm

PR #21709, opened 2024-10-21
{numlib}[GCC/13.3.0] LAPACK 3.12.0

jkuhl-uni:
Hey,
I needed LAPACK in the 2024a version, so I used try-toolchain for it :)

### DIFF
--- a/easybuild/easyconfigs/l/LAPACK/LAPACK-3.12.0-GCC-13.3.0.eb
+++ b/easybuild/easyconfigs/l/LAPACK/LAPACK-3.12.0-GCC-13.3.0.eb
@@ -1,0 +1,17 @@
+name = 'LAPACK'
+version = '3.12.0'
+
+homepage = 'https://www.netlib.org/lapack/'
+description = """LAPACK is written in Fortran90 and provides routines for solving systems of
+ simultaneous linear equations, least-squares solutions of linear systems of equations, eigenvalue
+ problems, and singular value problems."""
+
+toolchain = {'name': 'GCC', 'version': '13.3.0'}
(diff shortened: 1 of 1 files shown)


Micket:
Just curious as to what needs this. Obviously, one should *never* use this slow reference implementation for anything, and just rely on OpenBLAS, BLIS, MKL (which of course include the lapack interface as well).

smoors:
> I have a question: This does not include the LAPACKE headers. Can we configure it to also include those, as there is no LAPACKE easyonfig?

LAPACKE is included in the `FlexiBLAS` easyconfig,

jkuhl-uni:
Hi there, alright, thank you for the response. In fact, I noticed later, that the program I was compiling is linking against LAPACK directly. It is only an optional dependency for [GRID](https://github.com/paboyle/Grid/tree/develop). It uses it for very simple matmul operations.

Micket:
It should absolutely just link to flexiblas instead, like everything that needs lapack/blas

jkuhl-uni:
That may be. If I get a chance, I'll talk to a dev of it. I think no one has bothered as only a very small amount of execution time is spent in non-specialized matrix operations, but maybe it is worth a look into it, thanks!

Micket:
Normally we just tell the software to use the correct include and lib files during the configure step and doesn't allow for setting "-lopenblas" like normal then we often patch it to use the correct linking for EB.

jkuhl-uni:
Ah, alright. Thanks for the tip. I'll try to recompile the software with that base when I have to next time.
In that case, if you want the ec here, you can merge it, otherwise never mind. I would see the PR as closed, no worries :)

PR #21758, opened 2024-10-28
Add a github workflow that tags PRs with new/update/change + add diffs in comments

Micket:
I made a workflow that runs in a few seconds and can automatically adjust the new/update/change labels based on what is done in the PR. 
In addition, when software is updated, it prints diffs in summary blocks comparing them up to 3 other existing easyconfigs (based on how recently they were added to git, which I found to be a very good estimate of what the best comparisons are). 
The comment and labels are updated each time the workflow runs (reusing the existing comment as to not spam)


Micket:
I have some stuff to sort out which is difficult to test locally, so I'll make a few commits here to see how the bot reacts.

.github/workflows/tagbot.yml
```diff
@@ -0,0 +1,31 @@
+name: Tagbot
+on: [pull_request]
+
+permissions:
+  pull-requests: write
```
VRehnberg, line 5:
```suggestion
  contents: read
  pull-requests: write
```
don't you need read permissions as well?

Micket: I've tested this workflow separately in a junk repo and that was enough to specify. I'm just having some troubles getting the correct git checkout here when i have remote branches (working on it)

Micket:
> Failed to add labels: 403, {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/labels#add-labels-to-an-issue","status":"403"}
> Failed to post comment: 403, {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment","status":"403"}

I believe this is a security feature; PRs are limited to read permissions even if the yml specifies more. Apparently even so with PRs from maintainers i guess. Not sure if it's possible to manually rerun the workflow with elevated privileges, or one is actually forced to merge it first, which really sucks.

Micket:
Since I can't demonstrate the workflow here until it's merged; here are some examples of it in action

https://github.com/Micket/easybuild-easyconfigs/pull/21
https://github.com/Micket/easybuild-easyconfigs/pull/22

showing the tags and comment update

.github/workflows/tagbot.yml
```diff
@@ -0,0 +1,30 @@
+name: Tagbot
+on: [pull_request]
+
+jobs:
+  tagbot:
+    runs-on: ubuntu-20.04
```
ocaisa, line 6: Can we use a more recent OS, this is likely to be deprecated soon?

ocaisa (changes requested):
OS upgrade requested. I know you were also thinking of limiting the number of comments for various scenarios.

I'm ready to merge this, would be good to remove the easyconfig from the PR and add it in a follow-up instead.

.github/workflows/tagbot.yml
Micket: Done. I also reworked some code in order to limit the number of diffs. Will make some PRs on my own branch for testing to ensure i didn't mess it up, I'll comment again with links when I've had time to do that

.github/workflows/tagbot.py
```diff
@@ -0,0 +1,181 @@
...
+elif len(to_diff) > 10:
+    max_diffs_per_software = 1
+elif len(to_diff) > 5:
+    max_diffs_per_software = 2
+else:
+    max_diffs_per_software = 3
+
+if max_diffs_per_software > 0:
+    for new_file, neighbours in to_diff:
+        compare_neighbours = neighbours[:max_comparisons]
```
ocaisa, line 107:
```suggestion
        compare_neighbours = neighbours[:max_diffs_per_software]
```

Micket:
@ocaisa After an embarrassing number of typos and such

https://github.com/Micket/easybuild-easyconfigs/pull/22
https://github.com/Micket/easybuild-easyconfigs/pull/23

it finally works as intended.

ocaisa (approved):
LGTM, tiny issue with how paths are printed (see https://github.com/Micket/easybuild-easyconfigs/actions/runs/11615802064/job/32347253957?pr=22#step:6:18) but that's a detail not worth holding this back for (I'm sure there will be more requests once this is being used anyway)

PR #20997, opened 2024-07-11
{bio}[GCC/13.2.0] Bowtie2 2.5.4

fizwit:
(created using `eb --new-pr`)

PR #21860, opened 2024-11-15
Add easyconfig Bowtie2-2.5.1-GCC-13.2.0.eb

julianmorillo:


branfosj:
We merged `Bowtie2/2.5.4` in #20997 so I'll close this PR. If the RISCV5 patches are needed for that version then please add a PR against that version instead.

PR #21903, opened 2024-11-22
{vis}[GCCcore 12.3.0] JasPer 2.0.33 for WPS and WRF

asalois-msu:
Older JasPer on foss 2023a for WPS and WRF


ocaisa:
@stefan-wolfsheimer has actually included the patch in https://github.com/easybuilders/easybuild-easyconfigs/pull/21938/files#diff-d148887cb1e575c48524153439da2563aecb708279db2ccb8dbafc723d1251b8

PR #21684, opened 2024-10-17
{tools,vis}[GCCcore/13.2.0] Spyder v6.0.1, Deprecated v1.2.14, PyQt-builder v1.16.4, ...

pavelToman:
(created using `eb --new-pr`)

PR #21930, opened 2024-11-28
{vis,lang}[GCCcore/13.2.0] PyQt5 v5.15.11, SIP v6.8.6, PyQt-builder v1.16.4 w/ Python/3.11.5

zdemat:
PyQt5 for GCCore/13.2.0 with minor version updates.


zdemat:
@jhein32 I am actually waiting what will happen with #21684 (Spyder for 2023b) that they are actually reviewing already. I think we can adjust most of the `silx` and `nanomax-analysis-utils` stuff to use PyQt5 v5.15.10 from there.

For potential reviewers: If #21684 goes in, I can withdraw this PR.

PR #21968, opened 2024-12-04
Add easyconfig CGAL-5.6.1-GCCcore-13.2.0.eb

julianmorillo:


bedroge:
The version number in the title is different from the one in the easyconfig. But since an easyconfig for CGAL 5.6.1 is already available for 2023b, I'm closing this PR.

PR #21984, opened 2024-12-06
add collection of easyconfig templates organized per generation

lexming:
The goal of this collection of easyconfigs templates is to provide examples that are more ready to be used than the default `TEMPLATE.eb`. Having templates for common types of easyconfigs per generation has the advantage that many parameters can already be set in the template, such as versions of dependencies. This avoids wasting time looking up for that information every time a new easyconfig needs to be created.

We generally have two types of templates per type of easyconfig, a _minimal_ one and a _extensive_ one. The reason is that some people prefer to delete stuff from the template, while others prefer to start small and add stuff to it.

Templates are installed under `contrib/easyconfig-templates` alongside the example hooks in framework under `contrib/hooks`.

This PR also adds a README file about the template collection:
https://github.com/lexming/easybuild-easyconfigs/blob/templates/contrib/easyconfig-templates/README.rst


contrib/easyconfig-templates/2023a/PythonBundle-scipy.eb.tmpl
```diff
@@ -0,0 +1,39 @@
+# Template for a python bundle using SciPy-bundle on 2023a
```
smoors, line 1: can we add the toolchain name and version to the file name?

contrib/easyconfig-templates/2023a/PythonBundle-scipy.eb.tmpl
```diff
@@ -0,0 +1,39 @@
...
+dependencies = [
+    ('Python', '3.11.3'),
+    ('SciPy-bundle', '2023.07'),
+]
+
+use_pip = True
+
+exts_list = [
+    ('ext-normal', 'ext_version', {
+        'checksums': ['0000000000000000000000000000000000000000000000000000000000000000'],
```
smoors, line 23: usually checksums are added with `--inject-checksums` so this is not really needed

contrib/easyconfig-templates/2023a/PythonBundle-scipy.eb.tmpl
```diff
@@ -0,0 +1,39 @@
...
+    }),
+    ('ext-weird-name', 'ext_version', {
+        'modulename': 'import_name',
+        'checksums': ['0000000000000000000000000000000000000000000000000000000000000000'],
+    ('ext-wheel', 'ext_version', {
+        'source_tmpl': '%(name)s-%(version)s-py3-none-any.whl',
+        'checksums': ['0000000000000000000000000000000000000000000000000000000000000000'],
+    ('name-lower', 'ext_version', {
```
smoors, line 31: can we add a comment here that 'name-lower' should be the name used on PyPI, i.e. the name you use when you `pip install` it?

also, 'ext_version' can be replaced with `version`

contrib/easyconfig-templates/2022a/PythonBundle-scipy.eb.tmpl
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+toolchain = {'name': 'foss', 'version': '2022a'}
+
+dependencies = [
+    ('Python', '3.10.4'),
+    ('SciPy-bundle', '2022.05'),
+]
+
+use_pip = True
```
boegel, line 19: @lexming This is enabled by default in EasyBuild 5.0.0, so this line should be removed

boegel (changes requested):

contrib/easyconfig-templates/2022a/Bundle-common-components.eb.tmpl
```diff
@@ -0,0 +1,52 @@
+# Template for a bundle of components from a common organisation and using the same easyblock on GCCcore 11.3.0
+easyblock = 'Bundle'
+
+name = 'SoftwareName'
+version = '0.0.0'
+
+homepage = 'https://www.domain.org'
+description = """
+Description in 80 chars long column
```
verdurin, line 9: I thought the informal rule was keep to around 100 characters for description?

contrib/easyconfig-templates/2022a/Bundle-common-components.eb.tmpl
```diff
@@ -0,0 +1,52 @@
...
+    }),
+]
+
+sanity_check_paths = {
+    'files': ['bin/comp2', 'bin/comp3', f'lib/libcomp1.{SHLIB_EXT}'],
+    'dirs': ['include/comp1', 'include/comp2', 'include/comp3'],
+}
+
+sanity_check_commands = [
+    "comp2 -h",
```
verdurin, line 48: I think it would be useful to include the pattern where you need to grab the result of stderror from the command, which is quite common.

Thyre:
Do we need to add these EasyConfigs to the filtered ones in the test suite?
https://github.com/easybuilders/easybuild-easyconfigs/blob/b5baa0528782e82ac5fe14ce690a172b4b572f95/test/easyconfigs/easyconfigs.py#L222

verdurin:
Would be useful to have templates for binary/tarball ECs too.

contrib/easyconfig-templates/2023a/PythonBundle-scipy.eb.tmpl
lexming: I thought about this, but it is very messy because the names of the templates still need to have some description of what type of installation they cover. And mixing both results in very long filenames that are difficult to read.

Moreover, since the template has to be renamed in any case, there is little gain by already having the toolchain in the filename.

contrib/easyconfig-templates/2023a/PythonBundle-scipy.eb.tmpl
lexming: right, fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/21984/commits/42561bba065f26f74df737580b21abc6f09ed92f

contrib/easyconfig-templates/2023a/PythonBundle-scipy.eb.tmpl
lexming: We decided that we don't want comments in templates because they will probably stay in the final easyconfig.

So I tried to make this more clear without comments in https://github.com/easybuilders/easybuild-easyconfigs/pull/21984/commits/e43ce3d12109acce8aeaaffab5b194515bdaaa05

contrib/easyconfig-templates/2022a/PythonBundle-scipy.eb.tmpl
lexming: Fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/21984/commits/40cff25e1d5e98997b285b45520ee5c980b2133d

contrib/easyconfig-templates/2022a/Bundle-common-components.eb.tmpl
lexming: I prefer 80 because that is easier to read in the final module file. 100 characters is a rather long line.

Moreover, 80 is the default in `vi` to cut text in columns.

contrib/easyconfig-templates/2022a/Bundle-common-components.eb.tmpl
lexming: sure, added those in https://github.com/easybuilders/easybuild-easyconfigs/pull/21984/commits/c9c090497809baa09ea2e6725dc935b7c7bba397

lexming:
@Thyre those can indeed be removed, done in https://github.com/easybuilders/easybuild-easyconfigs/pull/21984/commits/ff1e21cbf23ca1e79044811cccb786dde2c8191d

@verdurin added templates for Tarballs in https://github.com/easybuilders/easybuild-easyconfigs/pull/21984/commits/bbf9699a09684d89b81e307ba17a240760cce0d3 and (Packed)Binary in https://github.com/easybuilders/easybuild-easyconfigs/pull/21984/commits/28270ef15205b266bf9d4bd5b9d31de53142593e

branfosj (approved):

PR #22002, opened 2024-12-10
Add easyconfig Critic2-1.2-foss-2023b.eb

julianmorillo:


Crivella:
I think this has been already added in

- https://github.com/easybuilders/easybuild-easyconfigs/pull/22677

Would we have a reason to have the newer version of libcint used for building?

julianmorillo:
I have no reason from my side so I will close the PR.

PR #22028, opened 2024-12-13
{lib}[foss/2024a] dm-haiku v0.0.13 w/ CUDA 12.6.0

ThomasHoffmann77:
(created using `eb --new-pr`)

- [ ] #21924
- [x] #22476


WilleBell:
Maybe this can use [jax 0.6.2](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/j/jax/jax-0.6.2-gfbf-2024a.eb) instead ?

PR #22052, opened 2024-12-18
sync with develop (20241219) + version bump to `5.0.0beta1`

boegel:

PR #22061, opened 2024-12-19
Fix test failure due to unresolved template

Flamefire:
Fixes e.g.
```
  File "test/easyconfigs/easyconfigs.py", line 1047, in test_r_libs_site_env_var
  if 'R_LIBS' in ec['ec'][key]:
  ...
easybuild.tools.build_log.EasyBuildError: 'Failed to resolve all templates in "/path/to/AlphaFold_DBs/%(versions)s" using template dictionary
```

Similar for other changes

Drive-by fix for `r_libs_ecs.append(ec['spec'])` that could append the same EC twice.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -988,37 +988,32 @@ class EasyConfigTest(TestCase):
             ]:
                 continue
 
-            if (('download_instructions' in ec and ec['download_instructions']) or ('crates' in ec and ec['crates']) or
-                    ('channels' in ec and ec['channels']) or ('source_urls' in ec and ec['source_urls'])):
-                continue
+            with ec.disable_templating():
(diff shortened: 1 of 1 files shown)


test/easyconfigs/easyconfigs.py
```diff
@@ -1027,7 +1022,7 @@ def test_sanity_check_paths(self):
         """Make sure specified sanity check paths adher to the requirements."""
 
         for ec in self.parsed_easyconfigs:
-            ec_scp = ec['ec'].get_ref('sanity_check_paths')
+            ec_scp = ec['ec'].get('sanity_check_paths', resolve=False)
```
Flamefire, line 1025: This and below is from the resolved conflict. Kept it as IMO `resolve=False` makes the intent clearer

Flamefire:
@boegel This is (now) a follow-up to https://github.com/easybuilders/easybuild-easyconfigs/pull/22052

boegel (approved):

PR #22069, opened 2024-12-19
archive easyconfigs and patches on the SYSTEM toolchain older than 5Y

lexming:
Cutoff set to 2019-12-31


easybuild/easyconfigs/g/Gubbins/Gubbins-2.4.0.eb
```diff
@@ -30,7 +30,7 @@ channels = ['r', 'conda-forge', 'bioconda']
 requirements = '%(namelower)s=%(version)s python=3.6'
 
 dependencies = [
-    ('Miniconda3', '4.7.10', '', SYSTEM)
+    ('Miniconda3', '4.8.3', '', SYSTEM)
```
Micket, line 33: Any value in keeping this old Gubbins easyconfig at all? This is very old, and we have a compiled version instead these days in Gubbins-3.3.5-foss-2022b.eb, so having a conda variant is not very valuable. I'd say it's OK to drop this one as a side-effect.

lexming: Ok, agreed. I just kept it because it is less than 5y old technically. But it can be removed indeed.

lexming: done in https://github.com/easybuilders/easybuild-easyconfigs/pull/22069/commits/16a3f00d76d28ed903f08f91ddbd5366c8e3409e

Micket (approved):
lgtm

PR #22088, opened 2024-12-21
disable input prompts in GO, Net-SSLeay and MIME-Lite extensions of Perl-bundle-CPAN

lexming:
(created using `eb --new-pr`)

The installation of `Perl-bundle-CPAN` is currently broken on `5.0.x`. The extension `GO`, `Net-SSLeay` and `MIME-Lite` ask for question during installation.

Example for `Net-SSLeay`:
```
eb-shell> export OPENSSL_PREFIX=$EBROOTOPENSSL &&  perl Makefile.PL PREFIX=/user/brussel/101/vsc10122/easybuild/install/skylake/software/Perl-bundle-CPAN/5.38.2-GCCcore-13.3.0 --help
Do you want to run external tests?
These tests *will* *fail* if you do not have network connectivity. [n] n
*** If there are build errors, test failures or run-time malfunctions,
    try to use the same compiler and options to compile your OpenSSL,
    Perl, and Net::SSLeay.
*** Found OpenSSL-3.2.2 installed in /user/brussel/101/vsc10122/easybuild/install/skylake/software/OpenSSL/3
Generating a Unix-style Makefile
Writing Makefile for Net::SSLeay
Writing MYMETA.yml and MYMETA.json
eb-shell> exit
```

Somehow this seems to have worked in EB 4.9, although I don't see any QA defined for these PerlModules. Anyhow, this PR fixes this issues without black magic, by just patching the `Makefile.PL` of those package to use fixed settings and not ask any questions.


lexming:
EB_BRANCH=5.0.x

boegel:
@lexming I'm not seeing any hangs or input prompts being produced at all, so I think it's worth triggering why you're hitting this while others are not...

easybuild/easyconfigs/p/Perl-bundle-CPAN/Perl-bundle-CPAN-5.36.1-GCCcore-12.3.0.eb
```diff
@@ -2056,22 +2071,6 @@ exts_list = [
         'source_urls': ['https://cpan.metacpan.org/authors/id/W/WS/WSNYDER'],
         'checksums': ['3973ebdc44682c9c15c776f66e8be242cb4ff1dd52caf43ff446b74d4dccca06'],
     }),
-    ('Sys::Info::Driver::Linux::Device::CPU', '0.7905', {
```
boegel, line 2059: @lexming This is being removed here because it's also included later again, but it really must be here, before `Sys::Info`, or the installation of that extension will break.

boegel, line 2059: @lexming This is being removed here because it's also included later again, but it really must be here, before `Sys::Info`, or the installation of that extension will break.

boegel (changes requested):

lexming:
Indeed, this is not the way forward. There are too many perl modules that actually use questions in their installs to be handled individually. Moreover, this should actually work with the existing easyconfigs, as it is the case in v4.9.

Closing to switch directions. Follow-up in #22094

PR #22116, opened 2025-01-05
test suite changes for EB5

branfosj:
* checksums are now checked in the fetch step
* `SYSTEM` toolchain check had a conditional on archived in EB5 toolchains


Micket (approved):
lgtm

PR #22137, opened 2025-01-09
added easyconfig for byobu 6.12 GCC 12.3.0

ctuni:
This is my first time contributing to the repo or creating an easyconfig file, so I would be thankful for any feedback on good practices when contributing to the code stack.

I took inspiration from the previous byobu eb file that existed and updated so it works on environments with newer GCC version.

### DIFF
--- a/easybuild/easyconfigs/b/byobu/byobu-6.12-GCC-12.3.0.eb
+++ b/easybuild/easyconfigs/b/byobu/byobu-6.12-GCC-12.3.0.eb
@@ -1,0 +1,56 @@
+# #
+# This is a contribution from Flomics Biotech SL (http://flomics.com)
+# Copyright::   Flomics Biotech SL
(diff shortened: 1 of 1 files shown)


ctuni:
Is there any way to be able to make the pending checks run? Just to be able to merge/modify if the tests pass or fail @Thyre thanks!

Thyre:
I would recommend syncing the branch e.g. via `eb --sync-pr-with-develop=22137`, since there were a few additions since you created the PR. Hopefully this then triggers the CI to run normally.

ctuni:
> I would recommend syncing the branch e.g. via `eb --sync-pr-with-develop=22137`, since there were a few additions since you created the PR. Hopefully this then triggers the CI to run normally.

Thank you for your help! I synced the branch and I think now it's way more up to date than how it was before but some checks remain with "pending" status

PR #22146, opened 2025-01-12
Adding cryolo v1.9.9's easyconfig

EricWong23:
Compiled and tested with EasyBuild 4.9.4 on Centos7.


Crivella:
This is not the way we usually install python packages with dependencies.
We use the `PythonBundle` easyblock.

See for example

- https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/a/aiohttp/aiohttp-3.9.5-GCCcore-13.2.0.eb
- https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/s/SciPy-bundle/SciPy-bundle-2023.07-gfbf-2023a.eb

boegel:

PR #19825, opened 2024-02-09
{math}[GCC/13.2.0,intel-compilers/2023.2.1] libcerf v2.4

SimonPinches:
(created using `eb --new-pr`)

PR #22162, opened 2025-01-16
{math}[GCC/13.2.0,intel-compilers/2023.2.1] libcerf-Fortran: fortran bindings for libcerf, 2023b toolchains

hattom:
Update of #17702 to newer libcerf version and newer toolchains.

Same as #19825, but in a separate package for Fortran.


easybuild/easyconfigs/l/libcerf-Fortran/libcerf-Fortran-2.4-GCC-13.2.0.eb
```diff
@@ -0,0 +1,43 @@
...
+    ('libcerf', version),
+]
+
+skipsteps = ['configure']
+start_dir = 'fortran/ccerflib_f95_interface'
+build_cmd = (
+    '$FC $FFLAGS -c use_libcerf_mod.f90 && '
+    '$FC -shared use_libcerf_mod.o -o libcerf-Fortran.so && '
+    'ar rc use_libcerf_mod.a use_libcerf_mod.o && '
+    'echo ""'
```
boegel: @hattom Why is this `echo` command here?

Is that some kind of hack to ignore exit code of the `ar` command? 🤔

boegel:

easybuild/easyconfigs/l/libcerf-Fortran/libcerf-Fortran-2.4-GCC-13.2.0.eb
hattom: It's a terrible hack because I've got a list of commands that need to run, but don't actually want to run `make`.
In the now-closed other PR, I added a comment along the lines of "not sure whether to patch a makefile or keep the lines that need to execute". Suggestions welcome.

hattom: Perhaps it does nothing. I don't remember, but I think that's *why* it ended up in there.

hattom: Plausibly to avoid the `-j4` being added to the `ar` command.

easybuild/easyconfigs/l/libcerf-Fortran/libcerf-Fortran-2.4-GCC-13.2.0.eb
```diff
@@ -0,0 +1,43 @@
...
+
+dependencies = [
+    ('libcerf', version),
+]
+
+skipsteps = ['configure']
+start_dir = 'fortran/ccerflib_f95_interface'
+build_cmd = (
+    '$FC $FFLAGS -c use_libcerf_mod.f90 && '
+    '$FC -shared use_libcerf_mod.o -o libcerf-Fortran.so && '
```
hattom: It's perhaps worth noting that I made up the name `libcerf-Fortran.so`. At the very least, I should have chosen the same name for the `.a` file and the `.so` file.
Whether these should be called `use_libcerf_mod.X` or `libcerf-Fortran.X` seems non-obvious to me.

hattom:
Suggest to change to `CmdCp` as I'd already done in the now-closed https://github.com/easybuilders/easybuild-easyconfigs/pull/17702.
The only open question then is the name of the `.so` and `.a` files.

PR #22258, opened 2025-02-04
unset `$BUILD_VERSION` set by `torchvision` easyblock in `preinstallopts` for `torchaudio` in easyconfigs for `PyTorch-bundle` 2.1.2

pavelToman:
There is problem with version of torchaudio - it has `version` instead of `2.1.2` in easyconfig, but it not get right version.
See: https://github.com/easybuilders/easybuild-framework/issues/4706

### DIFF
--- a/easybuild/easyconfigs/p/PyTorch-bundle/PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb
+++ b/easybuild/easyconfigs/p/PyTorch-bundle/PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb
@@ -101,7 +101,7 @@ exts_list = [
         ),
         'testinstall': True,
     }),
-    ('torchaudio', version, {
+    ('torchaudio', '2.1.2', {
         'installopts': "--no-use-pep517 -v",
         'patches': [
             'torchaudio-2.1.2_use-external-sox.patch',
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/p/PyTorch-bundle/PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb
```diff
@@ -101,7 +101,7 @@ exts_list = [
         ),
         'testinstall': True,
     }),
-    ('torchaudio', version, {
+    ('torchaudio', '2.1.2', {
```
boegel, line 104: @pavelToman this won't help, you need:
```python
'preinstallopts': "unset BUILD_VERSION && "

**Review by boegel — CHANGES_REQUESTED** [review]

File: easybuild/easyconfigs/p/PyTorch-bundle/PyTorch-bundle-2.1.2-foss-2023a.eb [hunk 2]
```diff
@@ -99,12 +99,13 @@ exts_list = [
         'testinstall': True,
     }),
     ('torchaudio', version, {
+        'preinstallopts': "unset BUILD_VERSION && ",
```
boegel, line 102: remove duplicate `preinstallopts`:
```suggestion
```

boegel (changes requested):

boegel:

boegel (approved):

PR #22260, opened 2025-02-05
Fiji archive file changed to .zip

wdlingit:
A minor change that Fiji archive files changed to *.zip and no more *.tar.gz

### DIFF
--- a/easybuild/easyconfigs/f/Fiji/Fiji-20240605-1417.eb
+++ b/easybuild/easyconfigs/f/Fiji/Fiji-20240605-1417.eb
@@ -1,0 +1,51 @@
+easyblock = 'PackedBinary'
+
+name = 'Fiji'
+version = '20240605-1417'
+
+homepage = 'https://fiji.sc/'
+description = """Fiji is an image processing package—a 'batteries-included' distribution of
+ ImageJ, bundling a lot of plugins which facilitate scientific image analysis.
+This release is based on ImageJ-2.1.0 and Fiji-2.1.1"""
+
+toolchain = SYSTEM
+
+source_urls = ['https://downloads.imagej.net/%(namelower)s/archive/%(version)s']
+sources = [{
+    'download_filename': '%(namelower)s-nojre.zip',
+    'filename': '%(namelower)s-nojre-%(version)s.zip',
+}]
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/f/Fiji/Fiji-20240605-1417.eb
```diff
@@ -0,0 +1,51 @@
+easyblock = 'PackedBinary'
+
+name = 'Fiji'
+version = '20240605-1417'
```
Micket, line 4: I think we should using this. Their github do tag releases:
https://github.com/fiji/fiji/tags
https://github.com/fiji/fiji/archive/refs/tags/fiji-2.16.0.tar.gz

so they do have actual version numbers

easybuild/easyconfigs/f/Fiji/Fiji-20240605-1417.eb
```diff
@@ -0,0 +1,51 @@
...
+toolchain = SYSTEM
+
+source_urls = ['https://downloads.imagej.net/%(namelower)s/archive/%(version)s']
+sources = [{
+    'download_filename': '%(namelower)s-nojre.zip',
+    'filename': '%(namelower)s-nojre-%(version)s.zip',
+}]
+checksums = ['d9329f79f548db1d74143146215b02f18a005549e57847cf8df3bedb2905ab12']
+
+dependencies = [('Java', '1.8', '', SYSTEM)]
```
Micket, line 20: 1.8 is the old Oracle java, please avoid using it. Based on earlier PRs, at least JAva 11 should work for this software. So, maybe 11 or 17?

Micket (changes requested):

easybuild/easyconfigs/f/Fiji/Fiji-20240605-1417.eb
MaximeVdB: I think the fiji repo just defines how it bundles ImageJ and its plugins. So to use those releases one would need to do the actual bundling to obtain the type of releases available at https://downloads.imagej.net/fiji/archive/. Not sure how difficult that would be, but at least there would be more to it than just changing the sources.

PR #22287, opened 2025-02-11
[Upd] Qt6 with snappy v1.1.10

maxim-masterov:
Update Qt6 to use snappy v1.1.10 instead of v1.2.1. Driven by the fact that many other packages in 2024a already depend on snappy v1.1.10.


maxim-masterov:
@Micket oh, great. to me it also makes more sense to use the newer version.

maxim-masterov:
I can confirm that at the moment only `Arrow-17.0.0-gfbf-2024a.eb` and `MariaDB-11.7.0-GCC-13.3.0.eb` have dependencies on `snappy-1.1.10`. @Micket will you create a PR?

Micket:
If you do it i can review it :)

maxim-masterov:
done: https://github.com/easybuilders/easybuild-easyconfigs/pull/22333

PR #22323, opened 2025-02-17
Add easyconfig jq-1.6-GCCcore-13.2.0.eb

julianmorillo:


bedroge:

bedroge:
This is a really old version (released in 2018), and while they hadn't released newer versions for several years, 1.7 was released in 2023. I think we should stick to 1.7.0/1.7.1 for this toolchain, and there's actually a PR for 1.7.1: https://github.com/easybuilders/easybuild-easyconfigs/pull/22979. So I'm closing this PR in favor of that one.

PR #22333, opened 2025-02-18
[Upd] [GCC/13.3.0,gfbf/2024a] Update Arrow and MariaDB with snappy v1.2.1

maxim-masterov:
As discussed here: https://github.com/easybuilders/easybuild-easyconfigs/pull/22287#issuecomment-2651399818


Micket (approved):
LGTM (for reference, this was the least painful change we needed to make to avoid a conflict between this and qt6)

Micket:

Micket:

PR #22386, opened 2025-02-25
Add missing OS dependency for static glibc to ccache & TopHat easyconfigs

Flamefire:
`static: True` requires static-glibc


boegel:

boegel:
Hmm, looks like several of those older `ccache` easyconfigs rely on (build) dependencies that are not compatible with Rocky Linux 9.x...

ocaisa:
Closing this since GCC 11.2.0 is deprecated or no longer supported, see https://docs.easybuild.io/policies/toolchains

Sorry for not getting back to this @Flamefire 

If this is still relevant, please consider opening a new pull request using a more recent toolchain.

PR #22259, opened 2025-02-04
{ai,tools}[foss/2023a] WhisperX v3.3.1, pyannote.audio v3.3.2 w/ CUDA 12.1.1

pavelToman:
(created using `eb --new-pr`)
resolves https://github.com/vscentrum/vsc-software-stack/issues/482
require:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/22258

PR #22135, opened 2025-01-08
{ai,devel,math,vis}[foss/2023a] whisper-ctranslate2 v0.5.2, ONNX-Runtime v1.19.2, PortAudio v19.7.0, ... w/ CUDA 12.6.0

pavelToman:
(created using `eb --new-pr`)
resolves https://github.com/vscentrum/vsc-software-stack/issues/483
requiers:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/22134

replaces: 
- https://github.com/easybuilders/easybuild-easyconfigs/pull/22124

PR #22134, opened 2025-01-08
{ai,tools}[GCCcore/12.3.0,foss/2023a] CTranslate2 v4.5.0, cpu_features v0.9.0 w/ CUDA 12.6.0

pavelToman:
(created using `eb --new-pr`)
dependency of WhisperX: https://github.com/vscentrum/vsc-software-stack/issues/482
dependency of whisper-ctranslate2: https://github.com/vscentrum/vsc-software-stack/issues/483

This PR replaces: https://github.com/easybuilders/easybuild-easyconfigs/pull/22119

PR #22394, opened 2025-02-25
{ai,lang}[foss/2023a] Scikit-HEP v2024.8.1, whisper-ctranslate2 v0.5.2 w/ CUDA 12.1.1

sassy-crick:
(created using `eb --new-pr`)

This PR creates a Python-bundle for the High Energy Physics Python meta [package](https://pypi.org/user/scikit-hep/).
The already existing Python-bundle was used as a template. 
One issue is that `awkward-cpp` is using `ninja` to build the software, but there is no job-control right now. That should be fixed.
Another restriction is the reliance on PyTorch, which is determining the toolchain-version and thus some of the version of the packages.


pavelToman:

pavelToman:

sassy-crick:

pavelToman:
Are you sure that the soundfile.patch is needed? I have something similar in my [PR](https://github.com/easybuilders/easybuild-easyconfigs/pull/22259) and the `soundfile-0.12.1` works after I added `libsndfile-1.2.2` and `cffi-1.15.1`.

easybuild/easyconfigs/s/Scikit-HEP/Scikit-HEP-2024.8.1-foss-2023a-CUDA-12.1.1.eb
```diff
@@ -0,0 +1,172 @@
...
+            {'soundfile-0.12.1.patch': 'aef2eed95780de6cc5350de0f6153edb12d75e9703ebf6ee66285f5e9103b7de'},
+        ],
+    }),
+    ('plu', '0.1.18', {
+        'checksums': ['c2f03582db2f4030e459325f233c5133e2f067b91c7e3e15e36e55d900d93200'],
+    }),
+    ('particle', '0.25.0', {
+        'checksums': ['8e2d5fa36555e6af218b66e97b9585b1d4f52085785d96c067736f0b2e57f5ad'],
+    }),
+    ('graphviz', '0.20.3', {
```
pavelToman, line 129: there is graphviz-python in EB repo: https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/g/graphviz-python/graphviz-python-0.20.1-GCCcore-12.3.0.eb

sassy-crick:
> Are you sure that the soundfile.patch is needed? I have something similar in my [PR](https://github.com/easybuilders/easybuild-easyconfigs/pull/22259) and the `soundfile-0.12.1` works after I added `libsndfile-1.2.2` and `cffi-1.15.1`.

I added the suggested `cffi-1.15.1` and removed the patch for the build. However, and that is the problem where the bot tipped over, that does not seem to work as expected:
```
$ python
Python 3.11.3 (main, Jan 19 2024, 22:53:25) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import _soundfile_data
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named '_soundfile_data'
>>> import soundfile
>>> import _soundfile_data
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named '_soundfile_data'
```
So I would suggest to keep the patch in for now. After all, my reading of the code was it is looking for `libsndfile_x86-64.so` which is not present. What is present is `libsndfile.so`. What is interesting is that on my machine the check 'works' whereas clearly not on the test-build.

easybuild/easyconfigs/s/Scikit-HEP/Scikit-HEP-2024.8.1-foss-2023a-CUDA-12.1.1.eb
sassy-crick: I did not see that. Given my version is newer, is there any harm in using the version I used?

pavelToman: I think the good  EB practice is to use already exist easyconfig - if you really need another version you can put it in exts_list as you did. There should be only one version per toolchain to avoid dependency hell.

pavelToman:
> > Are you sure that the soundfile.patch is needed? I have something similar in my [PR](https://github.com/easybuilders/easybuild-easyconfigs/pull/22259) and the `soundfile-0.12.1` works after I added `libsndfile-1.2.2` and `cffi-1.15.1`.
> 
> I added the suggested `cffi-1.15.1` and removed the patch for the build. However, and that is the problem where the bot tipped over, that does not seem to work as expected:
> 
> ```
> $ python
> Python 3.11.3 (main, Jan 19 2024, 22:53:25) [GCC 12.3.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import _soundfile_data
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ModuleNotFoundError: No module named '_soundfile_data'
> >>> import soundfile
> >>> import _soundfile_data
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ModuleNotFoundError: No module named '_soundfile_data'
> ```
> 
> So I would suggest to keep the patch in for now. After all, my reading of the code was it is looking for `libsndfile_x86-64.so` which is not present. What is present is `libsndfile.so`. What is interesting is that on my machine the check 'works' whereas clearly not on the test-build.

If I got it, the `_soundfile_data` is needed only to build a binary wheel not to run python-soundfile. It only needs `libsndfile`, `CFFI` and `NumPy`?
https://github.com/bastibe/python-soundfile?tab=readme-ov-file#building

sassy-crick:
> > > Are you sure that the soundfile.patch is needed? I have something similar in my [PR](https://github.com/easybuilders/easybuild-easyconfigs/pull/22259) and the `soundfile-0.12.1` works after I added `libsndfile-1.2.2` and `cffi-1.15.1`.
> > 
> > 
> > I added the suggested `cffi-1.15.1` and removed the patch for the build. However, and that is the problem where the bot tipped over, that does not seem to work as expected:
> > ```
> > $ python
> > Python 3.11.3 (main, Jan 19 2024, 22:53:25) [GCC 12.3.0] on linux
> > Type "help", "copyright", "credits" or "license" for more information.
> > >>> import _soundfile_data
> > Traceback (most recent call last):
> >   File "<stdin>", line 1, in <module>
> > ModuleNotFoundError: No module named '_soundfile_data'
> > >>> import soundfile
> > >>> import _soundfile_data
> > Traceback (most recent call last):
> >   File "<stdin>", line 1, in <module>
> > ModuleNotFoundError: No module named '_soundfile_data'
> >     
> >   
> > So I would suggest to keep the patch in for now. After all, my reading of the code was it is looking for `libsndfile_x86-64.so` which is not present. What is present is `libsndfile.so`. What is interesting is that on my machine the check 'works' whereas clearly not on the test-build.
> 
> If I got it, the `_soundfile_data` is needed only to build a binary wheel not to run python-soundfile. It only needs `libsndfile`, `CFFI` and `NumPy`? https://github.com/bastibe/python-soundfile?tab=readme-ov-file#building

From what I understand, `Scikit-HEP-2025.02-foss-2023a-CUDA-12.1.1.eb` wants to build `libsoundfile` itself, rather then falling back to any provided `libsoundfile.so`. Happy to be told otherwise but here I think it is best to leave the patch as I don't think it does any harm.

sassy-crick:
Test report can be found [here](https://gist.github.com/sassy-crick/2aec41c4ad8f0c11cfc1b61b60a603a1)

sassy-crick:

sassy-crick:

easybuild/easyconfigs/s/Scikit-HEP/Scikit-HEP-2024.8.1-foss-2023a-CUDA-12.1.1.eb
boegel: @sassy-crick I doubt the minor version difference is going to matter much, and I agree it's better to use the existing easyconfig as a proper dependency

sassy-crick: Done. I have replaced the `'graphviz', '0.20.3'` extension with the `'graphviz-python', '0.20.1'` we already have. I have also added `Ninja` although it will be pulled in anyway. 
The only annoying thing is: `awkward-cpp-35` uses `ninja` for building, and that uses all available resources. I could not work out where that is being dragged in.

sassy-crick:

easybuild/easyconfigs/s/Scikit-HEP/Scikit-HEP-2024.8.1-foss-2023a-CUDA-12.1.1.eb
```diff
@@ -0,0 +1,165 @@
...
+versionsuffix = '-CUDA-%(cudaver)s'
+
+homepage = 'https://python.org/'
+description = "Bundle of Python packages for HEP, based on Scikit-HEP meta package"
+
+toolchain = {'name': 'foss', 'version': '2023a'}
+toolchainopts = {'pic': True}
+
+builddependencies = [
+    ('binutils', '2.40'),
```
akesandgren: Don't add binutils to buildeps unless toolchain is GCCcore, please remove

akesandgren (changes requested):

easybuild/easyconfigs/s/Scikit-HEP/Scikit-HEP-2024.8.1-foss-2023a-CUDA-12.1.1.eb
sassy-crick: Thanks for finding this. I have removed that.

sassy-crick:

Crivella:
Deleted the test report to not block this, missed the need for GPU for some of the deps installations

easybuild/easyconfigs/w/whisper-ctranslate2/whisper-ctranslate2-0.5.2-foss-2023a-CUDA-12.1.1.eb
```diff
@@ -0,0 +1,54 @@
...
+    ('tokenizers', '0.15.2'),
+    ('PyAV', '11.0.0'),
+    ('ONNX-Runtime', '1.19.2', versionsuffix),
+    ('cffi', '1.15.1'),
+    ('PortAudio', '19.7.0'),
+]
+
+exts_list = [
+    ('ctranslate2', '4.5.0', {
+        'source_tmpl': '%(name)s-%(version)s-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl',
```
boegel, line 26: We can avoid using a pre-built (generic) binary wheel, see:
- #22134
- #22135

@sassy-crick Are you up for looking into that?
Should hopefully be easy based on the existing `CTranslate2` easyconfig for CUDA 12.6.0...

boegel (changes requested):

easybuild/easyconfigs/w/whisper-ctranslate2/whisper-ctranslate2-0.5.2-foss-2023a-CUDA-12.1.1.eb
sassy-crick: Thanks for finding that. Let me give it a try.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 2 out of 2 (total: 21 mins 6 secs) (2 easyconfigs in total)
node3902.accelgor.os - Linux RHEL 9.6, x86_64, AMD EPYC 7413 24-Core Processor (zen3), 1 x NVIDIA NVIDIA A100-SXM4-80GB, 590.48.01, Python 3.9.21
See https://gist.github.com/boegel/518bb36a9225f149d243c1602c3e0633 for a full test report.

akesandgren:
@sassy-crick See Kenneths comment

sassy-crick:
> @sassy-crick See Kenneths comment

@akesandgren I am now lost to be honest. :-(  Which comment do you mean?

akesandgren:
@sassy-crick  The one from 2026-02-11
"@sassy-crick Are you up for looking into that?
Should hopefully be easy based on the existing CTranslate2 easyconfig for CUDA 12.6.0..."

sassy-crick:
> @sassy-crick The one from 2026-02-11 "@sassy-crick Are you up for looking into that? Should hopefully be easy based on the existing CTranslate2 easyconfig for CUDA 12.6.0..."

@akesandgren  Thanks. The problem right now is the failing build-step for PR #25257 . Any ideas here? I think the only difference is the different `CUDA` version.

akesandgren:
No idea on the problem in #25257

easybuild/easyconfigs/w/whisper-ctranslate2/whisper-ctranslate2-0.5.2-foss-2023a-CUDA-12.1.1.eb
sassy-crick: It looks like this particular version of `ctranslate` and the `CUDA` version I need to use as I need to use `PyTorch` don't like each other as the builds fail (see PR #25257).
Thus, I will leave the `whl` file in there. A newer build of `Scikit-HEP` with a newer version of `PyTorch` will hopefully address this issue.

sassy-crick:

PR #22439, opened 2025-03-03
Use f-strings for MATLAB download instructions

Micket:
This wasn't updated from the 4.9.4 target when merged into 5.0.x, lets be consistent.


boegel (approved):
lgtm

PR #22457, opened 2025-03-04
update workflows to use Ubuntu 22.04

boegel:


.github/workflows/unit_tests.yml
```diff
@@ -10,27 +10,27 @@ concurrency:
 
 jobs:
   test-suite:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        python: [3.6, '3.11']
```
boegel, line 16: Removing testing with Python 3.6 for now, since that's not supported in Ubuntu 22.04

Once the problems with the custom workflow being added in https://github.com/easybuilders/easybuild-framework/pull/4790 are figured out, we can restore testing with Python 3.6...

lexming (approved):
LGTM

PR #22459, opened 2025-03-05
Update GitHub actions runners

Flamefire:
Don't use Ubuntu 20.04 anymore and update actions


boegel (approved):
lgtm

See also https://github.com/easybuilders/easybuild-easyconfigs/pull/22457 (for `5.0.x` branch)

PR #22472, opened 2025-03-06
Update GitHub actions workflows to use ubuntu-22.04 or ubuntu-24.04

Flamefire:
Rebase of https://github.com/easybuilders/easybuild-easyconfigs/pull/22459 on 5.0x

Use newer Ubuntu for linting to avoid having to update it too soon

Also test with Python 3.7 instead of 3.9 as minimum version of 5.0x is 3.6

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -12,10 +12,10 @@ jobs:
   python-linting:
     runs-on: ubuntu-24.04
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
 
     - name: set up Python
-      uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


.github/workflows/linting.yml
```diff
@@ -10,12 +10,12 @@ concurrency:
 
 jobs:
   python-linting:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
+    - uses: actions/checkout@v4
```
boegel, line 15: Using a specific commit is deliberate, out of abundance of care.

We'll only update to newer versions/commits when really needed, so please change back.

boegel (changes requested):

boegel (approved):

PR #2499, opened 2016-02-17
style fixes in n easyconfigs

boegel:
wrapper PR for #2392 by @wpoely86 (additional style fixes in NCO, ncurses, nettle easyconfigs)

PR #22497, opened 2025-03-07
Add easyconfig MetaEuk-6-GCC-13.2.0.eb

julianmorillo:


easybuild/easyconfigs/m/MetaEuk/MetaEuk-6-rv64gc.patch
```diff
@@ -0,0 +1,20 @@
...
++        set(RISCV64 1) 
+     else ()
+         message(WARNING "CPU without native SIMD instructions. Performance will be bad.")
+     endif ()
+@@ -113,6 +115,8 @@
+         set(MMSEQS_ARCH "-mcpu=native")
+     elseif (ZARCH)
+         set(MMSEQS_ARCH "-mcpu=native -mzvector")
++    elseif (RISCV64)
++        set(MMSEQS_ARCH "-march=rv64gc")
```
bedroge: https://github.com/easybuilders/easybuild-easyconfigs/pull/22499#pullrequestreview-2931985684 also applies here, we need some better way to inject the right flags (at the moment we only support generic RISC-V builds, but that may change in the near future). Maybe we can use `$OPTFLAGS` here?

julianmorillo: Applied the same solution as in #2499.

PR #22499, opened 2025-03-07
Add easyconfig MMseqs2-14-7e284-gompi-2023b.eb

julianmorillo:


bedroge:
@julianmorillo we already have `MMseqs2-15-6f452-gompi-2023b.eb` for 2023b, would it be possible to add your patches to that version?

edit: looks like the `cstdint` patches are no longer necessary for version 15, see https://github.com/soedinglab/MMseqs2/blob/15-6f452/src/commons/Util.h and https://github.com/soedinglab/MMseqs2/blob/15-6f452/src/prefiltering/Indexer.h.

easybuild/easyconfigs/m/MMseqs2/MMseqs2-14-7e284_riscv.patch
```diff
@@ -0,0 +1,20 @@
...
++	set(RISCV64 1)
+     else ()
+         message(WARNING "CPU without native SIMD instructions. Performance will be bad.")
+     endif ()
+@@ -113,6 +115,8 @@
+         set(MMSEQS_ARCH "-mcpu=native")
+     elseif (ZARCH)
+         set(MMSEQS_ARCH "-mcpu=native -mzvector")
++    elseif (RISCV64)
++        set(MMSEQS_ARCH "-march=rv64gc")
```
bedroge: We'll also need a way to handle this better, as this will always build generic binaries.

julianmorillo: I have committed a new patch that is much more generic and considers the ISA of the target machine, so if it is vector capable, it will automatically include the "v" option in the argument of "-march"

Crivella:
Test report by @Crivella
**SUCCESS**
See https://gist.github.com/Crivella/5ea64085a3f6573055aa77ae72481c3d for a full test report.

Crivella:
@julianmorillo Can we get a test report also form a RISC-V system (possibly also the `configure_step` part of the logs to show it is properly picking up the `-march`?

julianmorillo:
@Crivella , here is the log of a complete build in a RISC-V system:
[easybuild-MMseqs2-15-6f452-20251017.172702.log](https://github.com/user-attachments/files/22974866/easybuild-MMseqs2-15-6f452-20251017.172702.log)

Crivella:
It seems you used `--optarch=GENERIC` in your build
```
== 2025-10-17 16:57:05,031 compiler.py:375 INFO _set_optimal_architecture: using -march=rv64gc -mabi=lp64d as optarch for RISC-V-64/RISC-V.
```

which is causing 2 march options to appear

```
[  1%] Building C object lib/microtar/CMakeFiles/microtar.dir/microtar.c.o
cd /tmp/jmorillo/easybuild/build/MMseqs2/15-6f452/gompi-2023b/easybuild_obj/lib/microtar && /tmp/eb-3xy0zok1/tmpj7akgcx6/rpath_wrappers/gcc_wrapper/gcc --sysroot=/cvmfs/riscv.eessi.io/versions/20240402/compat/linux/riscv64  -I/tmp/jmorillo/easybuild/build/MMseqs2/15-6f452/gompi-2023b/MMseqs2-15-6f452/lib/zstd/lib -I/tmp/jmorillo/easybuild/build/MMseqs2/15-6f452/gompi-2023b/MMseqs2-15-6f452/lib/tinyexpr -I/tmp/jmorillo/easybuild/build/MMseqs2/15-6f452/gompi-2023b/MMseqs2-15-6f452/lib/microtar -O2 -ftree-vectorize -march=rv64gc -mabi=lp64d -fno-math-errno -O3 -DNDEBUG -fsigned-char -march=rv64imafdc -MD -MT lib/microtar/CMakeFiles/microtar.dir/microtar.c.o -MF CMakeFiles/microtar.dir/microtar.c.o.d -o CMakeFiles/microtar.dir/microtar.c.o -c /tmp/jmorillo/easybuild/build/MMseqs2/15-6f452/gompi-2023b/MMseqs2-15-6f452/lib/microtar/microtar.c
```

The second `-march=rv64imafdc` I assume is what is coming from your patch.

I am wondering if the patch itself would not be needed without running EB with `--optarch=GENERIC`.
In particular i think proper detection for this should be handled in framework and not on an EC/EB basis

julianmorillo:
Yes, I'm using `--optarch=GENERIC` because it is needed for the moment, as it is a build for EESSI (this will disappear in the near future). 

You are right, the second `-march=rv64imafdc` is the one coming from my patch, and it is absolutely needed, either you run with or without  `--optarch=GENERIC` because without the patch it would be `-march=native`, causing a compilation error as "native" is not a valid value in RISC-V builds.

So it is simply a problem in the CMakeLists.txt of the MMSeqs that it is not considering at all RISC-V architectures, and the flow falls into an 
```
else ()
         set(MMSEQS_ARCH "-march=native")
```
which serves for most of the architectures, but not for RISC-V. I do not think any change is needed in framework.

Crivella:
> which serves for most of the architectures, but not for RISC-V. I do not think any change is needed in framework.

What i meant is that it would be nice to pass the correct `-march=XXX` directly from framework when `'optarch': True` is used as a TC option similar to how we pass `-march=native` for the others

https://github.com/easybuilders/easybuild-framework/blob/0c10c1370eb6e41c371ceccd979a170edf77cb80/easybuild/toolchains/compiler/gcc.py#L91-L99

but that would be for another PR

> So it is simply a problem in the CMakeLists.txt of the MMSeqs that it is not considering at all RISC-V architectures, and the flow falls into an

Did not realize the problem was also in the CMake falling back to `native` at the end of the if/else chain, so the patch is needed regardless.
Just to check for future proofing (not an expert on risc-v): is the `isa` section of cpuinfo specifically targeted to match the `-march` flags for GCC?
Other worry might be that i assume cpuinfo is tied to the kernel so it could be possible to have a new enough kernel that we detect isa flags that are not yet supported by a version of GCC?

julianmorillo:
> > which serves for most of the architectures, but not for RISC-V. I do not think any change is needed in framework.
> 
> What i meant is that it would be nice to pass the correct `-march=XXX` directly from framework when `'optarch': True` is used as a TC option similar to how we pass `-march=native` for the others
> 
> https://github.com/easybuilders/easybuild-framework/blob/0c10c1370eb6e41c371ceccd979a170edf77cb80/easybuild/toolchains/compiler/gcc.py#L91-L99
> 
> but that would be for another PR

Oh, now I see what you meant... yes, it makes total sense. I will work on that and submit a PR.

> 
> > So it is simply a problem in the CMakeLists.txt of the MMSeqs that it is not considering at all RISC-V architectures, and the flow falls into an
> 
> Did not realize the problem was also in the CMake falling back to `native` at the end of the if/else chain, so the patch is needed regardless. Just to check for future proofing (not an expert on risc-v): is the `isa` section of cpuinfo specifically targeted to match the `-march` flags for GCC? 

Yes, the isa field in /proc/cpuinfo on RISC-V is intended to correspond closely to the -march string used by GCC and LLVM

>Other worry might be that i assume cpuinfo is tied to the kernel so it could be possible to have a new enough kernel that we detect isa flags that are not yet supported by a version of GCC?

I would say that it is very unlikely but yes, it could be possible, mainly for rare extensions.

PR #22545, opened 2025-03-12
{tools}[GCCcore/13.3.0] gettext-0.23, gettext-0.23.1

mdefende:
Added recipes for gettext 0.23 and 0.23.1 with both system and GCCcore/13.3.0 toolchains

### DIFF
--- a/easybuild/easyconfigs/g/gettext/gettext-0.23-GCCcore-13.3.0.eb
+++ b/easybuild/easyconfigs/g/gettext/gettext-0.23-GCCcore-13.3.0.eb
@@ -1,0 +1,34 @@
+easyblock = 'ConfigureMake'
+
+name = 'gettext'
+version = '0.23'
+
+homepage = 'https://www.gnu.org/software/gettext/'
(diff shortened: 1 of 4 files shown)


Crivella:
@mdefende 
I think you used the same template for making both the SYSTEM and GCCcore variants, but as far as i can see, the GCCcore versions that have already been merged use a different set of dependencies and configopts, EG:
- https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/g/gettext/gettext-0.22.5-GCCcore-13.3.0.eb
- https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/g/gettext/gettext-0.24-GCCcore-14.2.0.eb

Could you uniform your PR to those or add comments in the ECs to justify why that is being done differently?

mdefende:
If I remember correctly, I changed to setup in gettext 0.23.1 to be compatible with an mpifileutils config I was going to submit after this PR was approved. I don't remember the details though and can run back through it to make sure I'm remembering correctly. I'll make the necessary changes here to bring them in line with the already accepted gettext configs and then try to work on the mpifileutils config from there. Sorry about that

PR #22554, opened 2025-03-13
Add easyconfig OpenCV-4.8.1-foss-2023b-contrib.eb

julianmorillo:


Crivella:
We already have versions of these ECs for the 2023b tc in particular

- https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/o/OpenCV/OpenCV-4.11.0-foss-2023b-contrib.eb

Is there a reason to have an older version of OpenCV and GST-plugins-bad?
Also we would need a versionsuffixed variant of GTK4 if this is really needed

julianmorillo:
Not at all, most probably I did this PR before OpenCV-4.11.0 was merged. 
Closing this PR.

PR #22576, opened 2025-03-16
replace source URLs of Mesa to new location in archive.mesa3d.org

lexming:
Fixes failed test in https://gist.github.com/boegelbot/2645bd9f0653b47649fa39adae45a862


lexming:
EB_ARGS="--stop fetch --force-download"
EB_BRANCH=5.0.x

lexming:
Test for `libglvnd-1.3.2-GCCcore-10.2.0.eb` failed because the gitlab instance of freedesktop.org is ongoing a migration. Should be back in the future.

All tests for `Mesa` are positive, so this PR is good to go.

PR #22580, opened 2025-03-17
use all same checksums of libxc v6.2.2 in its easyconfigs

lexming:
Update all easyconfigs of libxc 6.2.2 with the checksums already in `libxc-6.2.2-GCC-13.3.0.eb`


lexming:
EB_ARGS="--stop fetch --force-download"
EB_BRANCH=5.0.x

boegel (approved):

PR #22602, opened 2025-03-18
Fix regression in pytest 8.2.2

Flamefire:
(created using `eb --new-pr`)

pytest-rerunfailures [explicitly disabled 8.2.2 ](https://github.com/pytest-dev/pytest-rerunfailures/pull/279)due to a [regression](https://github.com/pytest-dev/pytest-rerunfailures/issues/267) and won't work with that. Others might be affected by this too in that they fail to run properly

This applies [the fix](https://github.com/pytest-dev/pytest/pull/12438) which is backported to 8.2.x but without a release yet.


smoors (approved):
lgtm

smoors:
Going in, thanks @Flamefire!

PR #22608, opened 2025-03-19
Add easyconfig PyOpenGL-3.1.7-GCCcore-13.2.0.eb

julianmorillo:


Flamefire:
I created an easyconfig for 3.1.9 which I forgot to upload. Done that now: https://github.com/easybuilders/easybuild-easyconfigs/pull/22750
I'm referencing this in [another PR](https://github.com/easybuilders/easybuild-easyconfigs/pull/22643). Was there a specific reason for using 3.1.7 over 3.1.9? If so I'll try that other PR using 3.1.7 instead and close mine, otherwise I'd rather use the newer version

julianmorillo:
There is no specific reason for using 3.1.7; you can continue with 3.1.9. I only have a question: why is #22750 using the gfbf toolchain instead of GCCcore?

Flamefire:
> There is no specific reason for using 3.1.7; you can continue with 3.1.9. I only have a question: why is #22750 using the gfbf toolchain instead of GCCcore?

That was a mistake. At some point I added SciPy-bundle but removed the part that required it. When downgrading I saved the wrong file and didn't notice until EB "fixed" the filename during the PR creation. Fixed as #22751

Flamefire:
3.1.9 was just merged.

julianmorillo:
> 3.1.9 was just merged.

Good, then I close this PR.

PR #22626, opened 2025-03-20
Add easyconfig Clang-20.1.1-GCCcore-13.2.0.eb

julianmorillo:


Crivella:
We are stopping support for the clang easyblock.

See for reference
- https://github.com/easybuilders/easybuild-easyconfigs/pull/23028
- https://github.com/easybuilders/easybuild-easyconfigs/pull/23055
- https://github.com/easybuilders/easybuild-easyblocks/pull/3746

PR #22638, opened 2025-03-21
Mention filename for failing use_pip check

Flamefire:
The failure was direct instead of adding it to the list of failures which would then be printed with the filename.

I also fixed a few linter issues (most serious was check_https_url not closing the connection which might cause issues for many checks run) and added type hints to the `parsed/changed/ordered*` properties as it was confusing what each of them returned and also the linter couldn't check it.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -35,6 +35,7 @@ import shutil
 import stat
 import tempfile
 from collections import defaultdict
+from typing import List
 from unittest import TestCase, TestLoader, main, mock, skip
 from urllib.request import Request, urlopen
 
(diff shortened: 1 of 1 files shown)


Flamefire:
Thanks I haven't noticed this was a list-of-tuples to dict thing. Another reason for the change which makes it clearer

test/easyconfigs/easyconfigs.py
```diff
@@ -34,6 +34,7 @@
 import stat
 import tempfile
 from collections import defaultdict
+from typing import List
```
boegel, line 37: https://docs.python.org/3/library/typing.html#typing.List says this is deprecated since Python 3.9?

Flamefire: Yes, but we need to support 3.6 so unless they remove it I think there is no issue

Crivella: I do not think this will be removed anytime soon unless they want to break a huge amount of packages that started adopting typehints early

Crivella (approved):
LGTM

test/easyconfigs/easyconfigs.py
Crivella: Would say we will remove this in EB 6.0 where we plan to drop support for python < 3.9

Crivella:
Going in, thanks @Flamefire!

PR #22648, opened 2025-03-24
Improve failure message for download instruction CI check

Flamefire:
The current message starts with something about unequal lists which isn't really helpful


Micket (approved):
lgtm

PR #22683, opened 2025-04-01
Add openCARP v17.0 and dependencies

MarieHouillon:
- Add new version for openCARP (17.0), using foss-2024a
- Update dependencies not available with foss-2024a yet: gengetopt-2.23, PETSc-3.22.3, ParMETIS-4.0.3, pkg-config-0.29.2, SuperLU_DIST-8.2.1


MarieHouillon:
Converted to draft: `pkg-config` dependency should be replaced with `pkgconf`

boegel:
> Converted to draft: `pkg-config` dependency should be replaced with `pkgconf`

Is this the only change necessary you think?
If so, we can easily make that change to get this moving again...

PR #21199, opened 2024-08-16
{numlib}[gfbf/2023b,iimkl/2023b] SuiteSparse v7.7.0

appolloford:
(created using `eb --new-pr`)

PR #22696, opened 2025-04-03
{bio,numlib}[foss/2023b] openCARP v17.0, PETSc v3.22.5

MarieHouillon:
(created using `eb --new-pr`)


easybuild/easyconfigs/o/openCARP/openCARP-17.0-foss-2023b.eb
```diff
@@ -0,0 +1,37 @@
...
+source_urls = ['https://git.opencarp.org/openCARP/openCARP/-/archive/v%(version)s/']
+sources = ['openCARP-v%(version)s.tar.gz']
+checksums = [
+    '6439b6b8e6a265cd6cfe39f59507418739d936de6f6e059b2d0f468df816d26e',  # openCARP-v17.0.tar.gz
+]
+
+builddependencies = [
+    ('CMake', '3.27.6'),
+    ('gengetopt', '2.23'),
+    ('pkg-config', '0.29.2'),
```
boegel, line 21: `pkg-config` is no longer maintained, `pkgconf` is (mostly) a drop-in replacement for it, can we try using that instead?

```suggestion
    ('pkgconf', '2.0.3'),
```

(and then remove the `pkg-config` easyconfig file you're added, see https://docs.easybuild.io/integration-with-github/?h=github#github_new_update_pr_delete

boegel (changes requested):

easybuild/easyconfigs/o/openCARP/openCARP-17.0-foss-2023b.eb
```diff
@@ -0,0 +1,37 @@
...
+    ('pkgconf', '2.0.3'),
+]
+
+dependencies = [
+    ('PETSc', '3.22.3'),
+    ('zlib', '1.2.13'),
+]
+
+
+configopts = "-DSPACK_BUILD=ON "
```
ocaisa, line 30:
```suggestion
configopts = "-DSPACK_BUILD=ON "
```
I looked into what this implies and found https://git.opencarp.org/openCARP/openCARP/-/blob/master/cmake/postinst?ref_type=heads , there are hard-coded directories in there that are not portable and almost certainly do not exist...so I think this step will only print the final warning in that file.

MarieHouillon: It would also avoid include this file used for CPack packaging: https://git.opencarp.org/openCARP/openCARP/-/blob/master/cmake/CPackConfig.cmake?ref_type=heads

MarieHouillon: We are working on refactoring this installation part that is not portable in openCARP, but we use this workaround for now to be able to use the install target.

boegel: @MarieHouillon What's the impact of enabling this `SPACK_BUILD` option?

MarieHouillon: It disables some postinstall steps used for packaging

easybuild/easyconfigs/p/PETSc/PETSc-3.22.3-foss-2023b.eb
```diff
@@ -0,0 +1,45 @@
+##
+# Author:    Robert Mijakovic <robert.mijakovic@lxp.lu>
+# Author:    Jasper Grimm (UoY)
+##
+name = 'PETSc'
+version = '3.22.3'
```
boegel, line 6: @MarieHouillon Any particular reason you're sticking to PETSc 3.22.3 while 3.22.5 is already available for a while?

boegel:
Test report by @boegel
**SUCCESS**
node3523.doduo.os - Linux RHEL 9.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/boegel/7e884020a31098c8ac82531885928f6d for a full test report.

easybuild/easyconfigs/p/PETSc/PETSc-3.22.3-foss-2023b.eb
MarieHouillon: It's the version we use in our Docker image and for packaging currently, but we could certainly also use 3.22.5

easybuild/easyconfigs/o/openCARP/openCARP-17.0-foss-2023b.eb
boegel: OK, then a line of comment to mention that makes sense, because this looks very out of place in a build recipe for EasyBuild :)

MarieHouillon: I'll add it. We will get rid of this flag anyway in a near future.

easybuild/easyconfigs/p/PETSc/PETSc-3.22.3-foss-2023b.eb
ocaisa: I think if we are going to introduce a new PETSc into a toolchain, we should stick to including the latest version (particularly when it is patch release).

ocaisa: In fact 3.23.1 is available at this stage, thoughts @boegel ?

MarieHouillon: For us 3.22.5 is better, as 3.23 introduced changes that are not compatible with openCARP 17.0 without a patch.

MarieHouillon: I've updated the PETSc version to 3.22.5 as suggested, please let me know if anything else should be done.

boegel:
CI fails due to missing dependency:
```
Missing dependencies: SuiteSparse/7.7.0-foss-2023b-METIS-5.1.0 (no easyconfig file or existing module found)
```

@MarieHouillon I can't seem to find another PR that provides this, what am I overlooking?

easybuild/easyconfigs/p/PETSc/PETSc-3.22.5-foss-2023b.eb
```diff
@@ -0,0 +1,45 @@
...
+builddependencies = [('CMake', '3.27.6')]
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+    ('Boost', '1.83.0'),
+    ('METIS', '5.1.0'),
+    ('SCOTCH', '7.0.4'),
+    ('MUMPS', '5.6.1', '-metis'),
+    ('SuiteSparse', '7.7.0', '-METIS-5.1.0'),
```
boegel, line 30: The `versionsuffix` here should probably be dropped (just like it is in the existing `PETSc-3.20.3-foss-2023a.eb`):

```suggestion
    ('SuiteSparse', '7.7.0'),
```

boegel (changes requested):

easybuild/easyconfigs/p/PETSc/PETSc-3.22.5-foss-2023b.eb
boegel: Indeed, the `versionsuffix` was dropped in #21199, so we should change this accordingly here for the dependency to resolve

boegel:
Test report by @boegel
**SUCCESS**
node3539.doduo.os - Linux RHEL 9.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/boegel/cccee44315c757b4a6dd50026d1adfcf for a full test report.

boegel (approved):
lgtm

PR #22422, opened 2025-02-28
{numlib}[foss/2023b] DOLFINx v0.9.0, pugixml v1.14, FFCx v0.9.0, ...

tanmoy1989:
(created using `eb --new-pr`)

PR #22697, opened 2025-04-03
adding Basix and few dependencies for 2024a

moravveji:
This PR adds `Basix` which is one of the dependencies of `DOLFINx`. This PR depends on the easyblock from [this PR](https://github.com/easybuilders/easybuild-easyblocks/pull/3684).
Furthermore, these new easyconfigs are either created or a newer version is proposed:
- `scikit-build-core/0.11.0-GCCcore-13.3.0`
- `robin-map/1.4.0-GCCcore-13.3.0`
- `nanobind/2.6.1-GCCcore-13.3.0`

I also have to admit that this PR is inspired by PR #22422.

### DIFF
--- a/easybuild/easyconfigs/b/Basix/Basix-0.9.0-gfbf-2024a.eb
+++ b/easybuild/easyconfigs/b/Basix/Basix-0.9.0-gfbf-2024a.eb
@@ -1,0 +1,56 @@
+easyblock = 'EB_Basix'
+
+name = 'Basix'
+version = '0.9.0'
+
+homepage = 'https://github.com/FEniCS/basix'
+description = """Basix is a finite element definition and tabulation runtime library.
+Basix includes a range of built-in elements, and also allows the user to define their own custom elements.
+Basix is one of the components of FEniCSx, alongside UFL, FFCx, and DOLFINx."""
+
+toolchain = {'name': 'gfbf', 'version': '2024a'}
+
+source_urls = ['https://github.com/FEniCS/basix/archive/refs/tags/']
+sources = ['v%(version)s.tar.gz']
+patches = ['Basix-0.9.0-enable-distribution.patch']
+checksums = [
+    {'v0.9.0.tar.gz': '60e96b2393084729b261cb10370f0e44d12735ab3dbd1f15890dec23b9e85329'},
+    {'Basix-0.9.0-enable-distribution.patch': '902f672000ef1fa844208450568bc34d5c5f40ef3b8ae9ed3781c15579ff8f51'},
+]
+
+builddependencies = [
+    ('CMake', '3.29.3'),
+    ('PDM', '2.18.2'),
+    ('Ninja', '1.12.1'),
+    ('scikit-build-core', '0.11.0'),
+]
+
+dependencies = [
+    ('Python', '3.12.3'),
+    ('Python-bundle-PyPI', '2024.06'),
+    ('SciPy-bundle', '2024.05'),
+    ('nanobind', '2.6.1'),
+]
+
+parallel = 8
+
+exts_defaultclass = 'PythonPackage'
+exts_list = [('fenics-basix', version, {
+    'modulename': 'basix',
+    'nosource': True,
+    'download_dep_fail': True,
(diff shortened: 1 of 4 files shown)


sassy-crick:
It appears the`Basix-0.9.0-enable-distribution.patch` is missing?

moravveji:
@sassy-crick : I would like to revive this PR and bring it back to life. Are there still any steps I shall take to have it tested? Unfortunately, the easyblock cannot be automatically found, and few tests fail. How can I resolve this issue?

Thyre:
Are there any particular reasons why we include the newer versions of `nanobind`, `scikit-build-core` and `robin-map`?
We typically try to avoid having multiple versions for `dependencies` in a single toolchain. This would affect `nanobind` and `robin-map`.

easybuild/easyconfigs/b/Basix/Basix-0.9.0-gfbf-2024a.eb
```diff
@@ -0,0 +1,54 @@
...
+
+dependencies = [
+    ('Python', '3.12.3'),
+    ('Python-bundle-PyPI', '2024.06'),
+    ('SciPy-bundle', '2024.05'),
+    ('nanobind', '2.6.1'),
+]
+
+parallel = 8
+
```
Thyre: Is there a reason why we set this to `8`?
Typically, this would be handled by users by their `parallel` and `max-parallel` config options.

There might be a reason to set it to `1`, e.g. when a parallel build causes build failures. But I don't think this is the case here.

easybuild/easyconfigs/b/Basix/Basix-0.9.0-gfbf-2024a.eb
```diff
@@ -0,0 +1,54 @@
...
+exts_defaultclass = 'PythonPackage'
+exts_list = [('fenics-basix', version, {
+    'modulename': 'basix',
+    'nosource': True,
+    'download_dep_fail': True,
+    'use_pip': True,
+    'sanity_pip_check': True,
```
Thyre: I think those should be set by EB 5.x by default, so we don't need to specify them here anymore.

Thyre:
Based on the discussion in https://github.com/easybuilders/easybuild-easyblocks/pull/3684, trying `PythonPackage` or `PythonBundle` would be great.
We could then avoid having the separate EasyBlock. Then, the CI would also "work", since it knows all the EasyBlocks.

moravveji:
> Are there any particular reasons why we include the newer versions of `nanobind`, `scikit-build-core` and `robin-map`? We typically try to avoid having multiple versions for `dependencies` in a single toolchain. This would affect `nanobind` and `robin-map`.

This PR was pushed on 3 April whereas the older versions of `nanobind`, `scikit-build-core` and `robin-map` are merged to develop/master branches few months after that. At the time this PR was pushed and was waiting for a review, the proposed versions of these packages were the latest versions, and they could perfectly fit in, and fill up missing easyconfigs for 2024a toolchain. However, this PR has now collected a lot of "dust" and needs adjustment. Although that is gonna take me some more time, but I will try to accommodate it.

moravveji:
> We typically try to avoid having multiple versions for `dependencies` in a single toolchain. This would affect `nanobind` and `robin-map`.

The commit 6befba93ca retracts the proposed easyconfigs for those conflicting modules to adopt the existing versions from the 2024a toolchain.

easybuild/easyconfigs/b/Basix/Basix-0.9.0-gfbf-2024a.eb
moravveji: Cleaned up in 72d63e5fb6.

easybuild/easyconfigs/b/Basix/Basix-0.9.0-gfbf-2024a.eb
moravveji: Sure. I removed it in 72d63e5fb6.

moravveji:
Closing this PR in favor of #23838

PR #22701, opened 2025-04-03
Make Python a dependency of GObject-Introspection 1.80.1

stevenvdb:
For some reason, in `GObject-Introspection-1.80.1-GCCcore-13.3.0.eb` Python was marked as a builddependency, while in earlier versions it was a dependency. This causes issues, for example running the `g-ir-scanner` scripts will give an error:
```
ModuleNotFoundError: No module named 'giscanner._giscanner'
```
The reason is that `GObject-Introspection/1.80.1-GCCcore-13.3.0/lib/gobject-introspection/giscanner/_giscanner.cpython-312-x86_64-linux-gnu.so` is only picked up on condition that the correct Python interpreter is active. I think the correct solution here is to make Python a dependency also for this version of `GObject-Introspection`.

(created using `eb --new-pr`)


boegel:

boegel:
Going in, thanks @stevenvdb!

PR #22760, opened 2025-04-15
Python as dependency for Python templates instead of CMake

WilleBell:


jfgrimm (approved):

jfgrimm:
Going in, thanks @WilleBell!

PR #22833, opened 2025-05-02
adding easyconfigs: McStas-3.5.24-foss-2023a.eb

ebknudsen:
McStas is a neutron ray tracing package for Monte Carlo simulation of neutron scattering experiments and instrumentation.

It works by describing the geometry in a domain specific language. The description is then transformed by a lex/yacc-based code generator into a c-program, which in turn may be compiled into an executable.
Code-generation to a large part consists of concatenating pieces of code from various file, which is why the installed software library contains a set of .c and .h-files. This is by design.

PR #22834, opened 2025-05-02
{phys}[foss/2023a] adding easyconfigs: McXtrace-3.5.24-foss-2023a.eb

ebknudsen:
McXtrace is an X-ray ray tracing package for Monte Carlo simulation of X-ray scattering experiments and instrumentation. McXtrace is an X-ray port of its neutron counterpart McStas (#22833)

It works by describing the geometry in a domain specific language. The description is then transformed by a lex/yacc-based code generator into a c-program, which in turn may be compiled into an executable.
Code-generation to a large part consists of concatenating pieces of code from various file, which is why the installed software library contains a set of .c and .h-files. This is by design.


Thyre:
I _think_ you added `easybuild/easyconfigs/m/McStas/McStas-3.5.24-foss-2023a.eb` from #22833 here as well by mistake, right?

ebknudsen:
Yes indeed that was a mistake - I'll fix that. Thanks for noticing!

jfgrimm:

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
```diff
@@ -0,0 +1,42 @@
...
+toolchain = {'name': 'foss', 'version': '2023a'}
+
+github_account = 'mccode-dev'
+source_urls = ["https://github.com/mccode-dev/McCode/archive/refs/tags/"]
+sources = ['v%(version)s.tar.gz']
+checksums = ['b1958a24b6b7e06e65ff3fc302da6d306412fdaee15ac77fadf841e1521b773a']
+
+builddependencies = [
+    ('CMake', '3.26.3'),
+]
```
jfgrimm, line 21:
```suggestion
builddependencies = [
    ('CMake', '3.26.3'),
    ('Bison', '3.8.2'),
    ('flex', '2.6.4'),
]
```

jfgrimm:

Micket:
Test report by @Micket
**FAILED**
vera-icelake-build - Linux Rocky Linux 9.2, x86_64, Intel(R) Xeon(R) Silver 4316 CPU @ 2.30GHz, Python 3.9.18
See https://gist.github.com/Micket/fb75e3b0840f24abbf80f4600b376073 for a full test report.

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
```diff
@@ -0,0 +1,44 @@
...
+    ('flex', '2.6.4'),
+]
+
+dependencies = [
+    ('Python', '3.11.3'),
+    ('SciPy-bundle', '2023.07'),
+    ('PyYAML', '6.0')
+]
+
+configopts = '-DBUILD_MCXTRACE=ON -DENSURE_NCRYSTAL=off -DMCVERSION=%(version)s -DBUILD_TOOLS=off -DBUILD_MCXRUN=on'
```
Micket: For software like this, i would probably

1. hope their cmake scripts allow for an override to point to locally checkout sources
2. add all the sub-sources, possible with custom extract command or preconfigopts to put them in the right paths

failing 1. i would make a patch to comment out the downloading steps, and just place the sources where they are needed. You can use git_config source downloads with EB if you need particular commits. We strongly prefer to never just "fetch latest master" as it throws out all hope of reprodicibility.

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
```diff
@@ -0,0 +1,44 @@
...
+
+sanity_check_paths = {
+    'files': ["%(namelower)s/%(version)s/bin/mcxtrace"],
+    'dirs': [f"%(namelower)s/%(version)s/{dd}" for dd in ['samples', 'share', 'optics']]
+}
```
Micket:
```suggestion
sanity_check_paths = {
    'files': ["bin/mcxtrace"],
    'dirs': ['samples', 'share', 'optics'],
}
```

We really don't want those extra paths. I'd patch their cmakelists to not prepend that extra name + version to their installed material so that it properly respects the --prefix we pas to CMake automatically in CMakeMake easyblock.

Micket (changes requested):
copied over my feedback from the slack channel here.

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
ebknudsen: The way the install is intended to work normally is to put the software and its libs (which are not regular shared libraries) in /usr/share/mcxtrace/version/ etc. and then make symlinks to executables in e.g. /usr/bin. The env-variable MCXTRACE tells where the libs are to be found runtime. This to allow multiple versions to be installed but is of course not useful for an easybuild.

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
ebknudsen: In this case the eb does download a specific release (i.e. not just latest). The build-option MCVERSION is there to set that in some of the library files.
The upstream software is a bit unusual in the sense that it actually runtime assembles a c-code, so there are quite a few source-files in the library which are copied in (almost) verbatim.

That said, it _is_ probably better to do a build from a proper checkout.

ebknudsen: circumvented by 8cf815e

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
```diff
@@ -0,0 +1,45 @@
...
+]
+
+dependencies = [
+    ('Python', '3.11.3'),
+    ('SciPy-bundle', '2023.07'),
+    ('PyYAML', '6.0')
+]
+
+configopts = '-DBUILD_MCXTRACE=ON -DENSURE_NCRYSTAL=off -DENSURE_MCPL=off -DMCVERSION=%(version)s ' \
+    '-DBUILD_TOOLS=off -DBUILD_MCXRUN=on -DMCCODE_USE_LEGACY_DESTINATIONS=off'
```
Crivella: Considering the previous comments, not sure if this (and the dependencies) should be changed now that 

- #23224 

has been merged

Crivella:
Test report by @Crivella
**FAILED**
See https://gist.github.com/Crivella/5df12c947a2f7ee6f0053b0de6e3ea8e for a full test report.

Crivella:
```
--   Building mcpl with MCPL_PYPATH=mcxtrace/3.5.24/python
--   Building mcpl with CMAKE_INSTALL_LIBDIR=mcxtrace/3.5.24/libs/mcpl
--   Building mcpl with CMAKE_INSTALL_INCLUDEDIR=mcxtrace/3.5.24//libs/mcpl
[ 11%] Creating directories for 'mcpl-populate'
[ 22%] Performing download step (git clone) for 'mcpl-populate'
Cloning into 'mcpl-src'...
fatal: invalid reference: master
CMake Error at mcpl-subbuild/mcpl-populate-prefix/tmp/mcpl-populate-gitclone.cmake:49 (message):
  Failed to checkout tag: 'master'
```

Not sure if there is a way to tell it to use an external one instead of trying to redownload it.

Crivella:
I am really not an expert on the tool but MCPL might not need extra linking to McXtrace (it is an extra tool to write inputs?)
https://github.com/mccode-dev/McCode/discussions/2076

So not sure we need the `ENSURE_MCPL` which seems to just run another build of MCPL by looking at the CMakeList.txt

I will defer to someone who know how McXtrace is supposed to be used (or if examples of using it together with MCPL are provided so we can test if we can use the separate EConf build)

ebknudsen:
Yeah - me bad. The `ENSURE_MCPL`-flag really means exactly to checkout a local version of MCPL, which is not useful since MCPL now has its own EC-file

I _should_ really know exactly how McXtrace is built - I am one of the original authors in fact. This was a silly mistake.

ebknudsen:
To provide a bit more background. McXtrace is in essence a specialized c-code generator. When the generated code is to be compiled the mcpl-lib has to be linked. There's a python script in the distro that helps automate the process. To summarise: only in a late stage are the libs needed.

Crivella:
So the only important thing about MCPL w.r.t. McXtrace is that `libmcpl.so` is availalble on `LD_LIBRARY_PATH`?

ebknudsen:
That's almost enough - for it to work correctly in all cases it requires that the binary mcpl-config is available on the path as well.

Crivella:
Test report by @Crivella
**SUCCESS**
See https://gist.github.com/Crivella/4f0a6aef4a8be1bca330dd5a8dc59f58 for a full test report.

Crivella:
From the configure step

```
-- Generating done (0.0s)
CMake Warning:
  Manually-specified variables were not used by the project:

    ENSURE_NCRYSTAL
    PYTHON_EXECUTABLE
    Python3_EXECUTABLE
    Python_EXECUTABLE


-- Build files have been written to: /home/crivella/.local/easybuild/build/McXtrace/3.5.24/foss-2023a/easybuild_obj
```

It seems `ENSURE_NCRYSTAL` is actually not being used. I guess because it is only recognized in case of `BUILD_MCSTAS=ON` https://github.com/mccode-dev/McCode/blob/3e365f4d3cb8b1f41edb2f47507339d724871794/CMakeLists.txt#L38C1-L45C8

I am unsure if we should keep it considering you have a separate PR for McStas 

- https://github.com/easybuilders/easybuild-easyconfigs/pull/22833

(I guess we should also add a comment that McStas is built separately in easybuild to avoid people turning it on in the future without realizing it)

EDIT: an alternative might be to rename this to `McCode` and build both McXtrace and McStas together? Not sure if there was a reason they were originally split

ebknudsen:
Again a bit of background: McXtrace and McStas are very much similar and share some of the code base which is why they reside in the same repo. The former deals with X-rays, the latter with neutrons.
ncrystal is only valid for neutrons. (there's some talk of porting, but that project is not likely to happen in the foreseeable future).

Historically McStas is the older sibling, years later McXtrace was built as a port of it. Apart from some shared common code, they do not work together, and have always been considered two separate entities. The name mccode is almost exclusively used by the core development team. 

All-in-all, there's a practical case for having a single EC-file, but I would not recommend it - very few people would recognize it.

Is it possible to maintain a single mccode EC-file, but with shallow McXtrace and McStas "virtual" EC-files that depend on it? EDIT: That'd be using the "bundle"-easyblock I suppose

Crivella:

Crivella:
> Historically McStas is the older sibling, years later McXtrace was built as a port of it. Apart from some shared common code, they do not work together, and have always been considered two separate entities. The name mccode is almost exclusively used by the core development team.

Thanks for the info.
No need to overcomplicate it, my idea of merging them was because i thought they were 2 tools part of the same suite of code (as they are built through the same `CMakeLists.txt`).
If they are functionally separate I am fine keeping them split in EasyBuild

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
```diff
@@ -0,0 +1,46 @@
...
+
+name = 'McXtrace'
+version = '3.5.24'
+
+homepage = 'https://mcxtrace.org'
+description = "McXtrace is a Monte Carlo ray tracing X-ray instrumentaton and experiment simulation packages."
+
+toolchain = {'name': 'foss', 'version': '2023a'}
+
+github_account = 'mccode-dev'
```
Crivella:
```suggestion
```

This is not needed unless you are using a [template](https://docs.easybuild.io/version-specific/easyconfig-templates/#template-constants-that-can-be-used-in-easyconfigs) like `GITHUB_SOURCE`

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
```diff
@@ -0,0 +1,46 @@
...
+
+dependencies = [
+    ('Python', '3.11.3'),
+    ('SciPy-bundle', '2023.07'),
+    ('PyYAML', '6.0'),
+    ('mcpl', '2.2.0'),
+]
+
+configopts = '-DBUILD_MCXTRACE=ON -DENSURE_NCRYSTAL=off -DENSURE_MCPL=off DMCVERSION=%(version)s ' \
+    '-DBUILD_TOOLS=off -DBUILD_MCXRUN=on -DMCCODE_USE_LEGACY_DESTINATIONS=off'
```
Crivella:
```suggestion
configopts = ' '.join([
    '-DMCVERSION=%(version)s',
    '-DBUILD_MCXTRACE=ON',
    '-DBUILD_MCXRUN=ON',
    '-DENSURE_NCRYSTAL=OFF',
    '-DENSURE_MCPL=OFF',  # Use mcpl from dependency instead of rebuilding it
    '-DBUILD_TOOLS=OFF',
    '-DMCCODE_USE_LEGACY_DESTINATIONS=OFF'
])
```

Making this easier to edit and more future-proof
Feel free to add more comments if you think they could help clarify better

easybuild/easyconfigs/m/McXtrace/McXtrace-3.5.24-foss-2023a.eb
```diff
@@ -0,0 +1,46 @@
...
+
+sanity_check_paths = {
+    'files': ["bin/mcxtrace"],
```
Crivella:
```suggestion
    'files': ["bin/mcxtrace", "bin/mxrun"],
```

ebknudsen (commented):
LGTM

Crivella:
None of the required changes have been applied?

ebknudsen:
Ah - apologies - yet another silly mistake. I thought I accepted the changes by clicking resolved. "I owe you a beer at the airport" as we say in DK,

Crivella:

Crivella (approved):
LGTM

Crivella:
Going in, thanks @ebknudsen!

ebknudsen:
Thanks @Crivella! Much appreciated!

PR #22851, opened 2025-05-07
Use vars instead of paths which supports environment variables soon

Micket:
(created using `eb --new-pr`)

- [x] Needs @lexming 's fix https://github.com/easybuilders/easybuild-framework/pull/4855


lexming (approved):
LGTM

PR #22857, opened 2025-05-08
{vis}[system/system] Panoply v5.6.0

TopRichard:


easybuild/easyconfigs/p/Panoply/Panoply-5.6.0.eb
```diff
@@ -0,0 +1,32 @@
...
+checksums = ['0c4703f289750da18c7a3e7f4c61f3eceb63fd5d9b5e43fcbaa2bdf7615eaaa6']
+
+dependencies = [
+    ('Java', '11', '', SYSTEM),
+]
+
+postinstallcmds = [
+    'mkdir %(installdir)s/bin',
+    'mv %(installdir)s/panoply.sh %(installdir)s/bin',
+    'sed -i "s,jars,../jars,g" %(installdir)s/bin/panoply.sh',
```
boegel, line 23: @TopRichard Can you add a comment above this specific line to explain what it does/why it's needed?

The others are pretty obvious.

easybuild/easyconfigs/p/Panoply/Panoply-5.6.0.eb
```diff
@@ -0,0 +1,32 @@
+easyblock = 'PackedBinary'
```
boegel, line 1: `PackedBinary` is mostly interesting if you're using the custom `install_cmd` easyconfig parameter it supports.

That's not used here, so I would use `Tarball` easyblock instead

```suggestion
easyblock = 'Tarball'
```

easybuild/easyconfigs/p/Panoply/Panoply-5.6.0.eb
```diff
@@ -0,0 +1,32 @@
...
+
+name = 'Panoply'
+version = '5.6.0'
+
+homepage = 'https://www.giss.nasa.gov/tools/panoply'
+description = "Panoply plots geo-referenced and other arrays from netCDF, HDF, GRIB, and other datasets."
+
+
```
boegel, line 9: remove duplicate empty line
```suggestion
```

easybuild/easyconfigs/p/Panoply/Panoply-5.6.0.eb
```diff
@@ -0,0 +1,32 @@
...
+postinstallcmds = [
+    'mkdir %(installdir)s/bin',
+    'mv %(installdir)s/panoply.sh %(installdir)s/bin',
+    'sed -i "s,jars,../jars,g" %(installdir)s/bin/panoply.sh',
+    'ln -s %(installdir)s/bin/panoply.sh %(installdir)s/bin/panoply',
+]
+
+sanity_check_paths = {
+    'files': ['bin/panoply'],
```
boegel, line 28: Can we also have a sanity check command for this, like a `panoply --help` ?

boegel (changes requested):

easybuild/easyconfigs/p/Panoply/Panoply-5.6.0.eb
```diff
@@ -0,0 +1,32 @@
...
+
+homepage = 'https://www.giss.nasa.gov/tools/panoply'
+description = "Panoply plots geo-referenced and other arrays from netCDF, HDF, GRIB, and other datasets."
+
+
+toolchain = SYSTEM
+
+source_urls = ['https://www.giss.nasa.gov/tools/panoply/download/']
+sources = ['%(name)sJ-%(version)s.tgz']
+checksums = ['0c4703f289750da18c7a3e7f4c61f3eceb63fd5d9b5e43fcbaa2bdf7615eaaa6']
```
boegel, line 14: Download fails for me, looks like there's only `5.6.1` now, `5.6.0` can't be downloaded anymore? (ugh...)

Maybe we can reach out and ask if previous versions are still available somewhere?

TopRichard: > Download fails for me, looks like there's only `5.6.1` now, `5.6.0` can't be downloaded anymore? (ugh...)
> 
> Maybe we can reach out and ask if previous versions are still available somewhere?

WIP

TopRichard: Could not get an answer from the support, so i will close the PR

PR #22908, opened 2025-05-19
{tools}[system/system] gocryptfs v2.5.4

gkaf89:
Used by Apptainer in RHEL systems that do not provide a gocrytpfs package.

### DIFF
--- a/easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
+++ b/easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
@@ -33,7 +33,7 @@ buildopts = 'GOPATH=%(builddir)s/go'
 build_cmd_targets = ['build']
 
 # Needs the DESTDIR variable to not use a system directory
-install_cmd = 'make install DESTDIR=%(installdir)s'
+installopts = 'DESTDIR=%(installdir)s'
 
 sanity_check_paths = {
     'dirs': ['bin', 'share'],
--- a/easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
+++ b/easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
@@ -1,0 +1,47 @@
+easyblock = 'ConfigureMake'
+
+name = 'gocryptfs'
+version = '2.5.4'
+
+homepage = 'https://nuetzlich.net/gocryptfs/'
+description = "An encrypted overlay filesystem written in Go. gocryptfs is built on top the excellent go-fuse FUSE library."
+
+toolchain = SYSTEM
+
+source_urls = ['https://github.com/rfjakob/gocryptfs/releases/download/v%(version)s/']
+sources = ['gocryptfs_v%(version)s_src-deps.tar.gz']
+patches = ['gocryptfs-2.5.4_install.patch']
+checksums = [
+    {'gocryptfs_v2.5.4_src-deps.tar.gz': '0db47fe41f46d1ff5b3ff4f1cc1088ab324a95af03995348435dcc20a5ff0282'},
+    {'gocryptfs-2.5.4_install.patch': '0dc1dba0bac38c1cfd94d1c31a8d62c55a1167fff86bafa2b6da6ebd6e9cc6cd'},
+]
+
+builddependencies = [
+    ('Go', '1.23.6'),
+    ('Pandoc', '3.6.2'),
+]
+
+dependencies = [
+    ('OpenSSL', '3'),
+]
+
+# No configure script
+configure_cmd = ':'
+
+# Make builds Go modules; isolate the build environment
+build_cmd = 'make GOPATH=%(builddir)s/go'
+
+build_cmd_targets = ['build']
+
+# Needs the DESTDIR variable to not use a system directory
+install_cmd = 'make install DESTDIR=%(installdir)s'
+
+sanity_check_paths = {
+    'dirs': ['bin', 'share'],
+    'files': ['bin/%(name)s', 'bin/%(name)s-xray'],
+}
+
+# --help, --version, ... all exit with status code one, so we need to check the output
+sanity_check_commands = ['%(name)s --help 2>&1 | grep -q %(version)s']
+
+moduleclass = 'tools'
--- a/easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
+++ b/easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
@@ -25,8 +25,7 @@ dependencies = [
     ('OpenSSL', '3'),
 ]
 
-# No configure script
-configure_cmd = ':'
+skipsteps = ['configure']
 
 # Make builds Go modules; isolate the build environment
 build_cmd = 'make GOPATH=%(builddir)s/go'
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,51 @@
...
+    ('Go', '1.23.6'),
+    ('Pandoc', '3.6.2'),
+]
+
+dependencies = [
+    ('OpenSSL', '3'),
+]
+
+# No configure script
+configure_cmd = ':'
```
jfgrimm, line 29: if there is no configure, lets just skip the whole step instead
```suggestion
skipsteps = ['configure']
```

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,51 @@
...
+]
+
+builddependencies = [
+    ('Go', '1.23.6'),
+    ('Pandoc', '3.6.2'),
+]
+
+dependencies = [
+    ('OpenSSL', '3'),
+]
```
jfgrimm, line 26: pretty sure we also need an OS dep on `fuse`, no?

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,51 @@
...
+
+# No configure script
+configure_cmd = ':'
+
+# Make builds Go modules; isolate the build environment
+build_cmd = 'make GOPATH=%(builddir)s/go'
+
+build_cmd_targets = ['build']
```
jfgrimm, line 34: use `buildopts` to pass extra options to `make`
```suggestion
# Make builds Go modules; isolate the build environment
buildopts = 'GOPATH=%(builddir)s/go'

build_cmd_targets = ['build']
```

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,51 @@
...
+
+# Needs the DESTDIR variable to not use a system directory
+install_cmd = 'make install DESTDIR=%(installdir)s'
```
jfgrimm, line 37: similarly, use `installopts` to pass extra options to `make install`
```suggestion
# Needs the DESTDIR variable to not use a system directory
installopts = 'DESTDIR=%(installdir)s'
```

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,51 @@
...
+
+sanity_check_paths = {
+    'dirs': ['bin', 'share'],
+    'files': ['bin/%(name)s', 'bin/%(name)s-xray'],
```
jfgrimm, line 41: lets keep the order of these consistent with the vast majority of easyconfigs
no need to check for the `bin` directory, since we're checking for files within that dir anyway 
```suggestion
    'files': ['bin/%(name)s', 'bin/%(name)s-xray'],
    'dirs': ['share'],
```

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,51 @@
...
+version = '2.5.4'
+
+homepage = 'https://nuetzlich.net/gocryptfs/'
+description = "An encrypted overlay filesystem written in Go. gocryptfs is built on top the excellent go-fuse FUSE library."
+
+toolchain = SYSTEM
+
+source_urls = ['https://github.com/rfjakob/gocryptfs/releases/download/v%(version)s/']
+sources = ['gocryptfs_v%(version)s_src-deps.tar.gz']
+patches = ['gocryptfs-2.5.4_install.patch']
```
jfgrimm, line 13: might as well use templates for the names
```suggestion
sources = ['%(name)s_v%(version)s_src-deps.tar.gz']
patches = ['%(name)s-2.5.4_install.patch']
```

jfgrimm (changes requested):

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
gkaf89: The gocryptfs package [depends on](https://github.com/rfjakob/gocryptfs/blob/master/go.mod) the [go-fuse](https://github.com/hanwen/go-fuse) package which in turn provides native bindings for the FUSE kernel module.

It could be an overkill, but can we check whether kernel modules exist and are loaded with EasyBuild?

Micket: In GDRCopy we alert whoever installs this what's required on the host kernel level
See https://github.com/easybuilders/easybuild-easyconfigs/blob/1b01bab361aad37a7110318eba083900adc80008/easybuild/easyconfigs/g/GDRCopy/GDRCopy-2.4.4-GCCcore-14.2.0.eb#L23-L37

Could you please add similar instructions here?

gkaf89: Thank you, Instructions were added in the latest commit.

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,57 @@
+easyblock = 'ConfigureMake'
```
Micket, line 1: So, i'd really try to get this through with the GoPackage easyblock instead. It does more things here, and would potentially simplify a bit in the steps here

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,57 @@
...
+    ('Go', '1.23.6'),
+    ('Pandoc', '3.6.2'),
+]
+
+build_info_msg = """This easyconfig relies on the go-fuse library that requires the FUSE kernel module at runtime.
+
+Kernel module for FUSE: a userspace filesystem framework.
+https://www.kernel.org/doc/html/next/filesystems/fuse.html
+RHEL module name: 'fuse'
+Debian module name: 'fuse'"""
```
Micket, line 29: This is just 
```suggestion
osdependencies = [('fuse-dev', 'fuse-devel')]
```
?

easybuild/easyconfigs/g/gocryptfs/gocryptfs-2.5.4.eb
```diff
@@ -0,0 +1,57 @@
...
+skipsteps = ['configure']
+
+# Make builds Go modules; isolate the build environment
+buildopts = 'GOPATH=%(builddir)s/go'
+
+build_cmd_targets = ['build']
+
+# Needs the DESTDIR variable to not use a system directory
+installopts = 'DESTDIR=%(installdir)s'
+
```
Micket, line 44: shouldn't need this with GoPackage

The only extra thing would be that separate gocryptfs-xray subpackage. 
Maybe one can just make that happen with postinstallcmds, or update the easyblock to support multiple targets.
gocryptfs-xray contrib/statfs contrib/findholes contrib/atomicrename

(it sure would be a lot easier if this project just followed standard go build procedure)

Micket (commented):
seems i was tagged as reviewer here, not sure why, but here is some comments though i see it's still as draft

gkaf89:
@jfgrimm @Micket Closing this PR in favor of the newer #25936.

The new PR relies on a custom easyblock and uses the GoPackage easyblock for the Go packages.

PR #22909, opened 2025-05-19
add patch for PMIx 5.0.2 to fix support for IPv6

sassy-crick:
(created using `eb --new-pr`)

This addresses the issue with `IPv6` as mentioned in issue [22910](https://github.com/easybuilders/easybuild-easyconfigs/issues/22910)


easybuild/easyconfigs/p/PMIx/PMIx-5.0.2-GCCcore-13.3.0.eb
```diff
@@ -31,6 +35,7 @@ dependencies = [
     ('hwloc', '2.10.0'),
 ]
 
+preconfigopts = './autogen.pl && '
```
Thyre, line 38: If we run `autogen.pl`, we should also have `Autotools` as a builddependency:

https://docs.openpmix.org/en/latest/developers/autogen.html#running-autogen-pl

> You can now run PMIx’s top-level autogen.pl script. This script will invoke the GNU Autoconf, Automake, and Libtool commands in the proper order and do a bunch of component discovery and housekeeping to setup to run PMIx’s top-level configure script.

Thyre: Depending on how large the changes by the patch introduced are compared to the changes introduced by re-running `autogen.pl`, maybe we can just patch `configure` itself (even though way more painful to do...)

sassy-crick: Thanks for finding that. I was so happy I got it all working....

Thyre: The OpenPMIx `autogen.pl` script contains some sections to work around / patch weird behavior of Autotools, and don't seem to force a particular Autotools version as far as I have seen (like we do with Score-P). So I think this should be fine, if everything builds correctly :smile:

sassy-crick:

easybuild/easyconfigs/p/PMIx/PMIx-5.0.2-GCCcore-13.3.0.eb
sassy-crick: Good call! I added the `Perl` version using the `SYSTEM` toolchain, as it is only required for the running of `autogen.pl`. If that is not ok, let me know and I will change it. I thought that is the simplest way forward. 
It all seems to work on our `IPv6` cluster so I assume the installation is working ok.

jfgrimm:
EB_ARGS='--installpath=/tmp/$USER/pr22909'

boegel (approved):
lgtm

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 3 out of 3 (1 easyconfigs in total)
node3505.doduo.os - Linux RHEL 9.4, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.18
See https://gist.github.com/boegel/6c3a286243bb3ab895db9ae26bd49e2b for a full test report.

boegel:

PR #22954, opened 2025-05-26
Flag easyconfigs in 2020b toolchains as archived

Micket:


boegel (dismissed):
@Micket Archiving nowadays means *removing* from this repository and copying (with history preserved) to https://github.com/easybuilders/easybuild-easyconfigs-archive/

@lexming has more info on the details on that (we should document the procedure more clearly?)

Micket:
@lexming have I misunderstood the procedure. Should i just rm the files here in this PR or do we do that in a follow up once the directory has been integrated.

I figured we'd archive 2020b stuff, then archive old stuff, and maybe also archive the semi-dead Java 1.8 stuff, then follow up with just deleting the __archive__ again (and incorporating it in the other repo)

lexming:
@Micket this is good and needed, thx! This is the first (and harder) step in the archival process. Once everything is under `__archive__` it is trivial to migrate the files to the archive repo.

lexming (approved):
LGTM

PR #22955, opened 2025-05-26
Archive pre-2021 easyconfigs older than 4 years old

Micket:
I include the 2020b archiving here because i need to check that if it passes CI or if i slipped something through that was still in use.


Micket:
Alright, so we need to skip archiving:

binutils/2.32
BLAST/2.2.26-Linux_x86_64
FastQC/0.11.9-Java-11
flex/2.6.4
Gblocks/0.91b
GenomeThreader/1.7.3-Linux_x86_64-64bit
Go/1.14
Go/1.14.1
GPAW-setups/0.9.20000
Java/13.0.2
M4/1.4.18
Maven/3.6.3
MetaGeneAnnotator/20080819-x86-64
ncurses/5.9
Pilon/1.23-Java-11
pplacer/1.1.alpha19
Trimmomatic/0.39-Java-11
zlib/1.2.11

Micket:
This should also be good to be merged. I don't think there was any complications at all here, just took the old stuff out. Had to keep some things because things depended on it. 

Some things i pulled out a bit early, the things that required Java 13. but Java 13 isn't updated anymore, it's not an LTS, it's not build by adoptium so there isn't any cross platform builds I can download. It really shouldn't ever been used, but unfortunately, there were a couple of things that slipped in.

lexming (approved):
LGTM

PR #22968, opened 2025-05-28
Archive Java 1.8

Micket:
Old (oracle) Java 1.8 is very outdated now, *really* shouldn't be used for anything these days. Only the updated Java-8.eb wrapper should be used instead, as it's all openjdk based instead, with more cross platform builds.


branfosj:
Going in, thanks @Micket!

boegel:
@Micket @branfosj This shouldn't have been merged, rather than moving stuff into `__archive__`, it should go to https://github.com/easybuilders/easybuild-easyconfigs-archive/ ?

lexming:
> @Micket @branfosj This shouldn't have been merged, rather than moving stuff into `__archive__`, it should go to https://github.com/easybuilders/easybuild-easyconfigs-archive/ ?

@boegel Moving things into `__archive__` does not hurt. It is a necessary intermediate step actually.

PR #22975, opened 2025-05-29
Add patch to fix permission denied issue during PyTables installation. 

dagonzalezfo:
During the PyTable package installation, you will get:
```bash
PermissionError: [Errno 13] Permission denied: '<buildpath>/tables/libblosc2.so'
```
This is trigger by using shutil.copy, that create a copy of from `$EBROOTLIBLOSC2/lib/libblosc2.so` on `<buildpath>/tables/libblosc2.so`. However, shutil.copy enforces to have the same permissions as the origin, causing the issue. 

This can be solved by changing `shutil.copy` by `shutil.copyfile`.


Crivella:
Tried to install the same package without this patch, but it still works, which is weird considering the behavior of `shutil.copy` and `shutil.copyfile` should be the same (I mean across our 2 different tests) as they came from the EB-intalled python.

Are you doing your test on some more *problematic* filesystem (eg NFS or Lustre)?

dagonzalezfo:
This issue is related with installing this package using the EESSI machinery. I manage to reproduce this issue on using a local storage. Without this patch, I get the reported error. In any case, I will do some testing outside of EESSI and update this PR accordingly.

casparvl:
Maybe I'm not fully understanding, but would this still be needed if we have https://github.com/easybuilders/easybuild-easyconfigs/pull/23042 ? I think that makes the current PR obsolete, no?

dagonzalezfo:
As Caspar comment, this is covered by https://github.com/easybuilders/easybuild-easyconfigs/pull/23042

PR #22994, opened 2025-06-01
{numlib}[system/system] cuDNN v9.10.1.4 w/ CUDA 12.8.0

Micket:
(created using `eb --new-pr`)

PR #23004, opened 2025-06-02
Add easyconfig for cuDNN-9.10.1.4-CUDA-12.4.0

adammccartney:
Added this manually. Trying to build out the dependencies needed by some GPU focused software on a stack using EESSI as a base.


Crivella:
I see you added a possible source for `ppc64le` but on 

https://developer.download.nvidia.com/compute/cudnn/redist/cudnn/linux-ppc64le/

I do not see the version for `9.10.1.4`? The latest on the specified source+arch seems to be `cudnn-linux-ppc64le-9.0.0.312_cuda12-archive.tar.xz`

Micket:
cf. #22994 22994

PR #23008, opened 2025-06-03
Add easyconfig for SciPy-bundle-2023.11-foss-2023b

adammccartney:


Micket:
This is the wrong toolchain (foss instead of gfbf) and it's an update of already existing scipy bundle for 2023b. This will just be a world of pain to have duplicate versions of python bundles within one toolchain, so we can't accept this upstream.

PR #22418, opened 2025-02-27
{compiler}[GCCcore/13.3.0] dpcpp v6.0.0

rafbiels:
Add new easyconfig for the open-source DPC++ compiler version 6.0.0 (https://github.com/intel/llvm/releases/tag/v6.0.0). This build supports SYCL compilation for OpenCL, Level Zero and CUDA backends, allowing to run SYCL code on x86 CPUs, Intel GPUs and NVIDIA GPUs.

A patch file and configuration options are in place to ensure compatibility with a wide range of CUDA releases by removing the dependency on CUPTI. This build of DPC++ can be used with any CUDA version starting from 11.7+ or without CUDA at all (to target only OpenCL and Level Zero backends).

PR #23028, opened 2025-06-04
Move Clang easyconfigs to use LLVM easyblock

Flamefire:
This is intended to test https://github.com/easybuilders/easybuild-easyblocks/pull/3755

I'll reduce this to changes and move the new easyconfigs out using correct names and comments (I need a specific commit for Triton)

### DIFF
--- a/easybuild/easyconfigs/c/Clang/Clang-18.1.6-GCCcore-13.3.0-CUDA-12.6.0.eb
+++ b/easybuild/easyconfigs/c/Clang/Clang-18.1.6-GCCcore-13.3.0-CUDA-12.6.0.eb
@@ -2,7 +2,6 @@ name = 'Clang'
 version = "18.1.6"
 versionsuffix = '-CUDA-%(cudaver)s'
 easyblock = 'EB_LLVM'
-bootstrap = False
 
 homepage = 'https://clang.llvm.org/'
 description = """C, C++, Objective-C compiler, based on LLVM.  Does not
--- a/easybuild/easyconfigs/c/Clang/Clang-18.1.6-GCCcore-13.3.0-CUDA-12.6.0.eb
+++ b/easybuild/easyconfigs/c/Clang/Clang-18.1.6-GCCcore-13.3.0-CUDA-12.6.0.eb
@@ -1,5 +1,5 @@
 name = 'Clang'
-version = "18.1.8"
+version = "18.1.6"
 versionsuffix = '-CUDA-%(cudaver)s'
 easyblock = 'EB_LLVM'
 bootstrap = False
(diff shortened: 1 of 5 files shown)


Crivella:
I think we should really move on from `Clang` and start replacing it with `LLVM` whereever it was used as a dependency.
The new easyblock will always produce atleast both `clang` and `flang` which i feel makes the old name confusing.
Also to avoid duplication of effort in maintaining both sets of EasyConfigs

We should also use `lit` to run the full test suite, a huge part of the effort on the new easyblock was dedicated to making sure we actually are able to.

@Thyre For the CUDA part you definitely have more experience, do we actually need CUDA as a runtime dependency (or even a build dependency, maybe for running the tests?)
Especially with LLVM>=20 the `libomptarget-nvptx.bc` will end up being produced regardless and does also not need the compute capabilities specified anymore.

Thyre:
CUDA (and in the same sense ROCm) is a tiny bit complicated. We absolutely do not need it to build offload support.
We need it if we want to run the tests, and if we want to ensure that system versions are not picked up. The last part will be resolved with https://github.com/easybuilders/easybuild-easyblocks/pull/3747

I'm not against having it in the EasyConfigs, since that also makes absolutely clear that NVPTX support will be built. I see it more or less as a convenience.

I guess its similar to #22418 in a sense, where CUDA is required for the build, but doesn't actually matter during runtime. Only when you want to offload, you'll need to load CUDA.

Flamefire:
If we remove CUDA as a (build)dependency then the LLVM easyblock will not select the NVTPX backend. And manually specifying it will remove the other defaults.   
We could add a new parameter like `use_cuda` and `use_rocm` if the tests work without. That has the advantage of getting rid of the CUDA-suffixed ECs.

Thyre:
> If we remove CUDA as a (build)dependency then the LLVM easyblock will not select the NVTPX backend. And manually specifying it will remove the other defaults. We could add a new parameter like `use_cuda` and `use_rocm` if the tests work without. That has the advantage of getting rid of the CUDA-suffixed ECs.

LLVM selects the NVPTX backend when `cuda-compute-capabilities` are specified, see:

https://github.com/easybuilders/easybuild-easyblocks/blob/ac2b2f9a5069927a31079219aee732e169e0b05e/easybuild/easyblocks/l/llvm.py#L381-L388

```
            # If CUDA is included as a dep, add NVPTX as a target
            # There are (old) toolchains with CUDA as part of the toolchain
            if 'cuda' in self.deps or cuda_toolchain:
                self.log.info("CUDA dependency detected, adding NVPTX as a target")
                build_targets.append(BUILD_TARGET_NVPTX)
            elif cuda_cc_list:
                self.log.info("CUDA compute capabilities specified, adding NVPTX as a target")
                build_targets.append(BUILD_TARGET_NVPTX)
```

with a similar condition for AMD, though there's no config option for that yet (see https://github.com/easybuilders/easybuild-framework/pull/4860)

Crivella:
Also for LLVM >= 20  we have `build_targets = ['all']` in the EC since we can safely build for all targets by default as we do not neede to specify the CC and AMD_GFX anymore (if you try that with previous version the EasyBlock will complain that you need those set to build for NVPTX and AMDGPU)

Flamefire:
I've moved only the Clang 18 easyconfigs: https://github.com/easybuilders/easybuild-easyconfigs/pull/23055

Goal there was to have the same result as before for the existing easyconfigs. Newer ECs should use LLVM directly, so closing this.

PR #23042, opened 2025-06-05
Add a patch that enable to use libraries exposed via LIBRARY_PATH or

dagonzalezfo:
Enabling using LIBRARY_PATH exposed shared libraries will enhance user experience using local installed packages that relies on external libraries. Current solution requires using LD_LIBRARY_PATH or adding the neeeded libraries to python main installation. This is related with an EESSI reported issue. 
Further testing and refinement is needed.

### DIFF
--- a/easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
+++ b/easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
@@ -1,0 +1,52 @@
+diff -ruN Python-3.11.5.orig/Lib/ctypes/__init__.py Python-3.11.5/Lib/ctypes/__init__.py
+--- Python-3.11.5.orig/Lib/ctypes/__init__.py	2025-06-03 17:24:09.817225787 +0200
++++ Python-3.11.5/Lib/ctypes/__init__.py	2025-06-03 17:26:02.851166381 +0200
+@@ -13,6 +13,8 @@
+ from _ctypes import ArgumentError
+ 
+ from struct import calcsize as _calcsize
++from ctypes import util
++import re
+ 
+ if __version__ != _ctypes_version:
+     raise Exception("Version number mismatch", __version__, _ctypes_version)
+@@ -349,6 +351,11 @@
+             flags |= _FUNCFLAG_USE_ERRNO
+         if use_last_error:
+             flags |= _FUNCFLAG_USE_LASTERROR
++        if _os.name == "posix":
++            if name and name.endswith(".so"):
++                s = re.sub(r'lib', '', name)
++                s = re.sub(r'\..*', '', s)
++                self._name=util._findLib_ld(s) 
+         if _sys.platform.startswith("aix"):
+             """When the name contains ".a(" and ends with ")",
+                e.g., "libFOO.a(libFOO.so)" - this is taken to be an
+diff -ruN Python-3.11.5.orig/Lib/ctypes/util.py Python-3.11.5/Lib/ctypes/util.py
+--- Python-3.11.5.orig/Lib/ctypes/util.py	2025-06-03 17:24:09.816225753 +0200
++++ Python-3.11.5/Lib/ctypes/util.py	2025-06-03 17:28:52.212009753 +0200
+@@ -209,7 +209,7 @@
+             expr = os.fsencode(expr)
+ 
+             try:
+-                proc = subprocess.Popen(('/sbin/ldconfig', '-r'),
++                proc = subprocess.Popen((os.environ.get('EPREFIX') + '/sbin/ldconfig', '-r'),
+                                         stdout=subprocess.PIPE,
+                                         stderr=subprocess.DEVNULL)
+             except OSError:  # E.g. command not found
+@@ -301,7 +301,14 @@
+             # See issue #9998 for why this is needed
+             expr = r'[^\(\)\s]*lib%s\.[^\(\)\s]*' % re.escape(name)
+             cmd = ['ld', '-t']
+-            libpath = os.environ.get('LD_LIBRARY_PATH')
++            libpath = []
++            if os.getenv('EPREFIX'):
++                libpath.append(os.getenv('EPREFIX'))
++            if os.getenv('LD_LIBRARY_PATH'):
++                libpath.append(os.getenv('LD_LIBRARY_PATH'))
++            if os.getenv('LIBRARY_PATH'):
++                libpath.append(os.getenv('LIBRARY_PATH'))
++            libpath = ':'.join(libpath)
+             if libpath:
+                 for d in libpath.split(':'):
+                     cmd.extend(['-L', d])
--- a/easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
@@ -10,7 +10,11 @@ toolchainopts = {'pic': True}
 
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
-checksums = ['a12a0a013a30b846c786c010f2c19dd36b7298d888f7c4bd1581d90ce18b5e58']
+patches = ['Python-3.11.5-custom-ctypes.patch']
+checksums = [
+    {'Python-3.11.5.tgz': 'a12a0a013a30b846c786c010f2c19dd36b7298d888f7c4bd1581d90ce18b5e58'},
+    {'Python-3.11.5-custom-ctypes.patch': 'fd56e81986051f22304dfc88953f5a5aad0c060f8e8649faa3633de644f5a61d'},
+]
 
 builddependencies = [
     ('UnZip', '6.0'),


casparvl:
I'll try to test this patch in a bit, but @ocaisa already remarked it's a "scary big hammer", since it changes a very fundamental standard python library. Previously, when we patched `ctypes.util.find_library`, we did this as selectively as we could, i.e. only when `LD_LIBRARY_PATH` [is filtered](https://github.com/easybuilders/easybuild-easyblocks/blob/0f4157641cd21688f48532cd75153ebe65aecb92/easybuild/easyblocks/p/python.py#L355). This means it needed to be done at the `python` easyblock level, because it involves logic. I think it makes sense to do that again this time - but let me first test if it resolves the issues I'm seeing. In the EESSI support meeting, it came up that LAMMPS with GPU support ran into similar issues, and there was some concern about the `libcuda.so` stubs library being picked up as a result of this patch, that's another thing that we should test for.

casparvl:
The good news is: the hammer does seem to solve my particular problem.

Before, I got this issue:

```
  * Found HDF5 headers at ``/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/HDF5/1.14.3-gompi-2023b/include``, library at ``/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/HDF5/1.14.3-gompi-2023b/lib``.
  .. WARNING:: Could not find the HDF5 runtime.
     The HDF5 shared library was *not* found in the default library
     paths. In case of runtime problems, please remember to install it.
  * Found LZO 2 headers at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/LZO/2.10-GCCcore-13.2.0/include``, the library is located in the standard system search dirs.
  .. WARNING:: Could not find the LZO 2 runtime.
     The LZO 2 shared library was *not* found in the default library
     paths. In case of runtime problems, please remember to install it.
  * Skipping detection of LZO 1 since LZO 2 has already been found.
  * Found bzip2 headers at ``/usr/include``, library at ``/usr/lib64``.
  * Found blosc headers at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc/1.21.5-GCCcore-13.2.0/include``, the library is located in the standard system search dirs.
  .. WARNING:: Could not find the blosc runtime.
     The blosc shared library was *not* found in the default library
     paths. In case of runtime problems, please remember to install it.
  * Run 'blosc2_find_directories_hook'
  * Found blosc2 headers at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc2/2.13.2-GCCcore-13.2.0/include``, library at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc2/2.13.2-GCCcore-13.2.0/lib``.
    * Copying blosc2 runtime library to 'tables' dir because it was not found in standard locations
```

Now, PyTables builds succesfully, and I find this in the logs:

```
  * Using Python 3.11.5 (main, Jun 10 2025, 13:20:30) [GCC 13.2.0]
  * Found cython 3.0.4
  * USE_PKGCONFIG: True
  * Found HDF5 headers at ``/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/HDF5/1.14.3-gompi-2023b/include``, library at ``/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/HDF5/1.14.3-gompi-2023b/lib``.
  * Found LZO 2 headers at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/LZO/2.10-GCCcore-13.2.0/include``, the library is located in the standard system search dirs.
  * Skipping detection of LZO 1 since LZO 2 has already been found.
  * Found bzip2 headers at ``/usr/include``, library at ``/usr/lib64``.
  * Found blosc headers at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc/1.21.5-GCCcore-13.2.0/include``, the library is located in the standard system search dirs.
  * Run 'blosc2_find_directories_hook'
  * Found blosc2 headers at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc2/2.13.2-GCCcore-13.2.0/include``, library at ``/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc2/2.13.2-GCCcore-13.2.0/lib``.
  Compiling tables/utilsextension.pyx because it changed.
  [1/1] Cythonizing tables/utilsextension.pyx
  Compiling tables/hdf5extension.pyx because it changed.
  [1/1] Cythonizing tables/hdf5extension.pyx
  Compiling tables/tableextension.pyx because it changed.
  [1/1] Cythonizing tables/tableextension.pyx
  Compiling tables/linkextension.pyx because it changed.
  [1/1] Cythonizing tables/linkextension.pyx
  Compiling tables/_comp_lzo.pyx because it changed.
  [1/1] Cythonizing tables/_comp_lzo.pyx
  Compiling tables/_comp_bzip2.pyx because it changed.
  [1/1] Cythonizing tables/_comp_bzip2.pyx
  Compiling tables/lrucacheextension.pyx because it changed.
  [1/1] Cythonizing tables/lrucacheextension.pyx
  Compiling tables/indexesextension.pyx because it changed.
  [1/1] Cythonizing tables/indexesextension.pyx
  [06/10/25 13:33:50] WARNING  pyproject.toml does not contain a tool.setuptools_scm    setuptools.py:119
```

Also, before, I got:

```
module load Blosc2/2.13.2-GCCcore-13.2.0
module load Python/3.11.5-GCCcore-13.2.0
$ python -c "import ctypes; ctypes.CDLL('libblosc2.so')"
  File "/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/__init__.py", line 376, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: libblosc2.so: cannot open shared object file: No such file or directory
```

Whereas now I get:

```
module load Blosc2/2.13.2-GCCcore-13.2.0
module load Python/3.11.5-GCCcore-13.2.0
$ python -c "import ctypes; print(ctypes.CDLL('libblosc2.so'))"
<CDLL '/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc2/2.13.2-GCCcore-13.2.0/lib/libblosc2.so', handle 13748a0 at 0x145c9d1e9990>
```

As expected

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,52 @@
...
+                e.g., "libFOO.a(libFOO.so)" - this is taken to be an
+diff -ruN Python-3.11.5.orig/Lib/ctypes/util.py Python-3.11.5/Lib/ctypes/util.py
+--- Python-3.11.5.orig/Lib/ctypes/util.py	2025-06-03 17:24:09.816225753 +0200
++++ Python-3.11.5/Lib/ctypes/util.py	2025-06-03 17:28:52.212009753 +0200
+@@ -209,7 +209,7 @@
+             expr = os.fsencode(expr)
+ 
+             try:
+-                proc = subprocess.Popen(('/sbin/ldconfig', '-r'),
++                proc = subprocess.Popen((os.environ.get('EPREFIX') + '/sbin/ldconfig', '-r'),
```
casparvl, line 33: I think this is another reason why we should resolve things at the EasyBlock level: there, we can query if EasyBuild was configured with a `sysroot` build option. If so, we get _that_ value, instead of `EPREFIX` (which I believe is set by Gentoo-prefix, and thus specific to that OS?). That makes this more generic. It also excludes any possible name collisions (someone could set `EPREFIX` as an environment for a totally different purpose - unrelated to Gentoo-Prefix, but `build_option('sysroot')` can _only_ mean one thing: someone wanted to set a sysroot).

ocaisa:
@casparvl Can you try
```
python -c "import ctypes; print(ctypes.CDLL('libcuda.so'))"
```
with/without the CUDA module loaded on a system with a GPU? I'm curious to see if it picks up the "correct" one

casparvl:
@ocaisa correctly anticipated that this would be a problem:

```
$ module load CUDA/12.4.0
$ module load Python/3.11.5-GCCcore-13.2.0
$ python -c "import ctypes; print(ctypes.CDLL('libcuda.so'))"
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/intel/icelake/accel/nvidia/cc80/software/CUDA/12.4.0/stubs/lib64/libcuda.so', handle 13db710 at 0x15413f109c90>
```

Now, it may or may not actually bite us in practice. If a `libxyz.so` was built against `libcuda.so`, and if we import `libxyz.so` in Python through `ctypes`, all is _probably_ fine, because the runtime linker will load `libcuda.so` in the regular way (i.e. no Python shenanigans going on). Since the `stubs` dir are _not_ on the `LD_LIBRARY_PATH`, this means the runtime linker will correctly find the driver in the default location (`/usr/lib/libcuda.so.1` or something). This (probably) only causes trouble if a code directly loads `libcuda.so` through the `ctypes` interface.

One thing we could do - but this is even more invasive - is that we patch `ctypes` some more and do

```
library_path = os.getenv('LIBRARY_PATH')
# some code to clean up `stubs` prefixes from `library_path`. Or maybe more specific: clean up any directory matching `CUDA/*/stubs/lib64`
library_path = remove_stubs_from_library_path(library_path)
libpath.append(library_path)
```

casparvl:
Lol, I was already doing that @ocaisa :D just needed time for the write-up (and got distracted by something on Slack :P)

ocaisa:
For EESSI, this is actually less of a problem as we drop CUDA to a build dependency. This means that at runtime the stubs `libcuda.so` won't be available via `LIBRARY_PATH` either.

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,52 @@
...
+     raise Exception("Version number mismatch", __version__, _ctypes_version)
+@@ -349,6 +351,11 @@
+             flags |= _FUNCFLAG_USE_ERRNO
+         if use_last_error:
+             flags |= _FUNCFLAG_USE_LASTERROR
++        if _os.name == "posix":
++            if name and name.endswith(".so"):
++                s = re.sub(r'lib', '', name)
++                s = re.sub(r'\..*', '', s)
++                self._name=util._findLib_ld(s)
```
casparvl, line 21: I'm not sure I understand what's happening here. What issue is this actually resolving? Also, invoking _findLib_ld means we need to make absolutely sure that _that_ searches the correct path (though I guess that's ensured by the changes below as well)

casparvl: oooh, wait, this is where you overwrite the name with the full path already? I.e. this, together with the change below, makes sure we get something like `/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Blosc2/2.13.2-GCCcore-13.2.0/lib/libblosc2.so` back, and then the rest of the `CDLL` call just acts on that full path?

In other words: this is where we are actually _pretending_ the calling code called the function with the full path, thus avoiding the whole issue?

dagonzalezfo: Yes, by overriding `self._name`

dagonzalezfo:
I will encapsulate all changes using sysroot as condition, as you commented previously.

casparvl:
Summarizing, my suggestion would be:

1. Comment out the original patch for `find_library` at https://github.com/easybuilders/easybuild-easyblocks/blob/0f4157641cd21688f48532cd75153ebe65aecb92/easybuild/easyblocks/p/python.py#L351
2. Replace that with your fix, but done as part of the easyblock
3. When `LD_LIBRARY_PATH` is filtered, add `libpath.append(os.getenv('LIBRARY_PATH'))`
4. When sysroot is set, do `proc = subprocess.Popen(build_option('sysroot') + '/sbin/ldconfig', '-r'),` and ` libpath.append(os.path.join(build_option(`sysroot`), "path", "to", "default", "libs"))`
5. Test if this fix indeed makes the original patch for `find_library` obsolete (i.e. check that the original issue for that is still resolved). If not, restore that patch.

dagonzalezfo:
Patch migrated to https://github.com/easybuilders/easybuild-easyblocks/pull/3798.
Thus, I will close this PR

PR #21639, opened 2024-10-14
{toolchain}[system/system] nvofbf v2024.9, nvompi v2024.9, OpenMPI v5.0.3

sassy-crick:
(created using `eb --new-pr`)

PR #23054, opened 2025-06-06
{data}[nvofbf/2024.9] HDF5 v1.14.5

maxim-masterov:
HDF5-1.14.5 with MPI

Depends on:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/21639

### DIFF
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.14.5-nvompi-2024.09.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.14.5-nvompi-2024.09.eb
@@ -1,0 +1,32 @@
+name = 'HDF5'
+# Note: Odd minor releases are only RCs and should not be used.
+version = '1.14.5'
+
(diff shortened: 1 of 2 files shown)


WilleBell:
This easyconfig uses nvompi toolchain. Maybe this should be closed since we have [HDF5-1.14.5-NVHPC-25.1.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/h/HDF5/HDF5-1.14.5-NVHPC-25.1.eb), would that make sense?

maxim-masterov:
> This easyconfig uses nvompi toolchain. Maybe this should be closed since we have [HDF5-1.14.5-NVHPC-25.1.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/h/HDF5/HDF5-1.14.5-NVHPC-25.1.eb), would that make sense?

Is this PR still relevant? I mean, it's already more than one year old. IMO, it's better to close it.

PR #23100, opened 2025-06-16
restore test configurations

boegel:
reverts changed made in #23099, should only be merged after merge sprint of Mon 16 June 2025


branfosj:
We should also re-add the branch protection rules (Settings -> Branches) for these three when they are re-enabled in the testing.

PR #23103, opened 2025-06-16
Clarify use of download_instructions parameter in test_downloadable_or_instructions

WilleBell:


test/easyconfigs/easyconfigs.py
```diff
@@ -995,7 +995,10 @@ def get_deps_for(ec):
             self.fail('Should not have multiple variants of dependencies.\n' + multi_dep_vars_msg)
 
     def test_downloadable_or_instructions(self):
-        """Make sure the sources are downloadable or there are instructions for how to download them."""
+        """
+        Make sure the sources are downloadable or there are instructions for how to download them
+        using the download_instructions parameter, including a link.
```
lexming: Links do not always exist. It is not so uncommon to get sources that are only distributed between a closed group of researchers, with no link for documentation or anything else.
```suggestion
        using the download_instructions parameter, including a link if one exists.
```

lexming:
Merging, thanks @WilleBell !

PR #23111, opened 2025-06-17
util-linux: fix broken meson code

akesandgren:
Can't do += on find_library


akesandgren:
I managed to make the PR against the wrong base repo...

PR #23146, opened 2025-06-19
remove archived toolchains from `gcc_tc_gen_map` and correct comments

branfosj:


branfosj:
This code is included in #23177, so closing.

PR #23162, opened 2025-06-20
Update R-4.5.1 and update packages for gfbf 2023b 

w34sel:
Update R-4.5.1 and update packages for gfbf 2023b


verdurin:
@w34sel thanks for the contribution, and welcome to EasyBuild.

Could I ask why you're targeting an older toolchain with a very recent version of `R`?

Also, generally we try now to stick to a single `R` version within what you might think of as a "generation" of toolchains.
Yes, we have been a bit loose about this in the past...

w34sel:
Thanks! We're still running CentOS 7 and currently planning our upgrade path. Because of that, we decided to stick at foss 2023b for the time being.

PR #12687, opened 2021-04-20
improve check for multi-variant dependencies per generation of easyconfigs (for now, only for 2025a & newer)

boegel:
Issues fixed:

* also check dependency variants for easyconfigs using a `GCC` toolchain;
* also check dependency variants for easyconfigs using a `GCCcore-10.*` toolchain;
* check dependency variants per *generation* of easyconfigs (fixes #9846);

PR #23177, opened 2025-06-21
restore toolchain multideps test

branfosj:
#12687 switched to testing for multidep variants per toolchain generation, but this is only enabled for 2023b and newer.

We should still test for multidep variants per toolchain for the older generations.

The previous check (`r'GCCcore-([7-9]|[1-9][0-9])\.[0-9]'`) missed `GCC` toolchains and I've retained that due to there being multidep variant failures in `GCC` toolchains. The check on the newer generations will catch these problems, so the historical ones will get archived away over time.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -742,7 +742,7 @@ class EasyConfigTest(TestCase):
                 genver = gen.split('-', 1)[1]
                 is_recent_gen = LooseVersion(genver) >= LooseVersion('10.2')
             else:
-                raise EasyBuildError("Unkown type of toolchain generation: %s" % gen)
+                raise EasyBuildError("Unknown type of toolchain generation: %s" % gen)
 
             if is_recent_gen:
                 py2_dep_vars = [x for x in dep_vars.keys() if '; versionsuffix: -Python-2.' in x]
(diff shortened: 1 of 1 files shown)


test/easyconfigs/easyconfigs.py
```diff
@@ -957,23 +963,85 @@ def get_deps_for(ec):
...
 
-        # some software also follows <year>{a,b} versioning scheme,
-        # which throws off the pattern matching done below for toolchain versions
-        false_positives_regex = re.compile(r'^MATLAB(-Engine)?-20[0-9][0-9][ab]')
+        multi_dep_vars_msg = ''
+        # restrict to checking dependencies of easyconfigs using common toolchains
+        # and GCCcore subtoolchain for common toolchains
+        patterns = [
+            # compiler-only subtoolchains GCCcore and GCC
+            r'GCC(core)?-1(0\.3|[12]\.[0-9])',  # GCCcore 10.3 to 12.3
```
branfosj:
```suggestion
            # compiler-only subtoolchains GCCcore, this pattern has never checked GCC toolchains
            # retain the check as it has always been in the old generation check
            r'GCCcore-1(0\.3|[12]\.[0-9])',  # GCCcore 10.3 to 12.3
```

test/easyconfigs/easyconfigs.py
```diff
@@ -996,17 +1065,11 @@ def get_deps_for(ec):
         multi_dep_vars_msg = ''
         # restrict to checking dependencies of easyconfigs using common toolchains,
         # and GCCcore subtoolchain for common toolchains;
-        # for now, we only enforce this for recent toolchain versions (2025a + GCCcore 14.x, and newer);
+        # for now, we only enforce this for recent toolchain versions (2023b + GCCcore 13.x, and newer);
         patterns = [
             # compiler-only subtoolchains GCCcore and GCC
-            # r'GCCcore-[7-9]\.[0-9]\.',
-            # r'GCC(core)?-1[0-9]\.[0-9]\.',  # GCCcore 10.x, etc.
             r'GCC(core)?-1[3-9]\.[0-9]\.',  # GCCcore 13.x & newer
```
Micket, line 1071: Do we also want to check for intel-compilers?

branfosj: Possibly, but I'd like to do that as a follow on and use this PR for restoring the previous check.

Micket (approved):
lgtm, but I think we should really just have 
```
def test_versions_per_toolchain(self):
```
that verifies that we only have one version of something per toolchain (with some exceptions), as the existence of these alterantive versions we occasionally have is just a net negative. Like two versions of giflib that tripped me up yesterday and what not. 

In particular for GCCcore level, just one version of things please (unless strongly motivated why not).

PR #17469, opened 2023-03-05
replace src include path with installation dir for HDF5

smoors:
(created using `eb --new-pr`)

HDF5 adds the following include path to $H5BLD_CPPFLAGS in `h5c++` and `h5pcc`
```
/tmp/HDF5/1.14.0/gompi-2022a/hdf5-1.14.0/src/H5FDsubfiling
```

this PR replaces the path with the one of installation directory
see also https://github.com/easybuilders/easybuild-easyconfigs/pull/17452

PR #23219, opened 2025-06-26
Update HDF5-1.14.0-iimpi-2022b.eb - update name of updated exec h5cc in postinstallcmds

pavelToman:
Discussed here:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/17469#issuecomment-3008824817


pavelToman:
`h5cc` is only in -serial versions of HDF5 - where 'usempi': False

PR #20945, opened 2024-07-01
{vis}[GCCcore/13.2.0,foss/2023b] Octave v9.2.0, GL2PS v1.4.2, gnuplot v6.0.1, ...

SimonPinches:
(created using `eb --new-pr`)

PR #22930, opened 2025-05-22
{vis}[GCCcore/12.3.0,GCCcore/13.3.0] Added `GraphicsMagick-1.3.45` EC for GCCcore-12.3.0 + fix for #22927

Crivella:
New EC files for `GraphicsMagick` and `foss/2023a` with potential fix for 

- #22927 
- #22929

If from the discussion on the PR/Issue it is decided this is the proper way to go we should apply the change to enable both static and shared builds (+ sanity checks)

```
# When building Octave, mixing rpathed shared libraries with libGraphicsMagick++.a will cause the check on the library
# to fail due to missing symbols. The default previously was to build the static library only.
configure_opts = '--enable-shared --enable-static'
```

also to

GraphicsMagick-1.3.45-GCCcore-13.3.0.eb

Issue #22929, opened 2025-05-22
Octave configure script fails to pick up statically compiled GraphicMagick library


Crivella:
When trying a new build of Octave, the detection of GraphicMagick in the configure scripts fails for 2 reasons.

The first is highlighted in

- #22927

The second might only be present when the dependencies relies on RPATHING and is due to a mixing of static and dynamic libraries.
This causes of failure in compiling a test program in the configure due to several undefined symbols, as the static version of the libraries does not carry the RPATH information with it.

Making sure GraphicMagick compiles with both `--enable-shared --enable-static` fixes this

Issue #22927, opened 2025-05-22
GraphickMagick `pkg-config` fails due to misnamed `libtiff-4.pc` file


Crivella:
I am unsure if this should be patched in GraphickMagick or in LibTIFF, but while trying a new build of Octave, GraphicMagick was not being picked up due to the command

`pkg-config --exists --print-errors GraphicMagick++` failing with the error

```
Package libtiff was not found in the pkg-config search path.
Perhaps you should add the directory containing `libtiff.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libtiff', required by 'GraphicsMagick', not found
```

The problem is that LibTIFF only creates a `libtiff-4.pc` files for `pkg-config` to pick up.
I am unsure if this should be patched for GraphicMagick or LibTIFF.

The latter is probably the simplest as it requires only creating a symlink from `libtiff-4.pc` -> `libtiff.pc` so that both schemes will work

PR #23229, opened 2025-06-28
{math}[intel/2023b] Octave v9.2.0, SUNDIALS v7.0.0, GraphicsMagick v1.3.43

SimonPinches:
(created using `eb --new-pr`)

This PR depends upon
- [x] #20945
- [x] #26712


Crivella:
Could you check this against
- https://github.com/easybuilders/easybuild-easyconfigs/pull/22930

I think you might be hitting the same error as in
- https://github.com/easybuilders/easybuild-easyconfigs/issues/22929

Where `Octave` still compiles but silently fails the check for `GraphicsMagick` which will lead to `imshow`/`imread` not working (either partially or fully)

Crivella:
Among the mentioned issues i had forgotten the nested
- https://github.com/easybuilders/easybuild-easyconfigs/issues/22927

It is also here we might have to use the patch for `libtiff` with the `-4`.
Could you check the configure script of the Octave build to see if `GraphicsMagick` is actually being used and if `imread` and `imshow` works as expected in Octave?

https://docs.octave.org/latest/External-Packages.html

Crivella:
From https://github.com/easybuilders/easybuild-easyconfigs/pull/20945#issuecomment-3083114050

Any particular reason in using arpack-ng==3.9.1 and Java==21, have not tested them in particular extensively when i made the Octave 10 EC, but also did not see anything in particular complaining in the configure step.

We are either going to have to use the same as other packages in the toolchain or add an exception in the tests that are failing the CI

Crivella:
Beside the CI problems similar to the other PR, i am also hitting a missing dep in the configure (should see if it is possible to use the mkl version 

```
configure: WARNING: FFTW3 library not found.  Octave will not be able to perform fft calculations.
configure: WARNING: FFTW3F library not found.  Octave will not be able to perform fft calculations.
```

and a segfault in the testsuite

```
/bin/bash: line 1: 3654689 Segmentation fault      (core dumped) ( /bin/bash ../run-octave --no-init-file --silent --no-history -p /home/crivella/.local/easybuild/build/Octave/9.2.0/intel-2023b/octave-9.2.0/test/mex /home/crivella/.local/easybuild/build/Octave/9.2.0/intel-2023b/octave-9.2.0/test/fntests.m /home/crivella/.local/easybuild/build/Octave/9.2.0/intel-2023b/octave-9.2.0/test "" )
make[4]: *** [Makefile:3202: check-local] Error 139
```

Crivella:

SimonPinches:
> Beside the CI problems similar to the other PR, i am also hitting a missing dep in the configure (should see if it is possible to use the mkl version
> 
> ```
> configure: WARNING: FFTW3 library not found.  Octave will not be able to perform fft calculations.
> configure: WARNING: FFTW3F library not found.  Octave will not be able to perform fft calculations.
> ```
> 
> and a segfault in the testsuite
> 
> ```
> /bin/bash: line 1: 3654689 Segmentation fault      (core dumped) ( /bin/bash ../run-octave --no-init-file --silent --no-history -p /home/crivella/.local/easybuild/build/Octave/9.2.0/intel-2023b/octave-9.2.0/test/mex /home/crivella/.local/easybuild/build/Octave/9.2.0/intel-2023b/octave-9.2.0/test/fntests.m /home/crivella/.local/easybuild/build/Octave/9.2.0/intel-2023b/octave-9.2.0/test "" )
> make[4]: *** [Makefile:3202: check-local] Error 139
> ```

I didn't manage to persuade Octave to use the MKL version of FFTW so added an explicit version which seems to solve the FFTW problem.

Crivella:
https://github.com/easybuilders/easybuild-easyconfigs/pull/23229#issuecomment-3106707230
It seems to be failing with the same segmentation fault i observed locally :thinking:

SimonPinches:
> [#23229 (comment)](https://github.com/easybuilders/easybuild-easyconfigs/pull/23229#issuecomment-3106707230) It seems to be failing with the same segmentation fault i observed locally 🤔

I can reproduce this segfault but am not sure of the cause 😢

boegel:
CORE_CNT=16
EB_ARGS="--parallel 8"

easybuild/easyconfigs/f/FFTW/FFTW-3.3.10-iimpi-2023b.eb
```diff
@@ -0,0 +1,27 @@
+name = 'FFTW'
+version = '3.3.10'
+
+homepage = 'https://www.fftw.org'
+description = """FFTW is a C subroutine library for computing the discrete Fourier transform (DFT)
+in one or more dimensions, of arbitrary input size, and of both real and complex data."""
+
+toolchain = {'name': 'iimpi', 'version': '2023b'}
```
akesandgren, line 8: FFTW belongs at the GCC/intel-compiler level and there is no version for intel-compiler/2023.2.1
So if this really is something you need to use you need to change this to intel-compiler level.

easybuild/easyconfigs/o/Octave/Octave-9.2.0-intel-2023b.eb
```diff
@@ -0,0 +1,87 @@
...
+    ('Qt6', '6.6.3'),
+    ('HDF5', '1.14.3'),
+    ('qrupdate', '1.1.2'),
+    ('SuiteSparse', '7.7.0'),
+    ('makeinfo', '7.1'),
+    ('libsndfile', '1.2.2'),
+    ('GraphicsMagick', '1.3.45'),
+    ('SUNDIALS', '7.0.0'),
+    ('RapidJSON', '1.1.0-20240409'),
+    ('FFTW', '3.3.10'),
```
akesandgren, line 53: It would probably be better if Octave could be taught to use MKL's fftw interface.
It just needs to use $LIBFFT instead of any hardcoded -lfftw3 or similar

easybuild/easyconfigs/f/FFTW/FFTW-3.3.10-iimpi-2023b.eb
SimonPinches: Now not needed and so removed from this PR.

easybuild/easyconfigs/o/Octave/Octave-9.2.0-intel-2023b.eb
SimonPinches: @akesandgren, I've managed to replace `FFTW` with `MKL`'s fftw interface as suggested.

SimonPinches:
@akesandgren, @Crivella,  since `qrupdate` at the `GCCcore` level doesn't work for `Octave` when built for Intel and using MKL, I had to create a `qrupdate` eb for `intel-compilers`.  I've left the orignal eb at the `GCCcore` level but I think it should move to `GCC`.  Since, as far as I can tell, `Octave` is the only easyconfig that depends upon `qrupdate` I don't think this should cause any problems but I wait for confirmation before doing anything drastic.

akesandgren:
If it doesn't work at GCCcore level then do indeed upgrade it to GCC/intel-compiler.
And since qrupdate does have calls to BLAS functions it HAS to be at GCC/intel-compiler level since that's where OpenBLAS is.

Please generate new qrupdate ECs (possible in a separate PR) and if possible ones that actually runs the tests and creates shared libs if you feel up to it.
And in that PR also get rid of the existing qrupdates at GCCcore level since Octave is the only user there is no problem here I'd say.

SimonPinches:
> If it doesn't work at GCCcore level then do indeed upgrade it to GCC/intel-compiler. And since qrupdate does have calls to BLAS functions it HAS to be at GCC/intel-compiler level since that's where OpenBLAS is.
> 
> Please generate new qrupdate ECs (possible in a separate PR) and if possible ones that actually runs the tests and creates shared libs if you feel up to it. And in that PR also get rid of the existing qrupdates at GCCcore level since Octave is the only user there is no problem here I'd say.

@akesandgren, see https://github.com/easybuilders/easybuild-easyconfigs/pull/26712.

easybuild/easyconfigs/q/qrupdate/qrupdate-1.1.2-intel-compilers-2023.2.1.eb
```diff
@@ -0,0 +1,34 @@
+easyblock = 'MakeCp'
```
Crivella: Should change this to follow the same as in https://github.com/easybuilders/easybuild-easyconfigs/pull/26712

Crivella (dismissed):

easybuild/easyconfigs/q/qrupdate/qrupdate-1.1.2-intel-compilers-2023.2.1.eb
SimonPinches: Done.

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 6 out of 6 (total: 51 mins 7 secs) (5 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), 1 x NVIDIA NVIDIA GeForce RTX 5060, 580.173.02, Python 3.11.15
See https://gist.github.com/Crivella/8bd71c69a40e3d61a4e2d930175a16f2 for a full test report.

easybuild/easyconfigs/g/gawk/gawk-5.3.0-intel-compilers-2023.2.1.eb
```diff
@@ -0,0 +1,23 @@
+easyblock = 'ConfigureMake'
+
+name = 'gawk'
+version = '5.3.0'
+
+homepage = 'https://www.gnu.org/software/gawk'
+description = """The awk utility interprets a special-purpose programming language that makes it possible to handle
+simple data-reformatting jobs with just a few lines of code."""
+
+toolchain = {'name': 'intel-compilers', 'version': '2023.2.1'}
```
akesandgren, line 10: I suggest adding this at GCCcore level instead since all newer gawk versions are at that level.
There is a 5.3.0 at GCC level but we can just ignore that and do the "right" thing here

akesandgren (changes requested):

easybuild/easyconfigs/g/gawk/gawk-5.3.0-intel-compilers-2023.2.1.eb
boegel: In theory you're right, but introducing `gawk-5.3.0-GCCcore-13.2.0.eb` next to the existing `gawk-5.3.0-GCC-13.2.0.eb` may actually cause trouble...

For builds with a `foss` (sub)toolchain, it may lead to situation where both `gawk/5.3.0-GCCcore-13.2.0` and `gawk/5.3.0-GCC-13.2.0` are included as dependencies, which could lead to failures (depending on how Lmod is configured).

So, I would stick to `intel-compilers` toolchain in this particular case. We've switched `gawk` to `GCCcore` from `2024a` generation onwards, trying to do that now still for older generations simply isn't worth the (potential) pain imho...

boegel:
CORE_CNT=16

easybuild/easyconfigs/g/gawk/gawk-5.3.0-intel-compilers-2023.2.1.eb
SimonPinches: @boegel, @akesandgren, thanks for all the input.  I'll leave it using the `intel-compilers` toolchain in this case then.

akesandgren (approved):
LGTM

akesandgren:
Test report by @akesandgren
**SUCCESS**
Build succeeded for 9 out of 9 (total: 1 hour 37 mins 51 secs) (5 easyconfigs in total)
b-cn1611.hpc2n.umu.se - Linux Ubuntu 22.04, x86_64, AMD EPYC 7313 16-Core Processor, 1 x NVIDIA NVIDIA A100 80GB PCIe, 580.173.02, Python 3.10.12
See https://gist.github.com/akesandgren/3be2a0574888e7eb4bb1928f056ed1f4 for a full test report.

akesandgren:

PR #23230, opened 2025-06-28
cleanup `icc`, `iccifort`, and `ifort` directories

branfosj:
With the archiving of `2020b` we no longer have old school Intel, so these can be cleaned up.


Micket:
as with the other cleanup, I have just a small concern about whether we want to add it into the archive repo.

@lexming ?

branfosj:
> as with the other cleanup, I have just a small concern about whether we want to add it into the archive repo.
> 
> @lexming ?

I've moved over the patches and I'm unconcerned about moving the README files.

We should merge #23231 as the patch being removed here is also there.

lexming:
I don't an issue in keeping these README files in the archive, they explain stuff about the archived easyconfigs.

Micket (approved):
lgtm, I'd welcome a PR to merge these into the archive

PR #23231, opened 2025-06-28
clean up old patches

branfosj:
using the code from #19827


Micket:
do we care about putting these into the easyconfig archive repo?

branfosj:
> do we care about putting these into the easyconfig archive repo?

Done: https://github.com/easybuilders/easybuild-easyconfigs-archive/pull/13

Micket (approved):
lgtm

PR #23246, opened 2025-07-01
{data}[GCCcore/13.3.0] lmodule v0.4.0 w/ Python 3.12.3

ysagon:
(created using `eb --new-pr`)


easybuild/easyconfigs/l/lmodule/lmodule-0.4.0-GCCcore-13.3.0-Python-3.12.3.eb
```diff
@@ -0,0 +1,31 @@
+easyblock = 'PythonPackage'
+
+name = 'lmodule'
+version = '0.4.0'
+versionsuffix = '-Python-%(pyver)s'
```
Micket: no reason to add this suffix as it's it's the default python version

easybuild/easyconfigs/l/lmodule/lmodule-0.4.0-GCCcore-13.3.0-Python-3.12.3.eb
```diff
@@ -0,0 +1,31 @@
...
+
+homepage = 'https://github.com/HPC-buildtest/lmodule'
+description = """lmodule is a Python 3 API for Lmod module system. The API can be used end-users and 
+system-administrators. End-users can utilize the API for interacting with module system to run their application. 
+System Administrators can utilize the API for testing their Software Stack and retrieve output from the Lmod spider 
+command."""
+
+toolchain = {'name': 'GCCcore', 'version': '13.3.0'}
+
+source_urls = [PYPI_SOURCE]
```
Micket: this is also the default already
```suggestion
```

easybuild/easyconfigs/l/lmodule/lmodule-0.4.0-GCCcore-13.3.0-Python-3.12.3.eb
```diff
@@ -0,0 +1,31 @@
...
+sources = [SOURCE_PY3_WHL]
+checksums = ['ba1e3317a25fcc0be9676a7a1b27c3fb6b20757d013de0d6fe7965cc23eb0d16']
+
+exts_defaultclass = 'PythonPackage'
+use_pip = True
+
```
Micket:
```suggestion
```

There are no extensions inside this easyconfig, it's not a (python)bundle, and even if it was, this would have been the default anyway

easybuild/easyconfigs/l/lmodule/lmodule-0.4.0-GCCcore-13.3.0-Python-3.12.3.eb
```diff
@@ -0,0 +1,31 @@
...
+options = {'modulename': 'lmod'}
+
+dependencies = [
+    ('Python', '3.12.3'),
+]
+
+download_dep_fail = True
+use_pip = True
+
```
Micket:
```suggestion
```

Micket (changes requested):
I do wonder if this makes more sense as a system level package? But this is fine to if people think it's useful after having other modules loaded already

Thyre:
You still have the old file (`easybuild/easyconfigs/l/lmodule/lmodule-0.4.0-GCCcore-13.3.0-Python-3.12.3.eb`) in your PR as well. Can you remove it?

PR #23153, opened 2025-06-19
{bio}[foss/2023a] humann v3.9

verdurin:
(created using `eb --new-pr`)

PR #23247, opened 2025-07-01
{bio}[GCC/12.3.0] humann v3.9

sassy-crick:
(created using `eb --new-pr`)


sassy-crick:
I would need a bit of guidance here. Compared with PR [#23153](https://github.com/easybuilders/easybuild-easyconfigs/pull/23153), which I did not see at the time, I am including `MinPath` from EasyBuild, enabled via the patch-file as else it is downloading it automatically, as seen from the log-file when building from the PR above:
```
  Installing minpath.
  Downloading http://omics.informatics.indiana.edu/mg/get.php?justdoit=yes&software=minpath1.2.tar.gz
```
compared with this PR:
```
  Found minpath install.
```
I have not included `SAMtools`, `MetaPhlAn` and `biom-format` as they are not needed to run the program, from what I understand are just companions. Removing them means we can go for `GCC` instead of the `foss` toolchain.
I am also running the test jobs which is not done in the above merged PR. 
@boegel @verdurin : Any suggestions here?

easybuild/easyconfigs/h/HUMAnN/humann-3.9.patch
```diff
@@ -0,0 +1,18 @@
+Use MinPath from EasyBuild, rather than downloaded one
```
verdurin, line 1: It's complaining here because you have the path `HUMAnN` so the test doesn't think it's related to this easyconfig.

verdurin (dismissed):

verdurin:
I think it may also be installing `DIAMOND` and `Boost`.
Don't have any objection in principle to adding this, and it is nice to add the test run.

@boegel is there precedent for changing the 'level' without changing the version, in the same toolchain generation?

easybuild/easyconfigs/h/HUMAnN/humann-3.9.patch
sassy-crick: I see. I originally followed the way it is printed on their documentation, I only later changed that to what we already have. I will simply delete that patch file, and upload it again. Hopefully that works. Sorry for that.

sassy-crick:
> I think it may also be installing `DIAMOND` and `Boost`. Don't have any objection in principle to adding this, and it is nice to add the test run.
> 
I am not sure about `Boost` but I see that in the log-file:
```
  Found glpk install at /home/sassy/apps/software/GLPK/5.0-GCCcore-12.3.0/bin
  Found diamond install at /home/sassy/apps/software/DIAMOND/2.1.8-GCC-12.3.0/bin
  Found bowtie2 install at /home/sassy/apps/software/Bowtie2/2.5.1-GCC-12.3.0/bin
```
As for `Boost`: It could well be we don't need it?
The `setup.py` contains:
```
    print("Installing boost.")
```
but in the final log file I can **not** find `Installing boost`. Thus, I am wondering if `Boost` is only needed if and when some other program, which we pull in, is build on-the-fly, maybe?

sassy-crick:
Test report by @sassy-crick
**SUCCESS**
hp - Linux Debian GNU/Linux 12, x86_64, 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz, Python 3.11.2
See https://gist.github.com/sassy-crick/f44eb3ecbd808597916d29385c8617ee for a full test report.

sassy-crick:

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 4 out of 4 (total: 3 mins 14 secs) (1 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), Python 3.11.14
See https://gist.github.com/Crivella/26563a42602a7841d28ef7fa831f0ec7 for a full test report.

easybuild/easyconfigs/h/humann/humann-3.9-GCC-12.3.0.eb
```diff
@@ -0,0 +1,58 @@
...
+        'checksums': [
+            {'humann-3.9.tar.gz': 'd018ddbbe45f5e3e11080620b4ae5f1737ba213d2c8cecdc8237859f853b2287'},
+            {'humann-3.9.patch': '7de8e5a82d238ea759c2a519048f26739e5c0aa3315cbc3ffc49d0020b9ce49d'},
+        ],
+    }),
+]
+
+sanity_check_commands = ['humann_test']
+
+moduleclass = 'bio'
```
Crivella, line 58:
```suggestion
        ],
        'testinstall': True,
        'runtest': 'humann_test',
    }),
]

moduleclass = 'bio'
```

We can run this as an actual test step

Crivella (dismissed):
Besides this i think this is fine. We just have to decide how to deal with the lowering of the dependencies

sassy-crick:
Test report by @sassy-crick
**SUCCESS**
hp - Linux Debian GNU/Linux 12, x86_64, 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz, Python 3.11.2
See https://gist.github.com/sassy-crick/afedd13c6316f815e0225f40868b3bed for a full test report.

easybuild/easyconfigs/h/humann/humann-3.9-GCC-12.3.0.eb
sassy-crick: Thanks for finding this, I have addressed the changes.

sassy-crick:

akesandgren:
> I think it may also be installing `DIAMOND` and `Boost`. Don't have any objection in principle to adding this, and it is nice to add the test run.
> 
> @boegel is there precedent for changing the 'level' without changing the version, in the same toolchain generation?

I do that every now and then when I find something that is at the wrong level, haven't seen any complaints so far :-)

akesandgren:
Test report by @akesandgren
**SUCCESS**
Build succeeded for 2 out of 2 (total: 2 mins 40 secs) (1 easyconfigs in total)
b-cn1611.hpc2n.umu.se - Linux Ubuntu 22.04, x86_64, AMD EPYC 7313 16-Core Processor, 1 x NVIDIA NVIDIA A100 80GB PCIe, 580.126.09, Python 3.10.12
See https://gist.github.com/akesandgren/7c4dd059d1b45e44b31f7bcca4a19980 for a full test report.

akesandgren (approved):
LGTM

akesandgren:

akesandgren:
@sassy-crick feel free to make PR's for the other toolchain versions to move this down to GCC there too.

PR #23281, opened 2025-07-02
{tools}[gfbf/2024a] Deprecated v1.2.18

fizwit:
(created using `eb --new-pr`)
switching back to foss toolchain because dependence of wrapt.

PR #23283, opened 2025-07-03
{bio}[foss/2024a] nf-core 3.3.1

fizwit:
(created using `eb --new-pr`)


fizwit:
requires  PR #23281  - {tools}[foss/2024a] Deprecated v1.2.18 
Deprecated is missing from 2024a, but nf-core is already in 2024a. Consider this a place holder

PR #23302, opened 2025-07-03
Add newer setuptools to deal with PEP 639

Micket:
(created using `eb --new-pr`)

Updated versions to deal with PEP639

PR #23305, opened 2025-07-03
{tools}[GCCcore/13.3.0] hatchling v1.27.0

Micket:
(created using `eb --new-pr`)

This is used to deal with the frequent license issues we hit as a result of PEP639, which requires 1.27.0 or newer to work, so i backported the one we had for 2025a 

* [x] #23302 (for newer `packaging`)


Micket:

easybuild/easyconfigs/h/hatchling/hatchling-1.27.0-GCCcore-13.3.0.eb
```diff
@@ -0,0 +1,56 @@
...
+
+homepage = 'https://hatch.pypa.io'
+description = """Extensible, standards compliant build backend used by Hatch,
+a modern, extensible Python project manager."""
+
+toolchain = {'name': 'GCCcore', 'version': '13.3.0'}
+
+builddependencies = [
+    ('binutils', '2.42'),
+]
```
branfosj, line 14:
```suggestion
    ('setuptools', '80.9.0'),
]
```

When installing the `hatchling` extension.
```
ModuleNotFoundError: No module named 'packaging.licenses'
```

easybuild/easyconfigs/h/hatchling/hatchling-1.27.0-GCCcore-13.3.0.eb
```diff
@@ -0,0 +1,57 @@
...
+
+homepage = 'https://hatch.pypa.io'
+description = """Extensible, standards compliant build backend used by Hatch,
+a modern, extensible Python project manager."""
+
+toolchain = {'name': 'GCCcore', 'version': '13.3.0'}
+
+builddependencies = [
+    ('binutils', '2.42'),
+    ('setuptools', '80.9.0'),
```
branfosj: This needs to be a runtime dependency.

```
hatchling 1.27.0 has requirement packaging>=24.2, but you have packaging 24.0.
```

Would it be better having #23302, #23306, and this combined in one easyconfig?

Micket: Well, we have kind of this pattern of hatchling, flit and other builddeps being separate packages already, so I kind of followed that trend. We are really only inconsistent with setuptools and flit-core that were bundled into Python itself (I guess to bootstrap the rest).

I'm also not sure what to call that bundle.

There is a shit ton of backends for some reason, like
1. flit-core
2. setuptools
4. hatchling
5. pdm-backend
6. SIP
7. uv_build
8. poetry-core
9. maturin
10. mesonpy
11. flit-scm

ugh

I don't mind merging flit-core into setuptools though, since it's sort-of what we do in the Python bundle itself. Otherwise I would suggest we keep them separate. I'm actually even hesitant to add setuptools as a runtime dep here, I'm thinking it might just be better to add packaging to the list of extensions; that way we are most consistent, and if some package needs a even newer setuptools + hatchling, they can pull that in themselves without module conflicts.

But, I think this quite rare, so, I just added the setuptools as a runtime dep now here. Either that or keep it as a builddep and add packaging as a extension here (out of principle...)

akesandgren:
Test report by @akesandgren
**SUCCESS**
Build succeeded for 2 out of 2 (1 easyconfigs in total)
b-cn1613.hpc2n.umu.se - Linux Ubuntu 22.04, x86_64, AMD EPYC 7313 16-Core Processor, Python 3.10.12
See https://gist.github.com/akesandgren/b1dc8f4ad85706686937d8e73f53910a for a full test report.

akesandgren:
Going in, thanks @Micket!

PR #23323, opened 2025-07-04
fix `PySCF` easyconfigs

branfosj:
(created using `eb --new-pr`)

This:
* removes `separate_build_dir = True` from `PySCF-1.7.6-gomkl-2021a.eb`
* deletes `PySCF-1.7.6-foss-2021a.eb` and `pyWannier90-2021-12-07-foss-2021a.eb` as we have `gomkl` versions of both
* fixes the `$EBROOTQCINT` to be `$EBROOTLIBCINT`


branfosj:
Toolchain confusion on my part. Closing this for #23325

PR #23429, opened 2025-07-16
LLVM fix for libffi

Crivella:
Companion PR to 

- https://github.com/easybuilders/easybuild-easyblocks/pull/3849

## Notes

- `libffi` was already an indirect dependency but this future proofs it and makes it clear it is also a dependency
- Removed the `skip_mod_files_sanity_check` option as it is now being set by the EasyBlock when needed
- Even if the linked issue is related to `offload` which has been introduced since `LLVM >= 19` the added configure options are also available for `18.1.8`

### DIFF
--- a/easybuild/easyconfigs/l/LLVM/LLVM-18.1.8-GCCcore-13.3.0.eb
+++ b/easybuild/easyconfigs/l/LLVM/LLVM-18.1.8-GCCcore-13.3.0.eb
@@ -43,6 +43,7 @@ builddependencies = [
 ]
 
 dependencies = [
+    ('libffi', '3.4.5'),
     ('libxml2', '2.12.7'),
     ('ncurses', '6.5'),
     ('zlib', '1.3.1'),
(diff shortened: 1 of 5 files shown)


Crivella:
EB_ARGS="--installpath /tmp/$USER/pr-23429 --include-easyblocks-from-pr 3849"
CORE_CNT=16

bartoldeman:
Test report for `LLVM-20.1.5-GCCcore-13.3.0.eb`
https://gist.github.com/bartoldeman/ad4001205949db39be05bf186e80f356

two notes:
* I tweaked the easyblock a bit to not include the GCC lib dir into LD_LIBRARY_PATH, as that breaks in our somewhat unusual setup (using GCC from Gentoo prefix).
* I raised the number of allowed test failures to 50 as a precaution and got:
Test suite completed with 29 failed tests (50 allowed)

ocaisa (approved):
LGTM

PR #23442, opened 2025-07-18
add patch to EasyBuild 5.1.1 easyconfig for bug that causes failures when copying readonly patches 

bedroge:
In PR https://github.com/easybuilders/easybuild-framework/pull/4939 I fixed an issue where patches of extensions would not be copied to the `easybuild` subdir, but it introduced a new bug that would basically add every patch to the list twice (only for easyconfigs with extensions that have patches). This causes every patch to be copied twice, where the second copy would just overwrite the first one. This in itself should not cause issues, except in the case where the patch file has read-only permissions. For anyone using `--read-only-installdir`, this may happen frequently though, as all the patches in the EasyBuild installation itself will have such permissions. The second copy will then result in a `Permission denied` error, as we're also observing in EESSI.

To solve the issue for anyone using that option, I'm adding the fix for the issue (https://github.com/easybuilders/easybuild-framework/pull/4960) as a patch to the EasyBuild 5.1.1 easyconfig, so they can just rebuild EB 5.1.1 with `--from-pr / --from-commit`.

### DIFF
--- a/easybuild/easyconfigs/e/EasyBuild/EasyBuild-5.1.1.eb
+++ b/easybuild/easyconfigs/e/EasyBuild/EasyBuild-5.1.1.eb
@@ -35,10 +35,13 @@ sources = [
         'extract_cmd': "tar xfvz %s && mv easybuild_easyconfigs-%(version)s easybuild-easyconfigs-%(version)s",
     },
 ]
+patches = ['EasyBuild-5.1.1_fix-failing-copy-of-readonly-patches.patch']
 checksums = [
     {'easybuild_framework-5.1.1.tar.gz': '4579274c758b5a01aa7996bc6e2652f6f7ff6e94320f93f22376f5b68c71d0bb'},
     {'easybuild_easyblocks-5.1.1.tar.gz': '8bc202cb203d296de7cfe1f9bb38e75c1cc7490c8bd43108469df0f7777ddd25'},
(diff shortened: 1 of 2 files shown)


bedroge:
With a dummy Python bundle containing one extension with a patch, I got the following error with EB 5.1.1 and using `--read-only-installdir`:
```
== Summary:
   * [FAILED]  Dummy-pythonbundle/2025.06-foss-2025a
ERROR: Installation of dummy-python.eb failed: "Failed to copy file /home/bob/copypatch/sj.patch to /home/bob/easybuildinstall/software/Dummy-pythonbundle/2025.06-foss-2025a/easybuild/sj.patch: [Errno 13] Permission denied: '/home/bob/easybuildinstall/software/Dummy-pythonbundle/2025.06-foss-2025a/easybuild/sj.patch'"
```

After rebuilding EB 5.1.1 with `--from-pr 23442`, I now get:
```
== Build succeeded for 1 out of 1
== Summary:
   * [SUCCESS] Dummy-pythonbundle/2025.06-foss-2025a
```

bedroge:
Test report by @bedroge
**SUCCESS**
bob-Latitude-5300 - Linux Ubuntu 24.04.2 LTS (Noble Numbat), x86_64, Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz, Python 3.12.3
See https://gist.github.com/bedroge/8cdcb6fd83cc9057970cecbe66140fbd for a full test report.

ocaisa:

PR #23497, opened 2025-07-25
{math}[gfbf/2023b] fenics-basix v0.9.0

jhale:
This an attempt at adding fenics-basix and fenics-basix-py builds.

I am currently having an issue with the Python package tests failing due to not being able to tell the Python builder block what the name of the installed Python module is (`basix`, not `fenics-basix` or `fenics-basix-py`). The current version has a hack which is to call the package `basix` but clearly this is not correct.

Advice on naming also welcome - here I used what we are using on Spack. We always use `fenics-basix` on pypi.

https://github.com/FEniCS/basix

### DIFF
--- a/easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
+++ b/easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
@@ -2,6 +2,7 @@ easyblock = 'PythonPackage'
 
 name = 'fenics-basix-py'
 version = '0.9.0'
+
 homepage = 'https://github.com/FEniCS/basix'
 description = "Basix is a finite element definition and tabulation runtime library - Python binding"
 


easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,34 @@
...
+
+easyblock = 'CMakeMake'
+
+# Version and source
+version = '0.9.0'
+source_urls = ['https://github.com/FEniCS/basix/archive']
+sources = ['v%(version)s.tar.gz']
+
+# Toolchain (adjust to match your target system)
+toolchain = {'name': 'GCC', 'version': "13.2.0"}
```
Thyre:
```suggestion
toolchain = {'name': 'gfbf', 'version': "2023b"}
```

The toolchain already provides FlexiBLAS 😄 
You then also need to change the filename to `fenics-basix-0.9.0-gfbf-2023b.eb`.

easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,34 @@
+name = 'fenics-basix'
+homepage = 'https://github.com/FEniCS/basix'
+description = "Basix is a finite element definition and tabulation runtime library"
+
+versionsuffix = ''
```
Thyre:
```suggestion
```

No need to provide the empty `versionsuffix`.

easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,34 @@
...
+
+easyblock = 'CMakeMake'
```
Thyre: I _think_ we always have this as the first line.

easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,34 @@
...
+
+# Build dependencies
+builddependencies = [
+    ('CMake', '3.27.6'),
+]
+
+# Runtime dependencies
+dependencies = [
+    ('FlexiBLAS', '3.3.1'),
+]
```
Thyre: You should be able to drop this if you switch to `gfbf`.

easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,36 @@
+# TODO: In Spack the package name is fenics-basix-py, but then the sanity checks
+# try to work with `fenics_basix_py`.
+name = 'basix'
```
Thyre:
```suggestion
name = 'fenics-basix-py'
```

With the Python package correctly detected, we can now use the name used for the folder 😄

easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,36 @@
...
+homepage = 'https://github.com/FEniCS/basix'
+description = "Basix is a finite element definition and tabulation runtime library"
+
+versionsuffix = ''
```
Thyre:
```suggestion
```

easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,36 @@
...
+
+easyblock = 'PythonPackage'
```
Thyre, line 1: Similarly to the other EasyConfig, we should move this to be the first line

easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,36 @@
...
+    ('fenics-basix', '0.9.0')
+]
+
+# Runtime dependencies
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+]
+
+use_pip = True
```
Thyre:
```suggestion
```

Since EasyBuild 5.0, this is the default.

easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,34 @@
+name = 'fenics-basix'
+homepage = 'https://github.com/FEniCS/basix'
+description = "Basix is a finite element definition and tabulation runtime library"
+
+versionsuffix = ''
+
+easyblock = 'CMakeMake'
+
+# Version and source
+version = '0.9.0'
```
Thyre: I think we typically have the version just below the name.

Thyre (commented):
A few comments 😄 
You also need to add a checksum to both EasyConfigs. You should be able to do this with `eb --inject-checksums --force <your-eb-files>`.

easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,36 @@
...
+install_src = "./python"
+
```
Thyre:
```suggestion
install_src = "./python"
modulename = "basix"

```

This option of `PythonPackage` tells EasyBuild for which name it should look in its sanity check.
This should hopefully resolve the one issue you saw when building the package.

easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-foss-2023b.eb
jhale: This isn't completely clear to me - should we pick 'the most minimal' set? Basix C++ only requires a BLAS/LAPACK, otherwise GCC set is sufficient.

Thyre: Hm, yeah.. `gfbf` would also include FFTW, which is not needed.
@branfosj, can you give a hint on how this is typically handled?

easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
jhale:
```
WARNING: Use of 1 unknown easyconfig parameters detected in fenics-basix-py-0.9.0-gfbf-2023b.eb: modulename
If these are just local variables please rename them to start with 'local_', or try using --fix-deprecated-easyconfigs to do this automatically.
For more information, see https://docs.easybuild.io/easyconfig-files-local-variables/ .
```

We're using EasyBuild 5.0.0.

easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-foss-2023b.eb
jhale: Our cluster has FlexiBLAS with GCC 13.2.0, so this seems correct.

easybuild/easyconfigs/f/fenics-basix-py/fenics-basix-py-0.9.0-gfbf-2023b.eb
Thyre: Yeah... I thought the option of `PythonBundle` simply translates to `PythonPackage`. 

This should work though:

```
options = {'modulename': "basix"}

install_src = "./python"
```

Thyre:

boegel:
@Thyre Since you've already reviewed this, what's the current status of this, should it be ready to go?

Thyre:
> @Thyre Since you've already reviewed this, what's the current status of this, should it be ready to go?

I'm still not sure regarding the toolchain for `fenix`. Is GCC with FlexiBLAS as a dependency okay, or should this be `gfbf` / `foss`? Aside from that, this looks good to go.

jhale:
There is still the open question of @Thyre on which set to pick.

We are currently working on the 'whole FEniCS stack' which we plan to make into a set of PRs - perhaps it would be best to wait until we have the stack ready, rather than merge each piece - it's possible we will have to tweak things as we go.

PR #23499, opened 2025-07-25
add patch to Python easyconfigs to fix ctypes when `$LD_LIBRARY_PATH` is not being set

dagonzalezfo:
Adding patch to provide support for library resolution on python

Edit by @casparvl: depends on:
* https://github.com/easybuilders/easybuild-easyblocks/pull/3860

### DIFF
--- a/easybuild/easyconfigs/p/Python/Python-3.10.4-GCCcore-11.3.0-bare.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.10.4-GCCcore-11.3.0-bare.eb
@@ -12,8 +12,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
@@ -11,8 +11,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.10.4-GCCcore-11.3.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.10.4-GCCcore-11.3.0.eb
@@ -11,8 +11,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.10.8-GCCcore-12.2.0-bare.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.10.8-GCCcore-12.2.0-bare.eb
@@ -12,8 +12,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.10.8-GCCcore-12.2.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.10.8-GCCcore-12.2.0.eb
@@ -11,8 +11,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.11.2-GCCcore-12.2.0-bare.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.11.2-GCCcore-12.2.0-bare.eb
@@ -12,8 +12,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.11.3-GCCcore-12.3.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.11.3-GCCcore-12.3.0.eb
@@ -11,8 +11,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.12.3-GCCcore-13.3.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.12.3-GCCcore-13.3.0.eb
@@ -12,8 +12,8 @@ source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 patches = ['Python-3.12.3_avoid-tkinter-build.patch']
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.13.1-GCCcore-14.2.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.13.1-GCCcore-14.2.0.eb
@@ -16,8 +16,8 @@ patches = [
     'Python-3.13.1_runshared-ld-preload.patch',
 ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.9.5-GCCcore-10.3.0-bare.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.9.5-GCCcore-10.3.0-bare.eb
@@ -12,8 +12,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.9.5-GCCcore-10.3.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.9.5-GCCcore-10.3.0.eb
@@ -11,8 +11,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.9.6-GCCcore-11.2.0-bare.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.9.6-GCCcore-11.2.0-bare.eb
@@ -12,8 +12,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [
--- a/easybuild/easyconfigs/p/Python/Python-3.9.6-GCCcore-11.2.0.eb
+++ b/easybuild/easyconfigs/p/Python/Python-3.9.6-GCCcore-11.2.0.eb
@@ -11,8 +11,8 @@ toolchainopts = {'pic': True}
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
 
-# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
-# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+# Like patches, but these will only be applied if EasyBuild is configured to filter $LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on $LD_LIBRARY_PATH to find libraries
 patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
 
 checksums = [


easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,122 @@
...
+         def _get_soname(f):
+@@ -283,7 +331,7 @@
+             abi_type = mach_map.get(machine, 'libc6')
+ 
+             # XXX assuming GLIBC's ldconfig (with option -p)
+-            regex = r'\s+(lib%s\.[^\s]+)\s+\(%s'
++            regex = r'\s+(lib%s\.[^\s]+)\s+\(%s\)\s+=>\s+(\S+)'
+             regex = os.fsencode(regex % (re.escape(name), abi_type))
+             try:
+                 with subprocess.Popen(['/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/sbin/ldconfig', '-p'],
```
casparvl: This is EESSI-specific. Fine for me in order to test the functionality of this patch, but just as a reminder to ourselves: once we port it to an EasyBlock, we should make sure that this loads the `ldconfig` from the `sysroot` (if a `sysroot` is configured for EaysBuild).

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,122 @@
...
+         def find_library(name):
+             # See issue #9998
+-            return _findSoname_ldconfig(name) or \
+-                   _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))
++            if os.path.exists(name):
++                return name
++            else:
++                return _findSoname_ldconfig(name) or \
++                   _findLib_gcc(name) or _findLib_ld(name) or \
++                   _findLib_gcc_fullname(name)
```
casparvl: I'm not 100% sure when we expect something to by found by `findLib_ld` and not by `_findLib_gcc*`, but: should we also implement a `_findLib_ld_fullname` for completeness? It seems to understand the full name syntax `l:<full_libname>` similar to `gcc`, e.g.
```
ld -t -L /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4/software/SDL2/2.28.5-GCCcore-13.2.0/lib -o /dev/null -l:libSDL2.so
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen4/software/SDL2/2.28.5-GCCcore-13.2.0/lib/libSDL2.so
```

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,122 @@
...
++
+     if sys.platform == "sunos5":
+         # use /usr/ccs/bin/dump on solaris
+         def _get_soname(f):
+@@ -283,7 +331,7 @@
+             abi_type = mach_map.get(machine, 'libc6')
+ 
+             # XXX assuming GLIBC's ldconfig (with option -p)
+-            regex = r'\s+(lib%s\.[^\s]+)\s+\(%s'
++            regex = r'\s+(lib%s\.[^\s]+)\s+\(%s\)\s+=>\s+(\S+)'
```
casparvl, line 110: We should probably put some descriptive comment here, e.g.
```
# Regular expression that captures the full path to the library in the second capture group
```
or something equivalent. Also, we should probably change the function name to `_findSopath_ldconfig(name)`, because it now returns a full path (and not just an `Soname`).

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
casparvl: Ah, silly comment: that's probably how this got generated: by the easyblock, with a regex-replacement. So, that _should_ be dealt with automatically. But, we need to ensure that whatever way we patch the code (once this patch works) can deal with the fact that this line can contain anything as a prefix for `ldconfig`.

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,122 @@
...
+             regex = os.fsencode(regex % (re.escape(name), abi_type))
+             try:
+                 with subprocess.Popen(['/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/sbin/ldconfig', '-p'],
+@@ -293,7 +341,7 @@
+                                       env={'LC_ALL': 'C', 'LANG': 'C'}) as p:
+                     res = re.search(regex, p.stdout.read())
+                     if res:
+-                        return os.fsdecode(res.group(1))
++                        return os.fsdecode(res.group(2))
```
casparvl, line 120: Again, we should make sure to insert some descriptive comment, something like
```
# Extract second capture group, which contains the full path to the library
```

casparvl:
@dagonzalezfo I'm putting some reminders, also for myself, that we should ensure to put in some descriptive strings of what we do where. For now, they help me to understand what's going on - but you don't have to implement them into this patch. The patch is just here so we can test the functionality of the change :)

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,122 @@
...
+             except OSError:
+                 pass
+ 
+@@ -301,7 +349,7 @@
+             # See issue #9998 for why this is needed
+             expr = r'[^\(\)\s]*lib%s\.[^\(\)\s]*' % re.escape(name)
+             cmd = ['ld', '-t']
+-            libpath = os.environ.get('LD_LIBRARY_PATH')
++            libpath = os.environ.get('LIBRARY_PATH')
```
casparvl, line 130: We get `LIBRARY_PATH` instead of `LD_LIBRARY_PATH` if `LD_LIBRARY_PATH` is filtered by the EB configuration. This is the closest we can get to RPATH-like behavior in a Python setting: where `RPATH` encodes the location where `gcc` finds the library _at compile time_, this still uses `gcc` to locate the libraries (i.e. same search paths), but at runtime. I.e. _if_ the `LIBRARY_PATH` at runtime is the same as at compile time, this will behave the same as `RPATH`.

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
dagonzalezfo: We had to do a check on it. It seems that gcc solution includes the ld case.

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,122 @@
...
+@@ -149,6 +149,54 @@
+             return os.fsdecode(file)
+ 
+ 
++    def _findLib_gcc_fullname(name):
++        # Run GCC's linker with the -t (aka --trace) option and examine the
++        # library name it prints out. The GCC command will fail because we
++        # haven't supplied a proper program with main(), but that does not
++        # matter.
++        expr = r'^.*%s.*$' % re.escape(name)
```
casparvl: Hmm, the error upon trying to load a non-existing library is very vague:
```
$ python -c "import ctypes; print(ctypes.CDLL('libfoo.so'))"
  File "/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/__init__.py", line 356, in __init__
    self._name = util.find_library(name)
                 ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/util.py", line 382, in find_library
    _findLib_gcc_fullname(name)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/util.py", line 196, in _findLib_gcc_fullname
    if not _is_elf(file):
           ^^^^^^^^^^^^^
  File "/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/util.py", line 99, in _is_elf
    with open(filename, 'br') as thefile:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/bin/ld: cannot find -l:libfoo.so: No such file or directory'
```
The reason is that the `gcc` call in `_findLib_gcc_fullname` now does:
```
$ gcc -Wl,-t -o /dev/null -l:libfoo.so
...
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/bin/ld: cannot find -l:libfoo.so: No such file or directory
...
collect2: error: ld returned 1 exit status
```

The problem is that this regex

```
expr = r'^.*%s.*$' % re.escape(name)
```

is now so general, that it also matches the error pattern.

A more minimal change compared to the original would be:

```
expr = os.fsencode(r'^[^\(\)\s]*%s[^\(\)\s]*' % re.escape(name))
```
Essentially, I added a start-of-line character `^`, and removed the literal `lib` (to avoid `liblibfoo.so`) and the literal `.` to avoid searching for `libfoo.so.` with a trailiing dot).

This is much more specific, as it states that anything before or after the library name (on the same line) may _not_ contain spaces, or `()` brackets. The _expected_ output is a full path `/my/path/to/libfoo.so`. The error message that `ld` caant find `libfoo.so` contains spaces, and therefore doesn't match. The result is much cleaner:

```
[casparl@tcn292 ~]$ python -c "import ctypes; print(ctypes.CDLL('libSDL2-2.0.so.0'))"
<CDLL 'b'/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2-2.0.so.0'', handle dfacb0 at 0x14b2f45539d0>
[casparl@tcn292 ~]$ python -c "import ctypes; print(ctypes.CDLL('libfoo.so'))"
<CDLL 'None', handle 147ad76472c0 at 0x147ad6953b10>
```

casparvl:
FYI: I want to test this a bit more to make sure it also still works for `librosa`, with which we had our original issue, but that requires me to locally build `librosa` for 2023b, so... takes a bit of time.

In the meantime, I created this test script:

```
$ cat test.sh
#!/bin/bash

module purge
module load EESSI
module load EESSI-extend
module load Python/3.11.5-GCCcore-13.2.0
module load SDL2/2.28.5-GCCcore-13.2.0

# Testing find_library by library name

for lib in foo c SDL2; do
  echo "Running find_library for $lib"
  python -c "import ctypes; from ctypes.util import find_library; print(find_library('$lib'))"
  echo ""
done

# Testing library loads by full name
for lib in libfoo libSDL2.so libSDL2-2.0.so.0 libSDL2-2.0.so.0.2800.5 libc.so libc.so.6; do
  echo "Instantiating CDLL object for $lib"
  python -c "import ctypes; print(ctypes.CDLL('$lib'))"
  echo "Running LoadLibrary for $lib"
  python -c "import ctypes; print(ctypes.cdll.LoadLibrary('$lib'))"
  echo "Running find_library for $lib"
  python -c "import ctypes; from ctypes.util import find_library; print(find_library('$lib'))"
  echo ""
done

# Testing library loads by full path
for lib in /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib64/libSDL2.so /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6; do
  echo "Instantiating CDLL object for $lib"
  python -c "import ctypes; print(ctypes.CDLL('$lib'))"
  echo "Running LoadLibrary for $lib"
  python -c "import ctypes; print(ctypes.cdll.LoadLibrary('$lib'))"
  echo ""
done
```

I may add some more libraries later, but I think this already covers a fair range of cases: from libs in the compat layer, to non-versioned libnames, to versioned libnames, to non-existing libnames. All seem to produce the expected result (with the regex change I suggested in my review).

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
dagonzalezfo: Should we add some error if the library is not found instead of returning a `None`?

casparvl: Hm, I'd prefer to mimic the original behavior. An unpatched python does this:

```
$ python -c "import ctypes; from ctypes.util import find_library; print(find_library('libfoo.so')); print(ctypes.CDLL('libfoo.so'))"
None
  File "/usr/lib64/python3.9/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libfoo.so: cannot open shared object file: No such file or directory
```
I.e. it tries the dlopen, and then just fails that `dlopen` call, reporting the name of the library that wasn't found.

The reason our behavior changed is that `find_library` returns `None` if a lib is not found, and in that case, `ctypes.CDLL` returns:

```
python -c "import ctypes; print(ctypes.CDLL(None))"
<CDLL 'None', handle 7f3dcf161210 at 0x7f3dceba4c10>
```

My suggestion would be to change:

```
+            if name:
+                self._name = util.find_library(name)
```

into 

```
+            if name:
+                fullpath = util.find_library(name)
+                if fullpath is not None:
+                    self._name = fullpath
```
so that we _only_ overwrite `self._name` if we can resolve the full path. Otherwise, we keep the `self._name` as it is, thus resulting in the original error of the form:
```
OSError: libfoo.so: cannot open shared object file: No such file or directory
```

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
casparvl: > It seems that gcc solution includes the ld case.

You mean: `gcc` in the end calls `ld` anyway?

I think you're right btw in that it's not needed: _normally_ `findLib_gcc` would trigger a search of `LIBRARY_PATH`, where `findLib_ld` would trigger a search of `LD_LIBRARY_PATH`. Since we overwrite `findLib_ld`'s behaviour to search `LIBRARY_PATH` instead, I don't think there are any scenarios in which `findLib_gcc_fullname` would _not_ resolve the library, while a `findLib_ld_fullname` (that crawls the same `LIBRARY_PATH`) _would_.

Summarizing: I think we can keep it like this.

casparvl (changes requested):
With the addition of https://github.com/easybuilders/easybuild-easyblocks/pull/3860 I think there are only a few things left:

- Accept my PR to your feature branch https://github.com/dagonzalezfo/easybuild-easyconfigs/pull/1 (if you agree, of course)

- The change of the regex on line 35 to
expr = os.fsencode(r'^[^\(\)\s]*%s[^\(\)\s]*' % re.escape(name))

- The change of overwriting self._name only if util.find_library() doesn't return None

And then, I will give it a final test drive (after rebuilding python with `--include-easyblocks-from-pr 3860` of course)

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
casparvl: Ok, this actually needs to be changed to it's original value, i.e. before the EasyBlock replaced it with the whole sysroot prefixed. The patch will be applied _before_ the easyblock makes that change, so in order to match, we need the original line here.

easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
```diff
@@ -10,7 +10,17 @@ toolchainopts = {'pic': True}
 
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
-checksums = ['a12a0a013a30b846c786c010f2c19dd36b7298d888f7c4bd1581d90ce18b5e58']
+checksums = [
+    {'Python-3.11.5.tgz': 'a12a0a013a30b846c786c010f2c19dd36b7298d888f7c4bd1581d90ce18b5e58'},
+    {'Python-3.11.5-custom-ctypes.patch': 'f145f6daf11f430da45c1da77f1795e4793ee170178a3e97a8286c3c5468f997'},
```
ocaisa:
```suggestion
```

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
ocaisa: So there is a scenario where this is not what we want, and that is `libcuda.so` (or rather the CUDA stubs in general which are listed in `LIBRARY_PATH` but not `LD_LIBRARY_PATH`, see https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/c/cuda.py#L113)

ocaisa: Perhaps we just add a conditional below, something like
```
             if libpath:
                 for d in libpath.split(':'):
                     cmd.extend(['-L', d]) if 'stubs' not in d.split('/')
```

dagonzalezfo: Thus, instead of overriding `LD_LIBRARY_PATH`, should we have something like `LD_LIBRARY_PATH:LIBRARY_PATH` as `libpath` ?

ocaisa: No, that doesn't help, you would still have the stubs in the list of paths and so they would be picked up over the ones coming from the OS

ocaisa: The real `libcuda.so` is picked up from the default paths of the linker (not from `LD_LIBRARY_PATH`)

easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
casparvl: my bad, that was wrong in my PR to @dagonzalezfo 's feature branch. Sorry...

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
dagonzalezfo: Hi @ocaisa , I just push a version that include your proposed code snip, do yo mind to confirm that it works for the cuda case?

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
dagonzalezfo: I added the coment on the patch for the regex change. About renaming the function, I propose to keep it as it is. If we rename this, we should also rename the gcc and ld ones.

dagonzalezfo:
Hi @casparvl, 

I included all the requested changes and test it locally and it seems to work correctly.  Do you mind to confirm?

Best,

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
ocaisa: It's always a good idea to add a comment to explain why that condition is there, especially when it is such a specific reason.

dagonzalezfo: Comment added. :)

casparvl: Edit: <snip>, removed, posted as review so that this comment is more easily findable in the future. I'm sure we may want to look back some day and see why we did what we did...

casparvl (commented):
Hmmm, two things:

I think the regex is a bit too generic: we're now assuming that any path containing stubs in the name can be filtered out. I think we should check that it also contains CUDA with a case insensitive match. You can check with a regular expression pattern = re.compile(r'cuda.*stubs', re.IGNORECASE) for example
If I do find_library('cuda'), it will actually find the library through findLib_gcc, and not findLib_ld. Since this (by design) will check which library would have been compiled against, it will return the stubs library. That is, essentially, by design of the find_library call.
The second can be fixed by two small changes to the code. The last lines of the _findLib_gcc(name) should be replaced with:

            file = os.fsdecode(file)
            if re.search(r'cuda.*stubs', file, re.IGNORECASE):
                continue
            return file
Similarly, the last lines of _findLib_gcc_fullname should be

            if re.search(r'cuda.*stubs', file, re.IGNORECASE):
                continue
            return file
So that if the stubs lib is picked up, it is just ignored and the search continues.

In the end, there is one issue: it does not pick up on the driver's libcuda.so.1 either. Normally, this would have been caught by the _findSonmame_ldconfig:

$ /sbin/ldconfig -p | grep libcuda.so
        libcuda.so.1 (libc6,x86-64) => /lib64/libcuda.so.1
        libcuda.so (libc6,x86-64) => /lib64/libcuda.so
But, since we are running with a sysroot, we are calling <sysyroot>/sbin/ldconfig instead. And that one doesn't know about /lib64/libcuda.so.

Note that for calls to ctypes.CDLL('libcuda.so'), this is actually not an issue: they will see that find_library(...) returns None, and thus call dlopen('libcuda.so') which (correctly!) opens the driver from the default search path.

So... we are almost at identical results as the regular ctypes with LD_LIBRARY_PATH, with find_library('cuda') being the one exception.

I Can come up with two solutions:

Add yet another function to this return statement:
return _findSoname_ldconfig(name) or \
                   _findLib_gcc(name) or _findLib_ld(name) or \
                   _findLib_gcc_fullname(name) or \
                   _findLib_host_ldconfig(name)
where _findLib_host_ldconfig(name) then uses the host's ldconfig from /sbin/ldconfig to locate libraries. The tricky thing is that this provides a fallback for any library that is not found in the compatiblity layer. That may lead to (unexpected) differences in behaviour of EESSI accross systems, as codes that try to load libfoo may work on some (because the host provides one) and not on others (because the compat layer and software layer don't).

Handle the cuda case explicitely in _findSoname_ldconfig. I.e. something like
if name == 'cuda':
    ldconfig = '/sbin/ldconfig'
else:
    ld_config = '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/sbin/ldconfig'
...
with subprocess.Popen([ldconfig, '-p'],
...
Note that this is tricky: it is only needed if a sysroot is set, which is something we don't know at the patch/easyconfig level, but only at the easyblock level. I.e. this change should not be part of the patch, but part of the easyblock logic that currently replaces the ldconfig path

What do you think @dagonzalezfo ?

dagonzalezfo:
Hi @casparvl , 

Thanks for having a look on it. 
About enhancing stub regex, I will add the changes later today. 

About libcuda.so, it is not exposed using a sort of `host-injections` on a EESSI related path?
Moreover, is this path used by ld to find the libraries or how are they actually exposed?
(by a fast check, it seems that is exposed using LD_PRELOAD and this are not recovered by the ldconfig command)
If this is the case, there is no need to add the `/sbin/ldconfig`  option. 

Do you mind to clarify?

dagonzalezfo:
I just realize that maybe defining a new function for `LD_PRELOAD` override could be helpful. Or adding it into the _findLib_ld  and reorder return:

Option 1: 
```
def _findLib_ld_preload ():
.
.
.
return _findLib_ld_preload () or\
            _findSoname_ldconfig(name) or \
            _findLib_gcc(name) or _findLib_ld(name) or \
            _findLib_gcc_fullname(name) 
```

Option 2: 

```
return _findLib_ld(name) or \
             _findSoname_ldconfig(name) or \
             _findLib_gcc(name) or \
             _findLib_gcc_fullname(name) 
```

Any opinion?

casparvl:
> About libcuda.so, it is not exposed using a sort of host-injections on a EESSI related path?
Moreover, is this path used by ld to find the libraries or how are they actually exposed?
(by a fast check, it seems that is exposed using LD_PRELOAD and this are not recovered by the ldconfig command)
If this is the case, there is no need to add the /sbin/ldconfig option.

Very good point actually, I did not think about that. What I _don't_ know is if that's on anyone's path. We may need @ocaisa to clarify this, but I guess the whole reason for symlinking the drivers was to expose them to the linker from the prefix? I have to admit, I tested this on a CPU node which _does have_ the driver installed (a bit weird, but ok). I figured not finding the `libcuda.so.1` was an issue with this PR, but maybe it's somehow related to the driver symlinking... I'll at the very least try to re-test on a GPU node. And see if I can see the `host_injections` in any search path of the linker.

casparvl:
> I just realize that maybe defining a new function for `LD_PRELOAD` override could be helpful. Or adding it into the _findLib_ld and reorder return:
> 
> Option 1:
> 
> ```
> def _findLib_ld_preload ():
> .
> .
> .
> return _findLib_ld_preload () or\
>             _findSoname_ldconfig(name) or \
>             _findLib_gcc(name) or _findLib_ld(name) or \
>             _findLib_gcc_fullname(name) 
> ```
> 
> Option 2:
> 
> ```
> return _findLib_ld(name) or \
>              _findSoname_ldconfig(name) or \
>              _findLib_gcc(name) or \
>              _findLib_gcc_fullname(name) 
> ```
> 
> Any opinion?

Hmm, I'm not sure I understand what you mean here. Would you like to make similar behavior to `LD_PRELOAD`, i.e. if `LD_PRELOAD` is set, the runtime linker doesn't even _try_ to load the library, since the library is already in memory?

I think that surpasses the 'normal' behavior. I.e. even without RPATH/sysroot, `LD_PRELOAD` isn't taken into account by `ctypes`. I wouldn't implement it here either. Although maybe you are thinking about the fact that `LD_PRELOAD` is sort of the only route that users can go if sites don't install the drivers in the `host_injections` dir? Even then, it's a very large deviation. I'm also wondering if it's needed: if you `LD_PRELOAD` it, doesn't `dlopen` just return succesfully, concluding that the lib was already in memory (without actually checking any search paths)? I'm not sure, I haven't tested. But I'd expect it might...

casparvl:
Trying to reconstruct how regular executables find `libcuda.so` from the `host_injections`, since I wasn't sure:

```
module load EESSI/2023.06
module load OSU-...
LD_DEBUG=libs /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/ld-linux-x86-64.so.2 --list $(which osu_bw) 2>&1 | less
```

Says

```
3391478:      search cache=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/etc/ld.so.cache
   3391478:      search path=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/glibc-hwcaps/x86-64-v4:/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/glibc-hwcaps/x86-64-v3:/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/glibc-h
wcaps/x86-64-v2:/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64:/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/glibc-hwcaps/x86-64-v4:/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/glibc-hwcaps/x86-64-v3:/cvmfs/softw
are.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/glibc-hwcaps/x86-64-v2:/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64:/cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib/glibc-hwcaps/x86-64-v4:/cvmfs/software.eessi.io/host_injec
tions/2023.06/compat/linux/x86_64/lib/glibc-hwcaps/x86-64-v3:/cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib/glibc-hwcaps/x86-64-v2:/cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib           (system search path)
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/glibc-hwcaps/x86-64-v4/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/glibc-hwcaps/x86-64-v3/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/glibc-hwcaps/x86-64-v2/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/glibc-hwcaps/x86-64-v4/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/glibc-hwcaps/x86-64-v3/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/glibc-hwcaps/x86-64-v2/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib/glibc-hwcaps/x86-64-v4/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib/glibc-hwcaps/x86-64-v3/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib/glibc-hwcaps/x86-64-v2/libcuda.so.1
   3391478:       trying file=/cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib/libcuda.so.1
```
Note that this is the _very_ last search option, i.e. after `RPATH`, `LD_LIBRARY_PATH`, `RUNPATH`, `ld.so.cache`, it searches the 'system search path'. Now, my AI friend tells me that's set in `/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/etc/ld.so.conf`, but while the first ones on the list are, the `host_injections` dirs are _not_. So how do they make it in there?

casparvl:
Ok, from chat, I understood that the host-injections search paths are baked into the GLIBC.

It seems like `ld` does _not_ look through these search paths by default, which is why:
```
$ ld -t -lcuda
ld: cannot find -lcuda: No such file or directory
```
doesn't work, even though e.g.
```
$ ld -t -lz
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/libz.so
ld: warning: cannot find entry symbol _start; not setting start address
```

Note that `ld-linux-x86-64.so.2` does search this search path:

```
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/ld-linux-x86-64.so.2 --help
...
Shared library search path:
  (libraries located via /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/etc/ld.so.cache)
  /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64 (system search path)
  /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64 (system search path)
  /cvmfs/software.eessi.io/host_injections/2023.06/compat/linux/x86_64/lib (system search path)
```

But that is called on executables, you can't "just" ask it to find a library for you (at least, not that I have found).

I'm out of ideas. I think we should just accept that `find_library` will not find the driver lib. But since `ctypes.CDLL('libcuda.so')` _does_ work, I think that's ok: I don't think calling `find_library` to find the CUDA driver is very common useage anyway. I say: let's see how far this patch gets us. If it proves to be an issue, we can always reconsider it later.

casparvl:
Just for later reference, in case we ever want to revisit this decision:
- The compile time linker `ld` is invoked by `ctypes`. This does _not_ search the `host_injections` dirs, and thus doesn't find `libcuda.so.1` there.
- The runtime linker `ld.so` _does_ search `host_injections`, but it's much harder to query this for the library it will load. The only way would be to actually link a binary to the lib you want to search (e.g. `libcuda.so`), and then invoke `ld.so --list <binary>` and grep for the result
- The issue of not being able to find libraries with `find_library` will apply not only to `libcuda.so`, but to all libs that might be shadowed under `host_injections`. E.g. also when trying to inject an MPI library from the host - this will _not_ be picked up correctly by any Python code.

easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
```diff
@@ -10,8 +10,16 @@ toolchainopts = {'pic': True}
...
-checksums = ['a12a0a013a30b846c786c010f2c19dd36b7298d888f7c4bd1581d90ce18b5e58']
+patches_filter_ld_library_path = ['Python-3.11.5-custom-ctypes.patch']
+checksums_filter_ld_library_path = [
+    {'Python-3.11.5-custom-ctypes.patch': '8ae3efc47dd202b600de8249df31b8be252814933b8339ba42fe5c2e0d9adb14'}
+]
+checksums = [
+    {'Python-3.11.5.tgz': 'a12a0a013a30b846c786c010f2c19dd36b7298d888f7c4bd1581d90ce18b5e58'},
+]
 
+# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
```
casparvl: This comment should be moved right before the `patches_filter_ld_library_path` keyword is defined

easybuild/easyconfigs/p/Python/Python-3.11.5-GCCcore-13.2.0.eb
```diff
@@ -10,8 +10,16 @@ toolchainopts = {'pic': True}
 
 source_urls = ['https://www.python.org/ftp/%(namelower)s/%(version)s/']
 sources = [SOURCE_TGZ]
-checksums = ['a12a0a013a30b846c786c010f2c19dd36b7298d888f7c4bd1581d90ce18b5e58']
+patches_filter_ld_library_path = ['Python-3.11.5-custom-ctypes.patch']
+checksums_filter_ld_library_path = [
+    {'Python-3.11.5-custom-ctypes.patch': '8ae3efc47dd202b600de8249df31b8be252814933b8339ba42fe5c2e0d9adb14'}
+]
```
boegel: I would simplify this, there's no need for *two* custom easyconfig parameters, and I would also make the naming more direct.

I think it's very unlikely that we'll have additional patches that should only be applied when `LD_LIBRARY_PATH` is included in the `filter-env-vars` EasyBuild configuration option.

Something like:
```python
# this patch will only be applied when EasyBuild is configured to not update $LD_LIBRARY_PATH
patch_custom_ctypes = 'Python-3.11.5-custom-ctypes.patch'
```

There's no need for `checksums_filter_ld_library_path` at all, having extra entries at the end of `checksums` should be just fine. If the patch specified via `patch_custom_ctypes` is skipped, there's no harm in having an extra entry in `checksums`?

casparvl: Ok, I'll change that.

casparvl: Note that we can't check as easily for this _specific_ checksum to be present if we go that route. I'll just skip that specific check / error reporting, and rely on the regular reporting that a checksum is missing.

casparvl: Wait, I can still do a check: the length of `sources` and all patches should be equal to checksums. If not, AND if LD_LIBRARY_PATH is filtered, I'll print an error with a hint that the user _may_ have forgotten the checksum for the custom ctypes patch.

casparvl:
Test report by @casparvl
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
tcn358.local.snellius.surf.nl - Linux RHEL 9.4 (Plow), x86_64, AMD EPYC 7H12 64-Core Processor (zen2), Python 3.11.4
See https://gist.github.com/casparvl/01035178573b9b8be6125f80e8f38950 for a full test report.

casparvl:
EB_ARGS="--include-easyblocks-from-pr 3860"

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,149 @@
...
+diff -ruN Python-3.11.5/Lib/ctypes/util.py Python-3.11.5.patched/Lib/ctypes/util.py
+--- Python-3.11.5/Lib/ctypes/util.py	2023-08-24 14:09:18.000000000 +0200
++++ Python-3.11.5.patched/Lib/ctypes/util.py	2025-08-07 16:37:14.337414617 +0200
+@@ -146,8 +146,60 @@
+             # shared objects. See bpo-41976 for more details
+             if not _is_elf(file):
+                 continue
+-            return os.fsdecode(file)
++            file = os.fsdecode(file)
++            if re.search(r'cuda.*stubs', file, re.IGNORECASE):
```
ocaisa: This really deserves a comment, you should explain _why_ this exception is necessary

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,149 @@
...
+             except OSError:
+                 pass
+ 
+@@ -301,10 +355,13 @@
+             # See issue #9998 for why this is needed
+             expr = r'[^\(\)\s]*lib%s\.[^\(\)\s]*' % re.escape(name)
+             cmd = ['ld', '-t']
+-            libpath = os.environ.get('LD_LIBRARY_PATH')
++            # use LIBRARY_PATH instead of LD_LIBRARY_PATH to use EB provided shared libraries
++            libpath = os.environ.get('LIBRARY_PATH')
```
ocaisa, line 130: Rather the have the two exceptions for the cuda stubs, why not clear the stubs directory from `libpath` at this point instead.

ocaisa: This might be a bad comment, I see you kind of do that below, but I don't really understand why you need the exception at line 39

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
dagonzalezfo: This is needed to avoid using stubs provided cuda libraries as we commented some time ago: 
https://github.com/easybuilders/easybuild-easyconfigs/pull/23499#discussion_r2242262893. 

I also add a comment on the patch as you suggest.

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
casparvl: Because there are multiple functions used to locate libraries, and neither of those should pick up the stubs. I.e. line 127 makes sure that `_findLib_ld` doesn't return anything from the CUDA stubs, while line 39 makes sure that _findLib_gcc` doesn't return anything from the CUDA stubs.

casparvl:
To re-test the latest changes...

EB_ARGS="--include-easyblocks-from-pr 3860"

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,150 @@
...
+             if not _is_elf(file):
+                 continue
+-            return os.fsdecode(file)
++            file = os.fsdecode(file)
++            # Prevent using CUDA libraries provided on stubs
++            if re.search(r'cuda.*stubs', file, re.IGNORECASE):
++                continue
++            return file
+ 
++    def _findLib_gcc_fullname(name):
```
boegel: I would avoid copy-pasting `_findLib_gcc` to only tweak a small things in it (the regex being used)

Instead, I would add an optional parameters to `_findLib_gcc`, like this:

```python
def _findLib_gcc(name, name_regex=None):
    ...
    expr = os.fsencode(...)
    if name_regex is not None:
      expr = name_regex % re.escape(name)
```

or, just make it optional in `_findLib_gcc` to add `lib` to the name in the regex being used, because that's basically what we're doing with `_findLib_gcc_fullname`?

```python
def _findLib_gcc(name, name_is_full_path=False):
    if name_is_full_path:
        # explain here why match needs to start at beginning
        regex_start = '^'
    else:
        # specified name is name of a library (without the 'lib' part)
        name = 'lib' + name
        regex_start = ''
    expr = os.fsencode(regex_start + r'[^\(\)\s]*%s\.[^\(\)\s]*' % re.escape(name))
```

If the regex used in `_findLib_gcc` is then improved, we'll pick up on that too automatically

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,150 @@
...
++            with proc:
++                trace = proc.stdout.read().decode()
++        finally:
++            try:
++                temp.close()
++            except FileNotFoundError:
++                # Raised if the file was already removed, which is the normal
++                # behaviour of GCC if linking fails
++                pass
++        res = re.findall(expr, trace, re.MULTILINE)
```
boegel: If the `re.MULTILINE` is really needed, this needs to be changed in `_findLib_gcc`

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,150 @@
...
+             expr = r'[^\(\)\s]*lib%s\.[^\(\)\s]*' % re.escape(name)
+             cmd = ['ld', '-t']
+-            libpath = os.environ.get('LD_LIBRARY_PATH')
++            # use LIBRARY_PATH instead of LD_LIBRARY_PATH to use EB provided shared libraries
++            libpath = os.environ.get('LIBRARY_PATH')
+             if libpath:
+                 for d in libpath.split(':'):
+-                    cmd.extend(['-L', d])
++                    # Needed to avoid searching on CUDA stubs
++                    if 'stubs' not in d.split('/'):
```
boegel: should not only check for "not stubs", but also check for "cuda"?

```python
parts = d.split('/')
if any('cuda' in p.lower() for p in parts) and 'stubs' not in parts:

File: easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch [hunk 15]
```diff
@@ -0,0 +1,150 @@
+Override ctypes behavior to enable support for EB provided libraries.
```
boegel: Maybe the description can be improved:
- pick up on `$LIBRARY_PATH` rather than `$LD_LIBRARY_PATH`
- also consider `foo` as full name of library in `_findLib_gcc`, not just `libfoo`
- ignore CUDA stub libraries (based on their path matching `cuda.*stubs`)

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,150 @@
...
+-                   _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))
++            # Redefine find_library function, it will return the provided name if
++            # path exist, else it will use the set of functions to find the full library path. 
++            # it will return the one that has a match. 
++            if os.path.exists(name):
++                return name
++            else:
++                return _findSoname_ldconfig(name) or \
++                   _findLib_gcc(name) or _findLib_ld(name) or \
++                   _findLib_gcc_fullname(name)
```
boegel:
```suggestion
                   _findLib_gcc(name, lib_prefix=False)

File: easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch (hunk 13 above)
Comment (casparvl) [reply easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch]: I figured it out: it is needed IF, and only if, the regex starts with `'^'`. This was needed for the full library path to avoid matching e.g 
```
gcc -Wl,-t -o bla -l:libfoo.so
...
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/bin/ld: cannot find -l:libfoo.so: No such file or directory
```
You don't want this to lead to a regex match, since we intend to direct if the library _is found_ (and this error tells us it _isn't_, so we don't want to match). Thus, with the starting `'^'`, we avoided that the `libfoo.so` in `-l:libfoo.so` in the above error was matched.

It _should_ not hurt to also use a starting `'^'` in the non-fullname variant of this function, but I'd prefer keeping that behavior as close as possible to upstream, so I'll keep it without `'^'` and without `'re.MULTILINE'` for the non-fullname variant.

File: easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch (hunk 14 above)
Comment (casparvl) [reply easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch]: Actually, you want:

```
                    if not (any('cuda' in p.lower() for p in parts) and 'stubs' in parts):
                        cmd.extend(['-L', d])
```
but yes :)

**Comment by casparvl** [comment]
Retested https://github.com/easybuilders/easybuild-easyblocks/pull/3860/commits/c622e77da9af44642cf49193525b71a7d6f44d51 and https://github.com/easybuilders/easybuild-easyconfigs/pull/23499/commits/f7e8762970e9d7af1ceb180a2b4b73b8050557ce with the same test script as before, results are as expected:

```
$ ./test.sh
Module for EESSI/2023.06 loaded successfully
-- Using /tmp/$USER as a temporary working directory for installations, you can override this by setting the environment variable WORKING_DIR
and reloading the module (e.g., /dev/shm is a common option)
Configuring for use of EESSI_USER_INSTALL under /home/casparl/eessi
-- To create installations for EESSI, you _must_ have write permissions to /home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2
-- You may wish to configure a sources directory for EasyBuild (for example, via setting the environment variable EASYBUILD_SOURCEPATH) to
allow you to reuse existing sources for packages.
Running find_library for foo
None

Running find_library for c
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6

Running find_library for SDL2
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2.so

Running find_library for cuda
None

Instantiating CDLL object for libfoo
  File "/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/__init__.py", line 385, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: libfoo: cannot open shared object file: No such file or directory
Running LoadLibrary for libfoo
  File "/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/__init__.py", line 463, in LoadLibrary
    return self._dlltype(name)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen2/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/ctypes/__init__.py", line 385, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: libfoo: cannot open shared object file: No such file or directory
Running find_library for libfoo
None

Instantiating CDLL object for libSDL2.so
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2.so', handle 117dff0 at 0x1530f9653a10>
Running LoadLibrary for libSDL2.so
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2.so', handle a1f0e0 at 0x146374f53350>
Running find_library for libSDL2.so
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2.so

Instantiating CDLL object for libSDL2-2.0.so.0
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2-2.0.so.0', handle 1dc1150 at 0x153ac372fa50>
Running LoadLibrary for libSDL2-2.0.so.0
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2-2.0.so.0', handle e902a0 at 0x14edbc957390>
Running find_library for libSDL2-2.0.so.0
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2-2.0.so.0

Instantiating CDLL object for libSDL2-2.0.so.0.2800.5
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2-2.0.so.0.2800.5', handle 210f120 at 0x150502e8fa90>
Running LoadLibrary for libSDL2-2.0.so.0.2800.5
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2-2.0.so.0.2800.5', handle b25200 at 0x14d31d04b3d0>
Running find_library for libSDL2-2.0.so.0.2800.5
/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib/../lib64/libSDL2-2.0.so.0.2800.5

Instantiating CDLL object for libc.so
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6', handle 14dbcef5cb70 at 0x14dbce28fa90>
Running LoadLibrary for libc.so
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6', handle 151c250fcb70 at 0x151c242cebd0>
Running find_library for libc.so
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6

Instantiating CDLL object for libc.so.6
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib/../lib64/libc.so.6', handle 14fafb794b70 at 0x14fafaa8fa90>
Running LoadLibrary for libc.so.6
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib/../lib64/libc.so.6', handle 1550d4ef7b70 at 0x1550d4233410>
Running find_library for libc.so.6
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib/../lib64/libc.so.6

Instantiating CDLL object for libcuda.so
<CDLL 'libcuda.so', handle 11b2fc0 at 0x15020d08fb10>
Running LoadLibrary for libcuda.so
<CDLL 'libcuda.so', handle a80110 at 0x15050f08f410>
Running find_library for libcuda.so
None

Instantiating CDLL object for /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib64/libSDL2.so
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib64/libSDL2.so', handle 1d45320 at 0x1486a1093210>
Running LoadLibrary for /cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib64/libSDL2.so
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/software/linux/x86_64/amd/zen2/software/SDL2/2.28.5-GCCcore-13.2.0/lib64/libSDL2.so', handle 10e15e0 at 0x14a9ba573bd0>

Instantiating CDLL object for /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6', handle 15294861db70 at 0x152947953210>
Running LoadLibrary for /cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6
<CDLL '/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/lib64/libc.so.6', handle 152ed6696b70 at 0x152ed5993a50>
```

**Comment by casparvl** [comment]
EB_ARGS="--include-easyblocks-from-pr 3860"

**Comment by casparvl** [comment]
EB_ARGS="--include-easyblocks-from-pr 3860"

**Comment by casparvl** [comment]
Test report by @casparvl
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
tcn300.local.snellius.surf.nl - Linux RHEL 9.4, x86_64, AMD EPYC 7H12 64-Core Processor, Python 3.9.18
See https://gist.github.com/casparvl/3d23093773075e1824cc22ed207f539e for a full test report.

**Comment by casparvl** [comment]
@boegel I've also added the patch to all older & newer python versions. Haven't tested yet, as I don't have all these toolchains locally - so will test with boegelbot. I'd propose that if the patch doesn't apply to (some of the) older/newer versions, I just remove those changes and we merge for the versions where it _does_ apply (and hopefully that covers all versions in EESSI 2023.06). That'll at least remove the immediate issue. We can then see what we need to do to backport the change to older versions, and if that seems worth the trouble.

**Comment by casparvl** [comment]
EB_ARGS="--include-easyblocks-from-pr 3860"

**Comment by casparvl** [comment]
EB_ARGS="--include-easyblocks-from-pr 3860"

File: easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch [hunk 17]
```diff
@@ -0,0 +1,162 @@
...
+-            # See issue #9998
+-            return _findSoname_ldconfig(name) or \
+-                   _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))
++            # Redefine find_library function, it will return the provided name if
++            # path exist, else it will use the set of functions to find the full library path.
++            # it will return the one that has a match.
++            if os.path.exists(name):
++                return name
++            else:
++                return _findSoname_ldconfig(name) or \
```
casparvl, line 157: @boegel I've noticed that the order here is something that _might_ also cause bugs in the future:

```
[casparl@tcn1190 LoSoTo]$ python -c 'from ctypes import util; print(util.find_library("libpython3.so"))'
/home/casparl/eessi/versions/2023.06/software/linux/x86_64/amd/zen4/software/Python/3.11.5-GCCcore-13.2.0/lib/../lib64/libpython3.so
[casparl@tcn1190 LoSoTo]$ python -c 'from ctypes import util; print(util.find_library("python3"))'
/cvmfs/software.eessi.io/versions/2023.06/compat/linux/x86_64/usr/lib64/libpython3.12.so.1.0
```

The first behavior is introduced by this PR: being able to call find_library with the full library name (including `lib` and `.so<.some_version>`) is only possible because `_findLib_gcc(name, name_is_fullname=True)` knows how to handle that case. The second behavior was _always there_, also in the unpatched versions. It's _also_ there in the case where EasyBuild does _not_ filter `LD_LIBRARY_PATH`. The reason is that `find_library` tries `findSoname_ldconfig` _first_. I'm not sure why the python devs ever chose this approach. I'm pretty sure this is _not_ the order that the runtime linker uses. And the result is that stuff from your OS (in EESSI: the compat layer) gets picked up _first_, before anything in  the EasyBuild installation prefix (in EESSI: the software layer). 

I wouldn't be in favor of changing it in this PR: it's not specific to filtering LD_LIBRARY_PATH, the existing issue with PyTables has already been around for way too long, and as you know, I have some time pressure to get PyTables in. But I'm wondering if you'd like to have this changed in a follow-up PR. Or should we just raise an issue on it so that it is 'known' if this ever does result in a real, practical issue...?

casparvl:
EB_ARGS="--include-easyblocks-from-pr 3860"

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
```diff
@@ -0,0 +1,162 @@
...
+             return result
+ 
+         def find_library(name):
+-            # See issue #9998
+-            return _findSoname_ldconfig(name) or \
+-                   _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))
++            # Redefine find_library function, it will return the provided name if
++            # path exist, else it will use the set of functions to find the full library path.
++            # it will return the one that has a match.
++            if os.path.exists(name):
```
casparvl: This has an unintended side effect: if `name` is e.g. `libpython3.so` and the current dir contains such a file, this evaluates to true and returns `libpython3.so` (instead of a full path!). We wanted this patch to only return full paths.

Solution is simple: wrap this in an extra condition that checks if `name` is a relative path or full path. I.e.

```suggestion
+            if os.path.isabs(name) and os.path.exists(name):
```

We don't want to pick up something just because it 'happens' to be in the current working dir.

I'm doing one more build to confirm this is indeed the issue. If so, I'll commit this change and try to rebuild.

casparvl: Yes, that's the issue. I (locally) modified the patch with some debugging output, so it looks like this:

```
        def find_library(name):
            print(f"find_library called from 'else'")
            # Redefine find_library function, it will return the provided name if
            # path exist, else it will use the set of functions to find the full library path.
            # it will return the one that has a match.
            if os.path.exists(name):
                print(f"file {name} exists in the current dir, returning without resolving full path")
                return name
            else:
                return _findSoname_ldconfig(name) or \
                   _findLib_gcc(name) or _findLib_ld(name) or \
                   _findLib_gcc(name, name_is_fullname=True)
```

and get the output:

```
== 2025-10-20 14:27:19,567 run.py:624 INFO Output of 'python ...' shell command (stdout + stderr):
find_library called from 'else'
file libpython3.so exists in the current dir, returning without resolving full path
libpython3.so
```

So the early return (because the file is found in the current working dir) is causing the issue.

easybuild/easyconfigs/p/Python/Python-3.10.4-GCCcore-11.3.0-bare.eb
```diff
@@ -11,7 +11,15 @@ toolchainopts = {'pic': True}
...
 sources = [SOURCE_TGZ]
-checksums = ['f3bcc65b1d5f1dc78675c746c98fcee823c038168fc629c5935b044d0911ad28']
+
+# Like patches, but these will only be applied if EasyBuild is configured to filter LD_LIBRARY_PATH
+# In that scenario, ctypes needs to be patched since it heavily relies on LD_LIBRARY_PATH to find libraries
+patch_ctypes_ld_library_path = 'Python-3.11.5-custom-ctypes.patch'
+
+checksums = [
+    {'Python-3.10.4.tgz': 'f3bcc65b1d5f1dc78675c746c98fcee823c038168fc629c5935b044d0911ad28'},
+    {'Python-3.11.5-custom-ctypes.patch': '477a9453395f5298a24f9929ec4b9ecf32259ee4e121cb15ae45e0ad0c195dbe'}
```
casparvl: Need to update this to `b29c22f47587460149e05296ff09b29bf790a83e2b3b13fb2f42f5f236ad8ea7`, but let's see if the latest adaptation works first.

casparvl: Ok, test worked, now need to update all checksums

casparvl:
Proof of success for an environment where this patch is relevant (i.e. that's using RPATH and filtering LD_LIBRARY_PATH): it works for EESSI https://github.com/EESSI/software-layer/pull/1238#issuecomment-3412445253 , which contains 4 of the Python's that are part of this PR.

casparvl:
Test report by @casparvl
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
tcn398.local.snellius.surf.nl - Linux RHEL 9.4, x86_64, AMD EPYC 7H12 64-Core Processor (zen2), Python 3.13.4
See https://gist.github.com/casparvl/37c8312bcde7eb0ed749c00ae52a1f00 for a full test report.

edit: on top of EESSI 2025.06

casparvl:
Test report by @casparvl
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
tcn424.local.snellius.surf.nl - Linux RHEL 9.4 (Plow), x86_64, AMD EPYC 7H12 64-Core Processor (zen2), Python 3.11.4
See https://gist.github.com/casparvl/742cd16e873963d15b7d0305863bbe5f for a full test report.

edit: on top of EESSI 2023.06

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
boegel: I would just open an issue for now, and thoroughly document this potential problem popping up, so it's easy to find when we start searching for error messages that may pop up due to this.

boegel (dismissed):
lgtm

boegel:
Patch applies to all Python versions being touched here, I've checked that manually.

That's important to check, since the patch specified in `patch_ctypes_ld_library_path` is only actually applied when EasyBuild is configured to not update `$LD_LIBRARY_PATH`...

Patch applies cleanly for Python 3.11.x, and with minor offsets for other versions, like:

* Python 3.9.5
  <details>

  ```
  patching file Lib/ctypes/__init__.py
  Hunk #2 succeeded at 350 (offset -2 lines).
  patching file Lib/ctypes/util.py
  ```
  </details>

* Python 3.13.1
  <details>

  ```
  patching file Lib/ctypes/__init__.py
  Hunk #1 succeeded at 15 (offset 1 line).
  Hunk #2 succeeded at 366 (offset 14 lines).
  patching file Lib/ctypes/util.py
  Hunk #1 succeeded at 111 with fuzz 2 (offset 12 lines).
  Hunk #2 succeeded at 134 (offset 12 lines).
  Hunk #3 succeeded at 158 (offset 12 lines).
  Hunk #4 succeeded at 173 (offset 12 lines).
  Hunk #5 succeeded at 313 (offset 12 lines).
  Hunk #6 succeeded at 324 (offset 12 lines).
  Hunk #7 succeeded at 333 (offset 12 lines).
  Hunk #8 succeeded at 364 (offset 12 lines).
  ```

  </details>

Also still applies for Python 3.14.0:
<details>

```
patching file Lib/ctypes/__init__.py
Hunk #1 succeeded at 16 (offset 2 lines).
Hunk #2 succeeded at 438 (offset 86 lines).
patching file Lib/ctypes/util.py
Hunk #1 succeeded at 205 with fuzz 2 (offset 106 lines).
Hunk #2 succeeded at 228 (offset 106 lines).
Hunk #3 succeeded at 252 (offset 106 lines).
Hunk #4 succeeded at 267 (offset 106 lines).
Hunk #5 succeeded at 407 (offset 106 lines).
Hunk #6 succeeded at 418 (offset 106 lines).
Hunk #7 succeeded at 427 (offset 106 lines).
Hunk #8 succeeded at 458 with fuzz 2 (offset 106 lines).
```

</details>

We may need to rework the patch file at some point, but we're good for now...

casparvl:
Test report by @casparvl
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
tcn398.local.snellius.surf.nl - Linux RHEL 9.4, x86_64, AMD EPYC 7H12 64-Core Processor (zen2), Python 3.13.4
See https://gist.github.com/casparvl/e38cb28126069845f5ca22bb766126a3 for a full test report.

edit: on top of EESSI 2025.06 (see also https://github.com/EESSI/software-layer/pull/1254)

casparvl:
Test report by @casparvl
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
tcn424.local.snellius.surf.nl - Linux RHEL 9.4 (Plow), x86_64, AMD EPYC 7H12 64-Core Processor (zen2), Python 3.11.4
See https://gist.github.com/casparvl/283c9b0b2677f935a6be0040b464cf19 for a full test report.

edit: on top of EESSI 2025.06 (see also https://github.com/EESSI/software-layer/pull/1238 + https://github.com/EESSI/software-layer/pull/1252)

boegel:
Test report by @boegel
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
node4201.shinx.os - Linux RHEL 9.6 (Plow), x86_64, AMD EPYC 9654 96-Core Processor, Python 3.13.4
See https://gist.github.com/boegel/4f76d3593fb14d3151c1edcd3b72372b for a full test report.

edit: on top of EESSI 2025.06 (see also https://github.com/EESSI/software-layer/pull/1254)

boegel:
CORE_CNT=16
EB_ARGS="Python-3.9.5-GCCcore-10.3.0-bare.eb Python-3.10.8-GCCcore-12.2.0.eb Python-3.11.3-GCCcore-12.3.0.eb Python-3.13.1-GCCcore-14.2.0.eb --include-easyblocks-from-pr 3860 --installpath /tmp/$USER/pr23499"

boegel:
Test report by @boegel
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
node4201.shinx.os - Linux RHEL 9.6 (Plow), x86_64, AMD EPYC 9654 96-Core Processor, Python 3.11.4
See https://gist.github.com/boegel/6125fe20360acc20c49f216b62c2f70c for a full test report.

edit: on top of EESSI 2023.06

boegel:
Test report by @boegel
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3860
**SUCCESS**
node3541.doduo.os - Linux RHEL 9.6 (Plow), x86_64, AMD EPYC 7552 48-Core Processor, Python 3.11.4
See https://gist.github.com/boegel/650f1c7b534d634de709c441d55d0d5b for a full test report.

edit: on top of EESSI 2023.06

easybuild/easyconfigs/p/Python/Python-3.11.5-custom-ctypes.patch
casparvl: Done:
* https://github.com/easybuilders/easybuild-easyconfigs/issues/24344

boegel:
problem with easyconfigs test suite reporting that there's one too many checksums will be fixed after merge of:
* https://github.com/easybuilders/easybuild-easyblocks/pull/3965

boegel:
Test report by @boegel
Using easyblocks from PR(s) https://github.com/easybuilders/easybuild-easyblocks/pull/3965
**SUCCESS**
node4201.shinx.os - Linux RHEL 9.6 (Plow), x86_64, AMD EPYC 9654 96-Core Processor, Python 3.13.4
See https://gist.github.com/boegel/d499bad837d1b12eee1d1e7df65613cb for a full test report.

boegel:
Going in, thanks @dagonzalezfo!

dagonzalezfo:
Thanks to @casparvl,  @ocaisa and @boegel for all the effort :1st_place_medal:

PR #23516, opened 2025-07-28
Mesa: fix broken source_urls for versions 23.x 

Carldkennedy:
To reflect the current structure of the [archive.mesa3d.org](https://archive.mesa3d.org/) website.

The old path:
`https://archive.mesa3d.org/older-versions/%(version_major)s.x/`
no longer resolves for versions 23.x
Replaced with:
`https://archive.mesa3d.org/%(version_major)s.x/`
which matches the live layout for 23.x tarballs.

Tested against:

Mesa-23.1.4-GCCcore-13.2.0.eb

Successful download and checksum match.

### DIFF
--- a/easybuild/easyconfigs/m/Mesa/Mesa-23.1.4-GCCcore-12.3.0.eb
+++ b/easybuild/easyconfigs/m/Mesa/Mesa-23.1.4-GCCcore-12.3.0.eb
@@ -20,7 +20,7 @@ toolchain = {'name': 'GCCcore', 'version': '12.3.0'}
 
 source_urls = [
     'https://archive.mesa3d.org/',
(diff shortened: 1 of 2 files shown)


lexming:
EB_ARGS="--stop=fetch --force-download=sources"

lexming (approved):
LGTM

lexming:
Merging, thanks @Carldkennedy !

boegel:
@Carldkennedy Hmm, looks like something changed (again), because https://archive.mesa3d.org/23.x doesn't exist (anymore)?

There's no pressing problem, since the source tarballs for Mesa 23.x are directly available via https://archive.mesa3d.org/, but that's bound to change soon (and I'd guess they'll go with the existing pattern of introducing https://archive.mesa3d.org/older-versions/23.x/ ) ?

PR #23529, opened 2025-07-30
fix hardcoded path to java binary in canu v2.3 + add OpenSSL dependency

wdlingit:
As the discussion in https://github.com/marbl/canu/issues/2359, the Failed Java version check problem also happened to me. This modification hard-codes the bundled java binary path to variable `$java` in `Defaults.pm`. This might be the simplest way to fix this issue according to the discussion thread.

### DIFF
--- a/easybuild/easyconfigs/c/canu/canu-2.3-GCCcore-13.3.0-Java-11.eb
+++ b/easybuild/easyconfigs/c/canu/canu-2.3-GCCcore-13.3.0-Java-11.eb
@@ -23,7 +23,6 @@ dependencies = [
     ('Java', '11', '', SYSTEM),
     ('Perl', '5.38.2'),
     ('gnuplot', '6.0.1'),
-    # mark next line if the environment has OpenSSL3
     ('OpenSSL', '3', '', SYSTEM),
 ]
 


easybuild/easyconfigs/c/canu/canu-2.3-GCCcore-13.3.0-Java-11.eb
```diff
@@ -23,10 +23,16 @@ dependencies = [
     ('Java', '11', '', SYSTEM),
     ('Perl', '5.38.2'),
     ('gnuplot', '6.0.1'),
+    # mark next line if the environment has OpenSSL3
```
smoors: this comment is unnecessary. if OpenSSL-3 is installed in the system the OpenSSL-3 installed `openssl` binary will be a symlink to the system installed one

easybuild/easyconfigs/c/canu/canu-2.3-GCCcore-13.3.0-Java-11.eb
```diff
@@ -23,10 +23,16 @@ dependencies = [
...
+    ('OpenSSL', '3', '', SYSTEM),
 ]
 
 start_dir = 'src'
 
+
+build_cmd = r'sed -i "s|java = .*command.*;|'
+build_cmd += 'java = \'${EBROOTJAVA}/bin/java\';|" pipelines/canu/Defaults.pm && make'
```
smoors: better to use prebuildopts for this:

```suggestion
prebuildopts = r'sed -i "s|java = .*command.*;|'
prebuildopts += 'java = \'${EBROOTJAVA}/bin/java\';|" pipelines/canu/Defaults.pm && '
```

wdlingit: @smoors Thank you for the comments. I committed your changes. The comment of SSL3 was also removed.

smoors:

smoors:
Going in, thanks @wdlingit!

PR #23304, opened 2025-07-03
{tools}[GCCcore/14.2.0] ROCm-LLVM v6.4.1

Thyre:
(created using `eb --new-pr`)

PR #23279, opened 2025-07-02
{lib}[GCCcore/14.2.0] amdsmi v25.4.2 w/ ROCm 6.4.1

Thyre:
(created using `eb --new-pr`)

----

One of the first EasyConfigs for the ROCm 6.4.1 stack. Doesn't require ROCm itself, therefore straight-forward.

Tests are not built as they require both a GPU and access to change hardware features, which is very unlikely to be available when built via EasyBuild. Many of these tests will be skipped with upcoming `amd-smi` versions due to requiring `root` access.

----

Sanity check cannot run `amd-smi`, as it fails on systems without the correct kernel modules loaded.
Unfortunately, this is true regardless of the command being passed to `amd-smi`...

PR #23542, opened 2025-07-31
Add easyconfigs for ROCm 6.4.1 using ROCm-LLVM toolchain

Zeldhoron:
This PR adds easyconfigs to EB for ROCm support.
An overview of everything that's added:

1. ROCm-LLVM toolchain 
2. various dependencies that were missing
3. ROCmInfo, AMDSMI, HIP (and ROCm dependencies)
4. some ROCm dependencies for the ROCm Validation Suite (WIP)

What's left to be added:

1. rocBLAS (with or without hipBLASLt)
2. ROCm Validation Suite

This PR builds further upon the following PRs:

- https://github.com/easybuilders/easybuild-framework/pull/4860
- https://github.com/easybuilders/easybuild-framework/pull/4914
- https://github.com/easybuilders/easybuild-easyblocks/pull/3823
- https://github.com/easybuilders/easybuild-easyblocks/pull/3824
- https://github.com/easybuilders/easybuild-easyblocks/pull/3855
- https://github.com/easybuilders/easybuild-easyblocks/pull/3861


Thyre:
There seem to be a few things missing still (like `rocRAND`), right?
A additional things...

- Should we bring my EasyBlock for ROCm components to a draft PR, so that we can use that? I found that it simplifies EasyConfigs quite a bit (see https://github.com/Thyre/easybuild-custom/blob/support-passing-amdgcn/easybuild/easyblocks/r/rocmcomponent.py), especially passing GPU architectures and so on.
- Theoretically, most of these EasyConfigs should work within GCCcore, with a few, e.g. relying on Boost, needing to be in GCC/LLVMtc. One can use the ROCm libraries, built with ROCm-LLVM with GCC. Don't know how we want to handle this.

easybuild/easyconfigs/r/ROCmInfo/ROCmInfo-6.4.1-LLVMtc-ROCm-6.4.1.eb
```diff
@@ -0,0 +1,38 @@
+easyblock = 'CMakeMake'
+
+name = 'ROCmInfo'
```
Thyre: We should stick with the upstream name `rocminfo`

easybuild/easyconfigs/r/ROCmInfo/ROCmInfo-6.4.1-LLVMtc-ROCm-6.4.1.eb
```diff
@@ -0,0 +1,38 @@
...
+configopts = "-DCMAKE_EXE_LINKER_FLAGS='-Wl,--disable-new-dtags'"
+configopts += " -DROCRTST_BLD_TYPE=Release"
+configopts += " -DROCM_DIR=%(installdir)s"
+
+sanity_check_paths = {
+    'files': ['bin/rocminfo', 'bin/rocm_agent_enumerator'],
+    'dirs': [],
+}
+sanity_check_commands = [
+    'rocminfo --help',
```
Thyre: `--help` does not exist, and the check will likely fail if no GPU is present (to be verified).

Thyre:
Generally, I think trying to sync the integration would make sense as well.
I have #23304 open as a first step to add ROCm-LLVM to `2025a` as well, and we should try to have a consistent style at least if we add this to two toolchains.

I might try to update my ECs to ROCm 6.4.2 (or add them to 14.3.0 to make them available on our AMD nodes), but don't know when I find the time for that.

Thyre:
You'll need to port the patch for ROCm-LLVM from #23304 as well, or else building will fail after the merge of https://github.com/easybuilders/easybuild-easyblocks/pull/3799. This fixes both the build failure and another issue when using the installation later on, e.g. for OpenMP offload.

easybuild/easyconfigs/a/AMDSMI/AMDSMI-6.4.1-LLVMtc-ROCm-6.4.1.eb
```diff
@@ -0,0 +1,39 @@
+easyblock = 'CMakeMake'
+
+name = 'AMDSMI'
```
Thyre:
```suggestion
name = 'amdsmi'
```

Use name of existing `amdsmi` EasyConfigs

easybuild/easyconfigs/a/AMDSMI/AMDSMI-6.4.1-LLVMtc-ROCm-6.4.1.eb
```diff
@@ -0,0 +1,39 @@
...
+version = '6.4.1'
+
+homepage = 'https://github.com/ROCm/amdsmi'
+description = """The AMD System Management Interface (AMD SMI) library offers a unified tool
+for managing and monitoring GPUs, particularly in high-performance computing environments.
+It provides a user-space interface that allows applications to control GPU operations,
+monitor performance, and retrieve information about the system's drivers and GPUs."""
+docurls = ['https://rocmdocs.amd.com/projects/amdsmi/']
+
+toolchain = {'name': 'LLVMtc', 'version': 'ROCm-6.4.1'}
```
Thyre:
```suggestion
toolchain = {'name': 'GCCcore', 'version': '13.3.0'}
```

There should be nothing that requires building `amdsmi` with ROCm-LLVM, see 

https://github.com/easybuilders/easybuild-easyconfigs/blob/856e35188b9ec014cc8f95e13044e14e5d7463f9/easybuild/easyconfigs/a/amdsmi/amdsmi-25.4.2-GCCcore-14.2.0-ROCm-6.4.1.eb

Thyre: Also note the `esmi` components, which are required (building may fail otherwise) and _should not_ be pulled in during the build. We can provide them ourselves.

easybuild/easyconfigs/c/CppHeaderParser/CppHeaderParser-2.7.4-GCCcore-13.3.0.eb
```diff
@@ -0,0 +1,35 @@
...
+
+toolchain = {'name': 'GCCcore', 'version': '13.3.0'}
+
+dependencies = [
+    ('binutils', '2.42'),
+    ('Python', '3.12.3'),
+]
+
+use_pip = True
+sanity_pip_check = True
```
Thyre:
```suggestion
```

Not required anymore with EB 5.0+

Thyre:
I feel like all components should use the actual version instead of simply specifying `6.4.1`, i.e. the ROCm version.
Each component has an actual version, and this could potentially allow us to provide a second version of something if there's a known issue being fixed later on.

We might need to figure out if all components provide a tarball / tag for "their" releases as well or only for ROCm releases though.

Thyre:
Short discussion results from the EESSI slack with @Zeldhoron:

- Build ROCm-LLVM toolchain using GCCcore 13.3.0
- Build ROCm components using ROCm-LLVM toolchain
  - Use package version for the component version
  - ROCm-LLVM toolchain adds ROCm version suffix
  - Use the ROCm component easyblock (see https://github.com/easybuilders/easybuild-easyblocks/pull/3861)
- Build rocm-smi  and amdsmi and any others in that category (i.e. without explicit reliance on ROCm-LLVM or packages built by it) using GCCcore
  - Add a ROCm version suffix manually (see e.g. https://github.com/easybuilders/easybuild-easyconfigs/pull/23279)
  - Use the CMakeMake easyblock

easybuild/easyconfigs/l/LLVMtc/LLVMtc-ROCm-6.4.1.eb
```diff
@@ -0,0 +1,24 @@
+easyblock = 'Bundle'
+
+name = 'LLVMtc'
+version = 'ROCm-6.4.1'
```
Crivella: I am wondering how does interact with the logic in the toolchain to filter out flags that are not supported by flang
https://github.com/easybuilders/easybuild-framework/blob/7b14ab77662dbead8f4eb0e08d925373ed6135db/easybuild/toolchains/compiler/llvm.py#L59-L66
https://github.com/easybuilders/easybuild-framework/blob/bb62f27df2d3873480b1afe1eba81d7c00db101a/easybuild/toolchains/compiler/llvm.py#L162-L181

We should check that this does not cause LooseVersion to Fail and possibly implement some way to keep the version of LLVM to know what to filter (or add a list of filters for ROCm versions)

Thyre: The first check is basically for flags which are not understood until LLVM 21? If so, then we probably have to wait for a first ROCm version based on LLVM 21.

We could also have a ROCmtc, based on LLVMtc. That's yet another toolchain, but would make handling such versioning things easier.

Crivella: I put 21 there as it was not out yet when  made the toolchain but have to test if the `flang` in 21 actually supports those flags or not

Crivella: > The first check is basically for flags which are not understood until LLVM 21? If so, then we probably have to wait for a first ROCm version based on LLVM 21.

Not necessarily those checks are to allow general flags to be passed to bot `clang` and `flang` but than filter them out if flang does not supported them.
The other way would be to split the flags from generic into C and F _OPTIONS but it would still require some level of version control

Crivella: I think maybe the better approach is to use the LLVM version as the version and then suffix it with `ROCm-<rocm_version>`, that way we can reuse the logic to check allowed options

Zeldhoron:
I've finished updating everything to match the results of the discussion we had. I didn't have the time to test all the changes I made, but there shouldn't be any major difficulties. There are still a couple of easyconfigs that need to be added (`hipBLASLt`, `rocBLAS`, `ValidationSuite`), but they are currently blocked on the `hipBLASLt` issue https://github.com/ROCm/rocm-libraries/issues/316. A previous version of these easyconfigs (including the ones that still need to be added) can be found at https://github.com/Zeldhoron/easyconfig-rocm/tree/master/easyconfigs.

boegel:
The problem with excessive memory usage is also being looked at by Gentoo, see https://bugs.gentoo.org/960513

In a [recent commit](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=d4cecd11415c26e7679aaeaa76b2c00d9ed18d9e) made in the context of that issue, they introduced using `-DTensile_CPU_THREADS` to control how many cores are being used, which could perhaps resolve the issue?

Thyre:
> The problem with excessive memory usage is also being looked at by Gentoo, see https://bugs.gentoo.org/960513
> 
> In a [recent commit](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=d4cecd11415c26e7679aaeaa76b2c00d9ed18d9e) made in the context of that issue, they introduced using `-DTensile_CPU_THREADS` to control how many cores are being used, which could perhaps resolve the issue?

If I remember correctly, @Zeldhoron was already using this variable, as I also have been doing for hipBLASLt. It just looks like that some GPU targets consume excessive amounts of memory compared to others. Building for `gfx1201` with 32GB of system memory went smooth for me with this as the only target, as did building for `gfx90a` on our MI250X node, though we had 512GB (or 256GB?) of system memory there.

I've not tried building for other targets so far. That will probably come once we're a bit further with the ROCm stack, as I don't want to open draft PRs for all components until we've at least the ROCm LLVM itself merged...

alexschroeter:
```
This was generated with the help of AI
```

**Update (2026-08-20):** filed the LD_LIBRARY_PATH part as a standalone issue
per the suggestion below — EESSI/software-layer#1589 — since PR comments can
get lost. That issue now also has a **validated fix**: a `buildenv/rocm-compilers`
easyconfig that gives `hipcc`-compiled binaries a correct rpath automatically
(the same mechanism EESSI already uses for other toolchains), tested end-to-end
on real gfx90a hardware. No modulefile change needed once that buildenv is
loaded. Full details in the linked issue.

The `module show`/`module help` Lua crash below is still open and
unaddressed — same root cause as easybuild-framework#5216 (unguarded
`os.getenv(...) .. "/suffix"` concatenation in a generated modulefile
footer), just a different package triggering it.

---

# HIP 6.4.1 easyconfig: missing `LD_LIBRARY_PATH` and a `module show`/`help` crash

Tested the `HIP-6.4.43483-LLVMtc-19.0.0-ROCm-6.4.1` module as deployed in EESSI 2025.06
(`accel/amd/gfx90a`) on real AMD Instinct MI210 (gfx90a) hardware. Compiling and running a
trivial HIP kernel with `hipcc` surfaced two issues in the generated module.

## 1. Module never sets `LD_LIBRARY_PATH`

```bash
module load HIP
hipcc --offload-arch=gfx90a vecadd.hip -o vecadd
./vecadd
# error while loading shared libraries: libamdhip64.so.6: cannot open shared object file
```

The module sets `LIBRARY_PATH` (via `prepend_path`) and exposes `HIP_LIB_PATH` as an env var,
but never prepends the lib directory to `LD_LIBRARY_PATH`. `LIBRARY_PATH` is consulted by the
compiler/linker at *build* time, not by the runtime loader — so any binary compiled and linked
against `libamdhip64` this way fails to find it at *run* time, even though the module clearly
intends the library to be discoverable (it exports `HIP_LIB_PATH` for exactly this purpose).

Workaround: `export LD_LIBRARY_PATH=$HIP_LIB_PATH:$LD_LIBRARY_PATH`. Suggest adding a
`prepend_path("LD_LIBRARY_PATH", ...)` (or the equivalent `modextrapaths`/`modluafooter` entry
in the easyconfig) for the `lib` directory, consistent with how most other EasyBuild-generated
runtime modules handle their shared libraries.

## 2. `module show HIP` / `module help HIP` crash with a Lua error

```
$ module show HIP
Lmod has detected the following error: Unable to load module because of error
when evaluating modulefile:
.../HIP/6.4.1-....lua:[string "help([==["]:61: attempt to concatenate a nil value
```

This comes from the `modluafooter` block:

```lua
setenv("DEVICE_LIB_PATH", os.getenv("EBROOTROCMMINLLVM") .. "/amdgcn/bitcode")
setenv("HIP_CLANG_INCLUDE_PATH", os.getenv("EBROOTROCMMINLLVM") .. "/include")
setenv("HIP_CLANG_PATH", os.getenv("EBROOTROCMMINLLVM") .. "/bin")
setenv("HSA_PATH", os.getenv("EBROOTROCMMINLLVM"))
```

This assumes `ROCm-LLVM` is already loaded and `EBROOTROCMMINLLVM` is set — which happens to be
true during `module load HIP`, because the toolchain (`LLVMtc-19.0.0-ROCm-6.4.1`) transitively
loads `ROCm-LLVM` as part of its own hierarchy by the time this line executes. But Lmod's
`show`/`help` modes don't execute those dependency/toolchain side effects, so
`os.getenv("EBROOTROCMMINLLVM")` returns `nil` there and the string concatenation throws,
making `module show HIP` and `module help HIP` unusable for end users.

Suggest guarding these lines against a `nil` toolchain root (e.g. only setting them if
`os.getenv("EBROOTROCMMINLLVM")` is non-nil, or sourcing the path a different way that doesn't
depend on a fully-loaded toolchain at modulefile-evaluation time).

## Environment

- EESSI version: 2025.06
- Software subdir: `x86_64/amd/zen2`, accelerator: `accel/amd/gfx90a`
- Hardware: 8x AMD Instinct MI210 (gfx90a), ROCm kernel driver 6.14.14 (host)
- Reproduced via: `export EESSI_ACCELERATOR_TARGET_OVERRIDE=accel/amd/gfx90a && source
  /cvmfs/software.eessi.io/versions/2025.06/init/bash && module load HIP`

alexschroeter:
```
This was generated with the help of AI
```

**Update (2026-08-20): the analysis below is incorrect.** Leaving it in place
for context — corrected version follows.

Checked every library in the dependency chain with `readelf -d`
(`libamdhip64.so.6`, `libamd_comgr.so.3`, `libhsa-runtime64.so.1`,
`librocprofiler-register.so.0`, `libgflags.so.2.2`, `libz3.so.4.13`) — **zero
`DT_RUNPATH` anywhere.** The DT_RPATH/DT_RUNPATH interaction described below
isn't what's happening.

**Actual root cause:** `hipcc`-compiled binaries carry no rpath at all.
Because glibc resolves each SONAME once per process and shares that binding
with every later consumer, and processes the main binary's own direct
dependencies before descending into `libamdhip64.so.6`'s own (correctly
rpathed) dependency tree, the unrpathed binary's own request for
`libstdc++.so.6` is the *first* one in the whole process — falls through to
the compat layer's old `libstdc++`, and that binding sticks for everyone,
regardless of `gflags`/`Z3`'s own correct RPATH entries.

**Validated fix:** built a `buildenv/rocm-compilers` easyconfig completing
the `HIP_CLANG_PATH` approach discussed in the AMD-ROCm Slack thread — gives
`hipcc` binaries the same rpath treatment EasyBuild's wrapped compilers give
everything else. Fixes this and the missing-`LD_LIBRARY_PATH` issue in one
shot, validated end-to-end on real gfx90a hardware (8x MI210). Full write-up
and working easyconfig: EESSI/software-layer#1589.

---

# ROCm/comgr workloads pick up the wrong libstdc++ via a DT_RPATH/DT_RUNPATH interaction, not a missing rpath entry

## Summary

Running a HIP kernel built against the EESSI 2025.06 `accel/amd/gfx90a` ROCm 6.4.1 stack fails
at runtime with:

```
libstdc++.so.6: version `CXXABI_1.3.15' not found (required by
.../software/gflags/2.2.2-GCCcore-14.2.0/lib/../lib64/libgflags.so.2.2)
```

(same for `Z3/4.13.4-GCCcore-14.2.0`). At first glance this looks like a missing rpath entry,
but it isn't — `gflags.so.2.2` and `libz3.so.4.13` **do** carry a correct RPATH listing
`.../GCCcore/14.2.0/lib64` (confirmed with `readelf -d`). The real cause is a DT_RPATH vs.
DT_RUNPATH interaction that silently defeats that correct entry, compounded by the EESSI
compat layer having its own separate `ld.so.cache`.

## Root cause

Per the standard glibc dynamic-linker rules: an old-style `DT_RPATH` entry is only honored
*transitively* down a dependency chain as long as **no object anywhere in that chain** carries
a `DT_RUNPATH`. The instant one does, `DT_RPATH` on every other object is ignored outright for
the rest of resolution (not merely deprioritized below `LD_LIBRARY_PATH` — dropped).

- `gflags.so.2.2`, `libz3.so.4.13`, and `libamdhip64.so.6` itself all use the old-style
  `DT_RPATH` tag (EasyBuild's rpath wrapper), and all three correctly list
  `.../GCCcore/14.2.0/lib64`.
- Something further down the ROCm/`comgr` dependency chain (most likely an LLVM-derived
  `.so` — modern clang/lld emit `DT_RUNPATH` by default) introduces a `DT_RUNPATH`.
- From that point on, `DT_RPATH` stops being consulted for the rest of the process's
  dependency graph, including for `gflags.so`/`libz3.so`'s own need for `libstdc++.so.6`.

Confirmed directly with an `LD_DEBUG=libs` trace — when resolving `libstdc++.so.6` for
`gflags.so`, the linker does **not** try `gflags.so`'s own RPATH at all. It goes straight from
`LD_LIBRARY_PATH` to:

```
search cache=/cvmfs/software.eessi.io/versions/2025.06/compat/linux/x86_64/etc/ld.so.cache
 trying file=.../compat/linux/x86_64/usr/lib/gcc/x86_64-pc-linux-gnu/13/libstdc++.so.6
```

i.e. it falls through to the **EESSI compat layer's own, separate `ld.so.cache`**, which only
knows about its bundled gcc-13 libstdc++ — GCCcore-14.2.0's newer copy isn't registered there
since it lives outside the compat prefix. That's the exact library swap that breaks the ABI.

This reproduces identically whether the top-level binary uses `LD_LIBRARY_PATH` alone or an
explicit `-Wl,-rpath` (confirmed both ways), so it isn't specific to a particular way of
building the test binary — it's inherent to the dependency graph itself.

## Why this likely matters beyond this one test

Any real ROCm workload that exercises `comgr` for runtime kernel loading/compilation (which is
most GPU codes, not just toy kernels) will pull in this same `gflags`/`Z3` → `libstdc++`
resolution path, and hit the same failure once something with `DT_RUNPATH` sits between the
main binary and those libraries — which, given clang/lld's modern defaults, seems likely to be
common rather than an edge case.

## Possible fixes (not sure which is "correct" from an EasyBuild/EESSI policy standpoint)

1. Build `gflags`/`Z3` (and other EasyBuild deps in this toolchain) with `DT_RUNPATH` instead
   of old-style `DT_RPATH`, so they're no longer vulnerable to being silently overridden by a
   RUNPATH appearing elsewhere in the chain.
2. Register `GCCcore-14.2.0`'s libstdc++ in a location the *compat layer's* `ld.so.cache` (or
   the process's effective default search path) resolves ahead of the compat layer's own copy
   — though this feels more like an EESSI compat-layer design question than something a single
   easyconfig can fix, so may be worth a cross-post to `EESSI/software-layer` if that's judged
   to be the better home for a fix.
3. Have the HIP/ROCm-LLVM modules set `LD_LIBRARY_PATH` to include
   `GCCcore-14.2.0`'s lib64 explicitly (band-aid, but would mask the underlying RPATH issue
   for this specific case without fixing it generally).

## Reproduction

```bash
export EESSI_ACCELERATOR_TARGET_OVERRIDE=accel/amd/gfx90a
source /cvmfs/software.eessi.io/versions/2025.06/init/bash
module load HIP
hipcc --offload-arch=gfx90a vecadd.hip -o vecadd   # minimal HIP vector-add kernel
export LD_LIBRARY_PATH=$HIP_LIB_PATH:$LD_LIBRARY_PATH
LD_DEBUG=libs ./vecadd   # observe libstdc++.so.6 resolution skip gflags.so's own RPATH
```

readelf evidence:

```bash
readelf -d $EBROOTGFLAGS/lib64/libgflags.so.2.2 | grep -E 'RPATH|RUNPATH'
# RPATH (not RUNPATH), correctly lists .../GCCcore/14.2.0/lib64
```

## Environment

- EESSI version: 2025.06
- Software subdir: `x86_64/amd/zen2`, accelerator: `accel/amd/gfx90a`
- Hardware: 8x AMD Instinct MI210 (gfx90a)
- Packages involved: `HIP/6.4.1-rocm-compilers-19.0.0-ROCm-6.4.1`,
  `gflags/2.2.2-GCCcore-14.2.0`, `Z3/4.13.4-GCCcore-14.2.0`

Thyre:
@alexschroeter, thanks a lot for the report of the issues :smile: 
This pull request implements an earlier version of the ROCm toolchain than what was merged in EasyBuild.
Some of the parts are being ported to EESSI as well, with hopefully everything being ported at some point. Right now, ROCm-LLVM, HIP and RCCL should be merged already, with a PR for rocBLAS being open at the moment: https://github.com/EESSI/software-layer/pull/1549

It would be great it you could check if these issues persist there. We tried to make sure that everything is rpathed correctly, but any kind of feedback is always appreciated. Especially https://github.com/easybuilders/easybuild-easyconfigs/pull/23542#issuecomment-5353019492 could be something that still occurs with the merged EasyConfigs.

Regarding `LD_LIBRARY_PATH`: That's by design in EESSI. However, we need to make sure that the correct paths are linked (and potentially rpathed).

If any issues persist, the `amd-rocm` channel in the EESSI slack is certainly a good place to ask.
Opening issues for the particular topics also helps a lot, since comments in PRs can get lost with time...

PR #23552, opened 2025-07-31
Increase `max_failed_tests` for PyTorch 2 ECs to 10

Flamefire:


boegel:
@Flamefire My idea was to bump the default in the PyTorch easyblock (along with cleaning up the custom value in the PyTorch easyconfigs)

Flamefire:
Ah ok, I was also thinking about that too but thought this way it was more visible and forgot to ask. Will do.

Flamefire:
Done:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3869
- https://github.com/easybuilders/easybuild-easyconfigs/pull/23577

PR #23566, opened 2025-08-03
{bio}[foss/2025a] GROMACS v2025.2

absrocks:


easybuild/easyconfigs/g/GROMACS/GROMACS-2025.2-foss-2025a.eb
```diff
@@ -0,0 +1,83 @@
...
+    ('CMake', '3.31.3'),
+    ('scikit-build-core', '0.11.1'),
+]
+
+dependencies = [
+    ('Python', '3.13.1'),
+    ('SciPy-bundle', '2025.06'),
+    ('networkx', '3.5'),
+    ('mpi4py', '4.1.0'),
+    ('pybind11', '2.13.6'),
```
smoors: this should probably a builddependency instead of a normal dependency

smoors (commented):
thanks a lot for your contribution @absrocks !

easybuild/easyconfigs/g/GROMACS/GROMACS-2025.2-foss-2025a.eb
Thyre: Hijacking this thread, since I've built the same version for the potential 2025b:

```
== FAILED: Installation ended unsuccessfully: `/data/EasyBuild-develop/software/Python/3.13.5-GCCcore-14.3.0/bin/python -m pip check` failed:
gmxapi 0.4.2 requires pybind11, which is not installed.
```

So we need a hard-dependency.

smoors: i don't think it's needed.

in the gmxapi sources on pypi (which are used here) it is listed as an install dependency, but in gmxapi sources included in the GROMACS sources it is only listed as a build dependency.

Thyre: So we might be able to work around this with a patch.

Thyre: Patch added in #23699

absrocks: Added patch from #23699  and added pybind in builddependency

Crivella:

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 3 out of 3 (1 easyconfigs in total)
See https://gist.github.com/Crivella/2acf85bf6a6fadd32dcacd28889ef5ff for a full test report.

Crivella:
Going in, thanks @absrocks!

PR #23577, opened 2025-08-05
Remove `max_failed_tests` in PyTorch easyconfigs where the used value was less or equal to the default

Flamefire:
See https://github.com/easybuilders/easybuild-easyconfigs/pull/23552#issuecomment-3140707728

edit: requires:
- https://github.com/easybuilders/easybuild-easyblocks/pull/3869


boegel (approved):
lgtm, makes sense since custom easyblock for PyTorch has been updated to allow max. 10 failing tests (see https://github.com/easybuilders/easybuild-easyblocks/pull/3869)

PR #23580, opened 2025-08-06
{data}[gompi/2025a] netCDF-C++ 4-4.3.1

stefan-wolfsheimer:


easybuild/easyconfigs/n/netCDF-C++4/netCDF-C++4-4.3.1-disable-nc-set-log-level.patch
```diff
@@ -0,0 +1,25 @@
+author: Stefan Wolfsheimer, SURF
+disable nc_set_log_level as it is not supported in netCDF/4.9.3-gompi-2025a
```
casparvl, line 2: Is there an upstream issue on this? If so, it's nice to have a reference here, so that we also know when this patch may become obsolete. 

At least https://github.com/Unidata/netcdf-c/pull/2931/files suggests to me that the `nc_set_log_level` is still supported, though from https://github.com/Unidata/netcdf-c/pull/2931/files I understand that being able to set the log level is configurable at build time through a CMake option (?)

stefan-wolfsheimer: The feature has been disabled in installed netCDF c library configured by [netCDF-4.9.3-gompi-2025a.eb](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/n/netCDF/netCDF-4.9.3-gompi-2025a.eb).
I ran into linking errors.
```
module load 2025
module load netCDF/4.9.3-gompi-2025a

grep  -B 1 nc_set_log_level  $EBROOTNETCDF/include/netcdf.h
EXTERNL int
nc_set_log_level(int new_level);

nm $EBROOTNETCDF/lib/libnetcdf.so  | grep nc_set_log_level
```
Should we look into enabling nc logging in the C-library?

casparvl: > Should we look into enabling nc logging in the C-library?

I'd personally think that would be preferable over patching it out here. I think the fact that it got disabled is because of the changes in https://github.com/Unidata/netcdf-c/pull/2931/files . Re-enabling it would mean that https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/n/netCDF/netCDF-4.9.3-gompi-2025a.eb has the same behavior for `nc_set_log_level` as, say https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/n/netCDF/netCDF-4.9.2-gompi-2024a.eb

stefan-wolfsheimer: Hello @casparvl ,

I tried to enable the logging in the netCDF c-library which caused a failure of the test suite compilation (same type of error). I haven't figured out yet how we can fix it but it might require a patch of the configuration of the test suite. 
The configure script of the C++ library does not detect correctly that the feature has been disabled in the C library.

stefan-wolfsheimer: Hi @casparvl ,

I created an upstream issue for this, as it seems that the root cause of the problem is the configuration of the C++ library not passing the detected configuration to the compilation process:
https://github.com/Unidata/netcdf-cxx4/issues/166

It seems that the logging facilities are not meant for the public facing API:
https://github.com/Unidata/netcdf-fortran/issues/313

Should it be enabled at all?

casparvl: > It seems that the logging facilities are not meant for the public facing API

I read that issue differently: it wasn't _designed_ for it, but it appears to be useful (and thus used!) by external programs. That's sufficient reason why it would be nice if we could have it enabled in the C-library build.

> I tried to enable the logging in the netCDF c-library which caused a failure of the test suite compilation (same type of error). I haven't figured out yet how we can fix it but it might require a patch of the configuration of the test suite.

So... when you enable logging in the configure command for the C-lib, this somehow triggers a test (for the C-lib) for the logging, which then fails? What _is_ the error you are seeing?

stefan-wolfsheimer:
This PR can be closed as the package has been merged in another PR, including the log level fix.

PR #23591, opened 2025-08-07
{geo}[foss/2025a] ESMF v8.8.1 

stefan-wolfsheimer:
depends on

- https://github.com/easybuilders/easybuild-easyconfigs/pull/23580

### DIFF
--- a/easybuild/easyconfigs/e/ESMF/ESMF-8.8.1-foss-2025a.eb
+++ b/easybuild/easyconfigs/e/ESMF/ESMF-8.8.1-foss-2025a.eb
@@ -1,0 +1,37 @@
+name = 'ESMF'
+version = '8.8.1'
+
+homepage = 'https://www.earthsystemcog.org/projects/esmf/'
+description = """The Earth System Modeling Framework (ESMF) is a suite of software tools for developing
+ high-performance, multi-component Earth science modeling applications."""
+
+toolchain = {'name': 'foss', 'version': '2025a'}
+toolchainopts = {'usempi': True, 'openmp': True, 'cstd': 'c++11', 'pic': True}
+
+source_urls = ['https://github.com/esmf-org/esmf/archive/']
+sources = ['v%(version)s.tar.gz']
+patches = ['ESMF-6.1.1_libopts.patch']
+checksums = [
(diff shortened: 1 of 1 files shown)


pavelToman:
Please, update the title to `{geo}[foss/2025a] ESMF v8.8.1`

easybuild/easyconfigs/e/ESMF/ESMF-8.8.1-foss-2025a.eb
```diff
@@ -0,0 +1,37 @@
...
+
+toolchain = {'name': 'foss', 'version': '2025a'}
+toolchainopts = {'usempi': True, 'openmp': True, 'cstd': 'c++11', 'pic': True}
+
+source_urls = ['https://github.com/esmf-org/esmf/archive/']
+sources = ['v%(version)s.tar.gz']
+patches = ['ESMF-6.1.1_libopts.patch']
+checksums = [
+    {'v8.8.1.tar.gz': 'b0acb59d4f000bfbdfddc121a24819bd2a50997c7b257b0db2ceb96f3111b173'},
+    {'ESMF-6.1.1_libopts.patch': '3851627f07c32a7da55d99072d619942bd3a1d9dd002e1557716158e7aacdaf4'},
```
laraPPr, line 16: update the checksums after bumping the version.

easybuild/easyconfigs/e/ESMF/ESMF-8.8.1-foss-2025a.eb
```diff
@@ -0,0 +1,37 @@
+name = 'ESMF'
+version = '8.8.1'
```
laraPPr, line 2: We would prefer if you used the latest release.
```suggestion
version = '8.9.0'
```

laraPPr (changes requested):

laraPPr:
Since https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/e/ESMF/ESMF-8.9.0-foss-2025a.eb is already merged we were wondering if you want this easyconfig to be added to EasyBuild. ESMF is a dependency for a couple other packages so we prefer not to have multiple versions of in the same toolchain.

boegel:
I agree with @laraPPr here, this is superseded by:
- #25367

PR #21427, opened 2024-09-16
{chem}[gfbf/2023b] CENSO v1.2.0

sassy-crick:
(created using `eb --new-pr`)

PR #23697, opened 2025-08-18
[foss/2023b] Delete CENSO-1.2.0-gfbf-2023b.eb

akesandgren:
It is causing problems with the "one-depency-version-per-toolchain-version" rule
A mistake was done when merging #21427 and #23584 to quickly after each other without forcing a rerun of the test-suite in between

edit:
```
found 2 variants of 'CREST' dependency in easyconfigs using '2023b' toolchain generation
* version: 2.12; versionsuffix:  as dep for {'CENSO-1.2.0-gfbf-2023b.eb'}
* version: 3.0.2; versionsuffix:  as dep for {'CENSO-2.1.4-gfbf-2023b.eb'}
```


boegel:
I this case, I think a better approach would've been to add an exception to the easyconfigs test suite to allow the diverging `CREST` version as dependency for CENSO 1.x?

PR #23721, opened 2025-08-20
{bio}[foss/2023a] RepeatModeler v2.0.7, LTR_retriever v2.9.0, TEsorter v1.4.7

nd996:
(created using `eb --new-pr`)
- Found in the `RepeatModeler` readme: _"require the 2.9.0 version of LTR retreiver"_ so switched to that version
- Added `TEsorter` dependency


verdurin:
@nd996 you would need to add exceptions to the tests for `LTR` and `python-isal`, as listed in the unit test reports.

verdurin:
You can see the current test exceptions at:

https://github.com/easybuilders/easybuild-easyconfigs/blob/56b34df5a72dab077e6697cba3a8d7a3b91385ba/test/easyconfigs/easyconfigs.py#L553

nd996:
I'll need to take another look at this, bear with me

PR #23727, opened 2025-08-21
{chem}[foss/2023a] NAMD v3.0.1

robertosussex:
(created using `eb --new-pr`)


robertosussex:
Why was the static analysis / python-linting cancelled?

Crivella:
> Why was the static analysis / python-linting cancelled?

We can retrigger the CI if needed, but any reason you are not using the [NAMD EasyBlock](https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/n/namd.py) like other installations do?

I would assume the only difference you would need for this EC is modify https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/n/NAMD/NAMD-3.0.1-foss-2024a-CUDA-12.6.0.eb to remove the cuda part and downgrade the TC to `2023a`?

robertosussex:
The standard NAMD easyblick assumes that charm++ needs installing this one uses charm++ installed with easybuild .Does it make sense?R.

Yahoo Mail: Search, organise, conquer 
 
  On Tue, 9 Sept 2025 at 13:16, Davide ***@***.***> wrote:   Crivella left a comment (easybuilders/easybuild-easyconfigs#23727)

Why was the static analysis / python-linting cancelled?


We can retrigger the CI if needed, but any reason you are not using the NAMD EasyBlock like other installations do?

I would assume the only difference you would need for this EC is modify https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/n/NAMD/NAMD-3.0.1-foss-2024a-CUDA-12.6.0.eb to remove the cuda part and downgrade the TC to 2023a?

Crivella:
Yeah that EB seems quite old and could use some revamping.

@boegel Not sure if we would want to push for the EB to be changed (e.g only build CHARM if it is not a dependency) and work-on/merge this in the meanwhile.
Or if we would would want to stick to the current approach of having a monolitich build

robertosussex:
Doesnt the monolitich build defeats a bit the purpose of using EasyBuild?
Charm++ IS an independent library , as such (in my opinion) we should have the possibility of installing it separately.

robertosussex:
... and install NAMD using CHARM++ as a necessary dependency...

robertosussex:
Obviously if we change the EasyBlock then the other easyconfigs may have issues?

Crivella:
> Obviously if we change the EasyBlock then the other easyconfigs may have issues?

We would have to do it in a way that it detects if `CHARM` is a dependency and in case it does not rebuild it but links to the existing one.
It might not be a set of trivial changes though from what i've seen by a quick look to the easyblock (in particular i do not think deps are available when sources are extracted but i might be wrong)

robertosussex:
This easyconfig would work as it is as there is already an easyconfig that can install CHARM++ separately

robertosussex:
In the meantime we could let this EConfig in as it does nor require any change in EBlocks, right?

PR #23777, opened 2025-09-02
Adds an EasyConfig for `fenics-dolfinx 0.9.0 python bindings`

MdHassanPappu:
This PR adds an EasyConfig for `fenics-dolfinx 0.9.0 python binding'

During building i'm encountering an issues , the built-in sanity check  
(`import dolfinx`) fails because `dolfinx` requires an initialized MPI environment. 

Even after setting `sanity_check_skip = True` or using:

```
sanity_pip_check = False
runtest = False
```
the automatic import (import dolfinx) still runs and causes the build to fail.

I would greatly appreciate any guidance on whether there is a way to fully bypass this automatic `import dolfinx` sanity check for MPI-bound Python packages? Or any Alternative way to build ? Thnaks in advance.

**getting issues :** 

```
>> running command 'srun -n 1 python -c "from mpi4py import MPI; import dolfinx"' ...
  >> result for command 'srun -n 1 python -c "from mpi4py import MPI; import dolfinx"': OK
== ... (took 5 secs)
== FAILED: Installation ended unsuccessfully: Sanity check failed: command 
"/opt/apps/easybuild/systems/aion/rhel810-20250803/2023b/epyc/software/Python/3.11.5-GCCcore-13.2.0/bin/python -c "import dolfinx"" failed;
output:
[aion-0242:3427192] OPAL ERROR: Unreachable in file ext3x_client.c at line 111
--------------------------------------------------------------------------
The application appears to have been direct launched using "srun",
but OMPI was not built with SLURM's PMI support and therefore cannot
execute. There are several options for building PMI support under
SLURM, depending upon the SLURM version you are using:
```


jhale:
We are still looking at how to execute a custom easy block for the final testing step here, which requires mpi4py to be imported before dolfinx.

easybuild/easyconfigs/f/fenics-dolfinx-py/fenics-dolfinx-py-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,56 @@
+# Dolfinx package for FEniCS
```
jhale: Remove

easybuild/easyconfigs/f/fenics-dolfinx-py/fenics-dolfinx-py-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,56 @@
...
+    ('nanobind', '2.5.0'),
+    ('setuptools', '64.0.3'),  
+    ('pkgconfig', '1.5.5','-python'),
+    ('fenics-dolfinx', '%(version)s'),
+]
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle','2023.11'),
+    ('HDF5', '1.14.3'),
```
jhale: Try moving HDF5 to a build dependency, as in Spack.

PR #23816, opened 2025-09-08
{devel}[GCCcore-13.2.0] Setuptools v80.9.0

MdHassanPappu:
Add `Setuptools` built with GCCcore 13.2.0

The previous Easyconfig  for `Setuptools` included multiple variants with different toolchains.
This PR adds a Easyconfig built with GCCcore 13.2.0.

PR #23810, opened 2025-09-05
{data}[GCCcore-13.2.0] pugixml v1.14

MdHassanPappu:
### Add `pugixml` built with `GCCcore 13.2.0`
The previous EasyBuild files for `pugixml` included multiple variants with different toolchains.
This PR adds a new variant `V1.14`  built  with `GCCcore 13.2.0.`

PR #23808, opened 2025-09-05
{lib}[GCCcore-13.2.0 ] spdlog v1.12.0

MdHassanPappu:
# Add PIC-enabled variant for spdlog 1.12.0

The existing [`spdlog-1.12.0-GCCcore-13.2.0.eb`](https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/s/spdlog/spdlog-1.12.0-GCCcore-13.2.0.eb) builds without PIC.  
However, PIC is required for DolfinX to build shared libraries that link against spdlog.

This PR introduces a second variant of `spdlog` with `versionsuffix='pic'`, enabling shared library support, while keeping the original non-PIC build intact.

PR #23581, opened 2025-08-06
{tools}[foss/2023b] petsc4py v3.22.5

gkaf89:
Add a Python bindings for PETSc of 2023b toolchain.

The version of petsc4py and PETSc must be the same. There was no release 3.22.5 for petsc4py in PyPI, but foss/2023b is using PETSc version 3.22.5. This commit originally built petsc4py from the bindings directory of the 3.22.5 release of PETSc.

Now the source code for the petsc4py v3.22.5 has been [released in PyPI](https://gitlab.com/petsc/petsc/-/issues/1799), and the package is being built from the PyPI source code.

PR #23838, opened 2025-09-10
{tools}[foss/2023b] fenics-dolfinx-python v0.9.0

gkaf89:
This PR adds the core easyconfigs for the FEniCS platform.

Requires:
 
- [ ] #23581
- [ ] #23805
- [ ] #23808
- [ ] #23810
- [ ] #23816


gkaf89:
Test report by @gkaf89
**SUCCESS**
aion-0101 - Linux RHEL 8.10 (Ootpa), x86_64, AMD EPYC 7H12 64-Core Processor, Python 3.11.13
See https://gist.github.com/gkaf89/58a0d42e83724f8df471b224534d8cba for a full test report.

gkaf89:
Test report by @gkaf89
**FAILED**
Build succeeded for 3 out of 7 (7 easyconfigs in total)
aion-0331 - Linux RHEL 8.10 (Ootpa), x86_64, AMD EPYC 7H12 64-Core Processor, Python 3.11.13
See https://gist.github.com/gkaf89/5934c24a1b61776542da99d647f1ed3a for a full test report.

gkaf89:
Test report by @gkaf89
**SUCCESS**
Build succeeded for 14 out of 14 (7 easyconfigs in total)
aion-0331 - Linux RHEL 8.10 (Ootpa), x86_64, AMD EPYC 7H12 64-Core Processor, Python 3.11.13
See https://gist.github.com/gkaf89/33efbca5e7cfb3cd180049cbd43ddede for a full test report.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,60 @@
...
+    ('slepc4py', '3.22.2'),
+]
+
+install_src = './python'
+
+# Disable built-in check for module extension:
+#
+# python -c "import dolfinx"
+#
+# The module extension provide the dolphin Python package that is tested by the sanity check.
```
jhale: provide -> provides, dolphin -> dolfinx

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,60 @@
+easyblock = 'PythonPackage'
+
+name = 'fenics-dolfinx-python'
+version = '0.9.0'
+
+homepage = 'https://github.com/FEniCS/dolfinx'
+description = "DOLFINx is the computational environment of FEniCSx - python binding "
```
jhale: python -> Python

easybuild/easyconfigs/f/FEniCS-Basix-Python/FEniCS-Basix-Python-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,41 @@
...
+]
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+]
+
+install_src = './python'
+
+# Provide custom name for the module extension built-in sanity check: the EB module name is different from the Python
```
jhale, line 30: Remove comment, standard technique.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,60 @@
...
+#
+options = {'modulename': False}
+
+sanity_check_paths = {
+    'files': ['lib/python%(pyshortver)s/site-packages/dolfinx/__init__.py'],
+    'dirs': [],
+}
+
+sanity_check_commands = [
+    'mpirun -N 1 python -c "from mpi4py import MPI; import dolfinx"',
```
jhale: mpiexec is the standard compliant way and portable across all MPI implementations.

gkaf89: I have been complacent with mpirun being effectively the same in OpenMPI, changing to `mpiexec`.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
gkaf89: Done!

easybuild/easyconfigs/f/FEniCS-Basix-Python/FEniCS-Basix-Python-0.9.0-gfbf-2023b.eb
gkaf89: Not that standard! My understanding is that this command does not actually change the name of the Python package, just the name used in the sanity check.

I am in favor of keeping this in because the interface of the sanity checks is not that clear yet, and until it stabilizes, we may have to revisit this sections.

easybuild/easyconfigs/f/fenics-basix/fenics-basix-0.9.0-GCC-13.2.0.eb
```diff
@@ -0,0 +1,31 @@
...
+]
+
+dependencies = [
+    ('FlexiBLAS', '3.3.1'),
+]
+
+srcdir = "cpp"
+
+sanity_check_paths = {
+    'files': ['lib/libbasix.so'],
```
jfgrimm:
```suggestion
    'files': [f'lib/libbasix.{SHLIB_EXT}'],
```

Crivella:
Would it make sense to also have a FEniCS bundle to import all the various subpackages at once?

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,60 @@
...
+install_src = './python'
+
+# Disable built-in check for module extension:
+#
+# python -c "import dolfinx"
+#
+# The module extension provides the dolfinx Python package that is tested by the sanity check.
+#
+# See v5.1.1: easybuild-framework/easybuild/framework/extension.py:313
+#
```
jfgrimm:
```suggestion
# Disable import sanity check for dolfinx, since we need to import MPI from mpi4py first
```

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,44 @@
...
+    ('ParMETIS','4.0.3'),
+    ('PETSc', '3.22.5'),
+    ('fenics-basix', '0.9.0'),
+    ('fenics-ufcx', '0.9.0'),
+    ('pugixml', '1.14'),
+    ('spdlog', '1.12.0', '-pic'),
+]
+
+srcdir = "cpp"
+
```
jfgrimm: re-order these to follow the order in which steps are executed by easybuild
```suggestion
builddependencies = [
    ('make', '4.4.1'),
    ('CMake', '3.27.6'),
    ('HDF5', '1.14.3'),
    ('pkgconf', '2.0.3'),
]

dependencies = [
    ('Boost','1.83.0'),
    ('ParMETIS','4.0.3'),
    ('PETSc', '3.22.5'),
    ('fenics-basix', '0.9.0'),
    ('fenics-ufcx', '0.9.0'),
    ('pugixml', '1.14'),
    ('spdlog', '1.12.0', '-pic'),
]

srcdir = "cpp"

configopts = (
    ' -DDOLFINX_UFCX_PYTHON=OFF '
    ' -DDOLFINX_BASIX_PYTHON=OFF '
)

```

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,44 @@
...
+sanity_check_paths = {
+    'files': ['lib/libdolfinx.so'],
```
jfgrimm:
```suggestion
    'files': [f'lib/libdolfinx.{SHLIB_EXT}'],
```

easybuild/easyconfigs/f/fenics-ffcx/fenics-ffcx-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,33 @@
...
+sources = ['v%(version)s.tar.gz']
+checksums = ['afa517272a3d2249f513cb711c50b77cf8368dd0b8f5ea4b759142229204a448']
+
+dependencies = [
+    ('cffi', '1.15.1'),
+    ('fenics-basix-python', '%(version)s'),
+    ('fenics-ufl', '2024.2.0'),
+    ('Python', '3.11.5'),
+    ('SciPy-bundle','2023.11'),
+    ('setuptools', '80.9.0'),
```
jfgrimm: this should not be a runtime dependency

easybuild/easyconfigs/f/fenics-ufl/fenics-ufl-2024.2.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,27 @@
...
+sources = ['%(version)s.tar.gz']
+checksums = ['d9353d23ccbdd9887f8d6edab74c04fe06d818da972072081dbf0c25bc86f5a7']
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle','2023.11')
+]
+
+sanity_check_paths = {
+    'files': ['lib/python%(pyshortver)s//site-packages/ufl/__init__.py'],
```
jfgrimm:
```suggestion
    'files': ['lib/python%(pyshortver)s/site-packages/ufl/__init__.py'],
```

jhale:
> Would it make sense to also have a FEniCS bundle to import all the various subpackages at once?

I don't think so - `fenics-dolfinx-python` effectively already does this through transitive dependencies, and it will be an extra file to maintain/update. Most of our users are already aware that dolfinx is 'at the top of the stack'.

easybuild/easyconfigs/f/fenics-basix-python/fenics-basix-python-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,41 @@
+# Basix package for FEniCS
```
jhale: Remove this first comment line.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,60 @@
...
+    ('scikit-build-core', '0.10.7'),
+    ('setuptools', '80.9.0'),
+]
+
+dependencies = [
+    ('PETSc', '3.22.5'),
+    ('Python', '3.11.5'),
+    ('SciPy-bundle','2023.11'),
+    ('cffi', '1.15.1'),
+    ('fenics-basix-python', '%(version)s'),
```
jhale: There is no guarantee we will match version numbers in the future - these should be hardcoded.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,60 @@
...
+    ('fenics-ffcx', '%(version)s'),
```
jhale: Same comment, no guarantee we will match version numbers in the future.

easybuild/easyconfigs/f/fenics-ffcx/fenics-ffcx-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,33 @@
...
+
+toolchain = {'name': 'gfbf', 'version': '2023b'}
+
+source_urls = ['https://github.com/FEniCS/ffcx/archive']
+sources = ['v%(version)s.tar.gz']
+checksums = ['afa517272a3d2249f513cb711c50b77cf8368dd0b8f5ea4b759142229204a448']
+
+dependencies = [
+    ('cffi', '1.15.1'),
+    ('fenics-basix-python', '%(version)s'),
```
jhale: No guarantee we will match version numbers in the future - should be hardcoded.

easybuild/easyconfigs/f/fenics-ffcx/fenics-ffcx-0.9.0-gfbf-2023b.eb
jhale: Indeed, probably a mistaken copy from Spack - this was only needed for old versions of FFCx <= 0.6.0.

easybuild/easyconfigs/f/fenics-basix-python/fenics-basix-python-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,41 @@
...
+
+easyblock = 'PythonPackage'
```
Crivella:
```suggestion
easyblock = 'PythonPackage'
```

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
```diff
@@ -0,0 +1,31 @@
...
+builddependencies = [
+    ('make', '4.4.1'),
+    ('CMake', '3.27.6'),
+]
+
+dependencies = [
+    ('FlexiBLAS', '3.3.1'),
+]
+
+srcdir = "cpp"
```
Crivella, line 24:
```suggestion
toolchain = {'name': 'gfbf', 'version': '13.2.0'}

source_urls = ['https://github.com/FEniCS/basix/archive']
sources = ['v%(version)s.tar.gz']
checksums = ['60e96b2393084729b261cb10370f0e44d12735ab3dbd1f15890dec23b9e85329']

builddependencies = [
    ('make', '4.4.1'),
    ('CMake', '3.27.6'),
]

srcdir = "cpp"
```

Since other FEniCS packages are already using it, we might as well use `gfbf` directly (we get the bonus of the extra variables setup by the TC to help find libraries) even if we are not usng FFTW

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
Crivella:
```suggestion
    ('fenics-basix-python', '0.9.0'),
```

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
Crivella:
```suggestion
    ('fenics-ffcx', '0.9.0'),
```

easybuild/easyconfigs/f/fenics-ffcx/fenics-ffcx-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,33 @@
...
+    ('fenics-ufl', '2024.2.0'),
+    ('Python', '3.11.5'),
+    ('SciPy-bundle','2023.11'),
+    ('setuptools', '80.9.0'),
+]
+
+# Provide custom name for the module extension built-in sanity check: the EB module name is different from the Python
```
Crivella, line 26:
```suggestion
checksums = ['afa517272a3d2249f513cb711c50b77cf8368dd0b8f5ea4b759142229204a448']

builddependencies = [
    ('setuptools', '80.9.0'),
]

dependencies = [
    ('cffi', '1.15.1'),
    ('fenics-basix-python', '0.9.0'),
    ('fenics-ufl', '2024.2.0'),
    ('Python', '3.11.5'),
    ('SciPy-bundle','2023.11'),
]

# Provide custom name for the module extension built-in sanity check: the EB module name is different from the Python
```

Addresses both the version and runtime dep comments

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
Crivella: Beside agreeing on shortening the comments, is this really necessary?
In my case i was able to run `import dolfinx` even without importing mpi4py before.

Or is there something that could fail by importing mpi4py before dolfinx if the latter is not built to support it?

Crivella (changes requested):
I am still wondering if we are not making it more complicated for users and us to have 4 different `PythonPackages` instead of a single FEniCS `PythonBundle`.

Since none of these packages are really heavy to install i am not sure it makes sense to split them up.
It will make maintining/updating this ECs much simpler for the future.

The only usecase where it might make sense to split them up if there is reason to have many version of every package that work independently

EDIT: also is there a difference from the GH sources than what is available on PyPI?

Crivella:
@gkaf89 I've created a PR to your PR branch with how we could be bundling the 4 python packages together

https://github.com/gkaf89/easybuild-easyconfigs/pull/3

@jhale Do you also feel this could be reasonable (this way it is easier for us to maintain and i think also easier for users)

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,44 @@
...
+
+builddependencies = [
+    ('make', '4.4.1'),
+    ('CMake', '3.27.6'),
+    ('HDF5', '1.14.3'),
+    ('pkgconf', '2.0.3'),
+]
+
+dependencies = [
+    ('Boost','1.83.0'),
```
Crivella:
```suggestion
    ('pkgconf', '2.0.3'),
]

dependencies = [
    ('HDF5', '1.14.3'),
    ('Boost','1.83.0'),
```

This should definitively be a runtime dep
```
crivella@crivella-desktop:~/.local/easybuild/software/fenics-dolfinx/0.9.0-foss-2023b/lib$ libtree -p libdolfinx.so | grep HDF5
├── /home/crivella/.local/easybuild/software/HDF5/1.14.3-gompi-2023b/lib64/libhdf5.so.310 [rpath]
```

gkaf89:
> @gkaf89 I've created a PR to your PR branch with how we could be bundling the 4 python packages together
> 
> [gkaf89#3](https://github.com/gkaf89/easybuild-easyconfigs/pull/3)
> 
> @jhale Do you also feel this could be reasonable (this way it is easier for us to maintain and i think also easier for users)

We would like to separate the Python components from the C++ components in different modules. FEniCs is not just a Python package. Other than that, I think we can merge thy Python packages.

@jhale Is there a case where one may want to use one of the Python packages in isolation, and how inconvenient is it if all the Python packages are present in the same environment?

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
gkaf89: There has been some [discussion](https://github.com/easybuilders/easybuild-easyconfigs/pull/23497#pullrequestreview-3055037068) about this. I am not sure that it is worth adding extra dependencies, and it signifies a dependency on FFTW which is not correct. I prefer to use the minimal dependency set.

What kind of extra setup is performed by the toolchain?

Crivella: You can check what the TC setup will do when you select one that is Linear Algebra enabled 
https://github.com/easybuilders/easybuild-framework/blob/develop/easybuild/tools/toolchain/linalg.py#L38
and in particular with FlexiBlas
https://github.com/easybuilders/easybuild-framework/blob/816718854b269f0e46713778f1cd76bd46f06b4a/easybuild/toolchains/linalg/flexiblas.py#L67
like the gfbf
https://github.com/easybuilders/easybuild-framework/blob/816718854b269f0e46713778f1cd76bd46f06b4a/easybuild/toolchains/gfbf.py#L39

jhale:
Well in short, very annoying; we have multiple downstream users of UFL, FFCx and Basix who don't use DOLFINx, and multiple commercial users who use DOLFINx C++ who may or may not use one of UFL or FFCx, but really do not want DOLFINx Python.

For the standard user things are very simple; they bring in Python DOLFINx and they are good to go.

Finally, we are about to uncouple all of our version numbers starting from 1.0.0, I.e. semantic versioning.

jhale:
I think we should have a meeting next week about this; there are some fundamental misunderstandings about what these FEniCS packages do and how they work together. The proposal to bundle them is fundamentally incorrect on many levels.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
jhale: You were lucky, `import dolfinx` doesn't initialise MPI. On our cluster it crashes. The user is responsible for initialising MPI using mpi4py's well designed interface https://mpi4py.readthedocs.io/en/stable/mpi4py.html prior to running `import dolfinx`.

Crivella:
> We would like to separate the Python components from the C++ components in different modules. FEniCs is not just a Python package. Other than that, I think we can merge thy Python packages.

> Well in short, very annoying; we have multiple downstream users of UFL, FFCx and Basix who don't use DOLFINx, and multiple companies who use DOLFINx without Python at all.

The changes i proposed in https://github.com/gkaf89/easybuild-easyconfigs/pull/3 would keep the C++ libraries as separate packages, only the python packages are being installed together.

From the current files it seems that only BasiX and UFL could be treated independently while FFCx depends on both and DOLFINx depends on all.

This is the footprint of the packages in the 2 modes
```
crivella@crivella-desktop:~$ du -sh .local/easybuild/software/fenics-* --exclude=easybuild
5.2M    .local/easybuild/software/fenics-basix
768K    .local/easybuild/software/fenics-basix-python
3.4M    .local/easybuild/software/fenics-dolfinx
6.2M    .local/easybuild/software/fenics-dolfinx-python
1.1M    .local/easybuild/software/fenics-ffcx
56K     .local/easybuild/software/fenics-ufcx
2.5M    .local/easybuild/software/fenics-ufl
crivella@crivella-desktop:~$ du -sh .local/easybuild/software/FEniCS/ --exclude=easybuild
11M     .local/easybuild/software/FEniCS/
```

Anyway this is only my suggestion and i do not consider it a blocker for this PR, also happy to join in on a call.

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
jhale: These packages are written by the FEniCS Project but they are fundamentally independent components. The `GCC` toolchain is the correct minimal dependency set.

jhale:
I understand your PR. You want to place all of the Python components in one meta-package, which seems like a nice idea on first glance. However, it is a misunderstanding of how these pieces of software work together, and how users, or other package maintainers of downstream dependencies, may wish to use them in the future.

In particular:

- It is contrary to our upcoming semantic versioning plans - there will be no "FEniCS x.x.x" in the future, so users will be thoroughly confused.
- It forces every FFCx, UFL and Python Basix user and/or downstream dependency to bring in at least one unwanted piece of Python software. For example, if you ever get Firedrake or Dune packages, which depend on UFL, every Firedrake and Dune user will have three unwanted packages in their Python namespace. We also have a number of users who are just using FFCx (and indeed, Basix and UFL as they are dependencies) and don't want any DOLFINx at all.
- is different to every other package manager that builds the FEniCS components, which I agree is not an argument per se, but generally users do like consistency.
- The issue you imagine with user confusion doesn't exist; all of our 'regular' users know they want Python DOLFINx, and so they just bring in `fenics-dolfinx-python` and they are setup and ready to go.

jhale:
You will need to explain the package footprint point - the space taken by the upper Python packages looks equal to the lower amount, which is what I would expect to see.

gkaf89:
We are getting the following error in the tests:
```
binutils or GCC is a build dep: ['make', 'CMake']
```
The make and CMake packages use `binutils-GCCcore` as a build dependency. Does this error mean that whenever we use a package with binutils as build dependency then binutils should be added explicitly in the build dependencies? Is this enforced to ensure that the same binutils is used everywhere?

easybuild/easyconfigs/f/fenics-ffcx/fenics-ffcx-0.9.0-gfbf-2023b.eb
gkaf89: Fixed, `setuptools` is now a build dependency.

easybuild/easyconfigs/f/fenics-basix-python/fenics-basix-python-0.9.0-gfbf-2023b.eb
gkaf89: Removed as requested.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
gkaf89: - Disassociated the versions of fenics-basix, fenics-dolfinx, fenics-ffcx, and fenics-ufcx.
- Only fenics-<X> and fenics-<X>-python have the same version numbers (X=dolfin,basix).

easybuild/easyconfigs/f/fenics-basix-python/fenics-basix-python-0.9.0-gfbf-2023b.eb
gkaf89: Fixed as requested.

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
gkaf89: Yes, I moved HDF5 to runtime dependencies. I also add the PETSc and SLEPc packages in the runtime dependencies to ensure that they are available in the LD_LIBRARY_PATH.

easybuild/easyconfigs/f/fenics-ffcx/fenics-ffcx-0.9.0-gfbf-2023b.eb
gkaf89: Moved `setuptools` to build dependencies.

ocaisa:
CI is failing because there are now two variants of `spdlog` (with and without `pic`) and there is no exception made for this in CI.

I wonder if it is really worth having the two variants though? If `pic` (and building shared libraries) is more or less the only difference, then I think it is fine to have that enabled by default as long as both static and shared are being built.

jfgrimm:
> If pic (and building shared libraries) is more or less the only difference, then I think it is fine to have that enabled by default as long as both static and shared are being built.

I agree

Crivella:
@ocaisa @jfgrimm Did not check that `spdlog` was not on EESSI yet so we don't have to go the extra mile to replace it.

Opened a PR for the combined static/shared build

- #24308

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,45 @@
...
+dependencies = [
+    ('Boost', '1.83.0'),
+    ('ParMETIS', '4.0.3'),
+    ('PETSc', '3.22.5'),
+    ('SLEPc', '3.22.2'),
+    ('HDF5', '1.14.3'),
+    ('fenics-basix', '0.9.0'),
+    ('fenics-ufcx', '0.9.0'),
+    ('pugixml', '1.14'),
+    ('spdlog', '1.12.0', '-pic'),
```
Crivella:
```suggestion
    ('spdlog', '1.12.0'),
```

Depends on

- #24308

easybuild/easyconfigs/f/fenics-basix-python/fenics-basix-python-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,39 @@
...
+checksums = ['60e96b2393084729b261cb10370f0e44d12735ab3dbd1f15890dec23b9e85329']
+
+toolchain = {'name': 'gfbf', 'version': '2023b'}
+
+builddependencies = [
+    ('make', '4.4.1'),
+    ('CMake', '3.27.6'),
+    ('scikit-build-core', '0.10.7'),
+    ('nanobind', '2.5.0'),
+    ('fenics-basix', '%(version)s')
```
Crivella: This also needs to be a runtime dep

```
crivella@crivella-desktop:~/.local/easybuild/software/fenics-basix-python/0.9.0-gfbf-2023b/lib/python3.11/site-packages/basix$ libtree -p _basixcpp.cpython-311-x86_64-linux-gnu.so 
_basixcpp.cpython-311-x86_64-linux-gnu.so 
└── /home/crivella/.local/easybuild/software/fenics-basix/0.9.0-GCC-13.2.0/lib64/libbasix.so [rpath]
    └── /home/crivella/.local/easybuild/software/FlexiBLAS/3.3.1-GCC-13.2.0/lib64/libflexiblas.so.3 [rpath]
        ├── /home/crivella/.local/easybuild/software/GCCcore/13.2.0/lib64/../lib64/libquadmath.so.0 [rpath]
        └── /home/crivella/.local/easybuild/software/GCCcore/13.2.0/lib64/../lib64/libgfortran.so.5 [rpath]
```

Crivella: Also i assume here the version should be the same between `fenics-basix` and `fenics-basix-python`

Crivella:
I've opened a draft PR for ADIOS2

- #24316

Which i see is an optional dependency needed for [VTXWriter](https://docs.fenicsproject.org/dolfinx/v0.5.0/cpp/doxygen/d1/dee/classdolfinx_1_1io_1_1VTXWriter.html) which is being used by users on our side.
Tried to add it to the current `fenics-dolfinx` and it seems to be working, but ADIOS2 itself needs more work to ensure we are doing everything properly

easybuild/easyconfigs/f/fenics-basix-python/fenics-basix-python-0.9.0-gfbf-2023b.eb
gkaf89: Yes, indeed.

- basix is a runtime dependency for basix-python. Will be fixed in new commit.
- The versions are always the same for the python package and the underlying library, this the use of `%(version)s` template.

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
jhale: I'm not sure what the use case for the non-PIC shared library is, or if there is one, it's pretty niche.

jhale:
> I've opened a draft PR for ADIOS2
> 
> * [{tools}[foss/2023b] ADIOS2 v2.10.2 #24316](https://github.com/easybuilders/easybuild-easyconfigs/pull/24316)
> 
> Which i see is an optional dependency needed for [VTXWriter](https://docs.fenicsproject.org/dolfinx/v0.5.0/cpp/doxygen/d1/dee/classdolfinx_1_1io_1_1VTXWriter.html) which is being used by users on our side. Tried to add it to the current `fenics-dolfinx` and it seems to be working, but ADIOS2 itself needs more work to ensure we are doing everything properly

It would be nice to have, but ADIOS2 is a tricky dependency.

gkaf89:
@Crivella @jfgrimm We are preparing a few tests for the installation. I will convert to a draft temporarily,

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
Crivella: The shared libraries are still PIC (not even sure you could make a shared library without, or atleast a working one).
What we did with #24308 is to build both the static `.a` and shared `.so` library so they are bot available within the same installation as it is the preferred approach within EB (to allow using the same dep also for packages that do not support dynamic linking).

Crivella:
@gkaf89 @jhale 

How sensitive is dolfinx to the version of PETSc?
Was trying the same application mentioned in https://github.com/easybuilders/easybuild-easyconfigs/pull/23838#issuecomment-3425642869 as a test, and after compiling it with ADIOS2 it was failing with 

```
RuntimeError: Failed to successfully call PETSc function 'KSPSolve'. PETSc error code is: 92, See https://petsc.org/release/overview/linear_solve_table/ for possible LU and Cholesky solvers                                                                                                   
```

I tried to rebuild everything using PETSc 3.22.2 (the same as what gets installed from conda) and did not hit this error.

EDIT:

> We are preparing a few tests for the installation. I will convert to a draft temporarily,

 was wondering if this is something we should be careful about and have a test for it

jhale:
As a matter of policy we do not fiddle with PETSc's default behaviour - this means that if PETSc is built differently (different version, different optional packages), you can end up with different PETSc solvers running, even if the DOLFINx code looks the same.

The important parameter is the `petsc_options` kwarg to `LinearSolver` https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py#L714 - what are you setting, if anything, there?

A safe and fixed default is, e.g.:

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py#L727

A common gotcha is that users don't specify `"pc_factor_mat_solver_type": "mumps"` - this means PETSc is left to pick any of the LU solvers it was built with, some of which are less robust/performant.

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
jhale: Non-PIC shared by default is just a slightly odd feature of spdlog. Your approach to build both static and shared versions with PIC is the right one for EB.

Crivella:
I think it is what is being done  through different syntax

https://github.com/VTT-ProperTune/AITW-Darcy-Infiltration-Model/blob/758b9e67f37273d2a4002a32b149c64a006c62c9/infiltration.py#L116-L132

They told me they could try different preconditioners.

Anyway from this PR side i think it is just good to keep in mind if we should make an effort to use the same version of PETSc (3.22.2), which might be slightly painful as right now we have `SLEPc 3.22.2` which depends on `PETSc 3.22.5`.
Simplest solution would be to add a versionsuffix to SLEPc

EDIT: Added link to code now that is public.
Also i was using this as a test since it is what i have to run the code.
I do not consider it a blocker especially as it seems more PETSc related and  you are already adding proper testing.

jfgrimm:
> Anyway from this PR side i think it is just good to keep in mind if we should make an effort to use the same version of PETSc (3.22.2), which might be slightly painful as right now we have SLEPc 3.22.2 which depends on PETSc 3.22.5.
Simplest solution would be to add a versionsuffix to SLEPc

@Crivella unless absolutely necessary, please avoid adding multiple versions of `PETSc/SLEPc`. There are other easyconfigs that already depend on the existing versions within this toolchain, and those modules would conflict.

gkaf89:
- ADIOS2 will be added in the 2024a version.
- Tests need a bit more work. They are much better integrated from version 0.10 onward.

easybuild/easyconfigs/f/fenics-dolfinx/fenics-dolfinx-0.9.0-foss-2023b.eb
gkaf89: We are now using the new version of spdlog that supports PIC. So, I am marking this issues as resolved.

jfgrimm:
Test report by @jfgrimm
**SUCCESS**
Build succeeded for 7 out of 7 (7 easyconfigs in total)
node029.viking2.yor.alces.network - Linux Rocky Linux 8.10, x86_64, AMD EPYC 7643 48-Core Processor, Python 3.6.8
See https://gist.github.com/jfgrimm/570b876a2eee1d1fdd5708882da22386 for a full test report.

jfgrimm:

jhale:
I'm happy to have this merged. We can work towards better testing in 0.10 or 0.11 releases of DOLFINx.

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,51 @@
+easyblock = 'PythonPackage'
+
+name = 'fenics-dolfinx-python'
```
boegel: If I'm nitpicking I would like to see these easyconfigs changed so they match our usual policy w.r.t. software names, i.e. using the official "spelling".

In this case, that would be `FEniCS-DOLFINx-Python`.

I can see the value in using the `FEniCS` prefix to keep things nicely "together", so that's OK imho.

Since this provides the Python bindings for `FEniCS-DOLFINx`, I guess the extra `-Python` makes sense.

Although I don't see a downside in collapsing `DOLFINx` and it's Python bindings into a single easyconfig?
So basically merging what's now named `fenics-dolfinx` and `fenics-dolfinx-python` into `FEniCS-DOLFINx`?

jhale:
We will make the naming changes.

For the merging the Python wrappers and C++ library this was discussed above; we have quite a number of users who use DOLFINx C++ without the Python wrappers and so we would like to keep these properly separated. There are a number of additional good reasons, as discussed above, although that was more related to a mega-FEniCS package. This is the same in spirit as PETSc and petsc4py.

ocaisa:

easybuild/easyconfigs/f/fenics-dolfinx-python/fenics-dolfinx-python-0.9.0-foss-2023b.eb
gkaf89: I renamed the easyconfigs to be closer to the conventions of EB.

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
```diff
@@ -0,0 +1,31 @@
...
+sources = ['v%(version)s.tar.gz']
+checksums = ['60e96b2393084729b261cb10370f0e44d12735ab3dbd1f15890dec23b9e85329']
+
+builddependencies = [
+    ('make', '4.4.1'),
+    ('CMake', '3.27.6'),
+]
+
+dependencies = [
+    ('FlexiBLAS', '3.3.1'),
```
boegel, line 21: `gfbf/2023b` would make more sense as toolchain if this requires BLAS/LAPACK?

easybuild/easyconfigs/f/FEniCS-DOLFINx-Python/FEniCS-DOLFINx-Python-0.9.0-foss-2023b.eb
```diff
@@ -0,0 +1,51 @@
...
+    ('nanobind', '2.5.0'),
+    ('pkgconfig', '1.5.5', '-python'),
+    ('scikit-build-core', '0.10.7'),
+    ('setuptools', '80.9.0'),
+]
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
+    ('cffi', '1.15.1'),
```
boegel: Isn't this a build-only requirement? (I'm not sure)

Also, there's a `cffi` already included in `Python-bundle-PyPI` (which gets pulled in via `SciPy-bundle`).

If there's a specific reason to use an older version here, I guess that's fine, but then a comment above this line to clarify why it's required would be nice

easybuild/easyconfigs/f/FEniCS-FFCx/FEniCS-FFCx-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,36 @@
...
+source_urls = ['https://github.com/FEniCS/ffcx/archive']
+sources = ['v%(version)s.tar.gz']
+checksums = ['afa517272a3d2249f513cb711c50b77cf8368dd0b8f5ea4b759142229204a448']
+
+builddependencies = [
+    ('setuptools', '80.9.0'),
+]
+
+dependencies = [
+    ('cffi', '1.15.1'),
```
boegel: Same suggestion/question as in `FEniCS-DOLFINx-Python` easyconfig

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 7 out of 7 (7 easyconfigs in total)
node4226.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/c387e849defdf2c49002203f52242347 for a full test report.

easybuild/easyconfigs/f/FEniCS-FFCx/FEniCS-FFCx-0.9.0-gfbf-2023b.eb
```diff
@@ -0,0 +1,36 @@
...
+    ('FEniCS-Basix-Python', '0.9.0'),
+    ('FEniCS-UFL', '2024.2.0'),
+    ('Python', '3.11.5'),
+    ('SciPy-bundle', '2023.11'),
```
boegel, line 21: nitpicking: move those up as first entries in `dependencies` (just a style thing, doesn't make a difference for the actual installation)

easybuild/easyconfigs/f/FEniCS-DOLFINx-Python/FEniCS-DOLFINx-Python-0.9.0-foss-2023b.eb
jhale: No, it's an optional runtime dependency, but one that is worth having. It can be removed and DOLFINx will still work. Users will then be expected to manually load it to access that functionality. I suppose this is a style/taste thing.

There should be no issue to use a more modern version, if it's available.

easybuild/easyconfigs/f/FEniCS-FFCx/FEniCS-FFCx-0.9.0-gfbf-2023b.eb
jhale: This is a required runtime dependency for FFCx. If there is a newer version it should work fine. I don't know if it should be transitive through `SciPy-bundle` - what if that changes in the future? It's up to you I suppose.

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
jhale: We discussed this above - this is the most minimal dependency set, as `gfbf/2023b` contains quite a number of unneeded things. Again, whether it's best to be minimal and explicit, or rely more on transitive dependencies, I'm unsure.

jhale:
> Does FEniCS have any test suite for these different components?
> 
> If so, it would be great if we could also make EasyBuild run those. That's especially useful when taking the next step for getting this installed in EESSI...

@gkaf89 Did some work on that.

FFCx and UFL should be easy (pytest). Basix does not have a test suite, it is tested via its Basix-Python wrapper exclusively (pytest).

DOLFINx and DOLFINx-Python are a bit more complicated, @gkaf89 and I are working on improvements targeted for the 0.11 release.

easybuild/easyconfigs/f/FEniCS-DOLFINx-Python/FEniCS-DOLFINx-Python-0.9.0-foss-2023b.eb
jhale: I'm inclined towards removing cffi - users can load it if they want (or, it can become available transitively through scipy-bundle). This is consistent with numba, another optional runtime dependency.

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
gkaf89: Yes, I prefer to be specific as well. Let's leave the easyconfig with GCC+FlexiBLAS.

easybuild/easyconfigs/f/FEniCS-DOLFINx-Python/FEniCS-DOLFINx-Python-0.9.0-foss-2023b.eb
gkaf89: Sorry, I am not very familiar with Python bindings for other languages. I leave this dependency to be imported implicitly through Python-bundle-PyPI->SciPy-bundle.

easybuild/easyconfigs/f/FEniCS-FFCx/FEniCS-FFCx-0.9.0-gfbf-2023b.eb
gkaf89: @jhale That's a good point. I remove it for the time being and I am making a note to discuss transient dependencies in the next conference call. We can always revert to the old design.

easybuild/easyconfigs/f/FEniCS-FFCx/FEniCS-FFCx-0.9.0-gfbf-2023b.eb
gkaf89: Fixed!

gkaf89:
> > Does FEniCS have any test suite for these different components?
> > If so, it would be great if we could also make EasyBuild run those. That's especially useful when taking the next step for getting this installed in EESSI...
> 
> @gkaf89 Did some work on that.
> 
> FFCx and UFL should be easy (pytest). Basix does not have a test suite, it is tested via its Basix-Python wrapper exclusively (pytest).
> 
> DOLFINx and DOLFINx-Python are a bit more complicated, @gkaf89 and I are working on improvements targeted for the 0.11 release.

Working on importing the tests for FFCx and UFL.

ocaisa:
> FFCx and UFL should be easy (pytest). Basix does not have a test suite, it is tested via its Basix-Python wrapper exclusively (pytest).

I realise this has been discussed already, but if the only way to test Basix is via the wrapper then don't we have to install the wrapper along with the package? Otherwise if a test fails, it will flag the wrapper installation as broken when in reality it is (more likely) a problem with the main installation. 

If you really want to, you can still have a separate easyconfig for the wrappers, but the main installation should ship with _a_ version just so you can run the tests.

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
jfgrimm: the `gfbf` toolchain, by definition, consists of: `GCC` + `FlexiBLAS` + `FFTW`
I wouldn't exactly call that "quite a number of unneeded things"

boegel:
> > FFCx and UFL should be easy (pytest). Basix does not have a test suite, it is tested via its Basix-Python wrapper exclusively (pytest).
> 
> I realise this has been discussed already, but if the only way to test Basix is via the wrapper then don't we have to install the wrapper along with the package? Otherwise if a test fails, it will flag the wrapper installation as broken when in reality it is (more likely) a problem with the main installation.
> 
> If you really want to, you can still have a separate easyconfig for the wrappers, but the main installation should ship with _a_ version just so you can run the tests.

You could also do a temporary installation of the Python wrappers (with `venv` & `pip` for example) so you can run the tests?
Not sure how "clean" that would be in an easyconfig file along (may need an easyblock), but would be OK imho.

jhale:
Does EasyBuild support test dependencies? Then it would be pretty straightforward - FEniCS-Basix could depend on the Python wrapper only for executing the test suite.

We have quite a number of additional test-only dependencies, which are documented here:

https://github.com/FEniCS/ufl/blob/main/pyproject.toml#L34
https://github.com/FEniCS/basix/blob/main/pyproject.toml#L27
https://github.com/FEniCS/ffcx/blob/main/pyproject.toml#L37
https://github.com/FEniCS/dolfinx/blob/v0.9.0.post1/python/pyproject.toml

jhale:
> > > FFCx and UFL should be easy (pytest). Basix does not have a test suite, it is tested via its Basix-Python wrapper exclusively (pytest).
> > 
> > 
> > I realise this has been discussed already, but if the only way to test Basix is via the wrapper then don't we have to install the wrapper along with the package? Otherwise if a test fails, it will flag the wrapper installation as broken when in reality it is (more likely) a problem with the main installation.
> > If you really want to, you can still have a separate easyconfig for the wrappers, but the main installation should ship with _a_ version just so you can run the tests.
> 
> You could also do a temporary installation of the Python wrappers (with `venv` & `pip` for example) so you can run the tests? Not sure how "clean" that would be in an easyconfig file along (may need an easyblock), but would be OK imho.

This would work OK for Basix as building the wrappers is very simple, for DOLFINx I'm not sure - the majority of our testing for DOLFINx is also done through Python, although we do have a very small C++ test suite, unlike Basix.

The underlying reason for all of this is we can write a lot more high-quality tests in Python than we could ever hope to do in C++.

easybuild/easyconfigs/f/FEniCS-Basix/FEniCS-Basix-0.9.0-GCC-13.2.0.eb
gkaf89: Still, one extra dependency is one too many.

boegel:
> Does EasyBuild support test dependencies? Then it would be pretty straightforward - FEniCS-Basix could depend on the Python wrapper only for executing the test suite.
> 
> We have quite a number of additional test-only dependencies, which are documented here:
> 
> https://github.com/FEniCS/ufl/blob/main/pyproject.toml#L34 https://github.com/FEniCS/basix/blob/main/pyproject.toml#L27 https://github.com/FEniCS/ffcx/blob/main/pyproject.toml#L37 https://github.com/FEniCS/dolfinx/blob/v0.9.0.post1/python/pyproject.toml

We make no distinction between build-only and test-only dependencies, so you can include dependencies you require to run the tests in `builddependencies`

gkaf89:
@jhale @boegel I added tests for any package where the addition is straight forward. Could we include this PR as is in 5.2?

There are modification in newer versions of FEniCS which make testing much easier, we can add more tests in future easyconfigs.

boegel:

boegel (approved):
lgtm

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 7 out of 7 (total: 8 mins 54 secs) (7 easyconfigs in total)
node3535.doduo.os - Linux RHEL 9.6, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.21
See https://gist.github.com/boegel/859e7772b5585ac45f2326945786a2af for a full test report.

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 7 out of 7 (total: 7 mins 56 secs) (7 easyconfigs in total)
node4245.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/b1af2d9e6b745c792b01ad76e768fdc4 for a full test report.

boegel:
Going in, thanks @gkaf89!

jhale:
Thanks for everyone's help on this - much appreciated. @gkaf89 and I will do 0.10 or 0.11, hopefully with better testing support, in due course.

PR #23860, opened 2025-09-15
{chem}[foss/2023a] OpenMolcas-25.06

pavelToman:
From @mboisson 
Original EasyConfig is there: https://github.com/ComputeCanada/easybuild-easyconfigs/blob/computecanada-main/easybuild/easyconfigs/o/OpenMolcas/OpenMolcas-25.06-foss-2023a.eb


pavelToman:

pavelToman:
From my attemp to install OpenMolcas with foss, there is a problem with GlobalArrays - it should not have `--enable-i8`. At least our `flexiblas` and `scalapack` seems not have ILP64 variant, only LP64, no libscalapack64, nor a flexiblas_ilp64. I can build GlobalArrays with `--enable-i8`  but I believe it is not working as it should.

PR #23875, opened 2025-09-15
Update point release for NAMD3, developers claim ..fixes critical bug…

jebalunode:
Update point release for NAMD3, developers claim .."fixes critical bugs in the Colvars module that affect not only NAMD 3.0 and 3.0.1, but also earlier NAMD 2 versions. See announce.txt for details regarding correctness issues in : eigenvector, orientationAngle, orientation, both coordNum and selfCoordNum with pairlist, and distanceZ with dynamic axis. We strongly encourage users of Colvars in ALL versions of NAMD to upgrade immediately to version 3.0.2"

- [ ] #23876


Micket:
actually, you included NAMD in the other pr already, so this PR isn't needed.

jebalunode:
The missing dependency tcsh-6.24.16-GCCcore-14.2.0.eb,  which is tcsh  6.24.16 for  GCCcore 14.2.0 has now been added.   #2387

PR #23906, opened 2025-09-18
{bio}[GCC/13.2.0] LRE-Finder v1.0.1

asp8200:
(created using `eb --new-pr`)

Needs `kma/1.6.6-GCC-13.2.0` from

- #23915


asp8200:
@Thyre : I see I also need to add a kma-module for tool-chain GCC 13.2.0 as that is a dependency for LRE-finder. I noticed that there is already an "officiel" `kma-1.4.12a-GCC-12.2.0.eb`, so perhaps I should instead make the easyconfig for the latest version kma for tool-chain GCC 13.2.0, and see if LRE-Finder/1.0.1-GCC-13.2.0 can work with that. (The docs are not clear on which version of [kma](https://bitbucket.org/genomicepidemiology/lre-finder/src/1.0.1/) it requires.)

Thyre:
I think trying with the latest version first sounds like a good idea. If it works, one can go ahead with this one, and use an older version if it doesn't.

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 1 out of 1 (total: 3 secs) (1 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), Python 3.11.14
See https://gist.github.com/Crivella/be402bf7e467bf325c731d0b2fb1393e for a full test report.

easybuild/easyconfigs/l/LRE-Finder/LRE-Finder-1.0.1-GCC-13.2.0.eb
```diff
@@ -0,0 +1,70 @@
...
+    "cat > %(installdir)s/bin/lre-finder <<'EOF'\n"
+    "#!/usr/bin/env bash\n"
+    "set -euo pipefail\n"
+    "ROOT=\"$(cd \"$(dirname \"${BASH_SOURCE[0]}\")/..\" && pwd)\"\n"
+    "DB_PREFIX=\"${LREFINDER_DB:-$ROOT/elmDB/elm}\"\n"
+    "export LREFINDER_DB=\"$DB_PREFIX\"\n"
+    "exec $(ebrootpython 2>/dev/null || command -v python3 || echo python) "
+    "\"$ROOT/LRE-Finder.py\" -t_db \"$LREFINDER_DB\" \"$@\"\n"
+    "EOF",
+    "chmod +x %(installdir)s/bin/lre-finder",
```
Crivella, line 53: Not sure the `$ROOT` here is being used properly, did you mean `$EBROOTLREMINFINDER` ?

Also isn't it easier to patch `LRE-finder.py` to initialize the `t_db` to the env variable and install it directly?

https://bitbucket.org/genomicepidemiology/lre-finder/src/fac445d190853cc90c1aed392a55102fe9df4376/LRE-Finder.py#lines-43

```python
t_db = os.getenv('LREFINDER_DB', '');
```

I think this would be preferable as we allow users to still use the upstream argument to point to a db instead of the env variable we are introducing in the ECs

easybuild/easyconfigs/l/LRE-Finder/LRE-Finder-1.0.1-GCC-13.2.0.eb
```diff
@@ -0,0 +1,70 @@
...
+]
+
+# Sanity: DB index files exist
+sanity_check_paths = {
+    'files': [
+        'bin/lre-finder',
```
Crivella, line 59: Change according to previous comment

easybuild/easyconfigs/l/LRE-Finder/LRE-Finder-1.0.1-GCC-13.2.0.eb
```diff
@@ -0,0 +1,70 @@
...
+    ('Python', '3.11.5'),
+    ('kma', '1.6.6'),
+]
+
+# MakeCp copies the whole tree; nothing extra to copy
+files_to_copy = ['*.*', '*']
+
+# Unpack DB, build KMA index, create a lightweight wrapper
+postinstallcmds = [
+    # Unpack DB if present
```
Crivella, line 30: Multiple possibilities here:
- specify that `LRE-finder.py` should end up in bin (possibly renamed)
- Add the installdir directly to PATH
- Use the postinstallcmd to move the file in bin after the copy

Also i do not think any other EC uses `MakeCp` to copy all files, we should be specific on what needs to be copied

easybuild/easyconfigs/l/LRE-Finder/LRE-Finder-1.0.1-GCC-13.2.0.eb
```diff
@@ -0,0 +1,70 @@
...
+    'files': [
+        'bin/lre-finder',
+    ],
+    'dirs': ['elmDB'],
+}
+
+# Put wrapper on PATH; users can also run python $EBROOTLREFINDER/LRE-Finder.py manually
+modextrapaths = {'PATH': ['bin']}
+
+# Make the DB prefix discoverable if users don’t use the wrapper
```
Crivella, line 67: I think `LREFINDER_DB` only has meaning for the wrapper? Atleast i cannot seem to find any reference to it in the upstream code

PR #23702, opened 2025-08-18
{data}[gompi/2025b] HDF5 v1.14.6

branfosj:
(created using `eb --new-pr`)

* [x]  #23700

We should switch to using `CMake` to build `HDF5`:
* The configure / make code is going away (1.16?) and has not been updated in years: https://github.com/HDFGroup/hdf5/pull/5358
* This PR deliberately does not use the easyblock, as I think this easyconfig is simple enough without it.

Notes on this:
* `libaec` is checked for in the CMake install instead of `Szip`.
* The CMake install provides a `.pc` per HDF5 component and these are much improved over the single `.pc` file we currently generate in the easyblock.
* As far as I can tell the `postinstallcmds` we currently have in the `ConfigureMake` installs are not needed.
* This moves all the sanity check from the easyblock into the easyconfig.
* `h5redeploy` is installed here via `postinstallcmds`, but this script will go away in the future: https://github.com/HDFGroup/hdf5/pull/5486

PR #23952, opened 2025-09-22
{data}[GCC/14.3.0-serial] HDF5 v1.14.6

AlvaroHuanay:
HDF5 1.14.6 with GCC 14.3.0 (serial build).

### DIFF
--- a/easybuild/easyconfigs/h/HDF5/HDF5-1.14.6-GCC-14.3.0-serial.eb
+++ b/easybuild/easyconfigs/h/HDF5/HDF5-1.14.6-GCC-14.3.0-serial.eb
@@ -7,7 +7,7 @@ description = """HDF5 is a data model, library, and file format for storing and 
  It supports an unlimited variety of datatypes, and is designed for flexible
  and efficient I/O and for high volume and complex data."""
 
-toolchain = {'name': 'GCCcore', 'version': '14.3.0'}
+toolchain = {'name': 'GCC', 'version': '14.3.0'}
 toolchainopts = {'pic': True}
 
 


Thyre:
This should probably be in `GCC` as the toolchain due to the HDF5 Fortran interface.
Fortran module files are typically not compatible between Fortran compilers.

AlvaroHuanay:
> This should probably be in `GCC` as the toolchain due to the HDF5 Fortran interface. Fortran module files are typically not compatible between Fortran compilers.

Change applied

Thyre:
There are quite a few differences compared to the MPI variant we already have in this toolchain.
@branfosj spent some time looking into this.

I'm unfortunately not familiar with building HDF5 at all, and I've used it the last time three years ago.
So I cannot really tell what is needed and what is not.

branfosj:
> There are quite a few differences compared to the MPI variant we already have in this toolchain. @branfosj spent some time looking into this.

In #23702 I switched to using CMake for the `2025b` generation builds of `HDF5`. This will be included in the next EasyBuild release. The motivation for this switch is explained in the PR.

I've opened #23959 with a serial build of `HDF5`, which should build a serial `HDF5` that is equivalent to this PR. Please test that build and software built using it.

AlvaroHuanay:
Hi @branfosj,

Thank you for opening #23959. On my side I already tested it with other dependencies that require it (like h5py and Qiskit) which worked.

@Thyre, should I close this PR? checks were succeded in #23959.

Thyre:
> @Thyre, should I close this PR? checks were succeded in #23959.

Yeah, I think the PR supersedes this one.

AlvaroHuanay:
This PR is not longer required. #23959 will be used.

PR #23969, opened 2025-09-23
improve reported errors in easyconfigs test suite

Flamefire:
- Use the new `return_conflicts` parameter to show the conflicts in the error message instead of printing it to stderr during the status output when the test is run.
- [x] Requires https://github.com/easybuilders/easybuild-framework/pull/5000
- Especially "binutils or GCC is a build dep: " is a misleading message when the test fails because it is not.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -330,7 +330,7 @@ class EasyConfigTest(TestCase):
 
         self.assertEqual(check_conflicts(self.ordered_specs, modules_tool(), check_inter_ec_conflicts=False,
                                          return_conflicts=True),
-                         "")
+                         [])
 
     def test_deps(self):
         """Perform checks on dependencies in easyconfig files"""


Flamefire:
@boegel Can we get this in please? Found another instance where this would have helped: https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/21939459431/job/63361191232?pr=25045

boegel:
@Flamefire This needs another extra tiny change because of change in return value?
```
FAIL: test_conflicts (test.easyconfigs.easyconfigs.EasyConfigTest.test_conflicts)
Check whether any conflicts occur in software dependency graphs.
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 331, in test_conflicts
    self.assertEqual(check_conflicts(self.ordered_specs, modules_tool(), check_inter_ec_conflicts=False,
AssertionError: [] != ''
```

Or maybe just `assertFalse` ?

test/easyconfigs/easyconfigs.py
```diff
@@ -328,8 +328,9 @@ def test_conflicts(self):
             print("(skipped conflicts test)")
             return
 
-        self.assertFalse(check_conflicts(self.ordered_specs, modules_tool(), check_inter_ec_conflicts=False),
-                         "No conflicts detected")
+        self.assertEqual(check_conflicts(self.ordered_specs, modules_tool(), check_inter_ec_conflicts=False,
+                                         return_conflicts=True),
+                         "")
```
Flamefire:
```suggestion
                         [])
```

Flamefire:
Yes, should be an empty list

PR #20745, opened 2024-06-05
add patches for failing LAPACK tests and RISC-V test segfaults to OpenBLAS 0.3.27

bedroge:
(created using `eb --new-pr`)

PR #19280, opened 2023-11-21
add patches for OpenBLAS v0.3.21 to disable -ftree-vectorize for netlib LAPACK

juntangc:
https://github.com/EESSI/software-layer/blob/2023.06/eessi-2023.06-known-issues.yml

PR #16406, opened 2022-10-14
enable running of LAPACK tests for recent OpenBLAS easyconfigs + add patch to fix failing LAPACK tests due to use of -ftree-vectorize

boegel:
(created using `eb --new-pr`)

cfr. https://github.com/easybuilders/easybuild-easyconfigs/issues/16380

requires:
* https://github.com/easybuilders/easybuild-easyblocks/pull/2801
* https://github.com/easybuilders/easybuild-easyconfigs/pull/16411

~~By disabling the use of `-ftree-vectorize`, failing LAPACK tests are dropping from:~~

<details>

```
                        -->   LAPACK TESTING SUMMARY  <--
SUMMARY                 nb test run     numerical error         other error
================        ===========     =================       ================
REAL                    1294329         1226    (0.095%)        0       (0.000%)
DOUBLE PRECISION        1302917         1197    (0.092%)        0       (0.000%)
COMPLEX                 756180          1208    (0.160%)        0       (0.000%)
COMPLEX16               761792          450     (0.059%)        0       (0.000%)

--> ALL PRECISIONS      4115218         4081    (0.099%)        0       (0.000%)
```

to

```
                        -->   LAPACK TESTING SUMMARY  <--
SUMMARY                 nb test run     numerical error         other error
================        ===========     =================       ================
REAL                    1308369         1       (0.000%)        0       (0.000%)
DOUBLE PRECISION        1316957         0       (0.000%)        0       (0.000%)
COMPLEX                 768036          38      (0.005%)        0       (0.000%)
COMPLEX16               768848          94      (0.012%)        0       (0.000%)

--> ALL PRECISIONS      4162210         133     (0.003%)        0       (0.000%)
```

(results collected on an AMD Rome system)

</details>

edit: thanks to patch from @bartoldeman (added in bb0269a693) + patched GCC (see #16411), the number of failing LAPACK tests is now down to a handful...

PR #23974, opened 2025-09-23
add patch for FlexiBLAS v3.3.1 to fix LAPACK errors triggered by `-ftree-vectorize` on zen4/zen5 microarchs

arielzn:
(created using `eb --new-pr`)

With the modification of the bundle.py easyblock to include the tests of components introduced on eb v5.1.1 [easyblocks PR#3748](https://github.com/easybuilders/easybuild-easyblocks/pull/3748) we started hitting errors when testing the install of FlexiBLAS 3.3.1 on zen4/zen5.

It segfaults as:

```
 Running: /dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/easybuild_obj/test/lapack-3.11.0/LIN/xlintstz
ARGS= OUTPUT_FILE;/dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/easybuild_obj/test/lapack-3.11.0/ztest.out;ERROR_FILE;/dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/easybuild_obj/test/lapack-3.11.0/ztest.out.err;INPUT_FILE;/dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/flexiblas-3.3.1/test/lapack-3.11.0/ztest.in
Test OUTPUT:

Test ERROR:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x14e1021fe5af in ???
#1  0x14e101e53bcc in zgbtrf_
        at /dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/flexiblas-3.3.1/contributed/lapack-3.11.0/SRC/zgbtrf.f:429
#2  0x4225a1 in zchkgb_
        at /dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/flexiblas-3.3.1/test/lapack-3.11.0/LIN/zchkgb.f:466
#3  0x41c26b in zchkaa
        at /dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/flexiblas-3.3.1/test/lapack-3.11.0/LIN/zchkaa.F:488
#4  0x40844e in main
        at /dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/flexiblas-3.3.1/test/lapack-3.11.0/LIN/zchkaa.F:1271

CMake Error at /dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/flexiblas-3.3.1/test/lapack-3.11.0/runtest.cmake:51 (message):
  Test
  /dev/shm/alozano/ebbuild/FlexiBLAS/3.3.1/GCC-12.3.0-test/easybuild_obj/test/lapack-3.11.0/LIN/xlintstz
  returned Segmentation fault

        Start  93: LAPACK-xlintstrfz_ztest_rfp_in
```

Applying the `-fno-tree-vectorize` option for the netlib LAPACK build as proposed in this  PR fixes the issue.

We've hit this kind of issues on the past for lapack code as in  https://github.com/easybuilders/easybuild-easyconfigs/pull/19280 

I don't know if we can go for disabling tree-vectorize fully as proposed with the patch, or for a more targeted fix, patching the lapack code as it was done on these https://github.com/easybuilders/easybuild-easyconfigs/pull/20745  https://github.com/easybuilders/easybuild-easyconfigs/pull/16406


arielzn:
I guess @bartoldeman  input on this would be nice 

For more context, I tested to backport FlexiBLAS 3.4.4 (which builds and tests fine with tree-vectorize on 2024a) into 2023a and it also fails on tests with the same kind of errors.

So the issue does not seem to be the FlexiBLAS/Lapack version, but that the GCC versions on 2023a,b is not producing proper vectorized builds of lapack code for zen4/zen5.

bartoldeman:
I'll have to look at this carefully. I'd prefer a patch but using `-ffp-contract=off` may do the trick more cheaply, as FMA is where the differences often come from, not vectorization, which modulo compiler bugs gives the same result (GCC does not do reductions using vectors by default, which would also affect it; for that you would need `-fassociative-math` which is off by default).

So please try `-ffp-contract=off` for now.

bartoldeman:
https://github.com/Reference-LAPACK/lapack/pull/1033 disabled FMAs in one place using strategic parentheses.
But it's not in any released FlexiBLAS yet.

boegel:
@bartoldeman I tried changing the patch to use `-ffp-contract=off` instead of `-fno-tree-vectorize`, and that doesn't fix the segfault, `LAPACK-xlintstz_ztest_in` fails with:
```
Program received signal SIGSEGV: Segmentation fault - invalid memory reference
```

boegel:
CORE_CNT=16
EB_ARGS="--installpath /tmp/$USER/pr23974"

boegel:
Test report by @boegel
**SUCCESS**
node4210.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/167a1893ff1563970ca19ebb5156d924 for a full test report.

akesandgren:
I re-iterate my standpoint from way back, lapack's test code and matgen MUST be built without optimization of any kind and with -O0 since that code isn't written to handle compiler optimizations.
That's where most of the problems with lapack tests comes from.
If you do that then one can most likely keep the vectorization flag.

easybuild/easyconfigs/f/FlexiBLAS/FlexiBLAS-3.3.1-GCC-12.3.0.eb
```diff
@@ -38,7 +38,14 @@ components = [
...
-        'checksums': ['bbeebf5e5a006924558fec43f49affbe1aaa4cbacfc472a9ff6066ffda142e18'],
+        'patches': [
+            'FlexiBLAS-3.3.1_disable-lapack-tree-vectorize.patch',
+        ],
+        'checksums': [
+            'bbeebf5e5a006924558fec43f49affbe1aaa4cbacfc472a9ff6066ffda142e18',  # flexiblas-3.3.1.tar.gz
+            'd05e9d48255299428f3a8e2127bab3a3f80fa6de99340cfce34b6ab09f9b5884',
+            # FlexiBLAS-3.3.1_disable-lapack-tree-vectorize.patch
+        ],
         'backends': local_backends,
```
bartoldeman, line 47: based on slack discussion we can fix this using an arch specific block + `-mprefer-vector-width=256`

```
+if ARCH == 'x86_64':
+    components[0][2]['patches'] = 'FlexiBLAS-3.3.1_lapack-prefer-vector-width-256.patch'
+    components[0][2]['checksums'].append('deadbeef')
```

then change the patch to use `-mprefer-vector-width=256` instead of `-fno-tree-vectorize`

bartoldeman:
A little bit more digging:
the SIGSEGV was caused by an unaligned memory access with an aligned move (avx2 `vmovapd` used on 16-byte, not 32-byte aligned memory). It looks like an issue with GCC 12.3 and 13.2, it's not present with 13.3.

```
diff -ur flexiblas-3.3.1.orig/contributed/lapack-3.11.0/SRC/zgbtrf.f flexiblas-3.3.1/contributed/lapack-3.11.0/SRC/zgbtrf.f
--- flexiblas-3.3.1.orig/contributed/lapack-3.11.0/SRC/zgbtrf.f	2023-04-03 14:07:31.000000000 +0000
+++ flexiblas-3.3.1/contributed/lapack-3.11.0/SRC/zgbtrf.f	2025-09-24 15:29:42.138091646 +0000
@@ -169,8 +169,8 @@
       COMPLEX*16         TEMP
 *     ..
 *     .. Local Arrays ..
-      COMPLEX*16         WORK13( LDWORK, NBMAX ),
-     $                   WORK31( LDWORK, NBMAX )
+      COMPLEX*16, ALLOCATABLE :: WORK13( :, : ),
+     $                   WORK31( :, : )
 *     ..
 *     .. External Functions ..
       INTEGER            ILAENV, IZAMAX
@@ -232,6 +232,8 @@
 *
 *        Use blocked code
 *
+         ALLOCATE(WORK13(LDWORK, NBMAX), WORK31(LDWORK, NBMAX))
+*
 *        Zero the superdiagonal elements of the work array WORK13
 *
          DO 20 J = 1, NB
```
allocates the relevant matrices which bypasses the miscompiled code here, so that's a source code based workaround.

arielzn:
> A little bit more digging: the SIGSEGV was caused by an unaligned memory access with an aligned move (avx2 `vmovapd` used on 16-byte, not 32-byte aligned memory). It looks like an issue with GCC 12.3 and 13.2, it's not present with 13.3.
> 
> ...
> allocates the relevant matrices which bypasses the miscompiled code here, so that's a source code based workaround.

ok, i was just testing your other proposed patch, so we prefer this source code modif then, right ?

bartoldeman:
> > A little bit more digging: the SIGSEGV was caused by an unaligned memory access with an aligned move (avx2 `vmovapd` used on 16-byte, not 32-byte aligned memory). It looks like an issue with GCC 12.3 and 13.2, it's not present with 13.3.
> > ...
> > allocates the relevant matrices which bypasses the miscompiled code here, so that's a source code based workaround.
> 
> ok, i was just testing your other proposed patch, so we prefer this source code modif then, right ?

yes I'd prefer the source code modification as it is more targeted.

bartoldeman (approved):
lgtm

arielzn:
I changed the fix in order to use the patch lapack code option.

@bartoldeman  for the `FlexiBLAS-3.3.1-GCC-12.3.0.eb` case your patch was enough, but when applying the same to `FlexiBLAS-3.3.1-GCC-13.2.0.eb`  i've still got a segfault. In this case the trace pointed to another lapack subroutine `ZPBTRF`. It seems gcc 12.3 and 13.2 still do some slightly different miscompilations.

I applied the same logic of making the work array dynamic on this other routine and it worked, so we have different patches for each toolchain.

arielzn:
finally this was fixed patching a gcc bug:
* #24073

PR #23975, opened 2025-09-23
use generic version of RAxML v8.2.13 as dependency for bcgTree and Parsnp

ocaisa:
A custom version of RAxML with AVX2 optimisation was being used as a dependency in some places. This doesn't work on non-`x86_64` so rather switch to a more generic build **but** using the EasyBuild `CFLAGS` which allows the compiler to use native CPU optimisations.

If there is a concern about performance loss here, `RAxML-NG` should really be used anyway in that scenario.

### DIFF
--- a/easybuild/easyconfigs/r/RAxML/raxml_8.2.13_respect_eb_settings.patch
+++ b/easybuild/easyconfigs/r/RAxML/raxml_8.2.13_respect_eb_settings.patch
@@ -21,7 +21,7 @@ diff -rupN standard-RAxML-8.2.13/Makefile.HYBRID.gcc standard-RAxML-8.2.13_new/M
  
 -CFLAGS = -D_WAYNE_MPI -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops -O2  -D_USE_PTHREADS $(ARCH_CFLAGS)
 +# Ignore ARCH_CFLAGS and use EasyBuild CFLAGS by default (i.e. just keep things that are not commonly set)
-+CFLAGS = -D_WAYNE_MPI -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops  -D_USE_PTHREADS 
++CFLAGS += -D_WAYNE_MPI -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops  -D_USE_PTHREADS 
  
  
 -LIBRARIES = -lm -pthread
@@ -53,7 +53,7 @@ diff -rupN standard-RAxML-8.2.13/Makefile.MPI.gcc standard-RAxML-8.2.13_new/Make
  
 -CFLAGS = -D_WAYNE_MPI -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops -O2 $(ARCH_CFLAGS)
 +# Ignore ARCH_CFLAGS and use EasyBuild CFLAGS by default (i.e. just keep things that are not commonly set)
-+CFLAGS = -D_WAYNE_MPI -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops
++CFLAGS += -D_WAYNE_MPI -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops
  
  
  LIBRARIES = -lm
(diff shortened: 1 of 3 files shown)


ocaisa:

easybuild/easyconfigs/r/RAxML/RAxML-8.2.13-gompi-2023a-avx2.eb
```diff
@@ -1,39 +0,0 @@
...
-toolchain = {'name': 'gompi', 'version': '2023a'}
-toolchainopts = {'usempi': True}
-
-source_urls = ['https://github.com/stamatak/standard-RAxML/archive/']
-sources = ['v%(version)s.tar.gz']
-checksums = ['28e500793324bd7d330b396ef27ea49c9186fa5e1edb3d5439036dc6c33e6067']
-
-buildopts = '-f Makefile.AVX2.gcc CC="$CC" && rm *.o && '
-buildopts += 'make -j %(parallel)s -f Makefile.AVX2.PTHREADS.gcc CC="$CC" && rm *.o && '
-buildopts += 'make -j %(parallel)s -f Makefile.AVX2.HYBRID.gcc CC="$CC"'
```
boegel, line 19: @ocaisa Did you check how these Makefiles differ from `Makefile.PTHREADS.gcc` and `Makefile.HYBRID.gcc`?

Is it really only changes in compiler options?

ocaisa: Yes I checked them and there are two things done differently:
```
avxLikelihood.o : avxLikelihood.c $(GLOBAL_DEPS)
		$(CC) $(CFLAGS) -mavx2 -D_FMA -march=core-avx2 -c -o avxLikelihood.o avxLikelihood.c

fastDNAparsimony.o : fastDNAparsimony.c $(GLOBAL_DEPS)
		   $(CC) $(CFLAGS) -mavx -c -o fastDNAparsimony.o fastDNAparsimony.c
```
The second is only an additional compilation flag.

The first is a complete file with optimisations which gets picked up via a make flag. However, that file is 10+ years old, I'm sure it does what is was intended for (AVX2 optimisations) but if you are really looking for performance RAxML-NG claims to have the upper hand.

ocaisa: I don't have any way to test the performance between the two builds, but I really wonder if it is worth our while to introduce dependency branching for something that (right now) doesn't even use our standard compiler optimisations.

boegel:
I totally forgot about @lexming's easyblock PR that tackles the same problem in a different way:
* https://github.com/easybuilders/easybuild-easyblocks/pull/2180

ocaisa:

ocaisa:
This is (I believe) more correct than what was there before. We now build 4 versions of the package:
* serial
* threaded
* mpi-enabled
* hybrid (MPI/threaded)

boegel:
@lexming Can you take a look at this?

easybuild/easyconfigs/r/RAxML/raxml_8.2.13_respect_eb_settings.patch
```diff
@@ -0,0 +1,128 @@
...
++# Ignore ARCH_CFLAGS and use EasyBuild CFLAGS by default (i.e. just keep things that are not commonly set)
++CFLAGS = -D_GNU_SOURCE -fomit-frame-pointer -funroll-loops -D_USE_PTHREADS #-Wall -Wunused-parameter -Wredundant-decls  -Wreturn-type  -Wswitch-default -Wunused-value -Wimplicit  -Wimplicit-function-declaration  -Wimplicit-int -Wimport  -Wunused  -Wunused-function  -Wunused-label -Wno-int-to-pointer-cast -Wbad-function-cast  -Wmissing-declarations -Wmissing-prototypes  -Wnested-externs  -Wold-style-definition -Wstrict-prototypes  -Wpointer-sign -Wextra -Wredundant-decls -Wunused -Wunused-function -Wunused-parameter -Wunused-value  -Wunused-variable -Wformat  -Wformat-nonliteral -Wparentheses -Wsequence-point -Wuninitialized -Wundef -Wbad-function-cast
```
ocaisa: Argh, this should be `CFLAGS +=`

ocaisa: Fixed in https://github.com/easybuilders/easybuild-easyconfigs/pull/23975/commits/1e8c66f2eb480c958ecbf8e58902195d6d51b035

ocaisa:

lexming:
Merging, thanks @ocaisa !

PR #23991, opened 2025-09-24
remove deprecated license classifier in `setup.py`

boegel:
see also https://github.com/easybuilders/easybuild/pull/922


smoors (approved):
lgtm

PR #23994, opened 2025-09-25
add patch for FlexiBLAS v3.2.1 to fix LAPACK errors triggered by -ftree-vectorize on zen4/zen5 microarchs

arielzn:
(created using `eb --new-pr`)

this is the same fix we converged for in #23974 backported for v3.2.1


arielzn:
finally this was fixed patching a gcc bug, see #24073

PR #24001, opened 2025-09-25
{tools} [GCCcore/13.3.0, GCCcore/13.2.0] JupyterLab-git v.0.51.2, JupyterLab-git v.0.51.1

moravveji:
This PR adds two versions of `jupyterlab-git` plugin for JupyterLab for 2023b and 2024a toolchains. A similar installation based on the modules in 2023a fails with some `hatchling` issues.

### DIFF
--- a/easybuild/easyconfigs/j/JupyterLab-git/JupyterLab-git-0.51.1-GCCcore-13.2.0.eb
+++ b/easybuild/easyconfigs/j/JupyterLab-git/JupyterLab-git-0.51.1-GCCcore-13.2.0.eb
@@ -1,0 +1,67 @@
+# Author: Ehsan Moravveji (VSC - KU Leuven)
+
+easyblock = 'PythonBundle'
+
+name = 'JupyterLab-git'
+version = '0.51.1'
+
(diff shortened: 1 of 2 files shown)


moravveji:
@Thyre : not sure how important the labels are and how they are consumed in the CI/CD/test steps, but the two proposed easyconfigs should be better labeled as 'new' instead of 'update'. Again, not sure if it even matters at all.

Thyre:
> @Thyre : not sure how important the labels are and how they are consumed in the CI/CD/test steps, but the two proposed easyconfigs should be better labeled as 'new' instead of 'update'. Again, not sure if it even matters at all.

The labels are mostly just there to keep an overview. I'm not sure if they're used in the release process to make it easier to sort all the PRs :smile: 

I changed the labels. Don't know if the GitHub bot changes them to https://github.com/easybuilders/easybuild-easyconfigs/labels/update with the next push again :sweat_smile: 
There's certainly some things we can try to improve with the labels.

Micket (approved):
lgtm

Micket:

PR #24073, opened 2025-10-01
add patch for GCCcore 10.3-13.2 to fix AVX-512 vectorizer bug (upstream PR 114566)

bartoldeman:
This fixes a segmentation fault in FlexiBLAS LAPACK tests when compiling for Zen4/Zen5

See also:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114566 and:
* #23974

### DIFF
--- a/easybuild/easyconfigs/g/GCCcore/GCCcore-10.1-11.4_fix-avx512-vect-crash.patch
+++ b/easybuild/easyconfigs/g/GCCcore/GCCcore-10.1-11.4_fix-avx512-vect-crash.patch
@@ -1,0 +1,120 @@
+From 327533790760809abd2549e613a676dde5f8cd93 Mon Sep 17 00:00:00 2001
+From: Jakub Jelinek <jakub@redhat.com>
+Date: Fri, 5 Apr 2024 14:56:14 +0200
+Subject: vect: Don't clear base_misaligned in update_epilogue_loop_vinfo
+ [PR114566]
+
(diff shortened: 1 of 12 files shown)


bartoldeman:
still need to double check with 13.2 (I know the patch applies but haven't confirmed flexiblas)

bartoldeman:

bartoldeman:
Confirmed that this works with unpatched `FlexiBLAS-3.3.1-GCC-12.3.0.eb` and `FlexiBLAS-3.3.1-GCC-13.2.0.eb` with `-march=znver4`

arielzn:
I tested to rebuild GCCcore from here for 13.2.0, 12.3.0 and 12.2.0 

they all worked to build the current unpatched FlexiBLAS easyconfigs, so I guess we can close #23974 and #23994

easybuild/easyconfigs/g/GCCcore/GCCcore-10.1-11.4_fix-avx512-vect-crash.patch
```diff
@@ -0,0 +1,120 @@
+From 327533790760809abd2549e613a676dde5f8cd93 Mon Sep 17 00:00:00 2001
```
boegel, line 1: I've opened a follow-up PR to add a reference to the GCC bug in these patch files:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/24157

Issue #24026, opened 2025-09-26
libreadline "Couldn't find file readline-x.x.tar.gz anywhere,..." also for other versions


baverhey:
https://tiswww.case.edu/php/chet/readline/rltop.html#TOCCurrentStatus
new paths: 
https://cgit.git.savannah.gnu.org/cgit/readline.git/
ftp://ftp.cwru.edu/pub/bash/readline-8.3.tar.gz

ERROR: Installation of libreadline-8.2-GCCcore-12.3.0.eb failed: "Couldn't find file readline-8.2.tar.gz anywhere, 
 and downloading it didn't work either... Paths attempted (in order):easybuild/easybuild/easyconfigs/l/libreadline/l/libreadline/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/l/libreadline/libreadline/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/l/libreadline/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/l/libreadline/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/libreadline/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/readline-8.2.tar.gz, 
 /tmp/eb-0_dh6hcs/files_pr19678/l/libreadline/readline-8.2.tar.gz, 
 /tmp/eb-0_dh6hcs/files_pr19678/libreadline/readline-8.2.tar.gz, 
 /tmp/eb-0_dh6hcs/files_pr19678/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/l/libreadline/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/libreadline/readline-8.2.tar.gz, 
easybuild/easybuild/easyconfigs/readline-8.2.tar.gz, 
 https://ftp.gnu.org/gnu/readline/readline-8.2.tar.gz, 
 https://sources.easybuild.io/l/libreadline/readline-8.2.tar.gz "

todo: make pr/test

graham-cs:
Encountered this issue as well and used mirror https://ftpmirror.gnu.org/readline/readline-8.2.tar.gz per https://directory.fsf.org/wiki/Readline

Micket:
Yes I think 
https://ftpmirror.gnu.org/readline/readline-8.2.tar.gz
is the better choice here. 

[...]

PR #24102, opened 2025-10-02
updated the new download url withoud access restrictions libreadline-8.2-GCCcore-12.3.0.eb

baverhey:
Pr for https://github.com/easybuilders/easybuild-easyconfigs/issues/24026#issuecomment-3357305585


Micket:
I'm going to suggest we use https://github.com/easybuilders/easybuild-easyconfigs/pull/24108 instead. It's (apart from the libreadline vs readline name) what we use for almost all other gnu software via GNU_SOURCE. Also, i apply it to all easyconfigs.

PR #24130, opened 2025-10-06
{tools}[gccore/13.2.0] Neovim 0.11.4

adammccartney:


easybuild/easyconfigs/n/neovim/neovim-0.11.4-gcccore-13.2.0.eb
```diff
@@ -0,0 +1,32 @@
+easyblock = 'CMakeMake'
+
+name = 'neovim'
```
jfgrimm: documentation refers to it as `Neovim`, so lets use that name
```suggestion
name = 'Neovim'
```

jfgrimm: the filename would need updating to match the change. It already needs changing anyway, since `gcccore` should be capitalised in the filename also (`GCCcore`)
If you're using the EB github integreation, you can locally make the changes, and update the PR with e.g.:
`eb --update-pr 24130 :neovim-0.11.4-gcccore-13.2.0.eb Neovim-0.11.4-GCCcore-13.2.0.eb` (which will treat it as a rename)

easybuild/easyconfigs/n/neovim/neovim-0.11.4-gcccore-13.2.0.eb
```diff
@@ -0,0 +1,32 @@
...
+    'git_config': {
+            "url": "https://github.com/neovim",
+            "repo_name": "neovim",
+            "commit": "cec0ecabd8f47ff81dcb52e8fc9003e365563a84",
+            'recursive': True,
+            'keep_git_dir': True,
+    },
+}]
+
+builddependencies = [('CMake', '3.27.6')]
```
jfgrimm: `binutils` should be a build-time dependency, when compiling software with the `GCCcore` toolchain
```suggestion
builddependencies = [
    ('binutils', '2.40'),
    ('CMake', '3.27.6'),
]
```

easybuild/easyconfigs/n/Neovim/Neovim-0.11.4-GCCcore-13.2.0.eb
```diff
@@ -0,0 +1,32 @@
...
+            "commit": "cec0ecabd8f47ff81dcb52e8fc9003e365563a84",
+            'recursive': True,
+            'keep_git_dir': True,
+    },
+}]
+
+builddependencies = [('CMake', '3.27.6')]
+
+skipsteps = ['configure']
+build_cmd = 'make CMAKE_BUILD_TYPE=Release CMAKE_EXTRA_FLAGS="-DCMAKE_INSTALL_PREFIX=%(installdir)s"'
```
jfgrimm, line 22: this looks odd to me -- usually we want to run the cmake configure ourselves (since then EasyBuild has more control over the build, and we can be more confident that options EB sets are respected)

easybuild/easyconfigs/n/neovim/neovim-0.11.4-gcccore-13.2.0.eb
```diff
@@ -0,0 +1,32 @@
...
+sources = [{
+    'filename': 'neovim-0.11.4.tar.gz',
+    'git_config': {
+            "url": "https://github.com/neovim",
+            "repo_name": "neovim",
+            "commit": "cec0ecabd8f47ff81dcb52e8fc9003e365563a84",
+            'recursive': True,
+            'keep_git_dir': True,
+    },
+}]
```
jfgrimm: do we really need:
* a recursive clone
* to keep the `.git` dir (which forces us to disable checksums)

It's generally preferable to use release source tarballs, since we can checksum these

easybuild/easyconfigs/n/Neovim/Neovim-0.11.4-GCCcore-13.2.0.eb
adammccartney: Yeah, I think maybe it's something of a peculiarity of the neovim build. I actually haven't looked that closely, but I think the wrap cmake config in a makefile. I don't know if this was for backward compatibility with vim's build system (which I think just uses plain make). The core is written in C, and seems to be fairly portable as a result (built on an arm machine at the weekend with no issues). 
I can look into it in a bit more detail if you like. There definitely used to be various optimizations that you could build vim with, I'm sure neovim allows for some of the same configuration.

easybuild/easyconfigs/n/neovim/neovim-0.11.4-gcccore-13.2.0.eb
adammccartney: Thanks for the feedback, I'll use the source tarball / checksum approach.

boegel:

jfgrimm:
Test report by @jfgrimm
**SUCCESS**
node064.viking2.yor.alces.network - Linux Rocky Linux 8.10, x86_64, AMD EPYC 7643 48-Core Processor, Python 3.6.8
See https://gist.github.com/jfgrimm/58d670f6ce0bc87863fe3456eaee2145 for a full test report.

jfgrimm:
Going in, thanks @adammccartney!

PR #24223, opened 2025-10-13
remove Kent_tools from whitelist in test that requires checksums

lexming:


branfosj:
There are six `Kent_tools` easyconfigs currently and all have checksums, so this definitely makes sense to require them!

PR #24273, opened 2025-10-16
adding easyconfigs: glslang-SPIRV-16.0.0-GCCcore-14.3.0.eb

garadar:


garadar:
Didn't see an older version was compiled with same toolchain :/

Thyre:
Theoretically, we could still update the package in `2025b`, as there was no release with that yet.
If we would do that, we would also need to update OpenGL, and test building the existing packages on top to ensure nothing breaks.

Don't know if that's worth the effort though.

Crivella:
> Don't know if that's worth the effort though.

Most likely not.

> Didn't see an older version was compiled with same toolchain :/

@garadar 
Is there any new feature that would warrant having 2 separate major versions in the same TC?

PR #24288, opened 2025-10-17
fix `python-igraph` easyconfigs to actually use `igraph` dependency rather than build one + update easyconfig for `synthcity` 0.2.11 accordingly (changed version for `ipython-igraph` dependency)

pavelToman:
(created using `eb --new-pr`)

`python-igraph` should use `igraph` as depepndency, but I it does not now and compiles its own `_igraph.abi3.so` which is not linked to the igraph's `libigraph.so`. Thus it needs CMake (and Clang?) during the installation.

In this PR I fix this in all recent `python-igraph` easyconfigs and `python-igraph` uses `igraph` as depepndency, thus it does not need the CMake and Clang in deps anymore.

The problem is with version for foss/2023b, where the `python-igraph` version is too new for `igraph` version, thus I change the version for this toolchain. It is used as a dependency only in `synthcity-0.2.11-foss-2023b.eb` where it is as test dependency, not runtime. I am gonna update this Synthcity either


pavelToman:

Micket (commented):
I can't make any sense of this PR. 

You are saying

> Thus it needs CMake (and Clang?) during the installation.

but, this PR *removes* CMake and clang etc. ???

pavelToman:
> I can't make any sense of this PR.
> 
> You are saying
> 
> > Thus it needs CMake (and Clang?) during the installation.
> 
> but, this PR _removes_ CMake and clang etc. ???

`python-igraph` should use `igraph` as depepndency, but now it does not, thus it needs the CMake and Clang now - it compiles its own `igraph` instead of use installed one.
With my update the `python-igraph` uses `igraph` as a dependency, thus it does not need the CMake and Clang in deps anymore.

But you are right - the description is confusing, I update it.

boegel:
CORE_CNT=16

boegel:
Test report by @boegel
**SUCCESS**
node3618.doduo.os - Linux RHEL 9.6, x86_64, AMD EPYC 7552 48-Core Processor (zen2), Python 3.9.21
See https://gist.github.com/boegel/daf54639e969ca69df4ccc125e838bbd for a full test report.

boegel:
Going in, thanks @pavelToman!

PR #24289, opened 2025-10-17
Fix source distribution CI check

Flamefire:
Due to a recent change in setuptools and/or distutils `python setup.py sdist` normalizes filenames by replacing dashes by underscores.
We need to detect this and refer to the file with or without underscore.

Also put output into groups instead of silencing it

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -170,8 +170,14 @@ jobs:
     - name: Inspect files included in source distribution
       working-directory: dist
       run: |
-          tar xfz easybuild-easyconfigs*tar.gz
-          cd easybuild-easyconfigs-*/
+          ls
+          if [ -f easybuild-easyconfigs*tar.gz ]; then
+            tar xzf easybuild-easyconfigs*tar.gz
+            cd easybuild-easyconfigs-*/
+          else
+            tar xzf easybuild_easyconfigs*tar.gz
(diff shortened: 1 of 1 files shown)


Micket (dismissed):
lgtm

.github/workflows/unit_tests.yml
```diff
@@ -120,9 +120,18 @@ jobs:
...
           # install easyconfigs via distribution package
-          python setup.py sdist > /dev/null
+          echo "::group::Create source distribution"
+          python setup.py sdist
           ls dist
-          pip install dist/easybuild-easyconfigs*tar.gz > /dev/null
+          echo "::endgroup::"
+          echo "::group::Install from source distribution"
+          # Depending on the version of setuptools/distutils the dash will be replaced by underscore
+          if [ -f dist/easybuild-easyconfigs*tar.gz ]; then
```
Micket, line 129:
```suggestion
          if compgen -G 'dist/easybuild-easyconfigs*tar.gz' > /dev/null; then
```
maybe..?

Flamefire:
There was a similar issue with the tar command used in another test

.github/workflows/unit_tests.yml
Micket: oh nevermind that wasn't the issue

Flamefire: Why? If the file exists it will be expanded and pass, if not it will fail the check

PR #19797, opened 2024-02-06
fix homepage + source URL for Gblocks 0.91b

pavelToman:
Update source url
Old one does not work anymore

PR #24297, opened 2025-10-19
fix homepage + source URL for Gblocks 0.91b

bedroge:
This was changed in https://github.com/easybuilders/easybuild-easyconfigs/pull/19797, but it no longer works, and apparently things changed back to the old URL now. See https://www.biologiaevolutiva.org/jcastresana/, which redirects to https://molevol-ibe.csic.es/.


bedroge:
EB_ARGS="--force-download"

PR #24376, opened 2025-10-27
add patch for EasyBuild v5.1.2 to fix long-lived bug with toolchain prepare when not using `$CPATH`

ocaisa:
The bug has been exposed when using the new options `EASYBUILD_MODULE_SEARCH_PATH_HEADERS=include_paths` and `EASYBUILD_SEARCH_PATH_CPP_HEADERS=include_paths` so that `CPATH` is no longer set with `EESSI/2025.06`


ocaisa:

bedroge (approved):
Lgtm. This fixes a serious bug affecting anyone using a non-default search path configuration, so patching the EasyBuild easyconfig makes sense to me.

bedroge:
Going in, thanks @ocaisa!

PR #24442, opened 2025-11-03
{lib}[gcccoreflexiblas-14.3.0-3.4.5] PyTorch v2.9.0

AlvaroHuanay:
This PR adds PyTorch 2.9.0 for Stages/2026

Update 5.10.2025:

I propose adding the some of the respective `aarch64` prebuilt files to have a more symmetric and complete PyTorch since I encountered issues with aarch64 easyconfigs that require these nvidia extensions as well when I load PyTorch. **Question:** Is it a good option?

Looks like PyTorch aarch64 does not have the same extension requirements as x86_64.

cusparselt removed from aarch64 because of the following outcome even though [there are aarch64](https://pypi.org/project/nvidia-cusparselt-cu12/0.7.1/#nvidia_cusparselt_cu12-0.7.1-py3-none-manylinux2014_aarch64.whl) releases:

```
== 2025-11-05 17:02:36,528 python.py:245 INFO Found 0 invalid packages out of 179 packages
== 2025-11-05 17:02:36,539 build_log.py:226 ERROR EasyBuild encountered an error (at easybuild/base/exceptions.py:126 in __init__): `/.../2026/software/Python/3.13.5-GCCcore-14.3.0/bin/python -m pip check` failed:
nvidia-cusparselt-cu12 0.7.1 is not supported on this platform
 (at easybuild/easyblocks/p/python.py:257 in run_pip_check)
== 2025-11-05 17:02:36,539 build_log.py:322 INFO ... (took 4 secs)
== 2025-11-05 17:02:36,539 filetools.py:2152 INFO Removing lock /.../easybuild/jupiter/software/.locks/_e_project1_XXXX_XXXX_easybuild_jupiter_software_PyTorch_2.9.0-gcccoreflexiblas-14.3.0-3.4.5-CUDA-13.lock...

```

Looking for an alternative in the meantime.

### DIFF
--- a/easybuild/easyconfigs/p/PyTorch/PyTorch-2.9.0-gcccoreflexiblas-14.3.0-3.4.5-CUDA-12.eb
+++ b/easybuild/easyconfigs/p/PyTorch/PyTorch-2.9.0-gcccoreflexiblas-14.3.0-3.4.5-CUDA-12.eb
@@ -25,10 +25,79 @@ dependencies = [
 ]
 
 
-# PyTorch on Grace Hopper nodes don't have the same wheels and most of the
-# stuff is from the system
+# PyTorch on Grace Hopper nodes don't have the same wheels
 if ARCH == 'aarch64':
     exts_list = [
+        ('nvidia_nvtx_cu12', '12.8.90', {
+            'modulename': False,
+            'source_tmpl': '%(name)s-%(version)s-py3-none-manylinux2014_%(arch)s.manylinux_2_17_%(arch)s.whl',
+            'checksums': ['d7ad891da111ebafbf7e015d34879f7112832fc239ff0d7d776b6cb685274615'],
+        }),
+        ('nvidia_nvshmem_cu12', '3.3.20', {
+            'modulename': False,
+            'source_tmpl': '%(name)s-%(version)s-py3-none-manylinux2014_%(arch)s.manylinux_2_17_%(arch)s.whl',
+            'checksums': ['0b0b960da3842212758e4fa4696b94f129090b30e5122fea3c5345916545cff0'],
+        }),
+        ('nvidia_nvjitlink_cu12', '12.8.93', {
+            'modulename': False,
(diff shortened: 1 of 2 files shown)


Thyre:
The idea of using wheels instead of building from source was discussed a fair bit in https://github.com/easybuilders/easybuild/issues/921, a dedicated Slack channel and in https://github.com/easybuilders/easybuild/wiki/PyTorch-working-group#action-points

I can't tell at all if this approach you're using works. I have no experience with the AI stack.
Generally though, this PR should adhere to the conventions of this repository (i.e. use `gfbf` instead of `gcccoreflexiblas` for example, or stick to one CUDA version in the toolchain if possible). Once this is figured out, we can try to start thinking about how to handle `aarch64` and `x86` in the same EasyConfig when dealing with wheels. 

Maybe @Flamefire can also give some input then. I'd say that he has extensive experience with PyTorch (and all its peculiarities...)

Flamefire:
The idea was to have at least a versionsuffix because we cannot mix PYPI and source-built PyTorch with other software.

Besides that this should be OK for the PYPI version although I strongly suggest to unify those 2 lists. e.g. `local_foo = 'this' if ARCH == x  else 'that' `

Otherwise it is hard to tell if e.g. the packages are the same, same versions, etc.

Thyre:
> The idea was to have at least a versionsuffix because we cannot mix PYPI and source-built PyTorch with other software.

So for now, we could add a versionsuffix like `-pypi`. Sites could still remove the suffix if they wish.

> Besides that this should be OK for the PYPI version although I strongly suggest to unify those 2 lists. e.g. `local_foo = 'this' if ARCH == x else 'that' `
> 
> Otherwise it is hard to tell if e.g. the packages are the same, same versions, etc.

Absolutely agreed...

Flamefire:
I just found that Triton 3.5 comes with ptxas 12.8 so either the CUDA module used needs to be 12.8 or higher or `TRITON_PTXAS_PATH` should point to the `ptxas` binary in the CUDA module.

However changing the ptxas version has lead to other issues in the past, so we better use a matching CUDA version if possible

PR #24476, opened 2025-11-06
remove unneccessary env variable which has nothing to do with mcpl

ebknudsen:
By a mistake the easyconfig sets the environment variable MCXTRACE which originates from a different easyconfig (McXtrace which can use mcpl). This PR removes that line.


Micket:

PR #24490, opened 2025-11-07
add (back) patch to make sure environment variables are defined for RStudio-Server v1.2.1335 terminal sessions

ocaisa:
This reintroduces a patch we previously applied to `RStudio-Server`. As seen in https://github.com/EESSI/software-layer/pull/1292#issuecomment-3503472425, without this patch the full environment is not passed through to a terminal shell started in the RStudio-Server instance (`PATH` and `LD_LIBRARY_PATH` are, so most things work). This means that the `module` command does not work (for example), and in the case of EESSI this causes explicit breakage due to our use of a Gentoo Prefix.


ocaisa:

boegel:
Test report by @boegel
**SUCCESS**
node4228.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/9793e657c166fe39344bd24869d1b0cb for a full test report.

boegel:
Going in, thanks @ocaisa!

PR #24567, opened 2025-11-14
CI: Enable flake8-comprehension and fix issues

Flamefire:
https://pypi.org/project/flake8-comprehensions/ has various useful checks, most importantly:

    C419 Unnecessary list comprehension in <any/all>() prevents short-circuiting - rewrite as a generator.

I.e. things like any([expensive_check(x) for x in collection]) runs the check after processing the full collection instead of stopping after the first "match" which costs performance.

To me the whole collection of checks look useful for performance and readability so enable them all.

See https://github.com/easybuilders/easybuild-framework/pull/5043


jfgrimm:
Going in, thanks @Flamefire!

PR #24568, opened 2025-11-14
{math}[foss/2025a] libxsmm v1.17

rmoretti:


jfgrimm:
@rmoretti why add a `foss`-level variant? As far as I can tell, there is no reason this should be at anything higher than `GCC`
It does support building with `BLAS`, but the documentation advises explicitly against this:
> Similarly, an application is free to choose any BLAS or LAPACK library (if the link model available on the OS supports this), and therefore linking GEMM routines when linking LIBXSMM itself (by supplying BLAS=1|2) may prevent a user from making this decision at the time of linking the actual application.

rmoretti:
@jfgrimm That's correct, thanks. I'll just close this.

PR #24449, opened 2025-11-04
add patch to force use of OpenGL dependency in easyconfig for gmsh 4.12.2

gkaf89:
(created using `eb --new-pr`)

PR #24594, opened 2025-11-17
add patch to make gmsh 4.12.2 + 4.14.0 use provided OpenGL dependency

gkaf89:
Add patch of #24449 in all new toolchains.


Micket:

PR #24602, opened 2025-11-17
add patch for double precision SVE bug to GROMACS 2024.1 easyconfig

bedroge:
See https://gitlab.com/gromacs/gromacs/-/merge_requests/4299, solves https://gitlab.com/gromacs/gromacs/-/issues/5057 which affects Arm CPUs with SVE instructions. The fix is included in GROMACS 2024.3 (and there are no easyconfigs for 2024.2). 2023 versions may also be affected, but the same patch does not automatically work for these versions.


bedroge:
CORE_CNT=16
EB_ARGS="--installpath /tmp/pr24602"

bedroge:
This is being tested on A64FX in https://github.com/EESSI/software-layer/pull/1308.

PR #24469, opened 2025-11-05
{bio,tools}[foss/2025b] arrow-R v22.0.0, R-bundle-Bioconductor v3.22 w/ R 4.5.2

branfosj:
(created using `eb --new-pr`)

* [x] #24318
* [x] #24463
* [x] #24465

## Removed because deprecated and not available

* zlibbioc: https://www.bioconductor.org/packages/3.21/bioc/html/zlibbioc.html
* hiAnnotator: https://bioconductor.org/packages/3.22/bioc/html/hiAnnotator.html
* ccmap: https://bioconductor.org/packages/3.22/bioc/html/ccmap.html
* Repitools: https://bioconductor.org/packages/3.22/bioc/html/Repitools.html
* rGADEM: https://bioconductor.org/packages/3.22/bioc/html/rGADEM.html

## Remove because they are not currently being distributed for Bioconductor 3.22

On the page for each see that the 'Source Package' is empty. If a `.tar.gz` appears there then they are available again.

* cummeRbund: https://bioconductor.org/packages/3.22/bioc/html/cummeRbund.html
* DeconRNASeq: https://bioconductor.org/packages/3.22/bioc/html/DeconRNASeq.html
* rols: https://bioconductor.org/packages/3.22/bioc/html/rols.html
* struct: https://bioconductor.org/packages/3.22/bioc/html/struct.html
* structToolbox: https://bioconductor.org/packages/3.22/bioc/html/structToolbox.html
* Maaslin2: https://bioconductor.org/packages/3.22/bioc/html/Maaslin2.html

PR #24647, opened 2025-11-21
deal with case where the tagbot comment is too long

branfosj:
In #24469 there is no comparison with existing easyconfigs because the comment is too long:
```
Failed to post comment: 422, {"message":"Validation Failed","errors":[{"resource":"IssueComment","code":"unprocessable","field":"data","message":"Body is too long (maximum is 65536 characters)"}],"documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment","status":"422"}
```


Micket:
oh crap, there is a bug, it should be len(comment)

PR #24713, opened 2025-11-28
{numlib}[GCC/13.3.0] LAPACK v3.12.1

rmoretti:


Micket:
So, i've actually declined to merge updates to LAPACK in the past, there is really no need for this to exist, as one should always use the FlexiBLAS/MKL/BLIS/OpenBLAS instead.
Supplying the slow reference implementation, especially under this name might fool users to accidentally use this (perhaps they recognize the name "LAPACK"), when in actuality, this module should never be used.

rmoretti:
OK, I see your point.

PR #24385, opened 2025-10-28
{bio}[foss/2023b] MinIONQC v1.4.2 w/ R 4.4.1

PhoenixEmik:
(created using `eb --new-pr`)
I'm not sure if I should add the R version suffix to the easyconfig.
A similar package that contain a single R scipt, [Phantompeakqualtools](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/p/Phantompeakqualtools), has no R suffix.
Appreciate if anyone can advise.

PR #24730, opened 2025-12-02
Revert "{bio}[foss/2023b] MinIONQC v1.4.2 w/ R 4.4.1"

PhoenixEmik:
Reverts easybuilders/easybuild-easyconfigs#24385

We just find that the codes downloaded from release page isn't the latest, and the version number is confusing as well.
I would like to open a new pull request of the latest MinIONQC codes.


Crivella:
I think as long as this is not used as a dep for other codes we can leave it as is and just add the EC for the newer version

EDIT:
I guess this relates to
- https://github.com/easybuilders/easybuild-easyconfigs/pull/24731

But i am wondering why it is versioned as `1.4.1-XXX` was there a bifurcation from 1.4.1 where 1.4.2 and the develop branch diverged? Otherwise we can just name the EC from https://github.com/easybuilders/easybuild-easyconfigs/pull/24731 as `1.4.2-XXX`

PhoenixEmik:
> I think as long as this is not used as a dep for other codes we can leave it as is and just add the EC for the newer version
> 
> EDIT: I guess this relates to
> 
>     * [{bio}[foss/2023b] MinIONQC v1.4.1-20221020 w/ R 4.4.1 #24731](https://github.com/easybuilders/easybuild-easyconfigs/pull/24731)
> 
> 
> But i am wondering why it is versioned as `1.4.1-XXX` was there a bifurcation from 1.4.1 where 1.4.2 and the develop branch diverged? Otherwise we can just name the EC from #24731 as `1.4.2-XXX`

Ok, I will update #24731 and close this.

PR #24736, opened 2025-12-02
{bio}[system/system] CompuCell3D 4.7.0

rsdmse:


Crivella:
I think we should be able to build this from source, no reason to rely on the Conda installation/recipes

Probably can use
https://github.com/CompuCell3D/CompuCell3D/blob/master/conda-recipes/build.sh
as a starting point for a CMake based recipe

PR #24747, opened 2025-12-04
Switch to setuptools and `find_packages`

Flamefire:
Any reasonably modern Python distribution includes setuptools.

it was useful for Python 2.x, is no longer necessary and can be removed making it consistent for all 3 EB packages:

https://github.com/easybuilders/easybuild-easyblocks/pull/4006
https://github.com/easybuilders/easybuild-framework/pull/5065

Also add a test with Python 3.12 which was missing for this test suite but present in the others

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,8 +33,8 @@ This script can be used to install easybuild-easyconfigs, e.g. using:
 """
 
 import os
-from distutils import log
-from distutils.core import setup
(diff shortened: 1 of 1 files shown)


boegel:
I'm not keen on this, see discussion in https://github.com/easybuilders/easybuild-framework/pull/5065

PR #24796, opened 2025-12-09
Remove PythonPackage defaults from template easyconfigs

Flamefire:
The values set here are the defaults now.


sassy-crick:
There is actually a typo in the `2024a` version: the toolchain-version is set to `2023a`, and also in the comment the wrong version is mentioned. Might be a good time to correct that, and check if the dependency-versions are correct too.

Flamefire:
Yeah, C&P. CMake was even listed twice. Fixed

PR #24819, opened 2025-12-11
in-place update for BBTools from v39.52 to v39.56

JarneRenders:
BBTools 39.52 contained a bug where due to race conditions when reading compressed files concurrently some threads would never get killed and Java would never exit properly. See bbushnell/BBTools#6. This was fixed in the latest version: 3.56 bbushnell/BBTools@75a7ccf

BBTools 39.52 is not yet in a release, so upgrading it seems best.


lexming:

lexming (approved):
LGTM

lexming:
Merging, thanks @JarneRenders !

PR #24884, opened 2025-12-15
Add MiXCR 4.7.0 and 4.7.0-254-develop (Java 17)

machavolu2025:
This PR adds EasyBuild easyconfigs for MiXCR:

- MiXCR 4.7.0 (stable)
- MiXCR 4.7.0-254-develop

Both builds use Java 17.

The develop version is required to support newer presets that are not
available in the stable release.

Builds were tested successfully on an HPC environment.
No existing easyconfigs were modified.

### DIFF
--- a/easybuild/easyconfigs/M/MiXCR/MiXCR-4.7.0-254-develop-Java-17.eb
+++ b/easybuild/easyconfigs/M/MiXCR/MiXCR-4.7.0-254-develop-Java-17.eb
@@ -1,0 +1,28 @@
+easyblock = 'Tarball'
+
+name = 'MiXCR'
+version = '4.7.0'
+versionsuffix = '-254-develop-Java-17'
+
+homepage = 'https://milaboratory.com/software/mixcr'
+description = """MiXCR is a universal software for fast and accurate extraction of T- and B- cell receptor repertoires
+from any type of sequencing data."""
+
+toolchain = SYSTEM
+
+source_urls = ['https://link.milaboratory.com/software/mixcr/']
+sources = ['mixcr-develop.zip']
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/M/MiXCR/MiXCR-4.7.0-254-develop-Java-17.eb
```diff
@@ -0,0 +1,28 @@
+easyblock = 'Tarball'
+
+name = 'MiXCR'
+version = '4.7.0'
+versionsuffix = '-254-develop-Java-17'
```
Thyre, line 5:
```suggestion
versionsuffix = '-254-develop-Java-%(javaver)s'
```

easybuild/easyconfigs/M/MiXCR/MiXCR-4.7.0-254-develop-Java-17.eb
```diff
@@ -0,0 +1,28 @@
...
+
+homepage = 'https://milaboratory.com/software/mixcr'
+description = """MiXCR is a universal software for fast and accurate extraction of T- and B- cell receptor repertoires
+from any type of sequencing data."""
+
+toolchain = SYSTEM
+
+source_urls = ['https://link.milaboratory.com/software/mixcr/']
+sources = ['mixcr-develop.zip']
+checksums = ['6bf8600d81df3b044df1043b937c03da9d30a04de55839faf466a704578c60be']
```
Thyre, line 15: This doesn't look like a stable tarball, and likely breaks on their next commit.
The full file name (`mixcr-4.7.0-254-develop-d0e53ee49e16bc94c7ffb238cfe8319004d84e01.zip`) matches the current commit hash on GitHub. So if we include this as an EasyConfig, we should use a stable URL instead.

PR #24944, opened 2025-12-20
archive easyconfigs using a toolchain from `2021*` era

Micket:
https://docs.easybuild.io/policies/toolchains/#current-situation

I had to drop use of UnZip from GCCcore 10.3.0 in AGeNT-3.0.6.eb. I replaced it with nothing, as we would typically require unzip from OS level anyway to deal with sources. It has a restrictive license, i can't test that software.

I also archived the unused 11.1.0  GCC since the next toolchain that's on the deprecated list was 11.2.0


boegel:
@Micket Archiving now means copying stuff over to [easybuild-easyconfigs-archive](https://github.com/easybuilders/easybuild-easyconfigs-archive) repo, and *removing* from `easybuild-easyconfigs` repo...

Micket:
https://github.com/easybuilders/easybuild-easyconfigs/pull/22954#pullrequestreview-2868828005 I know :)
but https://github.com/easybuilders/easybuild-easyconfigs/pull/22954#issuecomment-2915426588

jfgrimm (approved):
lgtm

PR #24945, opened 2025-12-20
Archive SYSTEM level easyconfigs from before 2020-12-19

Micket:
I'm going to let the CI check if any of these were in use in newer easyconfigs and move them back as needed

Should merge the toolchain PR first:
- [x] #24944


jfgrimm (approved):
lgtm

PR #25015, opened 2026-01-06
{tools}[foss/2023b] FEniCS-DOLFINx v0.10.0

gkaf89:
This commit contains all the easyconfigs required for building FEniCS-DOPHINx.


gkaf89:
We are facing an problem when building the Python packages. The FFCx and UFL packages require the `setuptools` package (as a build dependency). However, these packages also require the `SciPy-bundle` that fetches an older version of the `packaging` package than `setuptools`, and the packages fail to build.

The current solution it to add `setuptools` as a dependency after `SciPy-bundle`. Is there another way to resolve the problem?

In another a minor issue, DOFINx is using the find utility of CMake to link with SCHOTCH. Since SCOTCH is built with make in EasyBuild, it is not setup to work with the find utility of CMake. However, KaHIP is a preferable choice over SCOTCH for DOLFINx as far as I understand.

gkaf89:
Problem was due to mixing package installations with different `--prefer-python-search-path`.

PR #24998, opened 2026-01-05
Use CargoPythonBundle for polars-0.15.6-foss-2022a

Flamefire:
(created using `eb --new-pr`)

(likely) requires:
- [ ] https://github.com/easybuilders/easybuild-easyblocks/pull/3993

The `maturin` extension uses `cargo` and hence tries to write to `$HOME`

We could replace some extensions by the `maturin` easyconfig but then that would not be installed anymore. We possibly could make `maturin` a runtime dependency (even though it is not/might not be) to avoid that, but going with the minimal change. Especially as we don't have a maturin-0.14.6 EC but only newer ones

PR #25066, opened 2026-01-15
remove (broken) easyconfigs for polars 0.15.6 + infercnvpy 0.4.2 w/ `foss/2022a`

Flamefire:
Polars can no longer be built as the cargo crates have changed to be incompatible with the Rust version used.
infercnvpy depends on polars.

See https://github.com/easybuilders/easybuild-easyconfigs/pull/24998#issuecomment-3710429327


boegel:
Going in, thanks @Flamefire!

PR #25103, opened 2026-01-20
Update Cbc-2.10.12-foss-2025b.eb   name 'VERSION_TAR_GZ' is not defined

baverhey:
ERROR: Failed to process easyconfig /opt/easybuild/easybuild-easyconfigs/easybuild/easyconfigs/c/Cbc/Cbc-2.10.12-foss-2025b.eb: Parsing easyconfig file failed: name 'VERSION_TAR_GZ' is not defined (line 15)  Reverted to same of previous versions


baverhey:
found out its a new feature and had not updated yet to new version of eb

PR #25112, opened 2026-01-21
require setting `sanity_pip_list` for new PRs

smoors:
depends on easyblock PR:
- https://github.com/easybuilders/easybuild-easyblocks/pull/4049

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -1318,6 +1318,8 @@ class EasyConfigTest(TestCase):
             r'ReFrame.*',
         ]
 
+        pythonpackage_easyblocks = ['CargoPythonBundle', 'CargoPythonPackage', 'PythonBundle', 'PythonPackage']
+
         failing_checks = []
 
         python_default_urls = PythonPackage.extra_options()['source_urls'][0]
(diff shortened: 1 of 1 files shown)


boegel:
@smoors Before we require this, we should update a couple of easyconfigs (like recent `Python`, `Python-bundle-PyPI`, `SciPy-bundle`, `matplotlib`, etc.) to make sure that there's no problems already with these?

smoors:
> @smoors Before we require this, we should update a couple of easyconfigs (like recent `Python`, `Python-bundle-PyPI`, `SciPy-bundle`, `matplotlib`, etc.) to make sure that there's no problems already with these?

unlike `sanity_pip_check`, this only checks for extensions listed in the easyconfig that has `sanity_pip_list = True` set, so there is no rush to fix existing easyconfigs. but indeed would be good to fix the common ones to avoid surprises during updates.

boegel:
> > @smoors Before we require this, we should update a couple of easyconfigs (like recent `Python`, `Python-bundle-PyPI`, `SciPy-bundle`, `matplotlib`, etc.) to make sure that there's no problems already with these?
> 
> unlike `sanity_pip_check`, this only checks for extensions listed in the easyconfig that has `sanity_pip_list = True` set, so there is no rush to fix existing easyconfigs. but indeed would be good to fix the common ones to avoid surprises during updates.

No rush indeed, but having *some* idea of how large the impact would be if we start requiring this would be good...

smoors:
this is no longer needed as we will do the check as part of `sanity_pip_check`

PR #25125, opened 2026-01-22
{chem}[NVHPC/25.3-CUDA-12.8.0] QuantumESPRESSO v7.5, FFTW v3.3.10, HDF5 v1.14.5, ...

pavelToman:
(created using `eb --new-pr`)
resolves https://github.com/vscentrum/vsc-software-stack/issues/571
requires:
- https://github.com/easybuilders/easybuild-easyblocks/pull/4053
- https://github.com/easybuilders/easybuild-easyconfigs/pull/25147


easybuild/easyconfigs/f/FFTW/FFTW-3.3.10-NVHPC-25.3-CUDA-12.8.0.eb
```diff
@@ -0,0 +1,22 @@
...
+in one or more dimensions, of arbitrary input size, and of both real and complex data."""
+
+toolchain = {'name': 'NVHPC', 'version': '25.3-CUDA-%(cudaver)s'}
+toolchainopts = {'pic': True}
+
+source_urls = [homepage]
+sources = [SOURCELOWER_TAR_GZ]
+checksums = ['56c932549852cddcfafdab3820b0200c7742675be92179e59e6215b340e26467']
+
+dependencies = [('CUDA', '12.8.0', '', SYSTEM)]
```
Thyre: Why do we set an explicit dependency for CUDA when FFTW doesn't use CUDA?
I would argue that we should explicitly name the compiler instead and not rely on `%(cudaver)s` for it.

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.5-NVHPC-25.3-CUDA-12.8.0.eb
```diff
@@ -0,0 +1,112 @@
...
+            'clone_into': _name,
+        },
+    })
+
+local_gccver = '14.2.0'
+builddependencies = [
+    ('M4', '1.4.19', '', ('GCCcore', local_gccver)),
+    ('CMake', '3.31.3', '', ('GCCcore', local_gccver)),
+    ('pkgconf', '2.3.0', '', ('GCCcore', local_gccver)),
+    ('Python', '3.13.1', '', ('GCCcore', local_gccver)),  # required for tests
```
Thyre:
```suggestion
    ('M4', '1.4.19'),
    ('CMake', '3.31.3'),
    ('pkgconf', '2.3.0'),
    ('Python', '3.13.1'),  # required for tests
```

This should not be required if the toolchain works correctly.

pavelToman: Some update? It was not working before

boegel:
@Crivella It would be great if you could take a look at this, before we get it merged.

Not urgent from my side.

@pavelToman Can you share some details on the testing you've done with the installation, using the [input from the UEABS benchmark suite](https://repository.prace-ri.eu/git/UEABS/ueabs/-/tree/r2.2-dev/quantum_espresso)?

Crivella:
Yeah we have been in touch in Slack.
My main concern is making sure we are picking up as much in termos of libraries from NVHPC as possible.
In particular `cuFFT` since QE is very FFT heavy. It seems to be doing so even with FFTW as a dep, but  i think we should adjust the EBlock for when compiling with NVHPC to make sure we always get the right stuff

pavelToman:
I've run `UEABS Case A - 112 atoms` with https://repository.prace-ri.eu/git/UEABS/ueabs/-/blob/r2.2-dev/quantum_espresso/test_cases/small/ausurf.in
The results are here: https://github.com/vscentrum/vsc-software-stack/blob/1730799b4f8f413dd44427aebc44a67277d73638/571_QuantumESPRESSO/ausurf.out

lexming:

easybuild/easyconfigs/q/QuantumESPRESSO/QuantumESPRESSO-7.5-NVHPC-25.3-CUDA-12.8.0.eb
```diff
@@ -0,0 +1,112 @@
...
+    ('M4', '1.4.19'),
+    ('CMake', '3.31.3'),
+    ('pkgconf', '2.3.0'),
+    ('Python', '3.13.1'),  # required for tests
+]
+dependencies = [
+    ('CUDA', '12.8.0', '', SYSTEM),
+    ('FFTW', '3.3.10'),
+    ('HDF5', '1.14.5'),
+    ('ELPA', '2025.01.002'),
```
lexming: Version of ELPA dependency seems wrong
```suggestion
    ('ELPA', '2025.06.002'),
```

easybuild/easyconfigs/h/HDF5/HDF5-1.14.5-NVHPC-25.3-CUDA-12.8.0.eb
```diff
@@ -0,0 +1,34 @@
...
+
+toolchain = {'name': 'NVHPC', 'version': '25.3-CUDA-%(cudaver)s'}
+toolchainopts = {'pic': True, 'usempi': True}
+
+source_urls = ['https://github.com/HDFGroup/hdf5/archive']
+sources = [{'download_filename': 'hdf5_%(version)s.tar.gz', 'filename': SOURCE_TAR_GZ}]
+patches = ['HDF5-1.14.5_m4-aclocal-fix.patch']
+checksums = [
+    {'hdf5_1.14.5.tar.gz': 'c83996dc79080a34e7b5244a1d5ea076abfd642ec12d7c25388e2fdd81d26350'},
+    {'patch1.1-m4-hdf5.patch': 'ac6a634385528daf249aaa7183980d21f847f75697de4cf7098430e6b00af021'},
```
lexming: this name does not match the name of the file

lexming:

easybuild/easyconfigs/h/HDF5/HDF5-1.14.5-NVHPC-25.3-CUDA-12.8.0.eb
```diff
@@ -0,0 +1,34 @@
...
+homepage = 'https://portal.hdfgroup.org/display/support'
+description = """HDF5 is a data model, library, and file format for storing and managing data.
+ It supports an unlimited variety of datatypes, and is designed for flexible
+ and efficient I/O and for high volume and complex data."""
+
+toolchain = {'name': 'NVHPC', 'version': '25.3-CUDA-%(cudaver)s'}
+toolchainopts = {'pic': True, 'usempi': True}
+
+source_urls = ['https://github.com/HDFGroup/hdf5/archive']
+sources = [{'download_filename': 'hdf5_%(version)s.tar.gz', 'filename': SOURCE_TAR_GZ}]
```
lexming: This breaks the checksum check due to the change in capitalization
> Missing checksum for HDF5-1.14.5.tar.gz

I suggest to keep it simple:
```suggestion
sources = ['hdf5_%(version)s.tar.gz']
```

pavelToman:

easybuild/easyconfigs/l/libxc/libxc-7.0.0-NVHPC-25.3-CUDA-12.8.0.eb
```diff
@@ -0,0 +1,49 @@
+easyblock = 'CMakeMake'
+
+name = 'libxc'
+version = '7.0.0'
+
+homepage = 'https://libxc.gitlab.io'
+description = """Libxc is a library of exchange-correlation functionals for density-functional theory.
+ The aim is to provide a portable, well tested and reliable set of exchange and correlation functionals."""
+
+toolchain = {'name': 'NVHPC', 'version': '25.3-CUDA-%(cudaver)s'}
```
lexming: it seems that this can be lowered to `nvidia-compilers` because `libxc` does not use MPI or BLAS

pavelToman: updated, thanks!

pavelToman:
> I see that you put CUDA as a normal dependency on all easyconfigs, that should not be needed as CUDA is provided by the NVHPC toolchain, did you have any issue with that?
> 
> Apart from this question, looks good, I only have one final remark that needs fixing (see below).

I updated all the easyconfigs to not have CUDA in deps, sorry for that - first time creating ECs with NVHPC

lexming:

lexming:
Not sure why those tests are broken, I need to check that

lexming:
I found what is wrong with the tests. The `get_toolchain_hierarchy` for `NVHPC` toolchain results in `['GCCcore', 'NVHPC']`, instead of `['nvidia-compilers', 'NVHPC']`. I don't have a fix yet.

lexming:
@pavelToman got a fix, see https://github.com/easybuilders/easybuild-framework/pull/5140

lexming:

lexming:
Merging, thanks @pavelToman !

PR #25200, opened 2026-02-02
{vis}[foss/2023a] VAPOR v3.10.1 and dependencies

migueldiascosta:
(created using `eb --new-pr`)

based on https://github.com/spack/spack-packages/blob/acfa08aaabdc6814acd3aaa259a44b7c732606ec/repos/spack_repo/builtin/packages/vapor/package.py, but managed to avoid copying python dependencies into the tree


migueldiascosta:

boegel:
@migueldiascosta CI is unhappy due to multiple versions of `PROJ` now being used as dependency in `2023a` generation of easyconfigs?

PR #25221, opened 2026-02-05
{geo}[gmpich/2025.07] GeoDict v2026-2-1.SP1

smoors:
(created using `eb --new-pr`)


easybuild/easyconfigs/g/GeoDict/GeoDict-2026-2-1.SP1-gmpich-2026.02.eb
```diff
@@ -0,0 +1,52 @@
...
+     '852804c6cc3ff1bb72824e61ac0a1b0d99d7905b701b70267e0cb6c188e4d873'},
+]
+
+# checksums = [
+#     {'GeoDict2025-6-3-Linux-x86_64-ServicePack5.tar.gz':
+#      '25da49fcba3ff1db43e525695b9e821e791302a02223dad9f0c56de2548040e1'},
+#     {'GeoDict2025-6-3-Linux-x86_64-ServicePack5-Tools.tar.gz':
+#      'b5fb26496a8b0d465db396837ed6e31d841e13b17eaf515e38f5d63e533d70ca'},
+# ]
+
```
jfgrimm:
```suggestion
```

easybuild/easyconfigs/g/GeoDict/GeoDict-2026-2-1.SP1-gmpich-2026.02.eb
```diff
@@ -0,0 +1,52 @@
...
+dependencies = [
+    # ('X11', '20240607'),
```
jfgrimm:
```suggestion
```

easybuild/easyconfigs/g/GeoDict/GeoDict-2026-2-1.SP1-gmpich-2026.02.eb
```diff
@@ -0,0 +1,52 @@
...
+    ('X11', '20250521'),
+]
+
+modextrapaths = {'PATH': '.'}
```
jfgrimm:
```suggestion
modextrapaths = {'PATH': ''}
```

smoors:

jfgrimm (dismissed):

easybuild/easyconfigs/g/gmpich/gmpich-2026.02.eb
```diff
@@ -0,0 +1,19 @@
+easyblock = 'Toolchain'
+
+name = 'gmpich'
+version = '2026.02'
```
boegel: GCC 14.2.0 was released Aug'24, MPICH 4.3.0 in Feb'25, so it would be more correct to use `2025.02`.

However, we already have `gmpich/2025.04` which uses GCC 13.3.0 & MPICH 4.2.2 😅 

I don't want to be pedantic here, but we also have MPICH 4.3.2 on top of GCC 14.3.0 already, and MPICH 5.0.0 was released a couple of days ago, see https://github.com/pmodels/mpich/releases/tag/v5.0.0 (it's not listed as a release yet though, see https://github.com/pmodels/mpich/releases)

So to have some freedom to define additional `gmpich` versions with newer GCC & MPICH versions, we shouldn't use `2026.02` here...

smoors: which version do you suggest then? any date between `2025.04` and `2026.02`?

it sounds more useful to me to choose the date that the easyconfig was added, rather than some random date that doesn't mean anything?

boegel: The latter is very true, but in this case I think we should make an exception, and keep `2026.02` for the sensible policy of using release date of most recent component (MPICH 5.0.0 in this case).

Let's just go with `2025.07` for this one, and `2025.10` for one based on GCC 14.3.0 + MPICH 4.3.2 (which makes sense, the latter was released Oct'25).

smoors:

boegel:
Going in, thanks @smoors!

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 2 out of 2 (total: 4 mins 23 secs) (2 easyconfigs in total)
node4210.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/bd4db7dcbdf7915df14f6bedb4b95a22 for a full test report.

PR #25253, opened 2026-02-11
[tools][GCCcore/14.3.0] ipympl v0.10.0

AlvaroHuanay:
This PR adds ipympl/0.10.0 for Stages/2026. Compiled and tested on x86_64 and aarch64.

### DIFF
--- a/easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
+++ b/easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
@@ -29,12 +29,7 @@ dependencies = [
 
 exts_list = [
     ('ipympl', version, {
-        'source_urls': ['https://github.com/matplotlib/ipympl/archive/refs/tags/'],
-        'source_tmpl': 'v%(version)s.tar.gz',
-        'checksums': [
-            {'v%(version)s.tar.gz':
-             '28b22ec30ceaddab2e929baa70d599e107579f90d9dc19be22c4650a763bbc22'},
-        ],
+        'checksums': ['eda69602a010af2a42e8ebd069b0ee0dbe8df7fc69d7c1e8b99fece0a2fe613f'],
     }),
 ]
 


easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-GCCcore-14.3.0.eb
```diff
@@ -0,0 +1,38 @@
...
+version = '0.10.0'
+
+homepage = 'https://matplotlib.org/ipympl'
+description = """Leveraging the Jupyter interactive widgets framework, ipympl enables the
+interactive features of matplotlib in the Jupyter notebook and in JupyterLab.
+Besides, the figure canvas element is a proper Jupyter interactive widget which
+can be positioned in interactive widget layouts.
+"""
+
+toolchain = {'name': 'GCCcore', 'version': '14.3.0'}
```
Thyre:
```suggestion
toolchain = {'name': 'gfbf', 'version': '2025b'}
```

Since we're using `gfbf` to use `matplotlib`, we need to bump this to `gfbf` as well.

easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
```diff
@@ -0,0 +1,38 @@
...
+    ('hatch-jupyter-builder', '0.9.1'),
+]
+
+dependencies = [
+    ('Python', '3.13.5'),
+    ('IPython', '9.4.0'),
+    ('jupyter-server', '2.17.0'),
+    ('JupyterLab', '4.4.9'),   # for jlpm availability
+    ('nodejs', '22.17.1'),   # jlpm call requires nodejs
+    ('matplotlib', '3.10.5', '', ('gfbf', '2025b')),
```
Thyre:
```suggestion
    ('matplotlib', '3.10.5'),
```

easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
```diff
@@ -0,0 +1,38 @@
...
+
+toolchain = {'name': 'gfbf', 'version': '2025b'}
+
+source_urls = ['https://github.com/matplotlib/ipympl/archive/refs/tags/']
+sources = ['v%(version)s.tar.gz']
+checksums = ['28b22ec30ceaddab2e929baa70d599e107579f90d9dc19be22c4650a763bbc22']
+
+
+builddependencies = [
+    ('binutils', '2.44'),
```
Thyre:
```suggestion
```

easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
```diff
@@ -0,0 +1,37 @@
...
+    ('hatchling', '1.27.0'),
+    ('hatch-jupyter-builder', '0.9.1'),
+]
+
+dependencies = [
+    ('Python', '3.13.5'),
+    ('IPython', '9.4.0'),
+    ('jupyter-server', '2.17.0'),
+    ('JupyterLab', '4.4.9'),   # for jlpm availability
+    ('nodejs', '22.17.1'),   # jlpm call requires nodejs
```
Micket: that sounds to me like it should be used as a builddep and do whatever jlpm it needs as a postinstallcmds instead, like JupyterLab-git.

easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
```diff
@@ -0,0 +1,37 @@
+easyblock = 'PythonPackage'
```
Micket: Please change this back to a PythonBundle, we prefer to always use those. I wouldn't force this for something that had been a PythonPackage all along, but as we correctly switched this over to PythonBundle previously, lets not revert that improvement here.

AlvaroHuanay: Hi @Micket,

Thanks for the feedback. Change applied :+1: 

Regards.

easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
AlvaroHuanay: Yes. You are correct. Fixed!

Thanks.

easybuild/easyconfigs/i/ipympl/ipympl-0.10.0-gfbf-2025b.eb
```diff
@@ -0,0 +1,43 @@
...
+
+exts_list = [
+    ('ipympl', version, {
+        'source_urls': ['https://github.com/matplotlib/ipympl/archive/refs/tags/'],
+        'source_tmpl': 'v%(version)s.tar.gz',
+        'checksums': [
+            {'v%(version)s.tar.gz':
+             '28b22ec30ceaddab2e929baa70d599e107579f90d9dc19be22c4650a763bbc22'},
+        ],
+    }),
```
Crivella, line 33: Any particular reason we are not pulling this from PyPI but github instead?

AlvaroHuanay: Hi @Crivella,

Thanks for the feedback. Your recommendation fits perfectly. It is fixed now.

Regards.

AlvaroHuanay:
> Any particular reason to add another version to an exissting package to the 225b toolchain?

I wanted to add a more recent version of ipympl. Should we stick to 0.9.8?

Crivella:
Problem is this is already being used as a dep https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/a/abTEM/abTEM-1.0.8-foss-2025b.eb
For now it is only 1 package, but we will either need to update it there or have to deal with adding exception in the CI every time a new software is added

AlvaroHuanay:
Hi @Crivella,

Thank you for sharing this. I will follow ipympl/0.9.8 then. Is it ok if I close the PR?

Regards.

Crivella:
Closing this in favor of the existing version of ipympl in the same TC.
Thanks @AlvaroHuanay  for the contribution

PR #25256, opened 2026-02-11
{tools}[foss/2024a,foss/2025a,foss/2025b] jupyterview v0.8.0, jupyterlab-h5web v12.6.1, JupyterLab-git v0.51.4, hatch-nodejs-build v1.1.1, meshio v5.3.5

Micket:
(created using `eb --new-pr`)

PR #25265, opened 2026-02-12
fix incorrect homepage in existing JupyterLab-git easyconfigs

moravveji:


Micket:
cf. #25256

moravveji:
> cf. #25256

Cool to see your PR ;-)
Is it OK that you drop the JupyterLab-git from yours, so that they would be all included in this one?
Otherwise, I close this one (or keep it only for fixing the `homepage` attribute in the older versions).

PR #25270, opened 2026-02-13
move MultiQC down from foss to gfbf toolchain (where it belongs)

akesandgren:
MultiQC doesn't depend on anything above gfbf, so move it down.


akesandgren:
@boegelbot Please test @ jsc-zen3

boegel:

PR #25457, opened 2026-02-27
Add CI using Python 3.12

Flamefire:
Requires installing setuptools explicitely and skipping test_dep_graph similar to https://github.com/easybuilders/easybuild-framework/pull/5116

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -20,6 +20,9 @@ jobs:
           - python: '3.11'
             modules_tool: Lmod-8.6.8
             module_syntax: Lua
+          - python: '3.12'
+            modules_tool: Lmod-8.6.8
+            module_syntax: Lua
       fail-fast: false
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # 4.2.2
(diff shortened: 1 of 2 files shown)


Thyre:
https://github.com/easybuilders/easybuild-easyconfigs/pull/25413 intentionally reduced the number of test combinations, so I'm not sure if adding a new one is desirable.
Maintainers would need to decide this.

If one wants to keep the same number of test jobs, moving Python 3.11 to Python 3.12 sounds like the easiest solution.

Flamefire:
> moving Python 3.11 to Python 3.12 sounds like the easiest solution.

I guess so, yes. Although 3.12 was a switch that has caused issues. So one before and after the switch makes sense to me.

Thyre:
I mean, we still have the Python 3.7 job. Naively asking, shouldn't this catch the case before the switch?

Flamefire:
> I mean, we still have the Python 3.7 job. Naively asking, shouldn't this catch the case before the switch?

But not everything up to the switch. They could have introduced something in 3.9 which might hurt us.
But in general yes, it should work. Maybe it's also not so important for the EC repo, but more for the framework. I just wanted to be safe.

I'd rather fold the 2 `test-sdist` jobs into a step of the others. Sure they are <30s, but likely worth to not wait for a runner to pick those up and go through its setup

>  Maintainers would need to decide this.

Maybe wait then? Or shall we just drop 3.11?

boegel:
I don't think this makes sense for easyconfigs, where we have very little (if anything) that would only work on specific Python versions... It would significantly increase the time required to run the easyconfigs test suite.

Also, the check for `pygraph` is no longer relevant, thanks to:
- https://github.com/easybuilders/easybuild-framework/pull/5128

I'll go ahead and close this.

@Flamefire If you find a specific case of trouble in easyconfigs that we missed because we weren't running CI with Python 3.12 (or any other Python version, for that matter), please open a new PR...

PR #23900, opened 2025-09-18
{bio}[foss/2024a] BEDOPS v2.4.42

asp8200:
(created using `eb --new-pr`)

PR #25511, opened 2026-03-05
adding easyconfigs: BEDOPS-2.4.42-foss-2024a.eb

nicorap:
tested + fixed it; here is a follow-up PR

### DIFF
--- a/easybuild/easyconfigs/b/BEDOPS/BEDOPS-2.4.42-foss-2024a.eb
+++ b/easybuild/easyconfigs/b/BEDOPS/BEDOPS-2.4.42-foss-2024a.eb
@@ -40,11 +40,13 @@ files_to_copy = ['bin']
 
 sanity_check_paths = {
     'files': [
-        'bin/%s' % x for x in ['bam2bed', '%(namelower)s', 'convert2bed', 'unstarch'],
+        'bin/bam2bed',
+        'bin/%(namelower)s',
+        'bin/convert2bed',
+        'bin/unstarch',
     ],
     'dirs': [],
 }
-
 sanity_check_commands = ['%(namelower)s --help']
 
 moduleclass = 'bio'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/b/BEDOPS/BEDOPS-2.4.42-foss-2024a.eb
```diff
@@ -0,0 +1,52 @@
...
+# else build of jansson library fails with: 'configure: error: C compiler cannot create executables'
+prebuildopts = (
+    "unset LIBS && "
+    # show what we're about to change (optional but useful)
+    "grep -R --line-number -E '^[[:space:]]*SFLAGS[[:space:]]*=.*-static' applications/bed || true && "
+    # edit every Makefile that matches */src/Makefile under applications/bed
+    r"find applications/bed -type f -path '*/src/Makefile' -exec sed -i -E "
+    r"'s/^[[:space:]]*SFLAGS[[:space:]]*=[[:space:]]*-static[[:space:]]*$/SFLAGS =/g' {} \; && "
+    r"find applications/bed -type f -path '*/src/Makefile' -exec sed -i -E "
+    r"'s/^[[:space:]]*SFLAGS[[:space:]]*=[[:space:]]*-static[[:space:]]*\$\{?MEGAFLAGS\}?[[:space:]]*$/SFLAGS = ${MEGAFLAGS}/g' {} \; && "
```
verdurin: @nicorap the bot is complaining about the length of this line.
Also, could you explain why you've added these changes, ideally as an inline comment?

verdurin (changes requested):

sassy-crick:
There is also a similar PR #23900 open

verdurin:
@sassy-crick yes, this is actually fixes for the previous PR.

boegel:
The approach taken here is a lot more involved than what's done in:
- #25686

Same approach was used in:
- #25653

PR #25578, opened 2026-03-17
{tools}ANSYS v2025R2, ANSYS v2026R1

baverhey:
(created using `eb --new-pr`)


easybuild/easyconfigs/a/ANSYS/ANSYS-2025R2-foss-2025a.eb
```diff
@@ -0,0 +1,57 @@
+name = 'ANSYS'
+version = '2025R2'
+
+homepage = 'https://www.ansys.com'
+description = """Ansys offers a comprehensive software suite that spans the entire range of physics, providing access
+to virtually any field of engineering simulation that a design process requires."""
+
+toolchain = {'name': 'foss', 'version': '2025a'}
```
boegel: Using a non-system toolchain for ANSYS doesn't make much sense to me, necessary dependencies should come from OS in this case...

PR #25602, opened 2026-03-25
{mix}[NVHPC-25.9] HDF5 v1.14.6, imkl v2020.0.166, Wannier90 v3.1.0

hajgato:


easybuild/easyconfigs/i/imkl/imkl-2020.0.166-NVHPC-25.9-CUDA-12.9.1.eb
```diff
@@ -0,0 +1,39 @@
+# This is an easyconfig file for EasyBuild, see https://easybuilders.github.io/easybuild/
+
+name = 'imkl'
+version = '2020.0.166'
```
boegel, line 9: @hajgato This is very old, can you clarify why this version is used?

hajgato: fftw in imkl has fftw planning cache. In some calculations, it gives 30% speed increase compared to fftw-amd. 
 
Yes, 2020.0.166 is the best option, because I can chose avx2 or avx512 codepath on AMD. This still gives 10-15% speedup compared to recent imkl versions. (its about fftw, the same is not true for some of the most used BLAS routines like dgemm) See for example https://danieldk.eu/Software/Misc/Intel-MKL-on-AMD-Zen

easybuild/easyconfigs/i/imkl/imkl-2020.0.166-NVHPC-25.9-CUDA-12.9.1.eb
```diff
@@ -0,0 +1,39 @@
+# This is an easyconfig file for EasyBuild, see https://easybuilders.github.io/easybuild/
+
```
boegel, line 7:
```suggestion
#
# old imkl version, where we can still control AVX2 vs AVX-512 codepath on AMD CPUs via $MKL_DEBUG_CPU_TYPE,
# see also https://danieldk.eu/Software/Misc/Intel-MKL-on-AMD-Zen;
# can give 10-15% speedup compared to recent imkl versions for some workloads
# that heavily rely on FFTW interface of Intel MKL

```

boegel (changes requested):

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 3 out of 3 (total: 30 mins 53 secs) (3 easyconfigs in total)
node3900.accelgor.os - Linux RHEL 9.6, x86_64, AMD EPYC 7413 24-Core Processor (zen3), 1 x NVIDIA NVIDIA A100-SXM4-80GB, 590.48.01, Python 3.9.21
See https://gist.github.com/boegel/2e635efc2d4df0f35178845b71921f5f for a full test report.

boegel (approved):
lgtm

boegel:
Going in, thanks @hajgato!

PR #21108, opened 2024-08-01
Fix download URLs for old Intel software (2018-23)

bartoldeman:
Intel changed "irc_nas" to "IRC_NAS". Newer packages (2023.1+) already use that, but older ones need changing.

Mirrors https://github.com/spack/spack/pull/43286

Fixes #21104

PR #25632, opened 2026-03-27
{mix}[GCC/14.3.0,gfbf/2025b,gompi/2025b] DFT-D4 v4.0.2, libxc v7.0.0, multicharge v0.5.0 (incl. mctc-lib v0.5.1, mstore v0.3.0), imkl v2020.0.166

hajgato:
In an ideal case, the serial lib in Wannier90 would be included in the MPI version, with a name, lets say, `libwannier-serial.a`, but my EB knowledge is not good enough for this.


hajgato:
In an ideal case, the serial libs in Wannier90 would be included in the mpi version, with a name, lets say, `libwannier-serial.a`, but my EB knowledge is not good enough for this.

boegel:
@hajgato CI is failing because we're using a different version of `mctc-lib` and `mstore` compared to what's already there, how critical are these specific versions?
```
found 2 variants of 'mctc-lib' dependency in easyconfigs using '2025b' toolchain generation
* version: 0.5.0; versionsuffix:  as dep for {'mstore-0.3.0-GCC-14.3.0.eb', 'multicharge-0.4.0-gfbf-2025b.eb'}
* version: 0.5.1; versionsuffix: -dft-d4 as dep for {'mstore-0.3.0-GCC-14.3.0-dft-d4.eb', 'multicharge-0.5.0-gfbf-2025b-dft-d4.eb', 'DFT-D4-4.0.2-gfbf-2025b.eb'}

found 2 variants of 'mstore' dependency in easyconfigs using '2025b' toolchain generation
* version: 0.3.0; versionsuffix:  as dep for {'multicharge-0.4.0-gfbf-2025b.eb'}
* version: 0.3.0; versionsuffix: -dft-d4 as dep for {'multicharge-0.5.0-gfbf-2025b-dft-d4.eb', 'DFT-D4-4.0.2-gfbf-2025b.eb'}
```

hajgato:
`DFT-D4 4.0.2` requires `mctc-lib 0.5.1`. It is a hard requirement, if it is not present meson will install it.

hajgato:
Since `multicharge` and `mstore` depends on `mctc-lib`, We anyway need to make a new easyconfig. In a long term, we might have to consider to make a bundle, like X-Win, as this whole thing is developed by the same group, and tightly belongs to the same "ecosystem"

boegel:
> Since `multicharge` and `mstore` depends on `mctc-lib`, We anyway need to make a new easyconfig. In a long term, we might have to consider to make a bundle, like X-Win, as this whole thing is developed by the same group, and tightly belongs to the same "ecosystem"

I looked into exactly that, see:
- https://github.com/hajgato/easybuild-easyconfigs/pull/39

easybuild/easyconfigs/w/Wannier90/Wannier90-3.1.0-nvidia-compilers-25.9-CUDA-12.9.1-lib.eb
```diff
@@ -0,0 +1,36 @@
+easyblock = 'MakeCp'
+
+name = 'Wannier90'
+version = '3.1.0'
+versionsuffix = '-lib'
```
boegel: Maybe we should use `-serial-lib` here?

boegel: Actually, we can update the existing `Wannier90-3.1.0-NVHPC-25.9-CUDA-12.9.1.eb` to also have the serial library, we don't need a dedicated easyconfig for this

boegel:
CORE_CNT=16

easybuild/easyconfigs/i/imkl/imkl-2020.0.166-gompi-2025b.eb
```diff
@@ -0,0 +1,44 @@
...
+
+homepage = 'https://software.intel.com/mkl'
+description = """Intel Math Kernel Library is a library of highly optimized,
+ extensively threaded math routines for science, engineering, and financial
+ applications that require maximum performance. Core math functions include
+ BLAS, LAPACK, ScaLAPACK, Sparse Solvers, Fast Fourier Transforms, Vector Math, and more."""
+
+toolchain = {'name': 'gompi', 'version': '2025b'}
+
+source_urls = ['https://registrationcenter-download.intel.com/akdlm/irc_nas/tec/16533/']
```
boegel: Should be `IRC_NAS` (see also #21108), but download doesn't actually work anymore, which sort of makes sense for these old `imkl` versions...
```suggestion
source_urls = ['https://registrationcenter-download.intel.com/akdlm/IRC_NAS/tec/16533/']

**Review by boegel — CHANGES_REQUESTED** [review]

**Comment by boegel** [comment]
@boegelbot please test @ jsc-zen3
CORE_CNT=16
EB_ARGS="imkl-2020.0.166-gompi-2025b.eb"

**Comment by boegel** [comment]
```
Your system is protected with Security-enhanced Linux* (SELinux).
```

Not going to block this PR over this because on ancient version of `imkl` is failing to install at `jsc-zen3`...

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 4 out of 4 (total: 2 hours 23 mins 57 secs) (4 easyconfigs in total)
node4414.skiddo.os - Linux RHEL 9.6, x86_64, AMD EPYC 9755 128-Core Processor (zen5), Python 3.9.21
See https://gist.github.com/boegel/8b4b0576a36f634546d3642634e37aa1 for a full test report.

boegel (approved):

boegel:
Going in, thanks @hajgato!

PR #25607, opened 2026-03-25
{chem}[gompi/2023b] ORCA v6.1.1 w/ avx2

sassy-crick:
(created using `eb --new-pr`)

This is a bugfix which addresses issue [25553](https://github.com/easybuilders/easybuild-easyconfigs/issues/25553)
Also, `xtb` was removed as it is not required to run `ORCA` per se. It is better if the user loads the module from the same toolchain. That way, any changes to `xtb` etc, can be addressed without rebuilding `ORCA`.

PR #25691, opened 2026-04-08
remove easyconfig for ORCA 6.1.1 with `gompi/2024a`, since it is not compatible with OpenMPI 5.x

### LINKED ISSUE #25553: ORCA-6.1.1-gompi-2024a-avx2: wrong OpenMPI version
One of my users has made me aware there is a problem with `ORCA-6.1.1-gompi-2024a-avx2.eb`: the wrong OpenMPI version is being used!
From the ORCA download page:

```
[ORCA 6.1.1, Linux, x86-64, .tar.xz Archive](https://orcaforum.kofo.mpg.de/app.php/dlext/?view=detail&df_id=267)   
Dynamically linked serial & parallel binaries linked against OpenMPI 4.1.8
orca_6_1_1_linux_x86-64_shared_openmpi418.tar.xz
```
However, `gfbf-2024a` is using the `OpenMPI-5.x` version:
```
 eb -Dr ORCA-6.1.1-gompi-2024a-avx2.eb | grep OpenMPI
 * [x] $CFGS/o/OpenMPI/OpenMPI-5.0.3-GCC-13.3.0.eb (module: OpenMPI/5.0.3-GCC-13.3.0)
```
I remember pretty well at the launched it was said they had problems with `OpenMPI-5.x` and thus were sticking to the `OpenMPI-4.x` version.

The problem I got told is as follow:
```
 If you decide to do testing, this performance degradation is not very reliable (or at least it doesn't depend on anything I've been able to track). Sometimes it does not happen at all, and sometimes jobs that should take 2 hours take multiple days, ie you can only go by whether there are any outliers from a large number of identical jobs (run to run variance notwithstanding).
```
I think that has slipped through at PR #24751 
It might be a good idea if somebody could confirm this, independently from our site but it would be in-line with what the ORCA forum reads.

boegel:
(created using `eb --new-pr`)

This easyconfig is effectively replaced by one using `gompi/2023b` (which still has OpenMPI 4.x):
- https://github.com/easybuilders/easybuild-easyconfigs/pull/25607

fixes #25553


bartoldeman (commented):
lgtm

bartoldeman (approved):

PR #25697, opened 2026-04-08
archive easyconfigs: 2021b generation

Micket:
First step, move 2021b toolchain stuff into __archive__ for later deletion.

PR #25698, opened 2026-04-08
Archive old easyconfigs using `system` toolchain (2021b generation)

Micket:
Should rebased once #25697 is merged.

- [x] #25697 

I expect to find some stuff to move back once CI is done checking for me.


boegel:
I had to bring back 19 of the 59 easyconfigs you have archived, since they're still required as dependency somewhere...

I did so in individual commits, in the hope that this helps with future archiving efforts...

boegel (dismissed):

boegel (approved):

PR #25700, opened 2026-04-09
re-enable `test_dep_graph`

boegel:
reverts #25699

draft PR, since we should actually first fix the issue that made us disable this test, i.e. figure out why `dep_graph` is suddenly a lot slower after the changes made in https://github.com/easybuilders/easybuild-framework/pull/5128


boegel:
@Crivella Do you have any idea what could make `dep_graph` so slow?

See also description in:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/25699

boegel:
ready to merge now fix is merged in framework:
- https://github.com/easybuilders/easybuild-framework/pull/5171

smoors (approved):
lgtm

PR #25739, opened 2026-04-10
 Handle arch-specific dependencies in tests

Flamefire:
When a dependency version depends on the architecture we currently only test the single one for the host the test is running on, likely x86.

This makes tests fail when dependencies are removed as the parsed on dumped list of dependencies will have different lengths. Fix this by choosing another one, if possible.

Additionally always pick a random version when there are multiple versions such that multiple CI runs are likely to test all variants.

Alternative to #25737
Closes #25737


boegel (approved):
lgtm

PR #25760, opened 2026-04-14
{bio}[GCC/13.3.0] TransDecoder v5.7.1

JarneRenders:
(created using `eb --new-pr`)

Do not yet use version 6 in 2024a for compatibility with PASA pipeline.

PR #25761, opened 2026-04-14
Add TransDecoder plugin to PASA

JarneRenders:
(created using `eb --new-pr`)

TransDecoder plugin is missing from PASA since it is bundled as a git submodule. Symlink to the eb installation of TransDecoder instead. Solves a 'bug' in funannotate since its function `runPASAtrain` will call PASA with the `--TRANSDECODER` flag.

Depends on:
  - [x] #25760


smoors:

smoors (approved):
lgtm

smoors:
Going in, thanks @JarneRenders!

PR #25767, opened 2026-04-15
{bio,data}[gfbf/2025b] DIA-NN v2.5.0, polars v1.28.1

SethosII:
(created using `eb --new-pr`)

The EasyConfig for `polars` is just copied from an existing one. How are these Rust EasyConfigs handled?


verdurin:
@SethosII cf. https://docs.easybuild.io/writing-easyconfig-files/?h=rust+crate#module_extensions

PR #25877, opened 2026-04-29
Add BabelStream-5.0-rocm-compilers-19.0.0-ROCm-6.4.1.eb

zerefwayne:
https://github.com/uob-hpc/BabelStream

### DIFF
--- a/easybuild/easyconfigs/b/BabelStream/BabelStream-5.0-rocm-compilers-19.0.0-ROCm-6.4.1.eb
+++ b/easybuild/easyconfigs/b/BabelStream/BabelStream-5.0-rocm-compilers-19.0.0-ROCm-6.4.1.eb
@@ -4,6 +4,7 @@ easyblock = 'ROCmComponent'
 
 name = 'BabelStream'
 version = '5.0'
+_rocm_version = '6.4.1'
 
 homepage = 'https://github.com/uob-hpc/BabelStream'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/b/BabelStream/BabelStream-5.0-rocm-compilers-19.0.0-ROCm-6.4.1.eb
```diff
@@ -0,0 +1,34 @@
+# Author:   Aayush Joglekar <aayush.joglekar@surf.nl>
+
+easyblock = 'ROCmComponent'
```
Thyre, line 3: We should update the CMakeMake easyblock to handle these ROCm EasyConfigs. For CUDA EasyConfigs, we also just rely on that easyblock.

Right now, we're setting options that are not necessary for the software we're building.
The options in `ROCmComponent` are mostly important for the actual ROCm components.

See this block I implemented back when starting to work on the ROCm stuff for a first draft:
https://github.com/Thyre/easybuild-custom/blob/71a430408ef0ea530321d8159006749cc9bc0b26/easybuild/easyblocks/generic/cmakemake.py#L365-L382

This is not perfect, as this e.g. fails if both `cuda_compute_capabilities` and `amgcn_capabilities` are set. We might want to make this dependent on `CMAKE_HIP_PLATFORM`, which can only be set to one option at a time (which we then could do via an easyblock option).

PR #25888, opened 2026-04-29
{system}[GCCcore/13.2.0] NotoSansCJK-Bundle v2.004

stamenminkov:
AI usage: GPT-4 for the proper install and sanity check commands


stamenminkov:
I'm not sure if calling `matplotlib` like this is "correct", but since I need it only for testing that the fonts can be used (only for sanity), I'm leaving the recipe as `GCCcore` toolchain

PR #25910, opened 2026-05-04
Add rocm support to OSU-rompi

zerefwayne:
- [ ] Merge #25902 first


zerefwayne:
Will be added with https://github.com/easybuilders/easybuild-easyconfigs/pull/25902

PR #25919, opened 2026-05-06
Add PTX code for last/highest CUDA arch for easyconfigs using `TORCH_CUDA_ARCH_LIST`

Flamefire:
With https://github.com/easybuilders/easybuild-framework/pull/5144 it will add PTX code for the highest arch

See also https://github.com/easybuilders/easybuild-easyblocks/pull/4129

It avoids failures with the new CUDA sanity check checking for this.
No change to performance etc, it just causes slightly bigger binaries due to the addition (i.e. not replacement) of PTX code.

This assumes that `cuda_cc_semicolon_sep` is non-empty, which IMO is a safe assumption.

### DIFF
--- a/easybuild/easyconfigs/p/PyTorch-bundle/PyTorch-bundle-1.12.1-foss-2022a-CUDA-11.7.0.eb
+++ b/easybuild/easyconfigs/p/PyTorch-bundle/PyTorch-bundle-1.12.1-foss-2022a-CUDA-11.7.0.eb
@@ -45,12 +45,12 @@ exts_list = [
     }),
     ('torchdata', '0.4.1', {
         'source_urls': ['https://github.com/pytorch/data/archive'],
-        'sources': [{'download_filename': 'v%(version)s.tar.gz', 'filename': '%(name)s-%(version)s.tar.gz'}],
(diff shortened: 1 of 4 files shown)


Flamefire:
Test report by @Flamefire
**FAILED**
Build succeeded for 2 out of 4 (total: 3 hours 59 mins 51 secs) (4 easyconfigs in total)
i8019 - Linux Rocky Linux 9.6, x86_64, AMD EPYC 7352 24-Core Processor (zen2), 8 x NVIDIA NVIDIA A100-SXM4-40GB, 580.65.06, Python 3.9.21
See https://gist.github.com/Flamefire/8a2701ea27007e8a2963c4414638a62f for a full test report.

Flamefire:
The pip list check for Python packages errors on this:
> torchtext 0.13.1 (version in 'pip list' output: 0.13.1a0) (took 10 mins 59 secs)

Same for similar ones: The filename and tag are missing the a0 suffix
I would either remove suffixes like `(a|b|pre)\d+`  before comparison or provide a way for an easyconfig to specify the pip-version which could then here be %(version)sa0

PR #25951, opened 2026-05-10
Add mars-earth easyconfig

edithatogo:
Add an H0-only `mars-earth` easyconfig to easybuild-easyconfigs.

The recipe is scoped to source installability and keeps higher-level accelerator, MPI, and distributed claims out of scope.

Validation:
- Python syntax check on the new easyconfig
- `git diff --check`

### DIFF
--- a/easybuild/easyconfigs/m/mars-earth/mars-earth-1.0.4-gfbf-2024a.eb
+++ b/easybuild/easyconfigs/m/mars-earth/mars-earth-1.0.4-gfbf-2024a.eb
@@ -1,0 +1,24 @@
+easyblock = 'PythonPackage'
+
+name = 'mars-earth'
+version = '1.0.4'
+
+homepage = 'https://github.com/edithatogo/mars'
+description = """Pure Python implementation of multivariate adaptive regression splines
+with a Rust-backed runtime core."""
+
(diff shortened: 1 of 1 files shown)


boegel:
@edithatogo Can you provide some context on what `mars-earth` does, and why you're interested in adding support for installing it with EasyBuild?

edithatogo:
Thanks, and apologies for leaving this unanswered. `mars-earth` is the PyPI distribution providing the `pymars`/`Earth` interface for multivariate adaptive regression splines: fitting piecewise regression models with hinge functions, interaction terms and pruning, with a scikit-learn-style estimator API.

The intended EasyBuild use case is to install that estimator into an existing site-managed Python/module stack for regression and surrogate-modelling workflows, with a pinned source archive and explicit scientific-Python dependencies. It is not a claim that this implementation needs a cluster, or that particular sites have deployed it.

This PR is deliberately limited to a source-installation recipe (the “H0” label in the description). It does not request MPI, distributed execution or accelerator support, and it does not establish HPC performance or scaling. The relevant project is https://github.com/edithatogo/mars and its published distribution is https://pypi.org/project/mars-earth/ . If that scope does not fit the EasyBuild inclusion policy, I understand; the submission is a packaging proposal, not evidence of adoption.

PR #25960, opened 2026-05-11
add patch for Underworld 2.16.4 to increase timeout in tests

sassy-crick:
(created using `eb --new-pr`)

Without this patch, the tests in `EESSI` are failing due to a timeout. Increasing the timeout solves this problem. 

Required for `EESSI` PR: [1454](https://github.com/EESSI/software-layer/pull/1454)


sassy-crick:

sassy-crick:

boegel:
@sassy-crick Do you know the cause of the tests being slow?

Is it because of the overhead caused by the writable overlay that is used in the EESSI build environment?

sassy-crick:
> @sassy-crick Do you know the cause of the tests being slow?
> 
> Is it because of the overhead caused by the writable overlay that is used in the EESSI build environment?

Not really. I had the same issue when I was building it on my laptop on top of a native `EESSI` installation, i.e. not inside a container, without having a local proxy. My hunch is it might be just a wee bit too slow, and then the default timeout kicks in. Thus I was quite generous with the timeout, just in case. It all builds fine without the patch in a local build environment. Do you want me to dig deeper into that?

boegel:
> > @sassy-crick Do you know the cause of the tests being slow?
> > Is it because of the overhead caused by the writable overlay that is used in the EESSI build environment?
> 
> Not really. I had the same issue when I was building it on my laptop on top of a native `EESSI` installation, i.e. not inside a container, without having a local proxy. My hunch is it might be just a wee bit too slow, and then the default timeout kicks in. Thus I was quite generous with the timeout, just in case. It all builds fine without the patch in a local build environment. Do you want me to dig deeper into that?

No, it's fine, I don't think this is hugely controversial...

boegel:
Test report by @boegel
**SUCCESS**
Build succeeded for 1 out of 1 (total: 23 mins 57 secs) (1 easyconfigs in total)
node4219.shinx.os - Linux RHEL 9.6, x86_64, AMD EPYC 9654 96-Core Processor (zen4), Python 3.9.21
See https://gist.github.com/boegel/4d39ff16052a76a5f34e85e47756a144 for a full test report.

sassy-crick:
> > > @sassy-crick Do you know the cause of the tests being slow?
> > > Is it because of the overhead caused by the writable overlay that is used in the EESSI build environment?
> > 
> > 
> > Not really. I had the same issue when I was building it on my laptop on top of a native `EESSI` installation, i.e. not inside a container, without having a local proxy. My hunch is it might be just a wee bit too slow, and then the default timeout kicks in. Thus I was quite generous with the timeout, just in case. It all builds fine without the patch in a local build environment. Do you want me to dig deeper into that?
> 
> No, it's fine, I don't think this is hugely controversial...

I done a bit of testing: setting the timeout to 340 is still too low, the default is 300. Setting it to 350 works well it seems, at least on my machine. So as I suspected, it just needs a little bit longer. 
What I noticed it: it is using all available cores, virtual and real, during the tests. I had a look at `nbmake` but could not really find a way to limit that to the number of cores I am giving. 
Note: I run the jobs within the EasyBuild-Shell, that might be the real issue here.

boegel (approved):
lgtm

boegel:

PR #25914, opened 2026-05-04
{lib}[gompi/2023b] libseccomp v2.5.5

stamenminkov:
No AI usage

PR #25967, opened 2026-05-12
{tools}[gompi/2023b] Apptainer v1.5.0

stamenminkov:
No AI used

DEPENDS ON: https://github.com/easybuilders/easybuild-easyconfigs/pull/25914 (because of a `gompi`-`libseccomp`)

### DIFF
--- a/easybuild/easyconfigs/a/Apptainer/Apptainer-1.5.0-gompi-2023b.eb
+++ b/easybuild/easyconfigs/a/Apptainer/Apptainer-1.5.0-gompi-2023b.eb
@@ -3,6 +3,7 @@
 # Barcelona Supercomputing Center, 2026
 
 easyblock = 'ConfigureMake'
+
 name = 'Apptainer'
 version = '1.5.0'
 
@@ -26,7 +27,7 @@ dependencies = [
     ('libseccomp', '2.5.5'),
 ]
 
-osdependencies = [('cryptsetup')]
+osdependencies = [('cryptsetup', 'fuse2fs', 'mksquashfs', 'unsquashfs')]
 
 # configure to build without root
 configure_cmd = './mconfig'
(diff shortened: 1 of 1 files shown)


asp8200:
Hi @stamenminkov 👋  Thanks for making this PR. I wonder which OS-dependencies the resulting Apptainer load-module will have 🤔  Could it be that it will still require tools and libraries like fuse-overlayfs, fuse2fs, mksquashfs unsquashfs, squashfuse and fakeroot?

I'm asking because we recently experienced an issue where our Apptainer load-module wasn't working properly on some of our compute-nodes - seemingly because squashfuse and fuse-overlayfs where missing on those compute-nodes.

stamenminkov:
> Hi @stamenminkov 👋 Thanks for making this PR. I wonder which OS-dependencies the resulting Apptainer load-module will have 🤔 Could it be that it will still require tools and libraries like fuse-overlayfs, fuse2fs, mksquashfs unsquashfs, squashfuse and fakeroot?
> 
> I'm asking because we recently experienced an issue where our Apptainer load-module wasn't working properly on some of our compute-nodes - seemingly because squashfuse and fuse-overlayfs where missing on those compute-nodes.

Hi @asp8200, ty for pointing out this, we missed to include any of these in the recipe, since Apptainer worked on the first go :sweat_smile: 

I just checked and on one machine we have `fuse-overlayfs`, `fuse2fs`, `mksquashfs` and `unsquashfs`, while on another one we have the latter 3, `fuse2fs`, `mksquashfs` and `unsquashfs`. 

`squashfuse` and `fakeroot` are not present on any of our machines.

We haven't observed any significant difference between our 2 environments so far :thinking: 

I'll update the recipe with `fuse2fs`, `mksquashfs` and `unsquashfs`, and regarding `fuse-overlayfs` I'm not sure, since one of the module still works fine without it.

What do you think?

Cheers!

asp8200:
I see one of the CI-tests failing with
```
Missing dependencies: libseccomp/2.5.5-gompi-2023b (no easyconfig file or existing module found)
```

I guess [this](https://github.com/easybuilders/easybuild-easyconfigs/pull/25914) PR just needs to be merged before this can be approved and merged.

sassy-crick:
I don't think software like `apptainer` should be in EasyBuild, as it requires `root` access. I think that should really come from the OS. This also means it is the distros' responsibility to keep it safe.

boegel:
See also discussion in:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/6221

PR #25974, opened 2026-05-13
tweak configuration options and customize sanity check paths for Qt6 6.9.3 for RISC-V

julianmorillo:


julianmorillo:
Successful build here: https://github.com/EESSI/dev.eessi.io-riscv/pull/102#issuecomment-4600855840

easybuild/easyconfigs/q/Qt6/Qt6-6.9.3-GCCcore-14.3.0.eb
```diff
@@ -96,10 +98,27 @@ configopts += "-DPython3_ROOT_DIR=$EBROOTPYTHON "
...
 
-sanity_check_paths = {
-    'files': ['bin/qmake6', 'lib/libQt6Core.%s' % SHLIB_EXT, 'lib/libQt6WebEngineCore.%s' % SHLIB_EXT],
-    'dirs': ['include/QtCore', 'include/QtWebEngineCore'],
-}
+# disable building the Qt Quick 3D Physics submodule for RISC-V
+# avoid using the bundled FFmpeg
+# disable building QtWebEngine
+if ARCH == 'riscv64':
+    configopts += "-DBUILD_qtquick3dphysics=OFF "
```
casparvl, line 103: I'm confused: your patch seems to be for `qtquick3dphysics`, but then you seem to disable it here. Would _one_ of the two (i.e. either patching OR disabling the component) not be enough?

easybuild/easyconfigs/q/Qt6/Qt6-6.9.3-GCCcore-14.3.0.eb
```diff
@@ -96,10 +98,27 @@ configopts += "-DPython3_ROOT_DIR=$EBROOTPYTHON "
...
+    configopts += "-DQT_FEATURE_webengine_system_ffmpeg=ON "
+    configopts += "-DBUILD_qtwebengine=OFF "
```
casparvl, line 105: I think this is quite a commonly used component, isn't it? I'm assuming that this means there was no way to get this working on RISC-V?

Thyre: It certainly means that e.g. PySide6 will not build, blocking everything that relies on that. Don't know about PyQt6.

easybuild/easyconfigs/q/Qt6/Qt6-6.9.3-GCCcore-14.3.0.eb
julianmorillo: Removed the patch. 
It was just the first step to allow the building of the component in riscv64 but then it came out that the component is not implemented at all for RISC-V, and needed to be disabled.

easybuild/easyconfigs/q/Qt6/Qt6-6.9.3-GCCcore-14.3.0.eb
julianmorillo: @casparvl , unfortunately I didn't found the way to get those components working in RISC-V: basically they lack a RISC-V implementation, it is not a matter of patching here and there.

PR #25992, opened 2026-05-13
Add vg 1.74.0 easyconfig

machavolu2025:
Adds an EasyBuild easyconfig for vg 1.74.0 using the official upstream binary release.

Tested successfully:
- module load vg/1.74.0
- vg version
- vg help
- graph construction from FASTA
- graph validation
- GFA conversion
- XG/GCSA indexing
- FASTQ read mapping
- GAM statistics
- GAM to JSON conversion

### DIFF
--- a/easybuild/easyconfigs/v/vg/vg-1.74.0.eb
+++ b/easybuild/easyconfigs/v/vg/vg-1.74.0.eb
@@ -1,0 +1,27 @@
+easyblock = 'Binary'
+
+name = 'vg'
+version = '1.74.0'
+
+homepage = 'https://github.com/vgteam/vg'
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/v/vg/vg-1.74.0.eb
boegel: Source file should be renamed after download, or we'll have trouble when writing easyconfigs for other versions.

```suggestion
sources = [{'download_filename': 'vg', 'filename': 'vg-%(version)s}]

`install_cmd` below needs to be changed accordingly

boegel (changes requested):

PR #25994, opened 2026-05-14
{vis}[GCCcore/13.2.0] Meld v3.22.2, GtkSourceView4 v4.8.4, PyGObject-C v3.46.0, ... w/ GNOME

stamenminkov:
AI usage: `GPT-4` used extensively for correcting the easyconfigs, creating the `Meld` patch, and also finding out the necessity of using a `pygobject-C` library instead of just the `python` wrappers

`pygobject-GNOME` can also be named `pygobject-C` or whichever makes more sense


sassy-crick:

sassy-crick:
@stamenminkov 
There seems to be a problem with the downloads of `gtksourceview4-4.8.4.tar.xz` and `pygobject-c-3.46.0.tar.xz`.
Would you mind to check if the URLs are still working?

stamenminkov:
> @stamenminkov There seems to be a problem with the downloads of `gtksourceview4-4.8.4.tar.xz` and `pygobject-c-3.46.0.tar.xz`. Would you mind to check if the URLs are still working?

@sassy-crick ahhh they do exist but they're named differently:

`gtksourceview-4.8.4.tar.xz` and 
`pygobject-3.46.0.tar.xz`

I'm correcting this now, hopefully that will be the only issue:)

stamenminkov:
Btw, I'm installing `Meld` as a `PythonPackage` here, which worked well, though not sure if that's the "correct" way, as it's supposed to be a `MesonNinja` package :thinking:

easybuild/easyconfigs/m/Meld/Meld-3.22.2-GCCcore-13.2.0.eb
```diff
@@ -0,0 +1,62 @@
...
+    ('GTK3', '3.24.39'),
+    ('GtkSourceView4', '4.8.4'),
+]
+
+# Ensuring Meld install prefix for the Datadir from patch
+preinstallopts = 'export MELD_INSTALL_PREFIX=%(installdir)s && '
+
+sanity_check_commands = [
+    'python3 -c "import meld"',
+    'meld --version',
```
stamenminkov, line 54: Also this might not be the cleanest call, as it depends on having active display settings 🤔 

is it good to leave it? I mean the `'meld --version'` path

PR #25995, opened 2026-05-14
{data,lib}[foss/2023b] cfchecker v4.1.0, cfunits v3.3.7, cftime v1.6.4

stamenminkov:
No AI usage


sassy-crick:

easybuild/easyconfigs/c/cfunits/cfunits-3.3.7-foss-2023b.eb
```diff
@@ -0,0 +1,25 @@
...
+checksums = ['e271ae2c01765cc437e7eefd802846894367179e3208653ab9b520ade48286d6']
+
+dependencies = [
+    ('Python', '3.11.5'),
+    ('UDUNITS', '2.2.28'),
+    ('SciPy-bundle', '2023.11'),
+    ('netcdf4-python', '1.7.1.post2'),  # needed for cftime
+]
+
+moduleclass = 'lib'
```
Crivella:
```suggestion
easyblock = 'PythonBundle'

name = 'cfunits'
version = '3.3.7'

homepage = 'https://github.com/NCAS-CMS/cfunits'
description = """A Python interface to UNIDATA's UDUNITS-2 package with CF extensions.
 Store, combine and compare physical units and convert numeric values to different units."""

toolchain = {'name': 'foss', 'version': '2023b'}

dependencies = [
    ('Python', '3.11.5'),
    ('UDUNITS', '2.2.28'),
    ('SciPy-bundle', '2023.11'),
    ('netcdf4-python', '1.7.1.post2'),  # needed for cftime
]

exts_list = [
    (name, version, {
        'checksums': ['e271ae2c01765cc437e7eefd802846894367179e3208653ab9b520ade48286d6'],
    }),
]

moduleclass = 'lib'
```

Crivella (changes requested):
In general we are trying to move to using Bundles also for single packages.

I added an example to convert one of the 2 ECs to one, could you convert also the other?

easybuild/easyconfigs/c/cfunits/cfunits-3.3.7-foss-2023b.eb
stamenminkov: I see, I still have plenty of old `PythonPackage` recipes which I'll have to change.

Thanks!

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 2 out of 2 (total: 18 secs) (2 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), 1 x NVIDIA NVIDIA GeForce RTX 5060, 580.159.03, Python 3.11.15
See https://gist.github.com/Crivella/e50ee7f59ba6e443e55cfe83b07a7b2a for a full test report.

Crivella (changes requested):
There is already an EC for `cfunits` with the same version in 2024a.
https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/cfunits/cfunits-3.3.7-gfbf-2024a.eb

There `cftime` is added as an extension instead of through `ntecdf4-python` which allows to keep the TC to a lower level (`gfbf` instead of `foss`).

Is there a reason here it was done in a such a manner?
It not i would follow the same approach as the existing EC as avoids introducing a lot of not needed dependencies

stamenminkov:
> There is already an EC for `cfunits` with the same version in 2024a. https://github.com/easybuilders/easybuild-easyconfigs/blob/develop/easybuild/easyconfigs/c/cfunits/cfunits-3.3.7-gfbf-2024a.eb
> 
> There `cftime` is added as an extension instead of through `ntecdf4-python` which allows to keep the TC to a lower level (`gfbf` instead of `foss`).
> 
> Is there a reason here it was done in a such a manner? It not i would follow the same approach as the existing EC as avoids introducing a lot of not needed dependencies

@Crivella Ah I see, my thought was just to use an already installed package rather than installing it again as an extension.

It does make sense though to keep it down to `gfbf`, instead of `foss`, I agree.

In that case, though, would it make more sense to revert the dependencies? We install `cftime` here as an extension to `cfunits`, and then we call `cfunits` as a dependency to `netcdf4-python`, and this way we maintain the balance in the toolchain. What do you think? Or rather have `cftime` as a standalone module that can be called as a dependency to both directions. 

What is confusing to me is having the same library repeated multiple times as an extension across different modules, feels a bit unorganised and to me it doesn't really fit the idea of a "single" version per toolchain, as we can end up with different versions also across these multiple modules.

Crivella:
In general if a package is used multiple times as a dependency i agree it would be best to add it as a separate dependency for improved re-usability, and ensuring we do not ship multiple versions in a TC.

There are currently talks on how we can improve this, either through CI checks or other methods.

I think in general it is about finding a balance on how to granularize bundles (i think the original idea was to avoide having a huge amount of entries in the PATH-like environment variables)
See also https://github.com/easybuilders/easybuild-easyblocks/pull/4146#issuecomment-4671894771 for a similar conversation on Julia packages

In this case whether we want to split `cftime` as another EC or not we definitely want to keep this EC on the lowest TC level possible.

stamenminkov:
> In general if a package is used multiple times as a dependency i agree it would be best to add it as a separate dependency for improved re-usability, and ensuring we do not ship multiple versions in a TC.
> 
> There are currently talks on how we can improve this, either through CI checks or other methods.
> 
> I think in general it is about finding a balance on how to granularize bundles (i think the original idea was to avoide having a huge amount of entries in the PATH-like environment variables) See also [easybuilders/easybuild-easyblocks#4146 (comment)](https://github.com/easybuilders/easybuild-easyblocks/pull/4146#issuecomment-4671894771) for a similar conversation on Julia packages
> 
> In this case whether we want to split `cftime` as another EC or not we definitely want to keep this EC on the lowest TC level possible.

Okay, I think for the sake of this PR I'll separate `cftime` as another EC, and then adapt `cfchecker` and `cfunits`.

@Crivella Regarding `netcdf4-python`, I can add a corrected version also here in this same PR... or maybe we should open a new one that covers all the available tooclhains so the fix is applied everywhere?

Crivella:
I would not do it in this PR and keep it for a separate one

stamenminkov:
> I would not do it in this PR and keep it for a separate one

Perfect, I'll open it when I have some time;)

Crivella (changes requested):
Judging from `cftime` docs and the diff with the previous `cfunits` EC we will probably need Cython to make sure the build works atleast for `cftime`

EDIT:
```
== 2026-06-22 17:46:45,931 run.py:650 INFO Output of '/home/crivella/.local/easybuild/software/Python/3.11.5-GCCcore-13.2.0/bin/python ...' shell command (stdout + stderr):
Using pip 23.2.1 from /home/crivella/.local/easybuild/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/site-packages/pip (python 3.11)
Processing /home/crivella/.local/easybuild/build/cftime/1.6.4/gfbf-2023b/cftime/cftime-1.6.4
  Preparing metadata (pyproject.toml): started
  Running command Preparing metadata (pyproject.toml)
  Compiling /home/crivella/.local/easybuild/build/cftime/1.6.4/gfbf-2023b/cftime/cftime-1.6.4/src/cftime/_cftime.pyx because it changed.
  [1/1] Cythonizing /home/crivella/.local/easybuild/build/cftime/1.6.4/gfbf-2023b/cftime/cftime-1.6.4/src/cftime/_cftime.pyx
  running dist_info
```

I guess on my system it picked it up from somewhere else but we should definitely add it as a builddep

stamenminkov:
> Judging from `cftime` docs and the diff with the previous `cfunits` EC we will probably need Cython to make sure the build works atleast for `cftime`
> 
> EDIT:
> 
> ```
> == 2026-06-22 17:46:45,931 run.py:650 INFO Output of '/home/crivella/.local/easybuild/software/Python/3.11.5-GCCcore-13.2.0/bin/python ...' shell command (stdout + stderr):
> Using pip 23.2.1 from /home/crivella/.local/easybuild/software/Python/3.11.5-GCCcore-13.2.0/lib/python3.11/site-packages/pip (python 3.11)
> Processing /home/crivella/.local/easybuild/build/cftime/1.6.4/gfbf-2023b/cftime/cftime-1.6.4
>   Preparing metadata (pyproject.toml): started
>   Running command Preparing metadata (pyproject.toml)
>   Compiling /home/crivella/.local/easybuild/build/cftime/1.6.4/gfbf-2023b/cftime/cftime-1.6.4/src/cftime/_cftime.pyx because it changed.
>   [1/1] Cythonizing /home/crivella/.local/easybuild/build/cftime/1.6.4/gfbf-2023b/cftime/cftime-1.6.4/src/cftime/_cftime.pyx
>   running dist_info
> ```
> 
> I guess on my system it picked it up from somewhere else but we should definitely add it as a builddep

Emmmmmmm true, you're right:) Just added it now

That's quite confusing though, since early ECs for `netcdf4-python` don't use it, I see it used only after `2024a`; plus my builds succeed without calling it and I definitely don't have a system lib for `Cython` :thinking: So I don't understand haha

But it has to be there, checking the `setup.py` and `pyproject.toml` of the package, it's clearly stated there..

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 3 out of 3 (total: 37 secs) (3 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), 1 x NVIDIA NVIDIA GeForce RTX 5060, 580.159.03, Python 3.11.15
See https://gist.github.com/Crivella/f7e0ce1630b9b4010342faebd9db70f7 for a full test report.

Crivella:

Crivella (approved):
LGTM

Crivella:
Going in, thanks @stamenminkov!

PR #26002, opened 2026-05-15
{system}[GCCcore/13.2.0] Linux-PAM v1.5.3

stamenminkov:
Ai usage: `GPT-5` for polishing dependencies and configopts


sassy-crick:
Should that not be really coming from the OS, or is that a different `pam` ??

stamenminkov:
> Should that not be really coming from the OS, or is that a different `pam` ??

@sassy-crick Well I guess it's the same `PAM`, but I've encountered a few machines already which don't come with it by default, so just to not depend always on sysadmins I think it's a good option to have... :grin:  Unless there are some security concerns about installing it this way (?)

PR #26042, opened 2026-05-20
{lib}[GCCcore/14.2.0] pyodbc v5.2.0, psqlODBC v17.00.00005, unixODBC v2.3.14

stamenminkov:
No AI usage


stamenminkov:
Now added everything as an `EasyBuild` dependency

PR #26055, opened 2026-05-21
{lib}[GCCcore/13.2.0] PyJWT v2.8.0

stamenminkov:
No AI usage

PR #26053, opened 2026-05-21
{lib}[GCCcore/13.2.0] FastAPI v2024.05, rich v13.7.0

stamenminkov:
No AI usage

**Depends on:**

- [x] **https://github.com/easybuilders/easybuild-easyconfigs/pull/26055 for `PyJWT`**

PR #26049, opened 2026-05-21
{tools}[GCCcore/13.2.0] googleapis-python v2024.02

stamenminkov:
No AI usage, based off an existing recipe

PR #26056, opened 2026-05-21
{ai,tools}[foss/2023b,gfbf/2023b] litellm v1.20.0, azure-bundle-llms v2024.02

stamenminkov:
AI usage: most part of the files has been done manually, used `GPT-5` for optimization of the script, a bit more clarity, etc.

**IMPORTANT!
It depends on the following:**

- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/26055 for `PyJWT`, as part of `azure-bundle-llms` and `fastapi-bundle`

- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/26053 for `fastapi-bundle`
- [x] https://github.com/easybuilders/easybuild-easyconfigs/pull/26049 for `googleapis-python`


stamenminkov:
**Question:** the existing `openai-python` recipe for `2023b` is `foss`, however I think with `gfbf` it should be enough. That's the only reason why `litellm` here is also `foss`, but I could submit a `gfbf` version of `openai-python`... Should I?

PR #26080, opened 2026-05-23
{dataset}[system/system] CIFAR-10-data v20240104 w/ hf-2025b

VRehnberg:
(created using `eb --new-pr`)

requires:
- [ ] https://github.com/easybuilders/easybuild-easyblocks/pull/4141
- [ ] https://github.com/easybuilders/easybuild-easyconfigs/pull/26241


VRehnberg:
Test report: https://github.com/easybuilders/easybuild-easyblocks/pull/4141#issuecomment-4525659222

VRehnberg:
CI test failures seem to be because it doesn't find the easyblock.

VRehnberg:
This would need some Pillow version to load instances from the dataset as well. But, I'm not sure if Pillow and Pillow-SIMD interfere with each other or not.

Alternatively, you could change HF-Datasets to a builddep and instead adding the exact version to the versionsuffix instead.

VRehnberg:
After discussion with Micke I will change these datasets a bit.

PR #26015, opened 2026-05-18
{chem}[gompi/2025a] ORCA v6.1.1 w/ avx2

JarneRenders:
(created using `eb --new-pr`)

PR #24988, opened 2026-01-02
{chem}[gfbf/2025a] CPCM-X v1.1.0, dftd4 v3.7.0, mstore v0.3.0, multicharge v0.3.1, numsa v0.2.0, ORCA v6.1.1, Simple-DFTD3 v1.2.1, tblite v0.4.0, xtb v6.7.1 

stefan-wolfsheimer:

PR #26084, opened 2026-05-26
remove easyconfig for ORCA-6.1.1 w/ `gompi/2025a` toolchain, as performance degradation is seen with OpenMPI-5, see issue #25553

sassy-crick:
(created using `eb --new-pr`)

Related to already merged PR #26015 and also relevant to #24988


ocaisa:
Notifying @JarneRenders and @smoors as they were involved in the PR that introduced this.

ocaisa:
There is an issue that describes the performance problem: https://github.com/easybuilders/easybuild-easyconfigs/issues/25553

ocaisa (approved):
Based on the discussion in https://github.com/easybuilders/easybuild-easyconfigs/issues/25553, removing this seems like a good idea

JarneRenders:
Thanks @sassy-crick and @ocaisa! I missed this when submitting the PR. So far no complaints from our users, but I will update them on the situation.

ocaisa:
It _could_ be that the regression is fixed with the later version of OpenMPI...unfortunately without it being easily reproducible it is very hard to check!

PR #26088, opened 2026-05-26
{bio}[GCC/13.3.0] Cyrius v1.1.1

nicorap:
from --pr

### DIFF
--- a/easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
+++ b/easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
@@ -17,10 +17,6 @@ dependencies = [
     ('SciPy-bundle', '2024.05', '', ('gfbf', '2024a')),
 ]
 
-exts_default_options = {
-    'source_urls': [PYPI_SOURCE],
-}
-
 exts_list = [
     ('cyrius', version, {
         'installopts': '--no-deps',
--- a/easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
+++ b/easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
@@ -19,7 +19,6 @@ dependencies = [
 
 exts_default_options = {
     'source_urls': [PYPI_SOURCE],
-    'use_pip': True,
 }
 
 exts_list = [
--- a/easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
+++ b/easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
@@ -17,9 +17,6 @@ dependencies = [
     ('SciPy-bundle', '2024.05', '', ('gfbf', '2024a')),
 ]
 
-use_pip = True
-sanity_pip_check = True
-
 exts_default_options = {
     'source_urls': [PYPI_SOURCE],
     'use_pip': True,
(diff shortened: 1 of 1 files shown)


easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
```diff
@@ -0,0 +1,52 @@
...
+    ('SAMtools', '1.21'),
+    ('Pysam', '0.22.1'),
+    ('statsmodels', '0.14.4', '', ('gfbf', '2024a')),
+    ('SciPy-bundle', '2024.05', '', ('gfbf', '2024a')),
+]
+
+exts_default_options = {
+    'source_urls': [PYPI_SOURCE],
+}
+
```
boegel: No need for this, that's the default when using `PythonBundle`

```suggestion
```

easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
```diff
@@ -0,0 +1,52 @@
...
+exts_list = [
+    ('cyrius', version, {
+        'installopts': '--no-deps',
```
boegel: No need for this, that's included by default by `PythonPackage` easyblock (which is used to install extensions in a `PythonBundle` easyconfig)

easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
```diff
@@ -0,0 +1,52 @@
...
+        'checksums': ['18fe5ac94f0cbf0641ca76f40439d7186542bef1e9e10f0530d43c7e7fd9bb94'],
+    }),
+]
+
+postinstallcmds = [
+    "printf '%s\\n' '#!/usr/bin/env bash' "
+    "'exec cyrius \"$@\"' "
+    "> %(installdir)s/bin/star_caller.py",
```
boegel, line 30: This comes a bit out of nowhere, why does it make sense to have a script named `star_caller.py` which is actually a shell script?!

At the very least, a comment should be included in the easyconfig on top of `postinstallcmds` to explain this...

easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
```diff
@@ -0,0 +1,52 @@
...
+        'bin/cyrius',
+        'bin/star_caller.py',
+    ],
+    'dirs': [],
+}
+
+sanity_check_commands = [
+    'cyrius -h',
+    'star_caller.py -h',
+    'python -c "import pysam; print(pysam.__version__)"',
```
boegel: I don't see the point of doing this here, this belongs in `Pysam` dependency?

boegel (changes requested):

easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
nicorap: true. removed it.

easybuild/easyconfigs/c/Cyrius/Cyrius-1.1.1-GCC-13.3.0.eb
nicorap: star_caller.py is the upstream entry point; wrapped cyrius binary to preserve the name  and put a comment

PR #26136, opened 2026-06-01
Flux: adding binutils as build dependency for RiscV build

hugomeiland:
trying to build Flux:
$ eb Flux-0.80.0-GCC-14.3.0.eb --robot --optarch=-march=rv64imafdc

== 2026-06-01 05:49:08,381 build_log.py:233 ERROR EasyBuild encountered an error: get_software_root software root for binutils was not found in environment (at easybuild/tools/toolchain/toolchain.py:465 in _get_software_root)

but binutils is not listed as a dependency...


Thyre:
`binutils` should typically be pulled in by `GCC`, except in cases where it is filtered.
Is there any more info in the build log about what expects `binutils` to be present?
The Bundle EasyBlock shouldn't expect `binutils` to exist.

hugomeiland:
yes, i'm running now with EESSI-extend, this seems to solve it...closing...

PR #26139, opened 2026-06-01
adding easyconfigs: PharmCAT-3.1.1-GCC-13.3.0.eb

nicorap:
(created using eb --new-pr)

Note: the review fixes were prepared with AI assistance (Claude, Opus 4.8 and copilot). Sources were downloaded from the official PharmCAT v3.1.1 GitHub release and checksums computed locally; I've reviewed the result.

### DIFF
--- a/easybuild/easyconfigs/p/PharmCAT/PharmCAT-3.1.1-GCC-13.3.0.eb
+++ b/easybuild/easyconfigs/p/PharmCAT/PharmCAT-3.1.1-GCC-13.3.0.eb
@@ -1,0 +1,128 @@
+easyblock = 'Binary'
+
+name = 'PharmCAT'
+version = '3.1.1'
+
+homepage = 'https://pharmcat.clinpgx.org'
+description = """Pharmacogenomics Clinical Annotation Tool (PharmCAT): core Java CLI plus VCF preprocessor & optional pipeline."""
+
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/p/PharmCAT/PharmCAT-3.1.1-GCC-13.3.0.eb
```diff
@@ -0,0 +1,128 @@
...
+    f'pharmcat-preprocessor-{version}.tar.gz',
+    f'pharmcat-pipeline-{version}.tar.gz',
+]
+
+checksums = []
+
+dependencies = [
+    ('Java', '17', '', SYSTEM),
+    ('Python', '3.12.3'),
+    ('SciPy-bundle', '2024.05', '', ('gfbf', '2024a')),
```
jfgrimm: if you're depending on `gfbf`-level dependencies at runtime, then this easyconfig should also be using the `gfbf` toolchain, not `GCC`

jfgrimm (changes requested):

easybuild/easyconfigs/p/PharmCAT/PharmCAT-3.1.1-GCC-13.3.0.eb
nicorap: yes, I updated the config now.

PR #26141, opened 2026-06-01
{lang}[foss/2025b] RStudio-Server v2026.01.2

bedroesb:


verdurin:
@bedroesb what do you think is needed for this to be review-ready?

bedroesb:
@verdurin thanks a lot for reaching out, I am compiling it on our node to test it out, but it is taking ages. But I could take it out of draft so the bots can test it :)

easybuild/easyconfigs/r/RStudio-Server/RStudio-Server-2025.09.2+418-foss-2025b-R-4.5.2.eb
```diff
@@ -0,0 +1,107 @@
+easyblock = 'CMakeNinja'
+
+name = 'RStudio-Server'
+version = '2025.09.2+418'
```
ocaisa: Latest release is `2026.05.0+218`, we should bump to that

bedroesb: I though within the 2025b tool chain it would make sense to go to the latest version of RStudio in 2025. But I am happy to bump to a newer version too

ocaisa: We usually go with latest available at the time of the PR unless there is a good reason not to (like it requires an updated version of a dependency that already exists in the toolchain). I don't think that's the case here.

ocaisa:

bedroesb:
I removed the patch since it is patched upstream: https://github.com/rstudio/rstudio/commit/89d142b3a64caf56eca234c92bee3f5d0add5d1e

ocaisa:

ocaisa:
@bedroesb Looks like you hit a reason not to use the latest and greatest: https://gist.github.com/boegelbot/6ae498c21453a045b6b86f39f295f813#file-rstudio-server-2026-05-0-218-foss-2025b-r-4-5-2_partial-log-L472

ocaisa:
From https://github.com/rstudio/rstudio/commit/51dda8f940dce770cc449a16db005c7fad15f05d, I think you have to go back to 2026.01.2+418 (or patch...which feels like too much effort)

ocaisa:
Actually a simple sed string replace at https://github.com/rstudio/rstudio/commit/51dda8f940dce770cc449a16db005c7fad15f05d#diff-e3052ec6d01ba5f9ae8d014fd3bdd8653f7e759e704a77bbfdb99ada532923daR176 might be enough

bedroesb:
@ocaisa is it worthed to make 1.91 version of boost with GCC 14.3 available and update the deps of this recipe? or am I opening a can of worms with that?

bedroesb:
this would make us more ready for the future I feel.

ocaisa:

bedroesb:
@ocaisa this could be a workable solution, using v2026.01.2?

ocaisa:
I'm fine with that, there's a good reason due to the Boost requirement

ocaisa (approved):
LGTM

bedroesb:
assuming 2026a will use GCC 15.2.0, boost will already be good for the latest RStudio

PR #25988, opened 2026-05-13
{tools,vis}[GCCcore/13.2.0,gompi/2023b] Ferret v7.6.0, tcsh v6.24.13

stamenminkov:
No AI usage, based entirely on existing recipes.

PR #25027, opened 2026-01-09
add tcsh dependency to NCL 6.6.2 to avoid using system csh

Thyre:
(created using `eb --new-pr`)

See discussion in Slack: https://easybuild.slack.com/archives/C34UA1HT7/p1767969299695519

The EasyBlock currently doesn't support most of `pre[...]opts`, so a patch is the only way to fix this.
Avoids errors like:

```
# Shell for the command: './ymake -config $PWD'
# Use command history, exit to stop
eb-shell> ./ymake -config $PWD
bash: ./ymake: /bin/csh: bad interpreter: No such file or directory
```

PR #26198, opened 2026-06-09
{data}[GCCcore/13.2.0,foss/2023b] NCL v6.6.2, g2clib v1.9.0

stamenminkov:
AI usage: `GPT-4` for the `preconfig`, `build` and `preinstallopts` for `NCL`

**Depends on:**

- [x] **PR https://github.com/easybuilders/easybuild-easyconfigs/pull/25988 for `tcsh`**


sassy-crick:

easybuild/easyconfigs/n/NCL/NCL-6.6.2-foss-2023b.eb
```diff
@@ -0,0 +1,80 @@
...
+    {'NCL-6.6.2_PROJ-compatibility.patch': '5ce206e5b9edab082a8e192f081faa82e4812ca2ba5ff065b72624be04bb5b83'},
+]
+
+preconfigopts = (
+    "export NCARG_ROOT=%(installdir)s && "
+    "export NCARG=%(installdir)s && "
+    "export YMAKEROOT=%(installdir)s && "
+)
+
+buildopts = 'CFLAGS=-DH5O_INFO_BASIC'
```
boegel: Overwriting `$CFLAGS` is full is not a good idea, should be something like:
```suggestion
buildopts = 'CFLAGS="$CFLAGS -DH5O_INFO_BASIC"'
```

Would also be good to add a line of comment on top to explain why this is required

easybuild/easyconfigs/n/NCL/NCL-6.6.2-foss-2023b.eb
```diff
@@ -0,0 +1,80 @@
...
+    {'NCL-6.6.2_HDF5-110-API.patch': 'aaa3086411d40594db04f6c1800b89372d22e7bce0fb9bc1b8414f035158fe17'},
+    {'NCL-6.6.2_g2clib-link.patch': '9eb7ee9ea314721e0ced93470b26cf7114cca2b3557d02fd1c34f8a9474a0d68'},
+    {'NCL-6.6.2_fix-missing-Copyright.patch': 'e125f3af1a9a78388127215078538551b52153e564e2d316f815c6debf257248'},
+    {'NCL-6.6.2_PROJ-compatibility.patch': '5ce206e5b9edab082a8e192f081faa82e4812ca2ba5ff065b72624be04bb5b83'},
+]
+
+preconfigopts = (
+    "export NCARG_ROOT=%(installdir)s && "
+    "export NCARG=%(installdir)s && "
+    "export YMAKEROOT=%(installdir)s && "
```
boegel: Would be helpful if a comment is added on top of this block to explain why it's required, since existing NCL easyconfigs don't have this....

What bad things happen if these are not used?

boegel (changes requested):

easybuild/easyconfigs/n/NCL/NCL-6.6.2-foss-2023b.eb
stamenminkov: Uhhhhhhhhhh I realised that's not the correct file I've uploaded 😬 

Let me work a bit on that and see if I can find a patch - this is regarding the NCL issue where it mangles the path if the path includes a define statement, like the arch for example.

I think these lines will need to be removed anyway though.

easybuild/easyconfigs/n/NCL/NCL-6.6.2-foss-2023b.eb
stamenminkov: This is to make it compatible with newer versions of `HDF5`, I'll need to find what was the specific error I was hitting.

stamenminkov:
Btw, I do have a patch which changes all the headers from `CSH` to `TCSH`, making it compatible with the `tcsh` module, instead of relying on the system `CSH` (which might not be present).

I think I'll add it here as well.

stamenminkov:
> Btw, I do have a patch which changes all the headers from `CSH` to `TCSH`, making it compatible with the `tcsh` module, instead of relying on the system `CSH` (which might not be present).
> 
> I think I'll add it here as well.

Ahh nevermind, I see that it was already added in https://github.com/easybuilders/easybuild-easyconfigs/pull/25027

However this means now that this PR will depend on https://github.com/easybuilders/easybuild-easyconfigs/pull/25988 for `tcsh`

easybuild/easyconfigs/n/NCL/NCL-6.6.2-foss-2023b.eb
stamenminkov: Okay actually I'm dismissing all this...

easybuild/easyconfigs/n/NCL/NCL-6.6.2-foss-2023b.eb
stamenminkov: Dismissing this part also.

stamenminkov:
@boegel I'm uploading a new patch which I've managed to make, it addresses the issue I was facing with the arch symbols being replaced by a "`1`" when being a substring of the `PATH`...

Then removing the rest of the `configopts`, and just keeping this new patch.

I'm not an expert in CPP, so the patch logic could possibly be improved; I have to say however that I already tried making it a generalized script that would run over a list of possible architectures, but I couldn't achieve to make it work, so now just keeping a long list of the same few lines reapplied for each possible architecture.

Also not sure if all these archs are needed for `NCL`, or if I might have missed some others, but I think it's a fair beginning.

Already tested it over my environment, in which the PATH contains a "`x86_64`", and the substitution is no longer applied, and the installation completes with success.

PR #26217, opened 2026-06-10
{lib}[gfbf/2025b] Safetensors v0.8.0, ...

Micket:
(created using `eb --new-pr`)

Downgrading multiprocess to 0.70.18 to avoid dependency conflict

PR #26241, opened 2026-06-11
{data,lib}[GCCcore/14.3.0,gfbf/2025b] HF-Datasets v5.0.0, huggingface-hub v1.16.1

VRehnberg:
(created using `eb --new-pr`)

requires:

- [ ] https://github.com/easybuilders/easybuild-easyconfigs/pull/26217


VRehnberg:
Ah, saw that Micke snuck in the same multiprocess config yesterday. I should switch to point this at the same PR.

PR #26244, opened 2026-06-11
add alternative checksum for metadat and metafor libraries in R-bundle-CRAN 2025.10

thanghn90:


Micket (commented):
So for checksum updates, we should keep the old ones as well. Make a tuple., preferably only update them after someone has verified that there were no meaningful differences.

thanghn90:
Thanks @Micket , I'm a bit new to EasyBuild's checksum, and I was not aware EasyBuild supports tuple of checksums for each source. Will attempt to make it into a tuple and try it out. I think I might need to modify other R-bundle-CRAN eb files for the same metadat and metafor checksums (if they use the same version of the two libraries).

thanghn90:
Ah, now I see the eb file for R-bundle-CRAN 2025.11 (foss 2025b) has two checksums each for metadat and metafor. My luck to choose 2025.10 to begin with.

Micket:
Test report by @Micket
**SUCCESS**
Build succeeded for 1 out of 1 (total: 35 secs) (1 easyconfigs in total)
vera-icelake-build - Linux Rocky Linux 9.8, x86_64, Intel(R) Xeon(R) Silver 4316 CPU @ 2.30GHz, Python 3.9.25
See https://gist.github.com/Micket/3c9f63672c54b2a38b7405e3e9113b91 for a full test report.

Micket (approved):
lgtm
can't verify the old sources, didn't have them apparently. But these seem correct

PR #25229, opened 2026-02-06
{lib}[gfbf/2025b] SymEngine-python v0.14.0

AlvaroHuanay:
This PR adds SymEngine-Python/0.14.0 for Stages/2026. Compiled and tested on x86_64 and aarch64.

PR #25157, opened 2026-01-27
{tools}[foss/2025b] Firedrake v2025.10.2

sassy-crick:
(created using `eb --new-pr`)

Requires:
- [x] PR #25158
- [x] PR #24393
- [x] PR #25159
- [x] PR #25194

PR #26271, opened 2026-06-15
enhance easyconfig for SymEngine 0.14.0 to also install Python bindings

boegel:
(created using `eb --new-pr`)


Thyre:
We should drop the (just merged) separate EasyConfig then, right?

- https://github.com/easybuilders/easybuild-easyconfigs/pull/25229

sassy-crick:
> We should drop the (just merged) separate EasyConfig then, right?
> 
> #25229

I would guess so. The issue arises with my `Firedrake` PR #25157 and the old `CMake` 3 or 4 dance.

boegel:
> We should drop the (just merged) separate EasyConfig then, right?
> 
>     * [{lib}[gfbf/2025b] SymEngine-Python v0.14.0 #25229](https://github.com/easybuilders/easybuild-easyconfigs/pull/25229)

Oh, I totally missed that... Since `SymEngine-python` already existed (see [here](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/s/SymEngine-python)), it probably makes more sense to keep that, actually, and make `Firedrake` depend on that instead?

sassy-crick:
> > We should drop the (just merged) separate EasyConfig then, right?
> > ```
> > * [{lib}[gfbf/2025b] SymEngine-Python v0.14.0 #25229](https://github.com/easybuilders/easybuild-easyconfigs/pull/25229)
> > ```
> 
> Oh, I totally missed that... Since `SymEngine-python` already existed (see [here](https://github.com/easybuilders/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/s/SymEngine-python)), it probably makes more sense to keep that, actually, and make `Firedrake` depend on that instead?

Either way is fine with me.

boegel:
Using `SymEngine-python` as dependency for Firedrake makes most sense I think, see https://github.com/easybuilders/easybuild-easyconfigs/pull/25157/commits/d89de47a5e2a8b0e71e25e4a7c75d66f036c8f52, added to:
- https://github.com/easybuilders/easybuild-easyconfigs/pull/25157

PR #26346, opened 2026-06-21
Update analysis workflow to include PR in artifact

Micket:
workflow_run is tricky. I think this is the simplest way forward; we save the PR number in the artifact.


ocaisa (approved):
Worth a shot

PR #26380, opened 2026-06-26
Improve python kernel's interoperability from EB to generic jupyter servers

Crivella:
This is an improvement on the behavior of base python kernels that can be used in a generic jupyter server.

Currently the kernel installed with `jupyter-server` is only usable through a server started with EB's python/jupyter binaries.

It is also possible and i would say desirable to start a generic (non-tied to EB) jupyter server and allow to load the varius python kernels inside of it
[jupyterlmod](https://github.com/cmd-ntrf/jupyter-lmod) allows to dynamically load modules in a jupyter server allowing new kernels to be made available on-demand.

This PR address 2 main points:
- hardcodes the path to EB's python in the `kernel.json` file to ensure the correct python is used (maybe this could be improved in `jupyterlmod` but right now even if `PATH` is updated in the notebook, the server that starts the kernel still has the python from the `venv` and not the one from the loaded module)
- Ensure `JUPYTER_PATH` is also set as `EB_ENV_JUPYTER_ROOT` is only used at start-time and not dynamically by `KernelSpecManger` to find available kernels (the alternative would require us to patch also `KernelSpecManager`'s code but at this point i would prefer the alternative described below)

This has been tested by starting a server through a venv+pip installation of `jupyterlab` + `jupyterlmod` and than, loading the patched `jupyter-server` module makes the new kernel appear in the selection list

image

The previous image shows a server started from a venv with only `jupyterlab` and `jupyterlmod` installed through pip 
Without these changes only the original kernel (the one from python used to start the server) would appear, even after loading the modules

## Possible improvements

- Allow the original kernel to still appear/be used with loaded modules
- Change approach entirely and ship an EB specific extension with a custom `KernelSpecManager` to allow finding and dynamically loading a kernel from all installed Python's (at-least the EB ones) that support it (similar to what VSCode does for starting kernels in a jupyter notebook, where any python that can load `ipykernel` can be used)

### DIFF
--- a/easybuild/easyconfigs/j/jupyter-server/jupyter-server-2.17.0-GCCcore-14.3.0.eb
+++ b/easybuild/easyconfigs/j/jupyter-server/jupyter-server-2.17.0-GCCcore-14.3.0.eb
@@ -226,6 +226,17 @@ exts_list = [
     }),
 ]
 
+_python_bin = "${EBROOTPYTHON}/bin/python"
+_kernel_json = "%(installdir)s/share/jupyter/kernels/python3/kernel.json"
+_kernel_desc = "EasyBuild Python %(pyver)s (ipykernel)"
+
+postinstallcmds = [
+    # Hardcode the path to the python executable in the kernel.json file, so that it works even if another python
+    # takes precedence in the PATH
+    f'sed -i "/^ *\\\"python/ s|\\\"python\\\"|\\\"{_python_bin}\\\"|" {_kernel_json}',
+    f"sed -i 's|\"display_name\": .*|\"display_name\": \"{_kernel_desc}\",|' {_kernel_json}",
+]
+
 sanity_check_paths = {
     'files': ['bin/jupyter'],
     'dirs': ['share/jupyter', 'etc/jupyter'],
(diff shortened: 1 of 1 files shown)


PhoenixEmik:
Quite good. We are currently running JupyterLab/JupyterNotebook modules through Open OnDemand. It sounds like improve the flexibility.

Crivella:
Also added a small PoC for a custom kernel manager
https://github.com/Crivella/easybuild_jupyter_kernels

I would say the first approach has the advantage of reducing the risk of mixing packages from separate toolchains (the Python for the kernel must be loaded in jupyterlmod, hence loading eg SciPy for a different toolchain would override the module).
The downside is that the user has to know to load the `jupyter-server` module to enable a kernel and will only be able to see that.

With the custom manager, the user can see all available EB pythons at once (and jupyter-server does not need a rebuild with the new parameters proposed in this PR).
The downside is that is is possible to load a python module from eg the 2025b toolchain with Python 3.13.5 but start a kernel with 3.11 which will most likely lead to errors

Crivella:
Closing this in favor of https://github.com/Crivella/easybuild_jupyter_kernels

Was able to get the best of both worlds, by exposing all installed (through easybuild) kernels if not Python module is already loaded, or only the ones (in general only 1) compatible with the current python version.

This way a user can bot start their own kernel from scratch, and use other mechanism to load modules inside a notebook or they can preload modules from `jupyterlmod` and than only kernels compatible with them will appear as an option

PR #26410, opened 2026-07-02
NetLogo 7.0.4

nikishe:
net logo 7.0.4
  -new download link and an EB java version as opposes to a local host one


easybuild/easyconfigs/n/NetLogo/NetLogo-7.0.4-64.eb
```diff
@@ -0,0 +1,28 @@
...
+
+# NetLogo is precompiled and needs no particular toolchain.
+# It even comes with its own Java.
+toolchain = SYSTEM
+
+source_urls = ['https://downloads.netlogo.org/%(version)s/']
+sources = ['%(name)s-%(version)s%(versionsuffix)s.tgz']
+checksums = ['e64293fe89e6012ea56f2eabcba61318']
+
+dependencies = [('Java', '25.36')]
```
Micket:
```suggestion
dependencies = [('Java', '25')]
```

we use the wrapper module, so it can be updated

easybuild/easyconfigs/n/NetLogo/NetLogo-7.0.4-64.eb
```diff
@@ -0,0 +1,28 @@
...
+versionsuffix = '-64'
+
+homepage = 'https://ccl.northwestern.edu/netlogo/'
+description = """NetLogo is a multi-agent programmable modeling environment. It
+is used by tens of thousands of students, teachers and researchers worldwide.
+It also powers HubNet participatory simulations. It is authored by Uri Wilensky
+and developed at the CCL."""
+
+# NetLogo is precompiled and needs no particular toolchain.
+# It even comes with its own Java.
```
Micket, line 14: So, should we use Java as a dep then?

Micket (changes requested):

easybuild/easyconfigs/n/NetLogo/NetLogo-7.0.4-64.eb
nikishe: It includes java 17 but can run with any java over 11. I declared 25 so that they have a more uptodate java incase they are using any modern java features see this [link](https://docs.netlogo.org/requirements#linux)

`NetLogo should work on standard Debian-based and Red Hat-based Linux distributions. The NetLogo tarball includes a copy of the Java 17 runtime. NetLogo can run using a different Java installed on your system of version 11 or higher.`

Micket: It is best to first
1. confirm that it does make use of the external Java, if it is loaded.
2. if it does, can you move the comment down to the Java dep, make a note/rewrite it to highlight that it can also make use of external (newer) java.

PR #26417, opened 2026-07-02
Fix Triton test for CUDA families and CUDA CCs like `9.0a`

Flamefire:
Strip everything but digits from the passed CUDA CC

### DIFF
--- a/easybuild/easyconfigs/t/Triton/triton_test.py
+++ b/easybuild/easyconfigs/t/Triton/triton_test.py
@@ -16,6 +16,7 @@ src = triton.compiler.ASTSource(
 )
 
 cuda_cc = sys.argv[1].split(',')[-1]
-target = GPUTarget("cuda", int(cuda_cc.replace('.', '')), 32)
+cuda_cc_digits = [d for d in cuda_cc if d.isdigit()]  # Strip dots and suffixes, e.g. from 9.0a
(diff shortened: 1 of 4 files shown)


easybuild/easyconfigs/t/Triton/triton_test.py
```diff
@@ -16,6 +16,7 @@ def kernel(_a, _b):
 )
 
 cuda_cc = sys.argv[1].split(',')[-1]
-target = GPUTarget("cuda", int(cuda_cc.replace('.', '')), 32)
+cuda_cc_digits = [d for d in cuda_cc if d.isdigit()]  # Strip dots and suffixes, e.g. from 9.0a
```
VRehnberg, line 19:
```suggestion
cuda_cc_digits = [d for d in cuda_cc if d.isdigit()]  # Only keep digits, e.g. from 9.0a to 90
```

if you in fact only want to strip dots and suffixes then maybe regexp would have been better. Something like `(\d+(?:[.]\d+)?).*`.

Flamefire: "if you use regex to solve a problem you'll now have 2 problems" ;-)
Jokes aside, I just wanted to keep it simple. The code already states to "keep only digits", so the comment was meant for the "why"

PR #26436, opened 2026-07-10
OpenBLAS-0.3.30-GCC-14.3.0-U74.eb (SiFive U74 / RVA20 target)

hmeiland:
## Summary

Adds `OpenBLAS-0.3.30-GCC-14.3.0-U74.eb`, a build of OpenBLAS 0.3.30 targeting the
**SiFive U74** application core (RVA20 / `rv64gc`) — e.g. the StarFive JH7110 on the
VisionFive 2.

It is built with `buildopts = 'TARGET=U74 DYNAMIC_ARCH=0'` (dynamic-arch dispatch is
not applicable to a fixed in-order core). The `U74` target does not exist in the
OpenBLAS 0.3.30 release, so it is added via the accompanying source patch
`OpenBLAS-0.3.30_add-sifive-u74-target.patch`, which introduces:

- a new `U74` target (`TargetList.txt`, `getarch.c`, `cpuid_riscv64.c`, `param.h`,
  `Makefile.prebuild`, `Makefile.riscv64`, `kernel/riscv64/KERNEL.U74`)
- a hand-tuned scalar 4x4 DGEMM micro-kernel (`kernel/generic/kern_u74.S` +
  `gemmkernel_4x4_u74.c`)

This mirrors the upstream OpenBLAS PR OpenMathLib/OpenBLAS#5903.

## Validation

Built and tested on a StarFive VisionFive 2 (JH7110, 4x U74 @ 1.5 GHz):

- BLAS L3 correctness: full DGEMM/DSYMM/DTRMM/DTRSM/DSYRK/DSYR2K test suite PASSED
- DGEMM performance vs the stock generic `rv64gc` build: single-core **1.77 GFLOP/s**,
  4-core **~6.3 GFLOP/s** — roughly **2x** the ~3.08 GFLOP/s of the default generic
  OpenBLAS on this core
- HPL residual checks PASSED
- `eb --check-style` passes; parses cleanly against current `develop`

## Notes

- Opened as a **draft** to solicit maintainer guidance: is a hardware-target-specific
  easyconfig (`versionsuffix = '-U74'`, `DYNAMIC_ARCH=0`) the preferred vehicle, or would
  you rather U74 be handled via `--optarch` / a generic mechanism once the target lands in
  an OpenBLAS release?
- The source patch is interim: once OpenMathLib/OpenBLAS#5903 is merged and released, a
  stock version-bump easyconfig can build U74 with just `buildopts` (no patch needed).

[...]

PR #5476, opened 2017-12-06
{chem,data.lib,devel}[intel/2017b] AmberMini-16.16.0, LZO-2.09, mpi4py-3.0.0, netcdf4-python-1.3.1, numexpr-2.6.4, ParmEd-2.7.3, PyTables-3.4.2, pandas-0.21.0,  pymbar-3.0.3

tovrstra:
These are easyconfigs related to alchemical free energy calculations with OpenMM.

Full disclosure: I stole the patch file `from /apps/gent/CO7/sandybridge/software/PyTables/3.4.2-intel-2017a-Python-3.6.1/easybuild/pyTables-3.4.2-fix-libs.patch`.

PR #26444, opened 2026-07-11
OpenBLAS-0.3.30-GCC-14.3.0-x60.eb (SpaceMiT X60 / K1 RVV ZVL256B target)

hmeiland:
## OpenBLAS 0.3.30 for the SpaceMiT X60 (K1) — upstream RVV `RISCV64_ZVL256B` target

Companion to #26436 (SiFive U74). This adds an OpenBLAS easyconfig that builds the **already-upstream**
RVV kernels for the SpaceMiT X60 / K1 SoC (RVV 1.0, VLEN=256 — e.g. Orange Pi RV2, Banana Pi BPI-F3,
Milk-V Jupiter).

No custom target is added: `RISCV64_ZVL256B` is already in OpenBLAS. The stock riscv64 build simply
auto-detects the X60 as `RISCV64_GENERIC` (scalar), so this easyconfig forces
`buildopts = 'TARGET=RISCV64_ZVL256B DYNAMIC_ARCH=0'`. Requires GCC >= 14 (the SGEMM RVV kernel fails to
compile on GCC 13.x).

### Why the patch — `OpenBLAS-0.3.30_fix-riscv64-gemv_n-nan.patch`

OpenBLAS 0.3.30's RVV `kernel/riscv64/gemv_n_vector.c` zeroes its vector accumulator with `vfsub_vv`
applied to an **uninitialized** vector register (the `VFILL_ZERO_FLOAT` macro). When that register holds
NaN bit-patterns, `NaN - NaN = NaN`, so **`d/sgemv` can return all-NaN**. A standalone DGEMM benchmark
does not expose this (different kernel), but anything leaning on GEMV does — e.g. **HPL fails with
`residual = nan`** (and `mpirun -np 1` hangs), because HPL's panel factorization uses `dgemv` heavily.

Fixed upstream **after** v0.3.30 (first released in **v0.3.31**) by:
- OpenMathLib/OpenBLAS#5408 — the direct NaN-init fix (commit `72f082f31`)
- OpenMathLib/OpenBLAS#5476 — GEMV_N kernel rewrite, LMUL f64m4 -> f64m8 (commits `2d82d144e` + `36f9cb85b`)

The patch backports 0.3.30's `gemv_n_vector.c` to that fixed upstream state. The function signature is
unchanged, so it is a drop-in replacement.

### Validation — end-to-end, built from this PR (Orange Pi RV2 / SpaceMiT X60, 8 cores @ 1.6 GHz, GCC 14.3.0)

**The bug is live in the shipped EESSI stack — not only in custom builds.** The stock CVMFS
`OpenBLAS/0.3.30-GCC-14.3.0` (dev.eessi.io/riscv) is a `DYNAMIC_ARCH` build whose **LP64** library carries
the `ZVL256B` RVV kernels and dispatches them on the X60, so **stock EESSI HPL fails with `residual = nan`**
on this board — reproduced from a clean CVMFS-only shell (`module load HPL/2.3-foss-2025b`, N=8000, 1×8,
`-np 8`). The run even reports a healthy-looking ~8.5 GFLOP/s; only the residual check reveals the answer is
garbage. (Aside: the *ILP64* `libopenblas64_p` in that install is `RISCV64_GENERIC` — the vector cores are
in the LP64 lib, which is what HPL uses.)

**Built + tested straight from this PR.** `eb --from-pr 26444 --robot` on the board completed with **0
failures** in ~15.5 h — all three OpenBLAS interface iterations (LP64 + ILP64 + ILP64 symbol-suffix), each
running `make tests` (BLAS) **and** the full netlib LAPACK test suite (within
`max_failing_lapack_tests_num_errors`), then sanity check + module generation. The installed module loads
and its LP64 `libopenblas.so` resolves to the RVV `libopenblas_riscv64_zvl256bp`.

**HPL correctness.** With the installed module as the FlexiBLAS backend, HPL **PASSES** — scaled residual
`4.04e-03` (N=8000, 1×8) — versus `residual = nan` on the stock CVMFS backend, same HPL binary and input.
Single-variable proof: swapping **only** `gemv_n_vector.c` flips `dgemv` from all-NaN to correct and HPL
from FAILED to PASSED.

**Performance (this build).** Raw RVV `ZVL256B` DGEMM ~**16.8 GFLOP/s** 8-core (numerically correct). HPL
peaks at ~**10.5 GFLOP/s** (N=20000, NB=256, 2×4 grid; 9.7 at 1×8) — real double-precision vector
throughput in place of the scalar `RISCV64_GENERIC` fallback the X60 selects by default.

### Why 0.3.30 (and not a newer OpenBLAS)

0.3.30 matches the version EESSI currently ships, keeping this a minimal in-place fix. The `gemv_n` fix is
native in **≥ 0.3.31** — but **0.3.31–0.3.33 carry a *separate* RISC-V/K1 `ZVL256B` DGEMM regression**
(OpenMathLib/OpenBLAS#5811, non-PSD from AᵀA), fixed only in the **as-yet-unreleased 0.3.34**
(OpenMathLib/OpenBLAS#5815, merged to the 0.3.34 milestone). So **0.3.30 + this one backport is the correct
interim**; once EESSI moves to 0.3.34 both the version bump and dropping this patch become a clean
one-liner.

### Build notes

- Build with GCC >= 14. On EESSI / dev.eessi.io the C `-march` must be pinned to a value GCC 14.3.0
  accepts — the auto-optarch canonical expansion (`..._zaamo_zalrsc_...`) is rejected by GCC 14.3.0 and
  breaks `getarch`'s `Makefile.prebuild` c_check.
- `run_lapack_tests`/`max_failing_lapack_tests_num_errors` mirror the stock `OpenBLAS-0.3.30-GCC-14.3.0.eb`.

Opened as a **draft** alongside the U74 sibling (#26436); now validated end-to-end from this PR (see
Validation) — ready to flip to review.

## AI disclosure (per the [EasyBuild AI policy](https://docs.easybuild.io/policies/ai/))

This contribution was made with AI assistance.

- **Tool / model:** an OpenCode AI coding agent driving **Claude Opus 4.8** (model
  `claude-opus-4.8`, served via the GitHub Copilot model endpoint).
- **Extent:** the AI root-caused the RVV `gemv_n` NaN bug (via single-threaded differential BLAS
  testing), produced the backport patch (`OpenBLAS-0.3.30_fix-riscv64-gemv_n-nan.patch`, backporting
  upstream OpenMathLib/OpenBLAS#5408 + #5476), wrote this easyconfig, ran the end-to-end
  `eb --from-pr` build + HPL validation, and wrote this PR description.
- **Human in the loop:** a human directed the work, provided and operated the SpaceMiT X60 hardware
  (Orange Pi RV2), and reviewed/validated all results reported above (the from-PR build, the stock-CVMFS
  `residual = nan` reproduction, and the HPL residual `PASSED` checks).

### DIFF
--- a/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.3.30-GCC-14.3.0-x60.eb
+++ b/easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.3.30-GCC-14.3.0-x60.eb
@@ -1,0 +1,70 @@
+name = 'OpenBLAS'
+version = '0.3.30'
+versionsuffix = '-x60'
+
+homepage = 'https://www.openblas.net/'
+description = """OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD version.
+This variant targets the SpaceMiT X60 (SpaceMiT K1 SoC, RVV 1.0, VLEN=256 -- e.g. Orange Pi
+RV2, Banana Pi BPI-F3, Milk-V Jupiter) by building the UPSTREAM RISCV64_ZVL256B RVV kernels
+(the 8x8 vfmacc DGEMM kernel). No X60-specific source patch is required -- ZVL256B is already
+upstream; only the build TARGET is forced, because the stock riscv64 build auto-detects the
+X60 as RISCV64_GENERIC (scalar). Requires GCC >= 14 (the SGEMM RVV kernel fails to compile on
+GCC 13.x)."""
+
+toolchain = {'name': 'GCC', 'version': '14.3.0'}
+
+source_urls = [
+    # order matters, trying to download the large.tgz/timing.tgz LAPACK tarballs from GitHub causes trouble
+    'https://www.netlib.org/lapack/timing/',
+    'https://github.com/xianyi/OpenBLAS/archive/',
+]
+sources = ['v%(version)s.tar.gz']
+# NOTE: these are the STANDARD OpenBLAS build patches shipped with the stock easyconfig
+# (compiler/vectorization workarounds) -- NOT chip-specific. There is deliberately NO X60 patch.
+patches = [
+    ('large.tgz', '.'),
+    ('timing.tgz', '.'),
+    'OpenBLAS-0.3.15_workaround-gcc-miscompilation.patch',
+    'OpenBLAS-0.3.21_fix-order-vectorization.patch',
+    'OpenBLAS-0.3.30_revert-fix-out-of-bounds-access.patch',
+    # Backport of the upstream fix for the RISC-V RVV real-GEMV-N kernel. In v0.3.30,
+    # kernel/riscv64/gemv_n_vector.c zeroes its vector accumulator via `vfsub_vv` on an
+    # UNINITIALIZED vector register (VFILL_ZERO_FLOAT), so d/sgemv can emit NaN when that
+    # register holds NaN bits -> HPL fails with residual=nan (and -np 1 hangs). The kernel was
+    # rewritten upstream after v0.3.30 (clean f64m8 accumulation). REQUIRED for the RVV target:
+    # dgemv is exercised heavily by HPL's panel factorization. Verified: with this patch the
+    # eb-built 0.3.30 passes HPL (residual finite, PASSED); without it, dgemv is all-NaN.
+    'OpenBLAS-0.3.30_fix-riscv64-gemv_n-nan.patch',
+]
+checksums = [
+    {'v0.3.30.tar.gz': '27342cff518646afb4c2b976d809102e368957974c250a25ccc965e53063c95d'},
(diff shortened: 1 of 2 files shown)


easybuild/easyconfigs/o/OpenBLAS/OpenBLAS-0.3.30-GCC-14.3.0-x60.eb
```diff
@@ -0,0 +1,70 @@
+name = 'OpenBLAS'
+version = '0.3.30'
+versionsuffix = '-x60'
```
boegel, line 3: Rather than introducing an alternate easyconfig, why don't we just apply this patch file in the existing `OpenBLAS-0.3.30-GCC-14.3.0.eb` ?

PR #26451, opened 2026-07-13
{tools}[foss/2025b] llama.cpp ad8d821 w/ SpaceMiT IME (RISC-V X60)

hmeiland:
(created with local `git` + `gh` — `eb` is not available in this environment, so no `eb --new-pr` test report)

## Summary

Adds `llama.cpp-ad8d821-x60-ime-foss-2025b.eb` — llama.cpp with the **SpaceMiT IME** int8-matrix backend
(`GGML_CPU_RISCV64_SPACEMIT`) enabled for the **SpaceMiT X60** (K1 SoC; e.g. Orange Pi RV2 / Banana Pi BPI-F3),
on top of the existing `foss/2025b` (GCCcore-14.3.0) stack.

**GCCcore-14.3.0 is NOT rebuilt.** The only toolchain change is a build-time newer assembler:

```python
builddependencies = [
    ('CMake', '3.31.8'),
    ('binutils', '2.46.1'),   # newer `as` only
]
```

binutils 2.46 adds the RISC-V `xsmtvdot` vendor extension (the `smt.vmadot*` opcodes) that the IME backend emits as
inline asm. GCC invokes `as` from `PATH`, and the `binutils/2.46.1` module prepends its `bin/`, so the newer assembler
is used without touching GCC.

## The patch (`llama.cpp-x60-ime-upstream-binutils.patch`)

ggml's SpaceMiT backend was written for the **SpaceMiT vendor toolchain** (mnemonic `vmadot`, ext `xsmtvdotii`,
requires GCC ≥ 15). This 3-file patch makes the **IME1** path build against **mainline binutils 2.46 + GCC 14.3**:

1. `ime1_kernels.cpp`: rename vendor `vmadot` → upstream `smt.vmadot` (16×; **identical encoding `0xe210312b`**,
   verified) and inject `.option arch,+xsmtvdot` into the two asm macros so a plain `-march` works (no GCC ≥ 15 /
   `_xsmtvdotii` needed).
2. `cmake/FindSMTIME.cmake`: the IME1 detection probe uses `.option arch,+xsmtvdot` + `smt.vmadot`.
3. `CMakeLists.txt`: compile `ime2_kernels.cpp` only when IME2 is detected — upstream binutils 2.46 (`xsmtvdot` v1.0)
   is IME1-only (rejects `vpack.vv`/`vfwmadot`/`vmadot.hp`), and the X60 is IME1-only at runtime (`use_ime2=0`). IME2
   dispatch in `ime.cpp` is already `#ifdef`-guarded, so this is link-safe. Backward-compatible: on the vendor
   toolchain IME2 is still detected + compiled.

## Validation

- Cross-compiled (SpaceMiT v1.2.4) and **run on an Orange Pi RV2**: coherent output (`use_ime1: 1`), 264 `smt.vmadot`
  in `libggml-cpu.so`.
- Patch **compile-verified** under the exact target scenario: GCC + base `-march=rv64gcv_zfh_zvfh_zba_zicbop` (no
  `xsmtvdotii`) + `RISCV64_SPACEMIT_IME1` only → emits real `smt.vmadot`.
- Perf (Qwen2.5-0.5B Q4_0, llama-bench): IME ~**1.56× prompt-processing at t=4**. NB: plain RVV at `-t 8` beats IME on
  both pp512 and tg128 (IME is cluster-0-only and regresses at t=8) — this IME build targets the prompt-heavy /
  core-constrained case.

## Notes for reviewers (draft)

- Pins the **validated commit `ad8d821`** (has the ggml `spacemit/` backend + current `FindSMTIME.cmake`). Can bump to
  a `b<NNNN>` release tag once one lands that carries the IME backend.
- RISC-V + SpaceMiT hardware is **not testable in EB CI**; validation was on real X60 silicon.
- Dependencies all already present in easyconfigs: `binutils-2.46.1`, `CMake-3.31.8`, `cURL-8.14.1`, `foss-2025b`.

### DIFF
--- a/easybuild/easyconfigs/b/binutils/binutils-2.46.1-xsmtvdot.eb
+++ b/easybuild/easyconfigs/b/binutils/binutils-2.46.1-xsmtvdot.eb
@@ -1,0 +1,37 @@
+name = 'binutils'
+version = '2.46.1'
+versionsuffix = '-xsmtvdot'
+
+homepage = 'https://directory.fsf.org/project/binutils/'
+
+description = """binutils: GNU binary utilities. This variant adds the SpaceMiT RISC-V `xsmtvdot`
+vendor extension (the `smt.vmadot*` int8 matrix-dot opcodes used by the SpaceMiT X60 / K1 IME
+backend). Stock GNU binutils (incl. 2.46.1) does NOT yet carry `xsmtvdot` -- it is an unmerged
+sourceware mailing-list patch (binutils 2026-April/148941, "RISC-V: Add SpacemiT vendor extensions
+xsmtvdot"). This easyconfig applies that patch. Verified: the built `as` assembles
+`smt.vmadot v2,v3,v4` -> 0xe241b12b (matches the patch's own gas testsuite). Drop this variant once
+the extension lands in an upstream GNU binutils release."""
+
+toolchain = SYSTEM
+
+source_urls = [GNU_SOURCE]
+sources = [SOURCE_TAR_GZ]
+patches = ['binutils-%(version)s_add-spacemit-xsmtvdot.patch']
+checksums = [
+    '364c8faa19ea46c44089c2d59c1fee6eda9a273787d0fe8ab9dfb249069c6aee',  # binutils-2.46.1.tar.gz
+    '68e83c1e58fe5be8e8c4566afbc6d31e4c33ef4b5fc3a05e2546104150ff787c',  # binutils-2.46.1_add-spacemit-xsmtvdot.patch
+]
+
+builddependencies = [
+    ('flex', '2.6.4'),
+    ('Bison', '3.8.2'),
+    # zlib required, but being linked in statically, so not a runtime dep
+    ('zlib', '1.3.2'),
+]
+
+# avoid build failure when makeinfo command is not available
+# see https://sourceware.org/bugzilla/show_bug.cgi?id=15345
+buildopts = 'MAKEINFO=true'
+installopts = buildopts
+
+moduleclass = 'tools'
--- a/easybuild/easyconfigs/b/binutils/binutils-2.46.1_add-spacemit-xsmtvdot.patch
+++ b/easybuild/easyconfigs/b/binutils/binutils-2.46.1_add-spacemit-xsmtvdot.patch
@@ -1,0 +1,323 @@
+diff --git a/bfd/elfxx-riscv.c b/bfd/elfxx-riscv.c
+--- a/bfd/elfxx-riscv.c
++++ b/bfd/elfxx-riscv.c
+@@ -1229,6 +1229,8 @@ static const struct riscv_implicit_subset riscv_implicit_subsets[] =
+   {"xtheadvector", "+zicsr", check_implicit_always},
+   {"xtheadzvamo", "+zaamo", check_implicit_always},
+ 
++  {"xsmtvdot", "+zve32x", check_implicit_always},
++
+   {"v", "+zve64d,+zvl128b", check_implicit_always},
+   {"zvfh", "+zvfhmin,+zfhmin", check_implicit_always},
+   {"zvfhmin", "+zve32f", check_implicit_always},
+@@ -1644,6 +1646,7 @@ static const struct riscv_supported_ext riscv_supported_vendor_x_ext[] =
+   {"xmipscmov",		ISA_SPEC_CLASS_DRAFT,	1, 0, 0 },
+   {"xmipsexectl",	ISA_SPEC_CLASS_DRAFT,	1, 0, 0 },
+   {"xmipslsp",		ISA_SPEC_CLASS_DRAFT,	1, 0, 0 },
++  {"xsmtvdot",		ISA_SPEC_CLASS_DRAFT,	1, 0, 0 },
+   {NULL, 0, 0, 0, 0}
+ };
+ 
+@@ -3062,6 +3065,8 @@ riscv_multi_subset_supports (riscv_parse_subset_t *rps,
+       return riscv_subset_supports (rps, "xmipsexectl");
+     case INSN_CLASS_XMIPSLSP:
+       return riscv_subset_supports (rps, "xmipslsp");
++    case INSN_CLASS_XSMTVDOT:
++      return riscv_subset_supports (rps, "xsmtvdot");
+     default:
+       rps->error_handler
+         (_("internal: unreachable INSN_CLASS_*"));
+@@ -3365,6 +3370,8 @@ riscv_multi_subset_supports_ext (riscv_parse_subset_t *rps,
+       return "xsfvqmaccdod";
+     case INSN_CLASS_XSFVFNRCLIPXFQF:
+       return "xsfvfnrclipxfqf";
++    case INSN_CLASS_XSMTVDOT:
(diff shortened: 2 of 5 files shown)


hmeiland:
## ⚠️ Blocked — stock binutils 2.46.1 does not provide `xsmtvdot`

Self-review turned up a blocker that invalidates this PR as written. **GNU binutils 2.46.1 does not contain the `xsmtvdot` / `smt.vmadot` assembler support** the SpaceMiT IME backend needs:

- binutils-2.46.1 release source: **0** matches for `xsmtvdot` / `smt.vmadot` / `vmadot` / `spacemit`; the RISC-V vendor-extension table has `xthead*`, `xmips*`, `xcv*`, `xsf*`, but no `xsmt`.
- GitHub code search on `bminor/binutils-gdb` (master): `xsmtvdot` → 0 results.

The `xsmtvdot` support is currently only an **unmerged mailing-list patch** ([binutils 2026-April/148941](https://sourceware.org/pipermail/binutils/2026-April/148941.html), v1–v4, still under review). It *is* merged in **LLVM/clang** (llvm/llvm-project#151706) and carried in **SpaceMiT's downstream vendor binutils** (which is what the validated cross-build actually used) — but not in any released GNU binutils.

⇒ `('binutils', '2.46.1', '', SYSTEM)` yields an `as` that rejects `smt.vmadot`; a real build fails at the assembler, exactly like the original binutils-2.44 wall. **Marking this blocked.**

**Plan:** replace the stock-binutils build dependency with a `binutils-2.46.1` easyconfig that applies the SpaceMiT `xsmtvdot` patch (to be verified to actually assemble `smt.vmadot`), until the extension lands in an upstream GNU binutils release. Will update this PR.

hmeiland:
## ✅ Resolved via a patched binutils — verified

Pushed the Option-A fix. Instead of stock `binutils-2.46.1` (which lacks `xsmtvdot`), this now adds a
**`binutils-2.46.1-xsmtvdot`** easyconfig that applies the SpaceMiT `xsmtvdot` patch, and points the
llama.cpp-x60 IME build at it:

```python
# llama.cpp-ad8d821-foss-2025b-x60-ime.eb
builddependencies = [('CMake', '3.31.8'), ('binutils', '2.46.1', '-xsmtvdot', SYSTEM)]
```

**New files**
- `b/binutils/binutils-2.46.1-xsmtvdot.eb` — stock binutils 2.46.1 + the patch below.
- `b/binutils/binutils-2.46.1_add-spacemit-xsmtvdot.patch` — the SpaceMiT `xsmtvdot` v4 patch from the
  sourceware binutils list ([2026-April/148941](https://sourceware.org/pipermail/binutils/2026-April/148941.html)),
  reduced to the functional source hunks (`bfd/elfxx-riscv.c`, `gas/config/tc-riscv.c`, `opcodes/riscv-opc.c`,
  `include/opcode/riscv{,-opc}.h`, `opcodes/riscv-dis.c`, NEWS/texi).

**Verified locally** (not just applied): built binutils 2.46.1 + the patch as a `riscv64-unknown-linux-gnu`
cross-assembler and exercised the new `as`:

| test | result |
|---|---|
| `smt.vmadot v2, v3, v4` (via `.option arch,+xsmtvdot`) | `0xe241b12b` — **matches the patch's own `gas` testsuite** (`x-smt-vdot.d`) |
| `smt.vmadot v2, v0, v1` | `0xe210312b` — matches what ggml emits |
| `-march=rv64gcv_xsmtvdot` route | assembles ✓ |
| without the extension | correctly rejected ✓ |

**Caveat for reviewers:** this carries an **unmerged** binutils patch (the extension is still under review on
the mailing list; it *is* merged in LLVM/clang). It should be dropped once `xsmtvdot` lands in an upstream GNU
binutils release. Keeping this as a draft pending your call on whether an unmerged-vendor-patch binutils belongs
in easyconfigs vs. staying an EESSI-local / clang-based build.

hmeiland:
## ⚠️ End-to-end check: the mechanism works, but this packaging can't merge

I verified the underlying mechanism end-to-end — both patches are correct:

- Built **binutils 2.46.1 + the `xsmtvdot` patch** as a `riscv64` cross-assembler; it assembles
  `smt.vmadot v2,v3,v4` → `0xe241b12b` (byte-for-byte the patch's own gas test), rejects without the extension,
  and IME2 ops are correctly absent.
- **GCC** compiled the *patched* `ime1_kernels.cpp` (base march, `RISCV64_SPACEMIT_IME1` only) and that upstream
  `as` assembled its output → **32 real `smt.vmadot`** in the object. So GCC-inline-asm → upstream-binutils-2.46.1+patch
  works.

**But CI's `test_conflicts` (correctly) fails**, and it's a structural problem, not a typo:

```
Conflict found for dependencies of llama.cpp-ad8d821-foss-2025b-x60-ime:
  binutils-2.46.1-xsmtvdot  vs  binutils-2.44-GCCcore-14.3.0
```

The `foss/2025b` toolchain's `GCCcore-14.3.0` already brings `binutils-2.44`, and EasyBuild forbids two `binutils`
in one dependency graph. The `('binutils', ver, '', SYSTEM)` build-dependency pattern only works when `ver` **matches**
the toolchain's binutils (bootstrapping) — injecting a **newer/different** binutils as an app build-dep is
incompatible with EasyBuild's one-binutils-per-toolchain model. So there is **no clean app-level way** to hand this
build a newer assembler.

**The real paths are toolchain-level, not an app easyconfig:**
- patch the toolchain's own `binutils-2.44-GCCcore-14.3.0` in place (an EESSI-infra decision with a wide blast
  radius; and the 2.46 patch would need porting to the 2.44 tree), **or**
- a future EESSI toolchain generation whose binutils already carries `xsmtvdot` — the clean long-term path, once the
  extension merges into an upstream GNU binutils release, **or**
- a **clang/LLVM** RISC-V build (the `XSMTVDot` support is already merged in LLVM), **or**
- ship the cross-compiled IME binaries **outside** EasyBuild (verified working on the Orange Pi RV2).

Leaving this as a **blocked draft** — the RISC-V IME path is real and the patches are verified, but it needs a
toolchain-level home rather than an app build-dependency. The `binutils-2.46.1-xsmtvdot` easyconfig + patch here are
still useful as the verified patch artifact for whichever toolchain-level route EESSI takes.

hmeiland:
**AI assistance disclosure**

Parts of this contribution were developed with the help of an AI coding assistant, in particular:
- diagnosing the `FindSMTIME.cmake` IME1 probe failure (a `\n` inside the `check_c_source_compiles` test string was double-evaluated by CMake, corrupting the C string literal → `RISCV64_SPACEMIT_IME1` left undefined → `#error`) and its fix (two separate `__asm__` statements);
- the assembler-only override in the easyconfig (`-B<dir>/` in C/CXXFLAGS pointing at a symlinked `as`, instead of PATH-prepending the whole binutils `bin/`, which hijacked `ld` and broke `-lgcc_s` resolution);
- driving the native build and runtime validation on the Orange Pi RV2.

All changes were reviewed and **empirically validated on real SpaceMiT X60 hardware**: `eb` build completed `rc=0`, runtime `use_ime1=1` (4 cluster-0 cores), coherent Qwen2.5-0.5B Q4_0 inference, and `llama-bench` pp512 ≈ 97.6 t/s @ t=4.

PR #26544, opened 2026-07-27
Improve analyzer to drop specific long diffs

Micket:
still draft, need to test it out a bit

### DIFF
--- a/.github/workflows/analyze.py
+++ b/.github/workflows/analyze.py
@@ -85,29 +85,43 @@ print(f'Generating comment for {len(to_diff)} updates softwares')
 # Limit comment size for large PRs:
 if len(to_diff) > 20:  # Too much, either bad PR or some broad change. Not diffing.
     max_diffs_per_software = 0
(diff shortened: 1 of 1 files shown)


.github/workflows/analyze.py
```diff
@@ -85,29 +85,43 @@ def pr_ecs(pr_diff):
...
+                diff_content += gitrepo.git.diff(f'HEAD:{neighbour}', f'HEAD:{new_file}')
+                diff_content += '\n```\n</details>\n\n'
+            comment_diffs.append((diff_title, diff_content))
+
+# Drop largest diff until it fit in a github comment (about 65k)
+for _ in range(10):  # Give up after 10 drops
+    comment = ''.join(f'{a}{b}' for a, b in comment_diffs)
+    if len(comment) <= 65000:
+        break
+    # Find the index of the longest diff, just drop it.
```
boegel, line 119: rather than just dropping it entirely, can we trim it to say the first 1k character instead?

PR #26594, opened 2026-08-04
Add CI checks for PythonBundles

Flamefire:
Based on[ the last confcall ](https://github.com/easybuilders/easybuild/wiki/Conference-call-notes-20260729#qa--other)this adds CI checks to ensure the new rules:

- Single extensions should be PythonPackage
- Python packages which (usually) are build dependencies only should not be extensions of other bundles

EasyConfigs named `*-bundle*` are explicitely ignored.

Running a check over all existings ECs:
- 549 single-extension bundles
- 139 with build dependencies. Most commonly wheel, setuptools, packaging, pyproject-metadata, hatchling

As a companion I can enhance PythonPackage to support `package_name` defaulting to `name`. This will be used a) for `pip` checks and b) to add an extension to modules.
Required decision: I'd only add it as an extension when it is different to `name`.
- Avoids cluttering `module spider/show/av` which would show it twice as the module and as an extension of itself which I don't see any benefit
- Is somehow backwards compatible in not adding extensions for existing PythonPackage modules

I can also look into automatically converting the existing single-extension bundles to make it easier for contributors to do updates. I'd expect to be able to use a simple script for that to do all in one go.

Initial draft created by local AI agent including the list of build-dependencies.

### DIFF
--- a/test/easyconfigs/easyconfigs.py
+++ b/test/easyconfigs/easyconfigs.py
@@ -1334,7 +1334,7 @@ class EasyConfigTest(TestCase):
                             idx_match = idx
                             break
                     if idx_match:
-                        del checksums[idx]
+                        del checksums[idx_match]
                     else:
                         self.fail(f"No checksum found for {patch_ctypes_ld_library_path} in {ec_fn}")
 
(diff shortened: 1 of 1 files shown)


boegel:
I would like input from some other maintainers on this, primarily @smoors ...

smoors:
i don't have a strong opinion on the "Single extensions should be PythonPackage" rule if we have support for a `package_name` parameter, so i propose to add that first.

> Required decision: I'd only add it as an extension when it is different to name.

i agree it avoids showing them twice, but on the other hand i like having all python packages be visible as extensions (marked with `(E)`), just for consistency. but that's a matter of taste, no strong opinion here either.

here is an example to show what we're talking about. `psutil/5.9.8` is an extension of `Python-bundle-PyPI/2024.06-GCCcore-13.3.0`. the other versions are single extensions of their corresponding module.

```bash
$ ml spider psutil                                                                                                                                                                                     

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  psutil:
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
     Versions:
        psutil/5.9.8 (E)
        psutil/6.0.0-GCCcore-13.3.0
        psutil/6.0.0 (E)
        psutil/7.0.0-GCCcore-14.2.0
        psutil/7.0.0 (E)

Names marked by a trailing (E) are extensions provided by another module.
```

Flamefire:
> if we have support for a `package_name` parameter, so i propose to add that first.

I can do that once we agreed on the semantics

> i like having all python packages be visible as extensions (marked with (E)), just for consistency. but that's a matter of taste, no strong opinion here either.

For a HMNS you'll get
```
$ ml spider psutil                                                                                                                                                                                     

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  psutil:
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
     Versions:
        psutil/5.9.8 (E)
        psutil/6.0.0
        psutil/6.0.0 (E)
        psutil/7.0.0
        psutil/7.0.0 (E)

Names marked by a trailing (E) are extensions provided by another module.
```

So IMO less useful there. And it does look inconsistent to me here. But yes: A matter of taste

smoors:
> > if we have support for a `package_name` parameter, so i propose to add that first.
> 
> I can do that once we agreed on the semantics

i'd go for `extension_name`, which seems most clear: if it's not defined, then there is no extension?
alternatively, `package_name` could also work. should be generic so it can be used for Python/R/Perl/... packages.

Flamefire:
Dumping my thoughts:

> i'd go for `extension_name`, which seems most clear: if it's not defined, then there is no extension?
> alternatively, `package_name` could also work. should be generic so it can be used for Python/R/Perl/... packages.

I wanted to use it also and foremost for the pip check as e.g. all PyTorch easyconfigs now fail as the *package/module*  is called `torch` but renaming the easyconfig isn't useful.
So adding the "package" name resolves that. And additionally this can be used to add it as an extension so both "PyTorch" and "torch" would show up in the module list.

The "generic" part is a bit of an issue. We have `modulename` used for import checks, likely because Perl has "modules", but for Julia it is a "package"
For Python it seems mixed: You import [modules](https://docs.python.org/3/tutorial/modules.html) but I'd say the distribution is a package: PyPI: Python **Package** Index.

So I'm not sure we should add this to framework but rather to PythonPackage to avoid the confusion mentioned in the last confcall about "module". But a "package name" in PerlModule seems odd too.

As for `extension_name`: So when that is defined (manually) it will add an extension to the module? Then the pip-check issue needs another solution and we'd have 3 names: extension, import and pip-list names.
If we use that for the pip-list name it means we force adding it as an extension.
Not too bad actually given that you usually define it when it is different to the name and then you'd likely want it listed in the extensions.
Maybe similar semantics as for `modname`:
- Set in EC: Use for pip-check and add as extension
- Not set in EC: use name for pip-check, don't add an extension
- Set to False in EC: skip pip-check and don't add an extension

It just doesn't cover the case where you don't want an extension but correct the pip-check. But so does my `package_name`.

So `extension_name` is more generic and could go into framework but then it might be hard to explain why it's used for the pip check.
So maybe we need a `pip_name`? Then you could do `extension_name = pip_name = 'torch'` if you want both

smoors:
>It just doesn't cover the case where you don't want an extension but correct the pip-check.

why would you want that? doesn't seem useful to me?

i hope we can avoid adding 2 parameters if possible.

Flamefire:
> > It just doesn't cover the case where you don't want an extension but correct the pip-check.
> 
> why would you want that? doesn't seem useful to me?

I can't think of a reason but wanted to point it out in case someone else does

Flamefire:
We now have `extension_name` and https://github.com/easybuilders/easybuild-easyblocks/pull/4224 uses it in PythonPackage/Bundle. So with those merged I think we can enable this CI check.
It might be worth converting the single-extension pythonbundle-ECs from at least the 2 recent toolchains so contributors using them as a starting point won't be blocked.
It's a trivial conversion so not much work

PR #26689, opened 2026-08-21
remove M4 build in easyconfigs CI

branfosj:
Remove this to reduce the number of times our CI processes try to download config.guess.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -152,10 +152,6 @@ jobs:
           eb --search '^foss-2023a.eb' | tee eb_search_foss.out
           grep '/foss-2023a.eb$' eb_search_foss.out
 
-          # try installing M4 with system toolchain (requires ConfigureMake easyblock + easyconfig)
-          # use /tmp/sources because that has cached downloads (see cache step above)
-          eb --prefix /tmp/$USER/$GITHUB_SHA --sourcepath /tmp/sources M4-1.4.18.eb
-
   test-sdist:
     runs-on: ubuntu-22.04
     strategy:
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -152,6 +152,10 @@ jobs:
           eb --search '^foss-2023a.eb' | tee eb_search_foss.out
           grep '/foss-2023a.eb$' eb_search_foss.out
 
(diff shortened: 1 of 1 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -152,10 +152,6 @@ jobs:
           eb --search '^foss-2023a.eb' | tee eb_search_foss.out
           grep '/foss-2023a.eb$' eb_search_foss.out
 
-          # try installing M4 with system toolchain (requires ConfigureMake easyblock + easyconfig)
-          # use /tmp/sources because that has cached downloads (see cache step above)
-          eb --prefix /tmp/$USER/$GITHUB_SHA --sourcepath /tmp/sources M4-1.4.18.eb
```
ocaisa, line 157: I think the problem here is that the cache does not contain the required `config.guess` and is never updated with new sources. Trying to perform a full build has value and I'm a bit reluctant to remove it

branfosj: I am not sure the cache is working at all. If I look at the CI for #26694 I see that M4 download in both of them:
* https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32506391571/job/96847396885?pr=26694#step:8:17226
* https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32506391571/job/96847396676?pr=26694#step:8:238

branfosj: But something has to be done soon. Three of the four most recent PRs to this repo have CI failures for this (and the fourth is failing before getting to this check!).

So it is either:
1. merge this
2. someone else works out how to fix the cache

ocaisa: Ok, I'll merge this for now then and open an issue (I'm on leave next week so can't follow up right now)

ocaisa: Issue opened (https://github.com/easybuilders/easybuild-easyconfigs/issues/26695), I left the check in place but commented out

ocaisa (approved):
LGTM

.github/workflows/unit_tests.yml
ocaisa: Looking at https://github.com/easybuilders/easybuild-easyconfigs/actions/caches, it seems that every PR is getting it's own cache

PR #26699, opened 2026-08-24
reenable M4 build by using cache restore and save

branfosj:
https://github.com/actions/cache#using-a-combination-of-restore-and-save-actions

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -26,13 +26,6 @@ jobs:
       with:
         fetch-depth: 0 # Required for git merge-base to work
 
-    - name: Restore cache of source files in /tmp/sources
-      id: restore-cache-sources
-      uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
-      with:
-        path: /tmp/sources
-        key: eb-sourcepath
-
     - name: set up Python
       uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
       with:
@@ -68,6 +61,14 @@ jobs:
           git clone -b $BRANCH --depth 10 --single-branch https://github.com/easybuilders/easybuild-easyblocks.git
           cd easybuild-easyblocks; git log -n 1; cd -
           pip install $PWD/easybuild-easyblocks
+
+    - name: Restore cache of source files in /tmp/sources
+      id: restore-cache-sources
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
+      with:
+        path: /tmp/sources
+        # Ensure cache is invalidated when configuremake.py changes so we download latest version of config.guess
+        key: eb-sourcepath-${{ hashFiles('easybuild-easyblocks/easybuild/easybuild/easyblocks/generic/configuremake.py') }}
 
     - name: install modules tool
       run: |
@@ -191,10 +192,3 @@ jobs:
           else
             echo "No .git folders found in source tarball: OK"
           fi
-
-    - name: Save cache of source files in /tmp/sources
-      id: save-cache-sources
-      uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
-      with:
-        path: /tmp/sources
-        key: eb-sourcepath
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -47,7 +47,6 @@ jobs:
 
     - name: install EasyBuild framework
       run: |
-          cd $HOME
           # first determine which branch of easybuild-framework repo to install
           BRANCH=develop
           if [ "x$GITHUB_BASE_REF" = 'xmain' ]; then BRANCH=main; fi
@@ -68,17 +67,16 @@ jobs:
       with:
         path: /tmp/sources
         # Ensure cache is invalidated when configuremake.py changes so we download latest version of config.guess
-        key: eb-sourcepath-${{ hashFiles('**/easyblocks/generic/configuremake.py') }}
+        key: eb-sourcepath-${{ matrix.python }}-${{ hashFiles('easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.py') }}
 
     - name: install modules tool
       run: |
-          cd $HOME
           # use install_eb_dep.sh script provided with easybuild-framework
           export INSTALL_DEP=$(which install_eb_dep.sh)
           echo "Found install_eb_dep.sh script: $INSTALL_DEP"
 
           # install modules tool
-          source $INSTALL_DEP ${{matrix.modules_tool}} $HOME
+          source $INSTALL_DEP ${{matrix.modules_tool}} $PWD
 
           # changes in environment are not passed to other steps, so need to create files...
           echo $MOD_INIT > mod_init
@@ -94,13 +92,13 @@ jobs:
           if [ "x$GITHUB_BASE_REF" != 'xmain' ]; then git fetch -v origin ${GITHUB_BASE_REF}:${GITHUB_BASE_REF}; fi
 
           # initialize environment for modules tool
-          if [ -f $HOME/moduleshome ]; then export MODULESHOME=$(cat $HOME/moduleshome); fi
-          source $(cat $HOME/mod_init); type module
+          if [ -f $PWD/moduleshome ]; then export MODULESHOME=$(cat $PWD/moduleshome); fi
+          source $(cat $PWD/mod_init); type module
 
           # make sure 'eb' is available via $PATH, and that $PYTHONPATH is set (some tests expect that);
           # also pick up changes to $PATH set by sourcing $MOD_INIT
           WORKDIR=$GITHUB_WORKSPACE/easybuild-easyconfigs
-          export PATH=$WORKDIR/test/bin:$(cat $HOME/path)
+          export PATH=$WORKDIR/test/bin:$(cat $PWD/path)
           export PYTHONPATH=$WORKDIR
 
           # tell EasyBuild which modules tool is available
@@ -121,7 +119,7 @@ jobs:
 
           # run test suite
           # if tests failed, print error message that is picked up by boegelbot to determine end of test output
-          # python -O -m test.easyconfigs.suite || (echo "ERROR: Not all tests were successful" && exit 1)
+          python -O -m test.easyconfigs.suite || (echo "ERROR: Not all tests were successful" && exit 1)
 
           unset PYTHONPATH
 
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -64,7 +64,7 @@ jobs:
 
           cp easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.py $GITHUB_WORKSPACE/configuremake.check
 
-    - name: Restore cache of source files in /tmp/sources
+    - name: Cache source files in /tmp/sources
       id: restore-cache-sources
       uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
       with:
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -47,6 +47,7 @@ jobs:
 
     - name: install EasyBuild framework
       run: |
+          cd $HOME
           # first determine which branch of easybuild-framework repo to install
           BRANCH=develop
           if [ "x$GITHUB_BASE_REF" = 'xmain' ]; then BRANCH=main; fi
@@ -61,22 +62,25 @@ jobs:
           cd easybuild-easyblocks; git log -n 1; cd -
           pip install $PWD/easybuild-easyblocks
 
+          cp easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.py $GITHUB_WORKSPACE/configuremake.check
+
     - name: Restore cache of source files in /tmp/sources
       id: restore-cache-sources
       uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
       with:
         path: /tmp/sources
         # Ensure cache is invalidated when configuremake.py changes so we download latest version of config.guess
-        key: eb-sourcepath-${{ matrix.python }}-${{ hashFiles('easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.py') }}
+        key: eb-sourcepath-${{ matrix.python }}-${{ hashFiles('configuremake.check') }}
 
     - name: install modules tool
       run: |
+          cd $HOME
           # use install_eb_dep.sh script provided with easybuild-framework
           export INSTALL_DEP=$(which install_eb_dep.sh)
           echo "Found install_eb_dep.sh script: $INSTALL_DEP"
 
           # install modules tool
-          source $INSTALL_DEP ${{matrix.modules_tool}} $PWD
+          source $INSTALL_DEP ${{matrix.modules_tool}} $HOME
 
           # changes in environment are not passed to other steps, so need to create files...
           echo $MOD_INIT > mod_init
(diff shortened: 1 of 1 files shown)


Micket (dismissed):
lgtm

.github/workflows/unit_tests.yml
```diff
@@ -191,3 +191,10 @@ jobs:
           else
             echo "No .git folders found in source tarball: OK"
           fi
+
+    - name: Save cache of source files in /tmp/sources
+      id: save-cache-sources
+      uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
+      with:
+        path: /tmp/sources
+        key: eb-sourcepath
```
Crivella:
```suggestion
        key: eb-sourcepath-${{ hashFiles('easybuild/easyblocks/generic/configuremake.py') }}
        restore-keys: eb-sourcepath-
```

Why not do something like this since https://github.com/easybuilders/easybuild-easyblocks/blob/develop/easybuild/easyblocks/generic/configuremake.py#L61 dictates how `config.guess` is downloaded?

Crivella: Can also go more specific and enforce the hash on the restore part of the actions

Crivella: Most likely need to point to the actual path of `easybuild/easyblocks/generic/configuremake.py` after we setup the repo

.github/workflows/unit_tests.yml
```diff
@@ -26,9 +26,9 @@ jobs:
       with:
         fetch-depth: 0 # Required for git merge-base to work
 
-    - name: Cache source files in /tmp/sources
-      id: cache-sources
-      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # 5.0.5
+    - name: Restore cache of source files in /tmp/sources
```
Crivella: I think this should be moved below `install EasyBuild framework` and be something like

```
    - name: Restore cache of source files in /tmp/sources
      id: restore-cache-sources
      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
      with:
        path: /tmp/sources
        # Ensure cache is invalidated when configuremake.py changes so we download latest version of config.guess
        key: eb-sourcepath-${{ hashFiles('easybuild-easyblocks/easybuild/easybuild/easyblocks/generic/configuremake.py') }}
```

This way we should pick up on `configuremake.py` from the installed easyblocks which dictates how we download `config.guess`

Also we do not need the `restore/save` logic, we can let the top-level action handle the saving at the end of the workflow

.github/workflows/unit_tests.yml
Crivella: Not sure this is actually needed or that it would be useful at all since no cache is being restored in the `test-sdist` workflow and we are not doing any EB specific actions that would write to `/tmp/sources`?

branfosj: I forgot that we have two separate test sections in here and put this block too far down.

branfosj: It should be at https://github.com/easybuilders/easybuild-easyconfigs/pull/26699/changes#diff-c91058a32040621ac54d6ea4194b302a420b7849f6e8561920d23a0e97acf69aR157

branfosj: Removed in b59f4b2, as per the other comment.

.github/workflows/unit_tests.yml
branfosj: Done in b59f4b2

.github/workflows/unit_tests.yml
```diff
@@ -69,6 +62,14 @@ jobs:
           cd easybuild-easyblocks; git log -n 1; cd -
           pip install $PWD/easybuild-easyblocks
 
+    - name: Restore cache of source files in /tmp/sources
+      id: restore-cache-sources
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # 6.1.0
+      with:
+        path: /tmp/sources
+        # Ensure cache is invalidated when configuremake.py changes so we download latest version of config.guess
+        key: eb-sourcepath-${{ hashFiles('easybuild-easyblocks/easybuild/easybuild/easyblocks/generic/configuremake.py') }}
```
Crivella:
```suggestion
        key: eb-sourcepath-${{ matrix.python }}-${{ hashFiles('easybuild-easyblocks/easybuild/easybuild/easyblocks/generic/configuremake.py') }}
```

This will most likely help us prevent race conditions like the one in
https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32389108646/job/96490672189
vs
https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32389108646/job/96490672154

The price to pay is duplicate cache for sources but that should not be to bad.
The alternative would be to have a non-matrix dependent workflow to pre-fetch the sources and cache them

Crivella:
One thing to note is that we will need the new workflow to be ran on top of develop atleast once to generate the re-usable cache in the `refs/heads/develop` scope as PRs from forks cannot write there to avoid cache poisoning attacks [see docs](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-access-for-low-trust-workflow-triggers)

ocaisa:
Required action added to the org, closing and reopening to trigger ci

.github/workflows/unit_tests.yml
ocaisa: Looking at https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32934243915/job/98072310659?pr=26699#step:6:17, I don't think the hash is being calculated. I have a strong suspicion that what you can hash is very limited.

ocaisa: Hmm, read the docs and it seems like it _should_ work, maybe we should use a glob in the path
```suggestion
        key: eb-sourcepath-${{ hashFiles('**/easyblocks/generic/configuremake.py') }}
```

branfosj:
I reran the tests and we can see that the cache is being used as we do not downlod M4 source and `config.guess`:
* before, when setting up the cache for the first time: https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32934243915/job/98072310659?pr=26699#step:8:228
* with restored cache: https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32934243915/job/98107248416?pr=26699#step:8:229

.github/workflows/unit_tests.yml
branfosj: Changed in d9e4c28, where I also temporarily disable running the test suite to speed up checking how this works.

.github/workflows/unit_tests.yml
```diff
@@ -120,7 +121,7 @@ jobs:
 
           # run test suite
           # if tests failed, print error message that is picked up by boegelbot to determine end of test output
-          python -O -m test.easyconfigs.suite || (echo "ERROR: Not all tests were successful" && exit 1)
+          # python -O -m test.easyconfigs.suite || (echo "ERROR: Not all tests were successful" && exit 1)
```
branfosj:
```suggestion
          python -O -m test.easyconfigs.suite || (echo "ERROR: Not all tests were successful" && exit 1)
```

Temporarily disabled - to speed up checking this is working. **Re-enable before merge!**

.github/workflows/unit_tests.yml
Crivella: Beside the path being wrong there is also the problem we are downloading the repos in HOME while apparently `hashFiles` only works within the `GITHUB_WORKSPACE` subdir

https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#hashfiles

```
The path is relative to the GITHUB_WORKSPACE directory and can only include files inside of the GITHUB_WORKSPACE
```

Guess we need to change the CI to download the repos not under $HOME

Crivella:
Have been doing some tests on my repo

After committing the new CI to develop to generate the cache
https://github.com/Crivella/easybuild-easyconfigs/actions/runs/32948224322/job/98113539704

Opened a PR and the cache seems to be restored also across PR triggered workflows, now would need to test on PRs from forks
https://github.com/Crivella/easybuild-easyconfigs/pull/4

Also it seems that the `config.guess` is being properly reused
```

== configuring...
  >> running shell command:
	/tmp/sources/generic/eb_v5.4.1.dev0/ConfigureMake/config.guess
```

https://github.com/Crivella/easybuild-easyconfigs/actions/runs/32949155718/job/98116422044?pr=4

Testing a full test-suite rune
https://github.com/Crivella/easybuild-easyconfigs/actions/runs/32949906632/job/98118760549
if everything goes well will open a PR to your PR

Crivella:
```
FAIL: test_easyconfig_locations (test.easyconfigs.easyconfigs.EasyConfigTest.test_easyconfig_locations)
Make sure all easyconfigs files are in the right location.
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 1243, in test_easyconfig_locations
    self.fail("There should be no easyconfig files in %s, found %s" % (dirpath, easyconfig_files))
AssertionError: There should be no easyconfig files in /home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/easybuild-framework, found ['eb']
FAIL: test_easyconfig_name_clashes (test.easyconfigs.easyconfigs.EasyConfigTest.test_easyconfig_name_clashes)
Make sure there is not a name clash when all names are lowercase
  File "/home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/test/easyconfigs/easyconfigs.py", line 1288, in test_easyconfig_name_clashes
    self.fail("EasyConfigs with case-insensitive name clash: %s" % duplicates)
AssertionError: EasyConfigs with case-insensitive name clash: {'easybuild': ['easybuild', 'easybuild', 'EasyBuild'], 'deprecated': ['deprecated', 'Deprecated'], '__pycache__': ['__pycache__', '__pycache__'], 'term': ['term', 'term'], 'r': ['r', 'R']}
```

Umh i am wondering if it is too hackish to just copy `configuremake.py` into `GITHUB_WORKSPACE` for this it is probably simpler ...

branfosj:
> Also it seems that the `config.guess` is being properly reused

To check that we find `config.guess` in the cache we need to look in the `fetching files and verifying checksums` step. Here we report if the download was successful or failed:
```
>> download succeeded: https://git.savannah.gnu.org/cgit/config.git/plain/config.guess?id=28ea239c53a2d5d8800c472bc2452eaa16e37af2
```
If there is no entry for `config.guess` in that section then we used one from the cache.

.github/workflows/unit_tests.yml
```diff
@@ -69,6 +62,16 @@ jobs:
           cd easybuild-easyblocks; git log -n 1; cd -
           pip install $PWD/easybuild-easyblocks
 
+          cp easybuild-easyblocks/easybuild/easyblocks/generic/configuremake.py $GITHUB_WORKSPACE/configuremake.check
+
+    - name: Restore cache of source files in /tmp/sources
```
branfosj:
```suggestion
    - name: Cache source files in /tmp/sources
```

Crivella:
Testing on my fork after re-generating the cache on `develop`
https://github.com/Crivella/easybuild-easyconfigs/actions/runs/32954513584/job/98133108416

and retriggering CI on a local PR
https://github.com/Crivella/easybuild-easyconfigs/actions/runs/32954831574/job/98134097907?pr=4

the cache seems to be working
```
ls -l /tmp/sources ---
total 8
drwxr-xr-x 3 runner runner 4096 Aug 26 09:45 generic
drwxr-xr-x 3 runner runner 4096 Aug 26 09:44 m
--------
/tmp/sources/generic/eb_v5.4.1.dev0/ConfigureMake/config.guess
ls -la /tmp/sources/generic ---
total 12
drwxr-xr-x 3 runner runner 4096 Aug 26 09:45 .
drwxr-xr-x 4 runner runner 4096 Aug 26 09:45 ..
drwxr-xr-x 3 runner runner 4096 Aug 26 09:45 eb_v5.4.1.dev0
ls -la /tmp/sources/generic/eb_v5.4.1.dev0 ---
total 12
drwxr-xr-x 3 runner runner 4096 Aug 26 09:45 .
drwxr-xr-x 3 runner runner 4096 Aug 26 09:45 ..
drwxr-xr-x 2 runner runner 4096 Aug 26 09:45 ConfigureMake
ls -la /tmp/sources/generic/eb_v5.4.1.dev0/ConfigureMake ---
total 60
drwxr-xr-x 2 runner runner  4096 Aug 26 09:45 .
drwxr-xr-x 3 runner runner  4096 Aug 26 09:45 ..
-rwxr-xr-x 1 runner runner 50622 Aug 26 09:45 config.guess


...

== fetching files and verifying checksums...
  >> sources:
  >> /tmp/sources/m/M4/m4-1.4.18.tar.gz [SHA256: ab2633921a5cd38e48797bf5521ad259bdc4b979078034a3b790d7fec5493fab]
  >> patches:
  >> /home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.18_glibc_2.28.patch [SHA256: a613c18f00b1a3caa46ae4b8b849a0f4f71095ad860f4fcd6c6bb4ae211681fa]
  >> /home/runner/work/easybuild-easyconfigs/easybuild-easyconfigs/easybuild/easyconfigs/m/M4/M4-1.4.18_glibc_2.34.patch [SHA256: 75f0ccc981bf313f5eb4e203a9f8b1ef9e633d840064587405cf360107d4915a]
== ... (took < 1 sec)

```

Crivella:
Also the PR scoped caches have been created here in the last CI run.
After this is merged in develop the run on the merge commit should create the `develop`-scoped cache that we can test on other PRs

Crivella (approved):
LGTM

@ocaisa @boegel  maye you want to give a final check since i was involved in the PR content

ocaisa (approved):
LGTM...let's see it in action!

Crivella:
New caches have been created
```
{                                                                                                                                                                                                                                                                                               
  "total_count": 31,                                                                                                                                                                                                                                                                            
  "actions_caches": [                                                                                                                                                                                                                                                                           
    {                                                                                                                                                                                                                                                                                           
      "id": 7013263584,                                                                                                                                                                                                                                                                         
      "ref": "refs/heads/develop",                                                                                                                                                                                                                                                              
      "key": "eb-sourcepath-3.11-996b20f904e8ae45234c59243477504f91a260c3c19e1485ece729a5f72231be",                                                                                                                                                                                             
      "version": "904532957c69642264795f790b990c9c755a6d15eefd36b1cffec396de53911a",                                                                                                                                                                                                            
      "last_accessed_at": "2026-08-26T11:46:53.601591000Z",                                                                                                                                                                                                                                     
      "created_at": "2026-08-26T11:18:19.200523000Z",                                                                                                                                                                                                                                           
      "size_in_bytes": 2019244                                                                                                                                                                                                                                                                  
    },                                                                                                                                                                                                                                                                                          
    {                                                                                                                                                                                                                                                                                           
      "id": 7013320935,                                                                                                                                                                                                                                                                         
      "ref": "refs/heads/develop",                                                                                                                                                                                                                                                              
      "key": "eb-sourcepath-3.7-996b20f904e8ae45234c59243477504f91a260c3c19e1485ece729a5f72231be",                                                                                                                                                                                              
      "version": "904532957c69642264795f790b990c9c755a6d15eefd36b1cffec396de53911a",                                                                                                                                                                                                            
      "last_accessed_at": "2026-08-26T11:46:46.755121000Z",                                                                                                                                                                                                                                     
      "created_at": "2026-08-26T11:20:12.452260000Z",                                                                                                                                                                                                                                           
      "size_in_bytes": 2019194                                                                                                                                                                                                                                                                  
    },                                                                                                                                                                                                                                                                                          
    {                                                                                                                                                                                                                                                                                           
      "id": 363071,                                                                                                                                                                                                                                                                             
      "ref": "refs/heads/develop",                                                                                                                                                                                                                                                              
      "key": "eb-sourcepath",                                                                                                                                                                                                                                                                   
      "version": "904532957c69642264795f790b990c9c755a6d15eefd36b1cffec396de53911a",                                                                                                                                                                                                            
      "last_accessed_at": "2026-08-26T11:44:42.871985000Z",                                                                                                                                                                                                                                     
      "created_at": "2025-02-10T15:48:47.370453000Z",                                                                                                                                                                                                                                           
      "size_in_bytes": 2019783                                                                                                                                                                                                                                                                  
    },                                                                                                                                                                                                                                                                                          
    {  
...
```

and are being properly restored and used in PRs
https://github.com/easybuilders/easybuild-easyconfigs/actions/runs/32964780636/job/98164791823?pr=26713

PR #26712, opened 2026-08-25
use `gfbf` instead of `GCCcore` toolchain for qrupdate v1.1.2 + build shared lib & run testsuite

SimonPinches:
`qrupdate` calls BLAS/LAPACK routines, so it belongs at the GCC toolchain level where a BLAS library (FlexiBLAS) is available, not at GCCcore.

Also switch from `MakeCp` to `ConfigureMake` so that `make install` installs a properly versioned shared library (`libqrupdate.so.1.1` with soname `libqrupdate.so.1`) alongside the static one, and run the testsuite (needs `-std=legacy` with GCC >= 10).

Since Octave is the only consumer of `qrupdate`, the existing GCCcore-level easyconfigs are removed, as agreed in PR #23229.


easybuild/easyconfigs/q/qrupdate/qrupdate-1.1.2-GCC-11.3.0.eb
```diff
@@ -0,0 +1,46 @@
...
+sources = [SOURCE_TAR_GZ]
+patches = ['%(name)s-%(version)s_makeconf.patch']
+checksums = [
+    'e2a1c711dc8ebc418e21195833814cb2f84b878b90a2774365f0166402308e08',  # qrupdate-1.1.2.tar.gz
+    '4aa0e61a18bd14c76048f7b2f1f52875f7bac4bc3426ce9a37dfdb5b2b0d2379',  # qrupdate-1.1.2_makeconf.patch
+]
+
+dependencies = [
+    ('FlexiBLAS', '3.2.0'),
+]
```
akesandgren: Ah yes.... it needs BLAS, so should really be in gfbf not GCC, sorry didn't think of that.
And thus not need this dependency explicitly.

Crivella:
Can you change the PR to show that the files have been renamed + changed instead of deleted re-created.
Also if BLAS routines are needed this should live on top of `gfbf` at minimum

easybuild/easyconfigs/q/qrupdate/qrupdate-1.1.2-GCC-11.3.0.eb
```diff
@@ -0,0 +1,46 @@
...
+
+skipsteps = ['configure']
+
+local_blas_opts = 'BLAS="-lflexiblas" LAPACK="-lflexiblas"'
```
akesandgren: This should be
```
local_blas_opts = 'BLAS="$LIBBLAS" LAPACK="$LIBLAPACK"'
```
which is generic and doesn't need updating when changing to the iimkl toolchain

akesandgren (dismissed):

easybuild/easyconfigs/q/qrupdate/qrupdate-1.1.2-GCC-11.3.0.eb
SimonPinches: Will update accordingly...

SimonPinches:
> Can you change the PR to show that the files have been renamed + changed instead of deleted re-created. Also if BLAS routines are needed this should live on top of `gfbf` at minimum

Git/GitHub decides to show the files as changed versus deleted/added based on the fraction of content changed.  I tried to show all as renamed but it didn't quite work for all files :-(

Crivella:
Not sure how the current result was achieved but it is not correct (files from 2025b are being moved to 2022a) etc...

You can use `git mv` to rename a file within git, this is not only for the PR to display things nicely, but also to ensure that the history of the file can be properly tracked

SimonPinches:
> Not sure how the current result was achieved but it is not correct (files from 2025b are being moved to 2022a) etc...
> 
> You can use `git mv` to rename a file within git, this is not only for the PR to display things nicely, but also to ensure that the history of the file can be properly tracked

Hi @Crivella,

My first commit does exactly this `git mv`: https://github.com/easybuilders/easybuild-easyconfigs/pull/26712/changes/1726b0a202abe0b62d4a19771d7f0a7970b5eb68, and then the second one makes the edits: https://github.com/easybuilders/easybuild-easyconfigs/pull/26712/changes/ed547041470f09c8519060be89ab90a031b71140.

Where do you see 2025b being moved to 2022a?  They all look correct to me.

Note also that `git mv` is effectively just `git rm` and `git add` rolled into one.  See last section of https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository.

Crivella:
MB gave git more credit and thought a dedicated `mv` would keep better track of the metadata.
It looks like `14.3.0` is ending up in `2022a` from the renames shown in the file changed and git diff, guess as long as the final files are correct it is fine

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 7 out of 7 (total: 1 min 39 secs) (6 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), 1 x NVIDIA NVIDIA GeForce RTX 5060, 580.173.02, Python 3.11.15
See https://gist.github.com/Crivella/1a11e7aa1aa20dcae4065425d5d64a01 for a full test report.

Crivella:

Crivella:
Also running a local test for qeupdate + Octave with 2025b, report coming soon

Crivella:
Test report by @Crivella
**SUCCESS**
Build succeeded for 2 out of 2 (total: 30 mins 6 secs) (2 easyconfigs in total)
crivella-desktop - Linux Ubuntu 22.04.5 LTS (Jammy Jellyfish), x86_64, 13th Gen Intel(R) Core(TM) i9-13900K (skylake), 1 x NVIDIA NVIDIA GeForce RTX 5060, 580.173.02, Python 3.11.15
See https://gist.github.com/Crivella/4fcc7aa8b57135f2ae29e386b4ac6b3c for a full test report.

Crivella (approved):
LGTM

Crivella:

PR #25826, opened 2026-04-22
{devel}[GCC/14.3.0,intel-compilers/2025.2.0] texlive v20250308

SimonPinches:
(created using `eb --new-pr`)

Depends upon:
 - [x] #25824 
 - [x] #25825

PR #26745, opened 2026-08-28
{devel}[GCC/14.3.0] texlive v20250308 w/ teTeX

stamenminkov:
Based on https://github.com/easybuilders/easybuild-easyconfigs/pull/25826

Thought of adding this one as well, since for me it's been a life-saver. The general `texlive` installation is done with the `full` scheme, which is around 7-8GB of downloads (could be more). That's quite massive because it usually ends up being treated as a "standalone" process and it hits certain limits on our clusters, making the whole `texlive` installation very troublesome to deal with.

The `teTeX` scheme comes to be I think the second most complete scheme after the `full` one, and it comprises some 1.5-2GB of packages and fonts, making it a lighter yet still an effective option for environments that won't hold the full `texlive`.


smoors:

smoors (approved):
lgtm

smoors:
Going in, thanks @stamenminkov!

PR #26784, opened 2026-09-03
{tools}[foss/2025b] eccodes-python v2.43.0

stamenminkov:
No AI usage


Micket (approved):
lgtm

easybuild/easyconfigs/e/eccodes-python/eccodes-python-2.43.0-foss-2025b.eb
```diff
@@ -0,0 +1,42 @@
...
+    ('cffi', '1.17.1'),
+    ('ecCodes', version),
+]
+
+exts_list = [
+    ('findlibs', '0.1.2', {
+        'checksums': ['1f56d220c69686392ebdc4c65b32ee344818bca633643a8c97592957d1728122'],
+    }),
+    ('eckitlib', '1.32.4.8', {
+        'source_tmpl': '%(name)s-%(version)s-cp313-cp313-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl',
```
Micket, line 26: ah, actually i was to quick to glanced over these. 

A bit unfortunate to pick up (only x86) wheels. 
I understand it's more work, but 
https://github.com/ecmwf/eckit
https://github.com/ecmwf/fckit
https://github.com/ecmwf/eccodes
seems like it could be build from source. at least, we should try to cover aarch64 wheels these days

Micket (changes requested):

easybuild/easyconfigs/e/eccodes-python/eccodes-python-2.43.0-foss-2025b.eb
stamenminkov: @Micket let me look into that, I'll correct it.

I actually do have existing ECs for `ecKit` and `fcKit` already, I'm thinking I could review and add them here, I'm just not sure if to include Python there or to keep a separate EC for the Python bindings, same as it's been done with `ecCodes`... what do you think?

Also there's the fact that the releases on github are always one level above, for example while on pypi there are the three `1.32.4.8`, `1.32.4.9` and `1.32.4.11`, on github there's only a single major `1.32.4` release, so I guess we'll need to constrain it to this level as an EC as well.

I'm also genuinely confused by the fact that there's both:
https://pypi.org/project/eckit/
and
https://pypi.org/project/eckitlib/
which seems to be the same actual package.

PR #26803, opened 2026-09-07
Remove outdated `nvofbf` toolchain easyconfig (+ easyconfig for VASP using it)

Micket:
(created using `eb --new-pr`)

Lets clean this out, one should just use NVHPC toolchain now, this one just causes confusion


Micket:
Ah, so there was an old VASP in there. We really shouldn't be using this toolchain, and it's confusing enough as is with all the various names. It's premature according to policy, but I want this gone.

boegel:
@Micket Should we also deprecate the support for using this toolchain in framework?

Micket:
yes i was just going to stop using it first

jfgrimm (approved):

jfgrimm:
Going in, thanks @Micket!

Issue #23253, opened 2025-07-02
Pillow-SIMD-10.4.0-* isn't actually the SIMD fork, it's just plain Pillow


Micket:
Pillow hasn't rebased their changes since version 9.5.0. Our newer easyconfigs are just incorrectly picking up the plain Pillow branches and tags because it's a fork of the original Pillow. 

So, Pillow-SIMD-10.4.0-GCCcore-13.2.0.eb and Pillow-SIMD-10.4.0-GCCcore-13.3.0.eb are just lies.
The Pillow-SIMD repo isn't seing much action these days, so i doubt thy will start rebasing again. Last update was 9.5 (and pillow is on 11.3 now).

Clearly, noone has even noticed this, so perhaps this isn't needed anymore.

@pavelToman I think you introduced the 10.4.0 versions first, i assume you found it among the tags/forks on their github?

pavelToman:
Yes, there is v10.4.0 in Pillow-SIMD repo: https://github.com/uploadcare/pillow-simd/tree/10.4.0 so I used it

Micket:
Just to be extra clear in case someone is unsure;  that tag + branch is just inherited from the upstream original non-simd Pillow (one can tell from the link just look at the README which is the original Pillow readme instead of the SIMD information)
One must stick to the `simd/x.x.x`  branches (which hasn't been updated since the last 9.5.x rebase), or whatever releases that exist on PyPI.

Micket:
The question that remains is do we remove the faulty Pillow-SIMD's? Or just comment inside them that they are actually wrong? We need to make sure we don't keep adding new broken ones.

PR #26856, opened 2026-09-14
{vis}[GCC/14.3.0] Pillow-SIMD v12.3.0

sassy-crick:
(created using `eb --new-pr`)

It appears to be that `pybind11` is required, hence `GCC` instead of `GCCcore`


sassy-crick:

sassy-crick:

easybuild/easyconfigs/p/Pillow-SIMD/Pillow-SIMD-12.3.0-GCC-14.3.0.eb
```diff
@@ -0,0 +1,50 @@
...
+    ('LittleCMS', '2.17'),
+]
+
+# patch setup.py to prefix hardcoded /usr/* and /lib paths with value of %(sysroot) template
+# (which will be empty if EasyBuild is not configured to use an alternate sysroot);
+# see also https://gitlab.com/eessi/support/-/issues/9
+preinstallopts = """sed -i 's@"/usr/@"%(sysroot)s/usr/@g' setup.py && """
+preinstallopts += """sed -i 's@"/lib@"%(sysroot)s/lib@g' setup.py && """
+
+exts_defaultclass = 'PythonPackage'
```
smoors: this can be removed, it's already the default for `PythonBundle`

easybuild/easyconfigs/p/Pillow-SIMD/Pillow-SIMD-12.3.0-GCC-14.3.0.eb
```diff
@@ -0,0 +1,50 @@
...
+        'sources': ['%(version)s.tar.gz'],
+        'checksums': ['9dd8253265dca7c8ed2984ef8bd65b7e618d3c471f1f8cebd9152b248a25a380'],
+        'modulename': 'PIL',
+    }),
+]
+
+sanity_check_paths = {
+    'files': [],
+    'dirs': ['lib/python%(pyshortver)s/site-packages/PIL'],
+}
```
smoors: this is already the default, so the entire `sanity_check_paths` can be removed

easybuild/easyconfigs/p/Pillow-SIMD/Pillow-SIMD-12.3.0-GCC-14.3.0.eb
sassy-crick: Done. Thanks for pointing that out. The original EasyConfig was `PythonPackage`, so that is where that came from.

easybuild/easyconfigs/p/Pillow-SIMD/Pillow-SIMD-12.3.0-GCC-14.3.0.eb
sassy-crick: As above.

easybuild/easyconfigs/p/Pillow-SIMD/Pillow-SIMD-12.3.0-GCC-14.3.0.eb
smoors: it's still there ;)

sassy-crick: My wrong, I was only looking at line 32 + 33.
Should be all fixed now.

sassy-crick:

Micket:
So, the reason we stopped adding new Pillow-SIMD, was because it was unmaintained for some time, and we have actually just incorrectly taken the tags from the upstream fork, i.e. just good old normal Pillow. 

https://github.com/easybuilders/easybuild-easyconfigs/issues/23253

I'm a bit doubtful this is actually Pillow-SIMD. I think it might be just Pillow. No idea where you got the download link from, but one has to be very careful and pick the correct branch here.

Micket:
I'm not seing even a 12.3.0 tag for SIMD variant. 
I *think* the tag `v12.2.0.post0` might be the SIMD fork.
https://github.com/uploadcare/pillow-simd/tags and it seems like all these recent tags were added.. uh 3 hours ago? seems like quite a conicidence. Or a force push. I don't know if these tags should be trusted at all.
https://pypi.org/project/Pillow-SIMD/#history there is no release on PyPI for years, except the 10.0.1.post0 one he made.. uh 2 hours ago.

sassy-crick:
> So, the reason we stopped adding new Pillow-SIMD, was because it was unmaintained for some time, and we have actually just incorrectly taken the tags from the upstream fork, i.e. just good old normal Pillow.
> 
> #23253
> 
> I'm a bit doubtful this is actually Pillow-SIMD. I think it might be just Pillow. No idea where you got the download link from, but one has to be very careful and pick the correct branch here.

Thanks for finding this, I was not aware of the problems around that. I used the previously merged `Pillow-SIMD-10.4.0-GCCcore-13.3.0.eb` as base and update it to the latest version using the link from above EasyConfig file. 
Background: I want to bring `PyTorch-bundle` to the latest version, and the current version `PyTorch-bundle-2.1.2-foss-2023a-CUDA-12.1.1.eb` includes:
```
    ('Pillow-SIMD', '9.5.0'),  # for torchvision
```
So that is where that is all coming from. 
So, if that is all wrong, as your issue indicates, what to do?

As for the recent push you mentioned in the second comment: no idea. That was not me at all!

So thanks for your awareness here, much appreciated!

Micket:
You can just use Pillow for anyplace Pillow-SIMD has been used. `Pillow-SIMD-10.4.0-GCCcore-13.3.0.eb` is wrong.

smoors:
i quickly checked the sources of the `v12.2.0.post0` tag of pillow-simd, and it does contain extra SIMD code. not sure why this version is not (yet) on PyPI though.

Flamefire:
> It appears to be that pybind11 is required, hence GCC instead of GCCcore

This has been identified as a mistake already. Preparing https://github.com/easybuilders/easybuild-easyconfigs/pull/26864 to drop pybind11 to GCCcore

sassy-crick:
> i quickly checked the sources of the `v12.2.0.post0` tag of pillow-simd, and it does contain extra SIMD code. not sure why this version is not (yet) on PyPI though.

Not sure about why it is not in PyPI (yet), I don't know the process. 
*IF* I wait for PR #26864 to get merged, check again in PyPI to see if that finally has landed there, we could upgrade it to the `v12.2.0.post0` tag release @smoors  mentioned I would suggest.

Micket:
The developer is very actively working on them, doesn't mark releases on the github page at all, and given how hard it is to identify what even if the simd forks on the github page, I don't think we should use the github sources at all, ever.
Switching to them was the original mistake which got us into this situation. 

If the developer considers them released, it seems to be published on pypi, so we should treat that as the source of truth exclusively.
https://pypi.org/project/pillow-simd/#history
developer seems to be actively working through the releases. Up to 12.1.1.post0 now. 


Also, I recommend making sure never to actually use this as a direct dependency. Since we basically treat it as shadowing the original pillow-simd, we should
1. only ever do the same major-minor version as Pillow for the toolchain
2. need to always load it last, so, probably a bad idea to mix in as a dependency.. almost anywhere. No gauarante that it's first in the import priority if we do that.
It's also not a dependency for the newest python-bundle (even though it's old), so of course, it's not needed.

PR #26646, opened 2026-08-13
{data}[gfbf/2025b] polars v1.41.2

PhoenixEmik:
(created using `eb --new-pr`)
Created with the assistance of Codex, GPT-5.6.
Copied from 2026.1.

PR #25056, opened 2026-01-14
{bio}[foss/2025b] MultiQC v1.33

sxpl-DavidSchmidt:
(created using `eb --new-pr`)
depends on:
- #25058
- #25054

PR #26881, opened 2026-09-16
{bio}[foss/2025b] MultiQC v1.35

jpecar:
(created using `eb --new-pr`)

Requires #26646, obsoletes #25056


boegel:

easybuild/easyconfigs/m/MultiQC/MultiQC-1.35-foss-2025b.eb
```diff
@@ -0,0 +1,94 @@
...
+    }),
+    ('jmespath', '1.0.1', {
+        'checksums': ['90261b206d6defd58fdd5e85f478bf633a2901798906be2ad389150c5c60edbe'],
+    }),
+    ('botocore', '1.39.9', {
+        'checksums': ['02f141c2849e4589a79feea245ce4ecc478d48b7865572445af8aae3b041772d'],
+    }),
+    ('boto3', '1.39.9', {
+        'checksums': ['e3d3a6b617e1575e7ec854c820a882ab2e189a0421e74dc0dca2c9e13d4370a5'],
+    }),
```
boegel, line 74: these 4 extensions can be avoided if we just use `boto3-1.39.8-GCCcore-14.3.0.eb` as a dependency for this easyconfig?

boegel (changes requested):

easybuild/easyconfigs/m/MultiQC/MultiQC-1.35-foss-2025b.eb
jpecar: Very likely.

